PR #51, opened 2014-11-08
Add Sphinx skeleton and a few example docs.

dnouri:
Adds Sphinx docs.  To build them, make sure you have nntools and dependencies 'active' (as in `source bin/activate`), then navigate to the `docs/` directory and install Sphinx: `pip install requirements.txt`.  (If there's requirements missing there, let's add them, I haven't made 100% sure.)

Now you can build the docs with `make html`, and navigate with your browser to `_build/html/index.html` to take a look.

I've added a few example docs here and there.  Feel free to throw them away, they're just here to demonstrate some of the syntax, and how I like to structure my docs.

Once we have docs somewhat in shape, I suggest we look into using readthedocs.org for hosting.  It's an excellent service which, after you set up a simple Github hook, will compile docs and put them up at the readthedocs.org website.


docs/requirements.txt
```diff
@@ -0,0 +1,2 @@
+Jinja2==2.7.3
+Sphinx==1.2.3
```
f0k: It's been a long time, but what is the "Jinja2" dependency needed for? Won't Sphinx pull in some good version by itself?

PR #52, opened 2014-11-17
Allow Layers to be constructed from input shapes

f0k:
As discussed in #17, it seemed to be a good idea to allow Layer instances to be constructed from input shapes rather than other Layer instances. While `get_output()` would not work for such layers, they could still be used standalone with `get_output_for()` and `get_output_shape_for()` -- after all, a layer only needs to know about its input and input shape to function.

We thought that implementing this would only touch the `Layer` base class (and the `MultipleInputsLayer` base class). However, while implementing it, I found that several other layer implementations rely on `self.input_layer.get_output_shape()` to work. There are about four ways of solving it:
1. Distinguishing between `self.input_layer` being a shape tuple or a layer in each of these cases
2. Moving this distinction into a new `get_input_shape()` method in the base class
3. Relying on this distinction being present in `Layer.get_output_shape()` and using `super(XYZLayer, self).get_output_shape()` instead of `self.input_layer.get_output_shape()` (this only works for classes directly derived from `Layer`, because in this case the super `get_output_shape()` will just return the input shape, so that's both ugly and fragile)
4. Wrapping a shape tuple in something that implements `get_output_shape()`

It turns out that we have a ready-made class for option 4, which is `InputLayer`. By wrapping shape tuples given in the constructor in `InputLayer` instances, only the base classes need to be changed. However, it feels like an unnecessary / misplaced shortcut now... what do we gain compared to requiring the user to provide an `InputLayer` instance herself or himself? Should we possibly choose option 2 instead? Should we create an `InputShape` class that really just wraps a shape tuple with a `get_output_shape()` method? Or is this proposal a good solution anyway?

### DIFF
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -23,22 +23,22 @@ class Layer(object):
     network's output :class:`Layer` instance can double as a handle to the
     full network.
     """
-    def __init__(self, input_layer, name=None):
+    def __init__(self, incoming, name=None):
         """
         Instantiates the layer.
 
         :parameters:
-            - input_layer : a :class:`Layer` instance or a tuple
+            - incoming : a :class:`Layer` instance or a tuple
                 the layer feeding into this layer, or the expected input shape
             - name : a string or None
                 an optional name to attach to this layer
         """
-        if isinstance(input_layer, tuple):
-            self.input_shape = input_layer
+        if isinstance(incoming, tuple):
+            self.input_shape = incoming
             self.input_layer = None
         else:
-            self.input_shape = input_layer.get_output_shape()
-            self.input_layer = input_layer
+            self.input_shape = incoming.get_output_shape()
+            self.input_layer = incoming
         self.name = name
 
     def get_params(self):
@@ -246,22 +246,22 @@ class MultipleInputsLayer(Layer):
     It should be subclassed when implementing new types of layers that
     obtain their input from multiple layers.
     """
-    def __init__(self, input_layers, name=None):
+    def __init__(self, incomings, name=None):
         """
         Instantiates the layer.
 
         :parameters:
-            - input_layers : a list of :class:`Layer` instances or tuples
+            - incomings : a list of :class:`Layer` instances or tuples
                 the layers feeding into this layer, or expected input shapes
             - name : a string or None
                 an optional name to attach to this layer
         """
-        self.input_shapes = [input_layer if isinstance(input_layer, tuple)
-                             else input_layer.get_output_shape()
-                             for input_layer in input_layers]
-        self.input_layers = [None if isinstance(input_layer, tuple)
-                             else input_layer
-                             for input_layer in input_layers]
+        self.input_shapes = [incoming if isinstance(incoming, tuple)
+                             else incoming.get_output_shape()
+                             for incoming in incomings]
+        self.input_layers = [None if isinstance(incoming, tuple)
+                             else incoming
+                             for incoming in incomings]
         self.name = name
 
     def get_output_shape(self):
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -19,12 +19,26 @@ class Layer(object):
     The :class:`Layer` class represents a single layer of a neural network.
     It should be subclassed when implementing new types of layers.
 
-    Because each layer keeps track of the layer(s) feeding into it, a
-    network's output :class:`Layer` instance doubles as a handle to the full
-    network.
+    Because each layer can keep track of the layer(s) feeding into it, a
+    network's output :class:`Layer` instance can double as a handle to the
+    full network.
     """
     def __init__(self, input_layer, name=None):
-        self.input_layer = input_layer
+        """
+        Instantiates the layer.
+
+        :parameters:
+            - input_layer : a :class:`Layer` instance or a tuple
+                the layer feeding into this layer, or the expected input shape
+            - name : a string or None
+                an optional name to attach to this layer
+        """
+        if isinstance(input_layer, tuple):
+            self.input_shape = input_layer
+            self.input_layer = None
+        else:
+            self.input_shape = input_layer.get_output_shape()
+            self.input_layer = input_layer
         self.name = name
 
     def get_params(self):
@@ -74,10 +88,9 @@ class Layer(object):
 
         :note:
             When implementing a new :class:`Layer` class, you will usually
-            keep this unchanged and just override `get_output_shape_for()`.        
+            keep this unchanged and just override `get_output_shape_for()`.
         """
-        input_shape = self.input_layer.get_output_shape()
-        return self.get_output_shape_for(input_shape)
+        return self.get_output_shape_for(self.input_shape)
 
     def get_output(self, input=None, *args, **kwargs):
         """
@@ -108,6 +121,10 @@ class Layer(object):
         if isinstance(input, dict) and (self in input):
             # this layer is mapped to an expression or numpy array
             return utils.as_theano_expression(input[self])
+        elif self.input_layer is None:
+            raise RuntimeError("get_output() called on a free-floating layer; "
+                               "there isn't anything to get its input from. "
+                               "Did you mean get_output_for()?")
         else: # in all other cases, just pass the input on to the next layer.
             layer_input = self.input_layer.get_output(input, *args, **kwargs)
             return self.get_output_for(layer_input, *args, **kwargs)
@@ -230,17 +247,34 @@ class MultipleInputsLayer(Layer):
     obtain their input from multiple layers.
     """
     def __init__(self, input_layers, name=None):
-        self.input_layers = input_layers
+        """
+        Instantiates the layer.
+
+        :parameters:
+            - input_layers : a list of :class:`Layer` instances or tuples
+                the layers feeding into this layer, or expected input shapes
+            - name : a string or None
+                an optional name to attach to this layer
+        """
+        self.input_shapes = [input_layer if isinstance(input_layer, tuple)
+                             else input_layer.get_output_shape()
+                             for input_layer in input_layers]
+        self.input_layers = [None if isinstance(input_layer, tuple)
+                             else input_layer
+                             for input_layer in input_layers]
         self.name = name
 
     def get_output_shape(self):
-        input_shapes = [input_layer.get_output_shape() for input_layer in self.input_layers]
-        return self.get_output_shape_for(input_shapes)
+        return self.get_output_shape_for(self.input_shapes)
 
     def get_output(self, input=None, *args, **kwargs):
         if isinstance(input, dict) and (self in input):
             # this layer is mapped to an expression or numpy array
             return utils.as_theano_expression(input[self])
+        elif any(input_layer is None for input_layer in self.input_layers):
+            raise RuntimeError("get_output() called on a free-floating layer; "
+                               "there isn't anything to get its inputs from. "
+                               "Did you mean get_output_for()?")
         else: # in all other cases, just pass the network input on to the next layers.
             layer_inputs = [input_layer.get_output(input, *args, **kwargs) for input_layer in self.input_layers]
             return self.get_output_for(layer_inputs, *args, **kwargs)
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -16,10 +16,10 @@ __all__ = [
 
 
 class Conv1DLayer(Layer):
-    def __init__(self, input_layer, num_filters, filter_length, stride=1, border_mode="valid", untie_biases=False,
+    def __init__(self, incoming, num_filters, filter_length, stride=1, border_mode="valid", untie_biases=False,
                  W=init.Uniform(), b=init.Constant(0.), nonlinearity=nonlinearities.rectify,
                  convolution=conv.conv1d_mc0, **kwargs):
-        super(Conv1DLayer, self).__init__(input_layer, **kwargs)
+        super(Conv1DLayer, self).__init__(incoming, **kwargs)
         if nonlinearity is None:
             self.nonlinearity = nonlinearities.identity
         else:
@@ -93,10 +93,10 @@ class Conv1DLayer(Layer):
 
 
 class Conv2DLayer(Layer):
-    def __init__(self, input_layer, num_filters, filter_size, strides=(1, 1), border_mode="valid", untie_biases=False,
+    def __init__(self, incoming, num_filters, filter_size, strides=(1, 1), border_mode="valid", untie_biases=False,
                  W=init.Uniform(), b=init.Constant(0.), nonlinearity=nonlinearities.rectify,
                  convolution=T.nnet.conv2d, **kwargs):
-        super(Conv2DLayer, self).__init__(input_layer, **kwargs)
+        super(Conv2DLayer, self).__init__(incoming, **kwargs)
         if nonlinearity is None:
             self.nonlinearity = nonlinearities.identity
         else:
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -42,7 +42,7 @@ class Conv1DLayer(Layer):
             self.b = self.create_param(b, (num_filters,), name="b")
 
     def get_W_shape(self):
-        num_input_channels = self.input_layer.get_output_shape()[1]
+        num_input_channels = self.input_shape[1]
         return (self.num_filters, num_input_channels, self.filter_length)
 
     def get_params(self):
@@ -65,9 +65,9 @@ class Conv1DLayer(Layer):
 
     def get_output_for(self, input, input_shape=None, *args, **kwargs):
         # the optional input_shape argument is for when get_output_for is called
-        # directly with a different shape than the output_shape of self.input_layer.
+        # directly with a different shape than self.input_shape.
         if input_shape is None:
-            input_shape = self.input_layer.get_output_shape()
+            input_shape = self.input_shape
 
         filter_shape = self.get_W_shape()
 
@@ -119,7 +119,7 @@ class Conv2DLayer(Layer):
             self.b = self.create_param(b, (num_filters,), name="b")
 
     def get_W_shape(self):
-        num_input_channels = self.input_layer.get_output_shape()[1]
+        num_input_channels = self.input_shape[1]
         return (self.num_filters, num_input_channels, self.filter_size[0], self.filter_size[1])
 
     def get_params(self):
@@ -145,9 +145,9 @@ class Conv2DLayer(Layer):
 
     def get_output_for(self, input, input_shape=None, *args, **kwargs):
         # the optional input_shape argument is for when get_output_for is called
-        # directly with a different shape than the output_shape of self.input_layer.
+        # directly with a different shape than self.input_shape.
         if input_shape is None:
-            input_shape = self.input_layer.get_output_shape()
+            input_shape = self.input_shape
 
         filter_shape = self.get_W_shape()
 
--- a/lasagne/layers/corrmm.py
+++ b/lasagne/layers/corrmm.py
@@ -76,7 +76,7 @@ class Conv2DMMLayer(MMLayer):
         self.corr_mm_op = GpuCorrMM(subsample=self.strides, pad=self.pad)
 
     def get_W_shape(self):
-        num_input_channels = self.input_layer.get_output_shape()[1]
+        num_input_channels = self.input_shape[1]
         return (self.num_filters, num_input_channels, self.filter_size[0], self.filter_size[1])
 
     def get_params(self):
--- a/lasagne/layers/corrmm.py
+++ b/lasagne/layers/corrmm.py
@@ -31,10 +31,10 @@ class MMLayer(Layer):
 
 
 class Conv2DMMLayer(MMLayer):
-    def __init__(self, input_layer, num_filters, filter_size, strides=(1, 1), border_mode=None, untie_biases=False,
+    def __init__(self, incoming, num_filters, filter_size, strides=(1, 1), border_mode=None, untie_biases=False,
                  W=init.Uniform(), b=init.Constant(0.), nonlinearity=nonlinearities.rectify, pad=None,
                  flip_filters=False, **kwargs):
-        super(Conv2DMMLayer, self).__init__(input_layer, **kwargs)
+        super(Conv2DMMLayer, self).__init__(incoming, **kwargs)
         if nonlinearity is None:
             self.nonlinearity = nonlinearities.identity
         else:
--- a/lasagne/layers/cuda_convnet.py
+++ b/lasagne/layers/cuda_convnet.py
@@ -97,10 +97,10 @@ class Conv2DCCLayer(CCLayer):
 
     def get_W_shape(self):
         if self.dimshuffle:
-            num_input_channels = self.input_layer.get_output_shape()[1]
+            num_input_channels = self.input_shape[1]
             return (self.num_filters, num_input_channels, self.filter_size, self.filter_size)
         else:
-            num_input_channels = self.input_layer.get_output_shape()[0]
+            num_input_channels = self.input_shape[0]
             return (num_input_channels, self.filter_size, self.filter_size, self.num_filters)
 
     def get_params(self):
@@ -264,8 +264,7 @@ class NINLayer_c01b(Layer):
         self.num_units = num_units
         self.untie_biases = untie_biases
 
-        output_shape = self.input_layer.get_output_shape()
-        num_input_channels = output_shape[0]
+        num_input_channels = self.input_shape[0]
 
         self.W = self.create_param(W, (num_units, num_input_channels))
         if b is None:
--- a/lasagne/layers/cuda_convnet.py
+++ b/lasagne/layers/cuda_convnet.py
@@ -37,10 +37,10 @@ class CCLayer(Layer):
 
 
 class Conv2DCCLayer(CCLayer):
-    def __init__(self, input_layer, num_filters, filter_size, strides=(1, 1), border_mode=None, untie_biases=False,
+    def __init__(self, incoming, num_filters, filter_size, strides=(1, 1), border_mode=None, untie_biases=False,
                  W=init.Uniform(), b=init.Constant(0.), nonlinearity=nonlinearities.rectify, pad=None,
                  dimshuffle=True, flip_filters=False, partial_sum=1, **kwargs):
-        super(Conv2DCCLayer, self).__init__(input_layer, **kwargs)
+        super(Conv2DCCLayer, self).__init__(incoming, **kwargs)
         if nonlinearity is None:
             self.nonlinearity = nonlinearities.identity
         else:
@@ -155,10 +155,10 @@ class Conv2DCCLayer(CCLayer):
 
 
 class MaxPool2DCCLayer(CCLayer):
-    def __init__(self, input_layer, ds, ignore_border=False, strides=None, dimshuffle=True, **kwargs):
+    def __init__(self, incoming, ds, ignore_border=False, strides=None, dimshuffle=True, **kwargs):
         from pylearn2.sandbox.cuda_convnet.pool import MaxPool
 
-        super(MaxPool2DCCLayer, self).__init__(input_layer, **kwargs)
+        super(MaxPool2DCCLayer, self).__init__(incoming, **kwargs)
         if ds[0] != ds[1]:
             raise RuntimeError("MaxPool2DCCLayer only supports square pooling regions, but ds=(%d, %d)" % ds)
 
@@ -252,10 +252,10 @@ class NINLayer_c01b(Layer):
     axis arrangement instead of bc01. This reduces the number of shuffles
     and reshapes required and might be faster as a result.
     """
-    def __init__(self, input_layer, num_units, untie_biases=False,
+    def __init__(self, incoming, num_units, untie_biases=False,
         W=init.Uniform(), b=init.Constant(0.), nonlinearity=nonlinearities.rectify,
         **kwargs):
-        super(NINLayer_c01b, self).__init__(input_layer, **kwargs)
+        super(NINLayer_c01b, self).__init__(incoming, **kwargs)
         if nonlinearity is None:
             self.nonlinearity = nonlinearities.identity
         else:
--- a/lasagne/layers/dense.py
+++ b/lasagne/layers/dense.py
@@ -62,8 +62,7 @@ class DenseLayer(Layer):
 
         self.num_units = num_units
 
-        output_shape = self.input_layer.get_output_shape()
-        num_inputs = int(np.prod(output_shape[1:]))
+        num_inputs = int(np.prod(self.input_shape[1:]))
 
         self.W = self.create_param(W, (num_inputs, num_units), name="W")
         self.b = self.create_param(b, (num_units,), name="b") if b is not None else None
@@ -109,8 +108,7 @@ class NINLayer(Layer):
         self.num_units = num_units
         self.untie_biases = untie_biases
 
-        output_shape = self.input_layer.get_output_shape()
-        num_input_channels = output_shape[1]
+        num_input_channels = self.input_shape[1]
 
         self.W = self.create_param(W, (num_input_channels, num_units), name="W")
         if b is None:
--- a/lasagne/layers/dense.py
+++ b/lasagne/layers/dense.py
@@ -52,9 +52,9 @@ class DenseLayer(Layer):
         >>> l_in = InputLayer((100, 20))
         >>> l1 = DenseLayer(l_in, num_units=50)
     """
-    def __init__(self, input_layer, num_units, W=init.Uniform(), b=init.Constant(0.),
+    def __init__(self, incoming, num_units, W=init.Uniform(), b=init.Constant(0.),
         nonlinearity=nonlinearities.rectify, **kwargs):
-        super(DenseLayer, self).__init__(input_layer, **kwargs)
+        super(DenseLayer, self).__init__(incoming, **kwargs)
         if nonlinearity is None:
             self.nonlinearity = nonlinearities.identity
         else:
@@ -96,10 +96,10 @@ class NINLayer(Layer):
     Any number of trailing dimensions is supported, so NINLayer can be used to implement
     1D, 2D, 3D, ... convolutions.
     """
-    def __init__(self, input_layer, num_units, untie_biases=False,
+    def __init__(self, incoming, num_units, untie_biases=False,
         W=init.Uniform(), b=init.Constant(0.), nonlinearity=nonlinearities.rectify,
         **kwargs):
-        super(NINLayer, self).__init__(input_layer, **kwargs)
+        super(NINLayer, self).__init__(incoming, **kwargs)
         if nonlinearity is None:
             self.nonlinearity = nonlinearities.identity
         else:
--- a/lasagne/layers/dnn.py
+++ b/lasagne/layers/dnn.py
@@ -98,7 +98,7 @@ class Conv2DDNNLayer(DNNLayer):
             self.b = self.create_param(b, (num_filters,), name="b")
 
     def get_W_shape(self):
-        num_input_channels = self.input_layer.get_output_shape()[1]
+        num_input_channels = self.input_shape[1]
         return (self.num_filters, num_input_channels, self.filter_size[0], self.filter_size[1])
 
     def get_params(self):
--- a/lasagne/layers/dnn.py
+++ b/lasagne/layers/dnn.py
@@ -25,8 +25,8 @@ class DNNLayer(Layer):
 
 
 class Pool2DDNNLayer(DNNLayer):
-    def __init__(self, input_layer, ds, strides=None, mode='max', **kwargs):
-        super(Pool2DDNNLayer, self).__init__(input_layer, **kwargs)
+    def __init__(self, incoming, ds, strides=None, mode='max', **kwargs):
+        super(Pool2DDNNLayer, self).__init__(incoming, **kwargs)
         self.ds = ds # a tuple
         self.mode = mode
         self.strides = strides if strides is not None else ds
@@ -42,14 +42,14 @@ class Pool2DDNNLayer(DNNLayer):
 
 
 class MaxPool2DDNNLayer(Pool2DDNNLayer): # for consistency
-    def __init__(self, input_layer, ds, strides=None, **kwargs):
-        super(MaxPool2DDNNLayer, self).__init__(input_layer, ds, strides, mode='max', **kwargs)
+    def __init__(self, incoming, ds, strides=None, **kwargs):
+        super(MaxPool2DDNNLayer, self).__init__(incoming, ds, strides, mode='max', **kwargs)
 
 class Conv2DDNNLayer(DNNLayer):
-    def __init__(self, input_layer, num_filters, filter_size, strides=(1, 1), border_mode=None, untie_biases=False,
+    def __init__(self, incoming, num_filters, filter_size, strides=(1, 1), border_mode=None, untie_biases=False,
                  W=init.Uniform(), b=init.Constant(0.), nonlinearity=nonlinearities.rectify, pad=None,
                  flip_filters=False, **kwargs):
-        super(Conv2DDNNLayer, self).__init__(input_layer, **kwargs)
+        super(Conv2DDNNLayer, self).__init__(incoming, **kwargs)
         if nonlinearity is None:
             self.nonlinearity = nonlinearities.identity
         else:
--- a/lasagne/layers/merge.py
+++ b/lasagne/layers/merge.py
@@ -15,8 +15,8 @@ __all__ = [
 
 
 class ConcatLayer(MultipleInputsLayer):
-    def __init__(self, input_layers, axis=1, **kwargs):
-        super(ConcatLayer, self).__init__(input_layers, **kwargs)
+    def __init__(self, incomings, axis=1, **kwargs):
+        super(ConcatLayer, self).__init__(incomings, **kwargs)
         self.axis = axis
 
     def get_output_shape_for(self, input_shapes):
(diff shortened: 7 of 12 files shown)


benanne:
We should continue this discussion - but maybe it needs to be turned into an issue instead so as to be more visible. It would be good to figure out how to handle this before we merge all the recurrence-related stuff.

We've pretty much decided that the 'free-floating layer' use case should be supported, in particular to support RNNs with 'custom' connectivity patterns, but also in general.

So the question now is whether we want to support this by either:
1. reusing `InputLayer` for this purpose
2. creating a class that represents a missing input that will be filled in, i.e. `ShapeHolder` or `InputShape` or `VirtualInputLayer` or whatever you want to call it
3. making `input_shape` a layer attribute that is set at creation time, and allowing a shape to be passed in for the `input_layer` argument.

3 is the most drastic change, but I'm also starting to lean towards preferring that if there are no big obstacles. Now is the time to make such a change at any rate (before we properly release anything). I would also be okay with 2. I do not like 1 because I believe it will lead to hard-to-understand bugs (read the previous discussion in this pull request for details).

What does everyone think of these 3 options? Did I miss any alternative options?

craffel:
I like 3 the most too.  What drastic changes are required?

benanne:
Currently `self.input_layer.get_output_shape()` is used throughout the code, all of these instances need to be replaced with `self.input_shape` and the `Layer` class needs to be modified so it sets this property. In addition, any other references to `self.input_layer` throughout the library need to be checked to make sure they still make sense, because it will now be possible for layers not to have an input layer (previously this was only possible for `InputLayer`).

This change will also have to be communicated clearly to all our current developers and users, so they make sure it doesn't cause problems with their own code and so all submitted code uses `self.input_shape` henceforth.

craffel:
Got it.  Yeah, I think it's reasonable because although it might touch a lot of code it's an understandable/logical change.

f0k:
Obviously (from the [previous](https://github.com/benanne/Lasagne/pull/52#issuecomment-68263096) [posts](https://github.com/benanne/Lasagne/pull/52#issuecomment-68265451)), I'd prefer option 3 as well. @dnouri, what's your preference?

Another funny thing is `get_output_shape()`. Unlike `get_output()`, it does not allow you to compute the output shape using any other input shape than the shape of the layer it was linked to on construction. If we store `self.input_shape` on construction, then we could just as well compute and store `self.output_shape` on construction. Or -- and I think I'd prefer this -- we should make `get_output_shape()` symmetric to `get_output()`.

> We've pretty much decided that the 'free-floating layer' use case should be supported, in particular to support RNNs with custom connectivity patterns, but also in general.

RNNs with custom connectivity patterns (or anything that wants to allow replacing a dot product with one or more layers) don't need free-floating layers. On free-floating layers, we can only call `get_output_for(input)`, so this would only support single layers. To support stacks of layers, we still need to link them to a properly shaped `InputLayer` instance and call `get_output(input)` on the topmost layer.

However, free-floating layers would broaden the applicability of Lasagne, allowing you to easily use them as part of a Theano expression, or build a model class around it that takes care of the connections. And they would make explicit that most layers are actually created for a particular input shape.

dnouri:
Option 3 certainly sounds reasonable!

benanne:
> Another funny thing is `get_output_shape()`. Unlike `get_output()`, it does not allow you to compute the output shape using any other input shape than the shape of the layer it was linked to on construction. If we store `self.input_shape` on construction, then we could just as well compute and store `self.output_shape` on construction. Or -- and I think I'd prefer this -- we should make `get_output_shape()` symmetric to `get_output()`.

I think it makes sense to add an optional `input_shape` argument to `get_output_shape()`. What are some use cases for being able to compute output shapes for input shapes that are different than the one set at construction time? Since we allow `None` in shapes, I guess a basic use case is to compute the output shape for a fully-specified input shape. But is that important? Are there any others?

If there are no important use cases then replacing `get_output_shape()` with an `output_shape` property, computed at instantiation, would actually be cleaner I guess (and more symmetrical since we already have `input_shape`).

> RNNs with custom connectivity patterns (or anything that wants to allow replacing a dot product with one or more layers) don't need free-floating layers.

Weird, I though that's where this discussion stemmed from. Don't you need free floating layers for the recurrent connections, because there is no `InputLayer` instance to attach them to? I thought that was the point. 

> To support stacks of layers, we still need to link them to a properly shaped `InputLayer` instance and call `get_output(input)` on the topmost layer.
So that would mean this doesn't solve the problem with recurrent connections at all, and we'd still need some kind of `VirtualInputLayer` or `ShapeHolder` or `InputShape` or whatever? That's a bummer!

> However, free-floating layers would broaden the applicability of Lasagne, allowing you to easily use them as part of a Theano expression, or build a model class around it that takes care of the connections. And they would make explicit that most layers are actually created for a particular input shape.
That's true, I think that's useful to have.

f0k:
> I think it makes sense to add an optional `input_shape` argument to `get_output_shape()`. What are some use cases for being able to compute output shapes for input shapes that are different than the one set at construction time?

What are some use cases for being able to compute outputs for inputs that are different than the one set at construction time? :) I think for symmetry we should probably offer both or none.

> Weird, I though that's where this discussion stemmed from. Don't you need free floating layers for the recurrent connections, because there is no `InputLayer` instance to attach them to?

If you want to replace a dot product with a single layer, then yes, you can use a free-floating layer without an `InputLayer` connected to it, and call its `get_output_for()` method where you would have done the dot product.
If you want to replace a dot product with a deep network, then no, you need to attach the deep network to an `InputLayer`, otherwise you would need to propagate the data yourself and cannot use the data propagation built into `get_output()`. (This is due to integrating the data propagation into the `Layer` class instead of doing it externally with a common function.)

> So that would mean this doesn't solve the problem with recurrent connections at all, and we'd still need some kind of `VirtualInputLayer` or `ShapeHolder` or `InputShape` or whatever?

No. We always need a real `InputLayer`.

benanne:
> What are some use cases for being able to compute outputs for inputs that are different than the one set at construction time? :) I think for symmetry we should probably offer both or none.

Fair point. Then I'm all for it.

> If you want to replace a dot product with a single layer, then yes, you can use a free-floating layer without an `InputLayer` connected to it, and call its `get_output_for()` method where you would have done the dot product.
If you want to replace a dot product with a deep network, then no, you need to attach the deep network to an `InputLayer`, otherwise you would need to propagate the data yourself and cannot use the data propagation built into `get_output()`. (This is due to integrating the data propagation into the `Layer` class instead of doing it externally with a common function.)

Right.

> No. We always need a real `InputLayer`.

I disagree. `InputLayer` creates an internal shared variable that is not necessary for this use case and only serves to create confusion and hard-to-understand bugs. But we've had this discussion before :)

f0k:
>> No. We always need a real `InputLayer`.

> I disagree. `InputLayer` creates an internal shared variable that is not necessary for this use case and only serves to create confusion and hard-to-understand bugs. But we've had this discussion before :)

Ahaaa... I see. You wanted something that allows `get_output(myinput)` to be called on the top layer of the stack, but not `get_output()`. (I though we were just looking for something that allows `get_output_for(...)` to be called, but not `get_output(...)`.) Maybe an `InputLayer` with `input_var` set to `False`? Did I suggest that before?

Does that change anything for the `self.input_shape` discussion? I think not, `self.input_shape` would still make things cleaner, no matter if we add variable-less input layers or not.

benanne:
Right, that's what I meant. I guess I should just have said that more literally. Sorry for the confusion! `input_var = False` sounds like a good compromise.

I don't think it changes anything about the `self.input_shape` discussion. I think we're now in agreement that this is a good idea, so we should probably get to work on it :)

f0k:
OK, I could do both of this in this PR, overwriting the current commit (mainly because all the discussion is in here). Not today, but maybe tomorrow.

benanne:
Sounds good!

f0k:
OK, I've overwritten this PR with a commit that introduces `self.input_shape` along with `self.input_layer` (that may be `None` now) and uses it as appropriate.
If this looks good to you, I'll think about what tests to add.
Finally, I can add the `input_var=False` case to `InputLayer`, and add appropriate tests as well.

benanne:
Looks okay to me. I was expecting this to be a bigger change than it seems to be :) Feel free to continue, or alternatively we can get some tests in and then merge this, and keep the `get_output_shape(input_shape)` and `input_var=False` stuff for a separate PR. Up to you though!

f0k:
Now that we're discussing #104, I'd keep the `get_output_shape` extension out of this PR, in case we end up removing `get_output()` and `get_output_shape()`. I'll add some tests, though.

f0k:
Rebased and added tests for the base classes. I've split up the layer base class tests a bit into parts using different fixtures. Travis passes, I think this can be merged, but any feedback is welcome.

benanne:
Looks good to me. I'll leave this open for a bit in case anyone else wants to chime in, otherwise I'll merge it tomorrow evening.

craffel:
Looks cool to me.  I'd suggest we rename the `input_layer`/`input_layers` parameter to `input`/`inputs` (or better yet `inp` or `inpt` because [`input` is a Python built-in](https://docs.python.org/2/library/functions.html#input)) but that's not a big deal.

benanne:
I like `input_layer`, it's nice and explicit. `input` or an abbreviation of that could also be taken to mean a Theano expression or variable that serves as the input.

craffel:
> I like input_layer, it's nice and explicit.

Right, but it mostly makes sense when its value is a `Layer` instance, not a `tuple` denoting shape.  I can't think of a better name though.

benanne:
aah, good point. Yeah. I just think `input` is a little too generic. Technically all of the function arguments are "inputs" in a sense. It's just a very overloaded word :)

craffel:
Is there any reason not to have separate instantiation keyword arguments `input_layer` and `input_shape` (probably in that order), both defaulting to `None`, with `ValueError` raised when they are both `None` (i.e. neither were supplied)?

benanne:
I think it would be convenient to just be able to pass either of them as the first positional argument. That's the convention we've followed so far and I think it's clear enough. Pragmatism trumps explicitness here imo :)

craffel:
>I think it would be convenient to just be able to pass either of them as the first positional argument. That's the convention we've followed so far and I think it's clear enough.

Well, `input_layer` would still be the first positional argument, and we haven't allowed for a shape to be passed as input before right?

> Pragmatism trumps explicitness here imo :)

You know what they say: "explicit is better than implicit" ;)

benanne:
yes, but dogmas are dangerous :p I meant that we've been following the unwritten rule that the first argument to a layer constructor is a representation of its input. Note the `InputLayer` takes the input shape as its first argument. I think it's natural, and in this case implicit definitely is cleaner than explicit imo. Instantiating a layer is an extremely common use case and this saves a lot of typing without buying us much in the way of extra clarity. The first thing you put in a layer is its input, I think that's clear enough :)

f0k:
Well, an explicit name for the parameter would be `input_layer_or_shape`, but that's a tad long... I sometimes use a keyword argument even for required positional arguments just so my code spells out the meaning of the argument, and then the length of the argument name matters. (But also the name... `input_layer` makes writing `layer = DenseLayer(input_layer=(10,20), num_units=20)` a bit awkward.)

> Note that the `InputLayer` takes the input shape as its first argument.

Speaking of which, this is the only `Layer` that doesn't have an `input_shape` member now, because it was called `shape`. Should we change that in one go?

benanne:
Fair enough, we can change that to `input_shape` for consistency. Although I guess if you take it to mean 'the shape of the layer's input', it doesn't really make sense because an `InputLayer` has no input in that sense. But that's a detail :)

benanne:
So, can we come up with a short name for the argument that encompasses both possibilities (`Layer` instance or shape tuple), isn't too long and is a bit more specific than just `input`? :)

If not, I guess we could go with just `input` - I'm a bit worried about the confusion it may cause but it's so commonly used that it might not be a big issue. It would shadow the Python builtin of the same name, but I don't think that's a problem (unless somebody plans to request user input in a `Layer` constructor!). I would definitely prefer `input` over an abbreviation of it.

f0k:
> So, can we come up with a short name for the argument that encompasses both possibilities (`Layer` instance or shape tuple), isn't too long and is a bit more specific than just `input`? :)

`input_definition`? That's not actually more specific, though. I can't think of a good name.

> It would shadow the Python builtin of the same name, but I don't think that's a problem (unless somebody plans to request user input in a `Layer` constructor!).

Even if, that somebody could just use `__builtin__.input` for that.

craffel:
> I don't think that's a problem (unless somebody plans to request user input in a Layer constructor!). I would definitely prefer input over an abbreviation of it.

Yeah, but it causes Pylint warning "redefined-builtin (W0622)".  This isn't a good warning to ignore in general because it alerts you to insane (but valid) things like `False = True`.  It's pedantic, but Pylint is as pedantic as it is awesome.  If this is something other people care about we should probably create a separate issue/PR to change the `input` argument name in `get_output` and `get_output_for`.

>> So, can we come up with a short name for the argument that encompasses both possibilities (Layer instance or shape tuple), isn't too long and is a bit more specific than just input? :)

> input_definition? That's not actually more specific, though. I can't think of a good name.

I'm not coming up with anything.  Maybe `input_spec`?

f0k:
Actually, `input_spec` and `input_definition` are more suited to shapes than layers. As this name discussing is delaying everything, shall we just have a vote? What's the best name for the first constructor argument of a `Layer`, which could be a layer instance or a shape tuple?

1. `input_layer`: Just leave it as it is. This conveys that it's usually a layer instance, and a shape is a special use.
2. `input_layer_or_shape`: This clearly says that it can be a layer or a shape.
3. `input`: This is unspecific and clashes with the builtin `input`, but it's short.
4. `input_spec`: This conveys that you're giving some indirect specification of what the input is, rather than the input.
5. `incoming`: This is about as unspecific as `input`, but doesn't clash with any builtin, and doesn't clash with input data or an input layer.

I'll cast my vote for 5, which just came to my mind now when thinking about DAGs and incoming edges.

senbon:
I'll just chime in and say that I really like the sound of `incoming` and `input_spec`. They both feel like an intuitive way of specifying a directed connection. Neither should pose a problem as long as the documentation is clear enough.

benanne:
I think I'm fine with anything except `input_layer_or_shape`, that's just too long and cumbersome (even though it's the most clear one).

craffel:
:+1: `incoming`

f0k:
`input_layer` is dead, long live `incoming`! (I went with the grammatically dubious `incomings` for the `MultipleInputsLayer` so I can still do `for incoming in incomings`.) Rebased to account for the changes in #117, should be ready to merge if Travis passes.

benanne:
Sweet! Looks good to me, I'll merge this tomorrow if there are no additional comments.

benanne:
Alright, time to merge!

craffel:
I think there's an issue here-
```Python
l = lasagne.layers.DenseLayer((1, 2), 3)
lasagne.layers.helper.get_all_layers(l)
```
yields a list where one of the entries is `None`. This causes problems when you call, e.g., `helper.get_all_params`. I think that as `get_all_layers` traverses the layers, it should check if the `input_layer` attribute is `None`, and if it is, it shouldn't add it to its layer list. Right?

benanne:
That sounds bad. I think checking for `None` in `get_all_layers` is probably a good fix (@f0k already hinted at it in https://github.com/benanne/Lasagne/pull/117#issuecomment-71863674). Then those functions work as expected, and things working as expected is always good, I reckon.

craffel:
I created a pull request to fix this issue.  #124

craffel:
Ironically, I don't think this can actually be used with recurrent layers.  If I change https://github.com/craffel/nntools/blob/master/examples/recurrent.py#L50
to
```Python
# Construct vanilla RNN: One recurrent layer (with input weights) and one
# dense output layer
l_in = lasagne.layers.InputLayer(shape=(N_BATCH, LENGTH, X_val.shape[-1]))

# As we iterate over time steps, the input will be batch size x feature dim
l_input_to_hidden = lasagne.layers.DenseLayer((N_BATCH, X_val.shape[-1]),
                                              N_HIDDEN, nonlinearity=None)

# As above, we need to tell the hidden-to-hidden layer what shape to expect
l_hidden_to_hidden_1 = lasagne.layers.DenseLayer((N_BATCH, N_HIDDEN), N_HIDDEN,
                                                 nonlinearity=None,
                                                 b=lasagne.init.Constant(1.))
l_hidden_to_hidden_2 = lasagne.layers.DenseLayer(l_hidden_to_hidden_1,
                                                 N_HIDDEN, nonlinearity=None,
                                                 b=lasagne.init.Constant(1.))

l_recurrent = lasagne.layers.RecurrentLayer(l_in,
                                            l_input_to_hidden,
                                            l_hidden_to_hidden_2,
                                            nonlinearity=None)
l_reshape = lasagne.layers.ReshapeLayer(l_recurrent,
                                        (N_BATCH*LENGTH, N_HIDDEN))

l_recurrent_out = lasagne.layers.DenseLayer(l_reshape,
                                            num_units=y_val.shape[-1],
                                            nonlinearity=None)
l_out = lasagne.layers.ReshapeLayer(l_recurrent_out,
                                    (N_BATCH, LENGTH, y_val.shape[-1]))
```
yields the error
```
RuntimeError: get_output() called on a free-floating layer; there isn't anything to get its input from. Did you mean get_output_for()?
```
which is unsurprising.  I guess this makes me wonder -- is there any reason we need an `InputLayer` at all?  Can't we just treat a layer which gets a shape as its input as an end of the layer graph?

benanne:
The input layer also has a shared variable associated with it by default, which can be convenient: you can just call `get_output()` without any arguments and then refer to `layer.input_var` when compiling the Theano functions. I usually do this.

craffel:
Ohh, hm.  I didn't realize that, I've always just made the input variable myself.  At any rate, it seems like either we need to 
1) Create input layers for free-floating layers (which we were trying to avoid by making the changes in this PR)
2) Allow layers to be constructed from input shape, but still allow `get_output` to be called _provided that an input variable was supplied_
3) Do away with the `InputLayer` and its built-in shared variable

Any thoughts?

benanne:
I like not having to create any variables myself :) So I'm not a fan of option 3. If you create it yourself you're duplicating a piece of information: the dimensionality of the input array. `InputLayer` can derive it from the shape you specify. If you specify an input shape as well as a variable to use, you're specifying the dimensionality twice. It may seem like a minor thing but I prefer to avoid that.

So 2) seems like the best option, if it's not too messy to implement that.

f0k:
> Ironically, I don't think this can actually be used with recurrent layers. 

Yes, I've [mentioned that](https://github.com/benanne/Lasagne/pull/52#issuecomment-69560504) a while ago (but it's a pretty long thread here).

> yields the error

Where does it yield the error? You're not calling `get_output()` in that part of the code. You can call it in a way that doesn't trigger the error, if needed.

> I've always just made the input variable myself

That's not a problem, you can make it yourself and then either create your input layer as `InputLayer(theshape, thevariable)`, or you can call `get_output({myinputlayer: thevariable})` or even just `get_output(thevariable)` if there's just a single `InputLayer` in the graph.

> 2) Allow layers to be constructed from input shape, but still allow `get_output` to be called *provided that an input variable was supplied*

The kind of free-floating layers we allow now is just suited for one purpose: You can create a single free-floating layer and call `get_output_for` on it. If you have a stack of layers, you still need an `InputLayer` at the bottom so you can call one of the `get_output()` variants on it. The only feature we wanted to add for this use case is to optionally construct an `InputLayer` without an `input_var` linked to it, so you can only call `get_output(input)` or `get_output({inputlayer: input})`, but not accidentally call `get_output()` on the stack of layers.

craffel:
> If you specify an input shape as well as a variable to use, you're specifying the dimensionality twice.

Given this, what's the use case for passing a variable to `get_output` then?

> So 2) seems like the best option, if it's not too messy to implement that.

I think it's as simple as changing (at https://github.com/benanne/Lasagne/blob/master/lasagne/layers/base.py#L124)
```Python
        elif self.input_layer is None:
```
to
```Python
        elif self.input_layer is None and input is None:
```

Right @f0k?  Am I missing something?

benanne:
```
Given this, what's the use case for passing a variable to `get_output` then?
```
As I said, it's a personal preference. We also support specifying a custom variable when instantiating an `InputLayer`, or just passing it when calling `get_output()`. All of these approaches have their own upsides and downsides and support many different use cases.

craffel:
Oops, missed your comment before posting mine.

> but not accidentally call get_output() on the stack of layers.

Right, but I think the issue is that even if you call `get_output(input)` on a stack of free-floating layers it still doesn't currently work.  That's what I'm doing, I think; here's the full traceback with context: https://gist.github.com/craffel/255c50901bfaa9a437f9

f0k:
> Right @f0k?

No. If you do that, it will get into the `else` case, which will try to call a method on `self.input_layer`.

> Right, but I think the issue is that even if you call `get_output(input)` on a stack of free-floating layers it still doesn't currently work.

That's intentional. `input` is only handled in an `InputLayer`: https://github.com/benanne/Lasagne/blob/master/lasagne/layers/input.py#L52
This is the only layer that is supposed to return the given `input`, all others are supposed to pass it on to their incoming layers.

We could change `Layer.get_output()` to behave like it had an input layer when `self.input_layer` is `None`, but I don't see any point for doing so. How is:
```python
layer = DenseLayer(inputshape, num_units=100)
layer = DenseLayer(layer, num_units=1)
```
any easier than:
```python
layer = DenseLayer(InputLayer(inputshape), num_units=100)
layer = DenseLayer(layer, num_units=1)
```
?

Of course it's one class name and two brackets less, but I don't see any problem with requiring stacks of layers to be wired to an `InputLayer`. I think that's cleaner / less confusing than allowing implicit input layers. What do others think? Should layers created from a shape support `get_output(input)` as if their `self.input_layer` had been set to `InputLayer(self.input_shape)`?

craffel:
> No. If you do that, it will get into the else case, which will try to call a method on self.input_layer.

Ah, right, sorry.

> We could change Layer.get_output() to behave like it had an input layer when self.input_layer is None, but I don't see any point for doing so. 

I thought that was the whole point behind this pull request - to avoid having to create `InputLayers` for free-floating layers (e.g. those used in a recurrent net), but maybe I'm mistaken/the scope changed. 

> ```
layer = DenseLayer(InputLayer(inputshape), num_units=100)
layer = DenseLayer(layer, num_units=1)
```

I mean, I think this is fine (and works/does the right thing), my main issue with it is that it's a little hard to conceptualize what it means when creating the recurrent network (as is currently done in the example):
https://github.com/craffel/nntools/blob/master/examples/recurrent.py#L61
But if that's the intended usage pattern, I'm not sure I know why we allow for layers to be instantiated with input shapes -- true that now we can instantiate a single layer and use `get_output_for` to get its output expression, but we can do exactly the same thing by doing what your code example above does.

**Comment by f0k** [comment]
> to avoid having to create `InputLayers` for free-floating layers

Yes, that *was* the point, but now it's really just about single free-floating layers and not about stacks of layers as we want to support for connections in RNNs.

> we can do exactly the same thing by doing what your code example above does.

Hmm, good point. It's not exactly clear why we should support free-floating layers, but not free-floating stacks of layers. More opinions?

**Comment by craffel** [comment]
> It's not exactly clear why we should support free-floating layers, but not free-floating stacks of layers.

Yeah, and since the code for a free-floating single layer is marginally different from the current input-shape form, I'm not sure it's worth it.  I do think it's conceptually weird to make an input layer in the context of the custom RNN, but the custom RNN is a pretty weird construction anyways.

**Comment by craffel** [comment]
One sort of ugly option, if we really wanted to keep input shape-constructable layers, would be for the `RecurrentLayer` (and any other similar layer) to create input layers for its layers whenever they are supplied without any.

**Comment by f0k** [comment]
> I do think it's conceptually weird to make an input layer in the context of the custom RNN

Well, we're operating outside the original scope of Lasagne, which was to represent DAGs. There must be a clash of concepts somewhere when we use it to implement RNNs :)

> One sort of ugly option, if we really wanted to keep input shape-constructable layers

The shape-constructible layers were meant to stand alone, to give users a possibility of just instantiating a layer and calling `get_output_for()` on it. I think this is worth supporting. We don't actually prevent users to stack layers on top of a free-floating layer, though, and such stacks are currently not useful.

> would be for the `RecurrentLayer` (and any other similar layer) to create input layers for its layers whenever they are supplied without any.

Okay, I lost sight of the original problem in between. So the problem is that when a user wants to create a custom RNN that uses, say, convolution + max-pooling for the forward step and "same" convolution for the recurrent step, the user will have to create `InputLayer` instances for the forward graph and the recurrent graph, although the associated input variables will never get used.

Maybe we should really treat a layer with an `input_layer` of `None` as a valid endpoint for a `get_output(input)` call. That would render `InputLayer`s unnecessary for any single-input network you don't want to attach specific input variables to. As always, the change would just affect the base classes, but it could change usage patterns quite a bit. Do we want that? (I don't see any immediate problem with this.)

An alternative for single-input networks without attached input variables would be allowing `input_var=False` for an `InputLayer`, but then we should probably be using this for free-floating layers as well instead of setting `input_layer = None`.

**Comment by benanne** [comment]
> Maybe we should really treat a layer with an `input_layer` of `None` as a valid endpoint for a `get_output(input)` call. That would render `InputLayer`s unnecessary for any single-input network you don't want to attach specific input variables to. As always, the change would just affect the base classes, but it could change usage patterns quite a bit. Do we want that? (I don't see any immediate problem with this.)

One possible issue is the case of multiple inputs. If there are no `InputLayer` instances to represent these different inputs, how do you designate which variable/expression should go where? Currently we support this by allowing a dictionary mapping to be passed in. But if there are no `InputLayer` instances, what should the keys in this dictionary be?

We can't use the `Layer` instance with `None` input as the key, because using any layer instance as a key already means "replace this layer's output with this expression". And I don't think changing this behaviour to "use this expression as the input for this layer" is wise, because then this will be very confusing in situations where `InputLayers` are used (and as I said before, I personally like to use them because they take care of creating the symbolic Theano variable for me).

**Comment by f0k** [comment]
> `{None: myinput}` looks horrible imo :p

Yes, but I liked your idea, and it wouldn't require any changes :)

> Yes it is, we can't have it mean both "use this expression instead of the output of this layer" and "use this expression as the input to this layer" depending on the context, that would be really really confusing.

Ah, didn't see you were pondering this. I agree, that's something we can't do!

> If we could handle the case of recurrence separately and support it there, I would be for it

Well then, what about the solution of calling `get_output({None: inputvar})` on a free-floating stack of layers? Would that be worse than any other solution crafted for this use case? An interesting side-effect is that this would *require* a free-floating layer or a free-floating stack of layers to be passed, making it explicit that you absolutely need to pass something that has an open end when you want custom connections for a RNN. It would also allow you to pass a graph that has an `input_layer = None` *somewhere*, but has other input layers with defined inputs. `None` would basically be a marker for an open end, for an implicit `InputLayer` instance that everybody knows. Does that sound intriguing or alarming? :)
If we opt for this, I would just change the error message for `get_output()` to not only suggest `get_output_for()`, but also `get_output({None: ...})`.

**Comment by craffel** [comment]
> Does that sound intriguing or alarming? :)

Both.  What's the advantage to this syntax over implicitly assuming that when a layer has `input_layer = None`, it should use the provided input as its input var (is that even an option)?

**Comment by f0k** [comment]
> What's the advantage to this syntax over implicitly assuming that when a layer has `input_layer = None`, it should use the provided input as its input var (is that even an option)?

That's an option as well, but then you could also pass non-free-floating graphs for your custom RNN. The advantage of using `get_output({None: ...})` in the RNN would be that it would *require* users to pass a free-floating graph, and that it would allow users to pass a partly free-floating graph (because you would be specifying that only `None` should use your custom input, every other input layer would continue to use its `input_var`). I don't think that's an important use case, but it would be possible.
An alternative to achieve the same level of ~~awesomeness~~ ~~complexity~~ flexibility would be to require users not to only pass a subgraph, but also pass the `InputLayer` instance that constitutes the "open end" (instead of assuming the open end is `None`).

**Comment by craffel** [comment]
>  The advantage of using get_output({None: ...}) in the RNN would be that it would require users to pass a free-floating graph

Why do we want to do that?

>  it would allow users to pass a partly free-floating graph (because you would be specifying that only None should use your custom input, every other input layer would continue to use its input_var).

I think this would be possible when assuming `input_layer = None` means use provided input, e.g.
```
l_1 = DenseLayer((10, 20), 30)
l_2 = DenseLayer(l_1, 40)
x = T.matrix()
l_2.get_output(x)
```
would pass `x` to `l_1`, which would see that `self.input_layer == None` and then use `x` as its input.  Right?

>An alternative to achieve the same level of awesomeness complexity flexibility would be to require users not to only pass a subgraph, but also pass the InputLayer instance that constitutes the "open end" (instead of assuming the open end is None).

Isn't that basically the same as it was originally being done, except that we aren't passing the `InputLayer`, we're just using it as an input layer?
https://github.com/craffel/nntools/blob/68d3b1288c869cb56fd7fdbbbda69d19ab52af26/examples/recurrent.py#L50

**Comment by benanne** [comment]
> I think this would be possible when assuming `input_layer = None` means use provided input, e.g.
```
l_1 = DenseLayer((10, 20), 30)
l_2 = DenseLayer(l_1, 40)
x = T.matrix()
l_2.get_output(x)
```
would pass `x` to `l_1`, which would see that `self.input_layer == None` and then use `x` as its input.  Right?

I guess that would also work, yeah.

I like @f0k's idea of using `None` to signify the recurrent input, and allowing other input layers to be part of this graph. (For example when training an RNN conditioned on some other input that does not vary over time - not a very common use case but fairly easy to support in this setting.)

**Comment by craffel** [comment]
> I like @f0k's idea of using None to signify the recurrent input, and allowing other input layers to be part of this graph. (For example when training an RNN conditioned on some other input that does not vary over time - not a very common use case but fairly easy to support in this setting.)

Sorry, I'm not really following what you mean by any of this - but I trust you two to make a good decision here.

**Comment by f0k** [comment]
> would pass x to l_1, which would see that self.input_layer == None and then use x as its input. Right?

Yes. But now assume a second input layer:
```python
l_1 = DenseLayer((10, 20), 30)
l_2 = DenseLayer(l_1, 40)
l_3 = InputLayer((10, 40))
l_4 = ElemwiseSumLayer(l_2, l_3)
x = T.matrix()
l_4.get_output(x)  # will use x both as an input for l_1 and output of l_3
l_4.get_output({None: x})  # will use x as an input for l_1 and l_3.input_var as the output of l_3
```

This is probably somewhere between being elegant and being too clever. One issue with this is that you need to know that creating a layer from a shape will result in `input_layer` being `None`. The code doesn't explicitly state that. You could also do:
```python
l_4.get_output({l_1.input_layer: x})
```
This is easier to understand, but would require `l_1` to be known to whoever is calling `get_output`. Plus it would be more natural to use this form if `l_1` was actually linked to an `InputLayer` instance.

> Isn't that basically the same as it was originally being done, except that we aren't passing the InputLayer, we're just using it as an input layer?

No, I was thinking of something like: `rnnlayer = CustomRNNLayer(recurrent=(l_4, l_1))`, where `l_4` is your subgraph for the recurrent connection and `l_1` is the "open end" input layer to this subgraph, assuming `l_1` is a proper `InputLayer` instance this time, but not necessarily the only one, as above. `CustomRNNLayer` would then know it would have to call `l_4.get_output({l_1: x})` on that. If we state that the "open end" always is the layer that has `input_layer = None`, users don't need to pass `l_1` any more, they just need to ensure one of the layers in the `l_4` subgraph has `input_layer = None`.
As I said above, the downside is that this can currently only be achieved by creating a layer from a shape, not by anything that makes clear the input layer is set to `None`.

> Sorry, I'm not really following what you mean by any of this

Are you with us now? You're the RNN expert, it would be nice to have this approved by you :)

craffel:
> Are you with us now? You're the RNN expert, it would be nice to have this approved by you :)

Sorry, no, not really; I may be the guy who is implementing RNNs but I still get lost pretty quickly when lasagne internals come up.  Just to be clear about my intentions, the main thing I'm interested in is not having to create InputLayer instances for the input-to-hidden and hidden-to-hidden connections, as is done here:
https://github.com/craffel/nntools/blob/68d3b1288c869cb56fd7fdbbbda69d19ab52af26/examples/recurrent.py#L50
>

f0k:
> I still get lost pretty quickly when lasagne internals come up

Hmm, that's a bad sign! We want the internals to be easily graspable. But hopefully this is just a matter of missing documentation or a developer tutorial.

> (although you'd only need one? Not sure I understand this part)

What me and Sander were after is that somebody might want to supply a graph for the `input-to-hidden` connection that has more than just the required "open end" for *input* and an output for *hidden*, but additional input(s) with side information. Of course most often there will only be a single input that defines what I called the "open end", i.e., something were `CustomRNNLayer` will provide the input for, and yes, "you'd only need one". But we may want to support the additional-inputs-with-side-information case, and if we do that, `CustomRNNLayer` has to use the `get_output(dict)` form instead of the `get_output(variable)` form because the latter would associate `variable` with *all* input layers of the graph, not just with the "open end".

`get_output(variable)` really is limited to single-input graphs, `get_output()` is limited to graphs with `InputLayers`, and `get_output(dict)` is unlimited (although, if you have a graph with multiple `input_layer = None` layers, it runs into the same problem as `get_output(variable)`).

> The example code doesn't do this

When we're working on the documentation, we should include a multiple-inputs example.

> Is it something like this?

Yes, exactly. This seems to be what @benanne uses frequently. My usage is:
```python
var1, var2 = theano.tensor.matrices('var1', 'var2')
l1 = lasagne.layers.InputLayer((10, 20), var1)
l2 = lasagne.layers.InputLayer((10, 20), var2)
l_out = lasagne.layers.ElemwiseSumLayer([l1, l2])
f = theano.function([var1, var2], l_out.get_output())
```
So in my case, the one compiling the graph just needs to know the output layers and the input variables. In Sander's case, it probably knows the output layers and the input layers (a useful side-effect being that it knows the input shapes as well). Either is fine.

> Finally, thanks for being patient with me on this!

Sure! We need to figure out how to properly document this anyway.

---
Coming back to the main discussion, I'm still a bit unsure on whether marking open ends with `None` is a clean solution, mainly because this only happens implicitly when creating a layer from a shape. The alternatives are:
a) Using an `InputLayer` to mark the open end, and only allowing graphs with a single input layer for the custom RNN connections and wherever else this may come up
b) Using an `InputLayer` to mark the open end and supply both the output layer and the "open end" instance for the custom RNN connections
c) Using `None` to mark the open end (by creating each layer that directly reads the "open end" from a shape)
d) Creating a separate class to mark the open end (something like an `InputLayer`, but without an `input_var`)

Note that a custom input-to-hidden connection could also be of the form:
```
 hidden
 /   \
A     B
 \   /
 input
```
I.e., there could be multiple layers sitting directly on top of the graph's "open end".

craffel:
OK, I think I'm starting to follow along.  So, it appears there are the following options for setting input/getting output:

1) Using a tuple for the layer's input (via this PR).  In this case, you can only do `get_output_for`, not `get_output`, so you can't do this for layer stacks.
```Python
l = DenseLayer((10, 20), 30)
x = T.matrix()
output = l.get_output_for(x)
```
2) Use an `InputLayer` and `get_output()`.  This will infer the use of the built-in `l_in.input_var`:
```Python
l_in = InputLayer((10, 20))
l = DenseLayer(l_in, 30)
output = l.get_output()
```
3) Use an `InputLayer` and `get_output`, but explicitly supply the input variable of the input layer to `get_output`.  This is equivalent to the 2); it's just done implicitly in there and explicitly here.
```Python
l_in = InputLayer((10, 20))
l = DenseLayer(l_in, 30)
output = l.get_output(l_in.input_var)
```
4) Use an `InputLayer` with your own variable:
```Python
x = T.matrix()
l_in = InputLayer((10, 20), x)
l = DenseLayer(l_in, 30)
output = l.get_output() # or get_output(x)
```
5) Use a dict to map layers to their input variables:
```Python
x = T.matrix()
l_in = InputLayer((10, 20), x)
l = DenseLayer(l_in, 30)
output = l.get_output({l_in: x})
```

I may be alone here but I think we need to pare down these usages; I think this is largely what was confusing me (the only one I really knew about was 2).

I don't think we need 1) (what this PR is for); the same thing could be done about as verbosely by
```Python
x = T.matrix()
l = DenseLayer(InputLayer((10, 20), x), 30)
output = l.get_output(x)
```
with the added benefit that in this case we could stack layers.  I also think it's a good idea to suggest that `get_output_for` is not used outside of the context of `get_output`, i.e., by an end-user.  I'm sorry to have suggested "input layers initialized using shape tuples" in the context of RNNs only to later suggest we don't have them!

I'm also not seeing a reason we should support both explicit and implicit creation of `InputLayer`'s `input_var`, i.e., we should choose one of 
```Python
x = T.matrix()
l_in = InputLayer((10, 20), x)
```
and
```Python
l_in = InputLayer((10, 20))
# Now, use l_in.input_var
```
I would personally suggest the latter (it seems like this is one of the main reasons `InputLayer` should exist at all).

I also think we should either allow the user to not pass anything to `get_output` (implying "use the built-in inputs of all layers") or force them to use a dict.  The reason is that the following will chug along happily, even though it's doing something weird:
```Python
l_in_1 = lasagne.layers.InputLayer((10, 20))
l_in_2 = lasagne.layers.InputLayer((10, 20))
l_sum = lasagne.layers.ElemwiseSumLayer([l_in_1, l_in_2])
l_sum.get_output(l_in_1.input_var)
```
I think that the convenience of passing a variable to `get_output` for single-input networks is nice, but I also think it creates a lot of confusion/potential issues (as in this example) down the line.  I'm now understanding (thanks for explaining!) why you all brought up the use case where the input-to-hidden connection contains multiple input layers and why that implies that `CustomRecurrentLayer` needs to know not only the output layer of its input-to-hidden connection but also the input to use in its step function.  Now that I understand what's going on, I'd vote for b):
> b) Using an InputLayer to mark the open end and supply both the output layer and the "open end" instance for the custom RNN connections

It makes things a little more verbose but it also makes it general purpose, which is what `CustomRecurrentLayer` is meant to be.

Finally, I think `InputLayer` should be called `Input`.  This makes its usage less counterintuitive in the context of RNNs, and the idea of it being a layer doesn't make a lot of sense to me.  But this is probably an unpopular opinion purely about semantics, so feel free to ignore.

Anyways, sorry for the long-winded post; these are largely just a reflection of my experiences getting up to speed and are intended as a way to prevent other people from being confused/shooting themselves in the foot.

benanne:
> OK, I think I'm starting to follow along.  So, it appears there are the following options for setting input/getting output: [...]

Yep, that's all of them. Not very Pythonic to have so many ways to do a thing, but they all serve their own use cases.


> I may be alone here but I think we need to pare down these usages; I think this is largely what was confusing me (the only one I really knew about was 2).

I think we just need to document them carefully, explain what the use cases are of all the different approaches, and recommend a standard way that covers 90% of the use cases (which I think should be number 2 in your list).


> I don't think we need 1) (what this PR is for); [...]

I'm on the fence about this one because personally I don't think using a layer in 'standalone' mode is that important a use case. But @f0k thinks it is, so maybe he can clarify the need for this. Earlier in this PR he briefly mentioned the motivation for this:

> However, free-floating layers would broaden the applicability of Lasagne, allowing you to easily use them as part of a Theano expression, or build a model class around it that takes care of the connections. And they would make explicit that most layers are actually created for a particular input shape.

... I guess I agree that this is useful, but definitely not a crucial feature. So if it does end up complicating things too much, we should reconsider supporting this.


> I also think it's a good idea to suggest that `get_output_for` is not used outside of the context of `get_output`, i.e., by an end-user.  I'm sorry to have suggested "input layers initialized using shape tuples" in the context of RNNs only to later suggest we don't have them!

We should discuss this. Initially I figured it was okay to add this because it doesn't increase code complexity, but it seems that it does somewhat complicate usage.


> I'm also not seeing a reason we should support both explicit and implicit creation of `InputLayer`'s `input_var`, [...]

These two methods of constructing an `InputLayer` serve two different use cases. I usually create my `InputLayer`s without explicit `input_var`s, so I count on the layer constructor to create one for me. I like this, it saves me some typing (the layer already knows what the dimensionality of the variable should be anyway, so it can take care of everything). When I need it later I can easily access it as `layer.input_var`.

However, what if you want to use a Theano expression as the input to your network? Let's say you have some sort of preprocessing implemented in Theano (outside of Lasagne). Then passing this preprocessing expression to the `InputLayer` at construction time allows you to call `get_output()` without any arguments later on, so that's pretty convenient.

Personally I don't see any advantage to creating a simple Theano variable and passing that to an `InputLayer`, as @f0k proposes, but I guess some people prefer to create the variable explicitly.


> I also think we should either allow the user to not pass anything to `get_output` (implying "use the built-in inputs of all layers") or force them to use a dict.

I disagree. In the overwhelming majority of use cases, there will be a single input layer. Forcing users to call `get_output()` with a dictionary would add additional cognitive overhead to all of these use cases.

It's kind of like how you can call `theano.function` with a Theano expression for its output argument, or a list of Theano expressions, when the function has multiple outputs. Always having to construct a list isn't exactly more verbose (it's only two characters), but it adds cognitive overhead to the overwhelming majority of use cases (most functions will have a single output).


> The reason is that the following will chug along happily, even though it's doing something weird: [...]

I think this is a contrived example. Once we have some decent documentation, I don't think anybody would try something like this and expect it to work.


> I think that the convenience of passing a variable to `get_output` for single-input networks is nice, but I also think it creates a lot of confusion/potential issues (as in this example) down the line.

As I said, I'm not sure I agree. I think this is not just a matter of convenience, it's an important feature that reduces overhead in the vast majority of use cases.


> Finally, I think `InputLayer` should be called `Input`.  This makes its usage less counterintuitive in the context of RNNs, and the idea of it being a layer doesn't make a lot of sense to me.  But this is probably an unpopular opinion purely about semantics, so feel free to ignore.

We could consider renaming it, as you said it's not really a layer, and in typical neural network terminology, the input is not regarded as a layer of the network either. So I wouldn't be opposed to this.

craffel:
Thanks for covering my suggestions thoroughly.

> I'm on the fence about this one because personally I don't think using a layer in 'standalone' mode is that important a use case.

I'm not arguing whether the standalone layer is a useful use-case or not here; my point is that because the following are the same:
```
l = DenseLayer(InputLayer((10, 20)), 30)
output = l.get_output()
```
and 
```
l = DenseLayer((10, 20), 30)
output = l.get_output()
```
we don't need this PR.  And, there's with the added benefit that in the first case we could stack layers, e.g.
```
l_1 = DenseLayer(InputLayer((10, 20)), 30)
l_2 = DenseLayer(l_1, 40)
output = l_2.get_output()
```
Plus, the first case is also more explicit at a marginal character count cost.  I do think that allowing for free-floating layers is a potentially useful use-case, following @f0k's argument.

 > However, what if you want to use a Theano expression as the input to your network? Let's say you have some sort of preprocessing implemented in Theano (outside of Lasagne). 

Ah, I wasn't thinking of this use-case.  That makes sense, this is certainly something we should support.

> Forcing users to call get_output() with a dictionary would add additional cognitive overhead to all of these use cases.
> ...
> I think this is a contrived example. Once we have some decent documentation, I don't think anybody would try something like this and expect it to work.

My issue is that by making things implicit, we are allowing for broken behavior without any kind of warning/error.  I agree that convenience is important, but I think we need a mechanism to prevent things like what I posted - it's contrived because it's the simplest case where it would come up.  When there's something this simple which causes an issue, it's easy to imagine a user doing something more complex and having this cause a bug which is counterintuitive/hard to track down (people often don't read the docs, after all).  I think enforcing this could be as simple as requiring the user to pass a dict to `get_output` when a network's graph contains multiple `InputLayer`s.

> We could consider renaming it, as you said it's not really a layer, and in typical neural network terminology, the input is not regarded as a layer of the network either. So I wouldn't be opposed to this.

We should do this!

PR #91, opened 2015-01-06
[WIP] add pooling layer based on dnn_pool, to support strided pooling

benanne:
This needs some tests, but I don't have access to a machine with cudnn installed at the moment. Hopefully I'll get around to it shortly.

### DIFF
--- a/lasagne/layers/dnn.py
+++ b/lasagne/layers/dnn.py
@@ -29,8 +29,8 @@ class Pool2DDNNLayer(DNNLayer):
         
     def get_output_shape_for(self, input_shape):
         output_shape = list(input_shape) # copy / convert to mutable list
-        output_shape[2] = int(np.floor(float(output_shape[2] - self.ds[0]) / self.strides[0])) + 1
-        output_shape[3] = int(np.floor(float(output_shape[3] - self.ds[1]) / self.strides[1])) + 1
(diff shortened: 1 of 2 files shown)


lasagne/layers/dnn.py
```diff
@@ -0,0 +1,42 @@
...
+    def __init__(self, input_layer, ds, strides=None, mode='max'):
+        super(Pool2DDNNLayer, self).__init__(input_layer)
+        self.ds = ds # a tuple
+        self.mode = mode
+        self.strides = strides if strides is not None else ds
+        
+    def get_output_shape_for(self, input_shape):
+        output_shape = list(input_shape) # copy / convert to mutable list
+        output_shape[2] = int(np.floor(float(output_shape[2] - self.ds[0]) / self.strides[0])) + 1
+        output_shape[3] = int(np.floor(float(output_shape[3] - self.ds[1]) / self.strides[1])) + 1
```
f0k: I think `output_shape[2] = (input_shape[2] - self.ds[0]) // self.strides[0] + 1` would be easier to read.

benanne: Agreed. I copy pasted this from the regular MaxPool2DLayer (which also handles the ignore_border=False case) and neglected to clean it up.

Issue #90, opened 2015-01-06
Allow `border_mode` to be a tuple of integers for `Conv2DLayer`


benanne:
So I had a go at implementing the `Conv2DDNNLayer`, and I realised that our `Conv2DLayer` already has a `convolution` keyword argument, which allows for `theano.tensor.nnet.conv2d` to be replaced with `theano.sandbox.cuda.dnn_conv`.

The only thing it doesn't support is passing a tuple of integers to the `border_mode` argument, which is kind of the main reason we'd want a separate layer for `dnn_conv` to begin with.

So instead, it might make more sense to modify `Conv2DLayer` so tuples can be passed in, and then make it complain about that only if the default Theano convolution is used. What does everyone think?

Related to that, the current `Conv2DCCLayer` has two arguments, `border_mode` and `pad`. The former can be used to pass in the strings 'valid', 'same' or 'full', and the latter can be used to pass an integer (cuda-convnet does not support different amounts of padding in the x and y directions so it was pointless to make this a tuple). These two arguments cannot be used together.

So it might be better to modify this code as well to merge `border_mode` and `pad` into a single argument (called `border_mode`), for consistency, which supports both string values and tuples of integers. This will break existing code that uses the `pad` argument, but at this stage I suppose that's not such a big deal (or we can leave it in an deprecate it, but I'm not sure if we want to start cluttering our codebase like that already).

Thoughts?

We could also consider implementing manual padding in `Conv2DLayer` for the case where a tuple is specified and the default Theano convolution is used.

f0k:
> So it might be better to modify this code as well to merge `border_mode` and `pad` into a single argument (called `border_mode`), for consistency, which supports both string values and tuples of integers.

Yes, that's exactly what should happen upstream (in Theano, I mean: https://github.com/Theano/Theano/issues/2118#issuecomment-58160895).

> I realised that our `Conv2DLayer` already has a `convolution` keyword argument

It will call that with `image_shape` and `filter_shape` keyword arguments, though. I think Theano's `conv2d()` is the only one that supports these.

dimatura:
By the way, Conv2DMM layer causes deprecation warnings from theano's `BaseGpuCorrMM` op about not using `pad`, and to use `border_mode` instead.

FWIW, I wrote a simple Conv2D layer using the `dnn` stuff from the Theano sandbox. I definitely got a performance boost relative to the cuda-convnet layer.

benanne:
Cool! Feel free to send a pull request :)

[...]

PR #93, opened 2015-01-07
Conv2DDNN layer using cudnn

dimatura:
Hello, here's a first stab at the `cudnn`-based layer, as suggested in #90 . It works for me, and the mnist example works fine as well. But I haven't tested different versions of `border_mode` and `pad`. I have just noticed there is already another branch with a `cudnn`-based pooling layer already there. I guess they should be merged before further work? Not sure the best way to proceed.


dimatura:
Actually I'm pretty sure there's something wrong in the `border_mode`/`pad` logic -- working on it now.

benanne:
This looks great! It's good that you've mimicked the interface of `Conv2DCCLayer` where possible. Maybe we should call the submodule `dnn` instead of `cudnn` though, just to be consistent with Theano's naming convention.

Unfortunately this currently seems to break all tests on Travis because importing `theano.sandbox.cuda.dnn` fails when no GPU is present. I am having the same issue in #91 (which wraps `dnn_pool`, so it complements this work nicely).

I'm not certain of this, but it should be fine to import this whenever a GPU device is available, even if the cudnn library isn't (we should probably test this assumption).

Theano itself only imports theano.sandbox.cuda when `theano.config.device.startswith("gpu")`: https://github.com/Theano/Theano/blob/master/theano/__init__.py#L94
I guess we should do something similar here.

benanne:
I added a check to ensure we're running on the GPU, and whether cudnn is available. This fixes it: https://github.com/benanne/Lasagne/commit/acabd27fd41dbab04c2e9353bc52a94e610a5570

Maybe I'll merge that into master and then you can rebase on top of it? Let me know if that's okay.

dimatura:
Sounds good!

benanne:
done :)

dimatura:
Things seem to be ok except for the fact the example fails in travis, when the "dnn not available" exception is thrown. I guess there should be a special case for that, as with the pylearn2 import.

benanne:
@dimatura on a somewhat unrelated note, are you using this code with cudnn R1 or R2? I'm having a hard time getting R2 to work with Theano: https://groups.google.com/forum/#!topic/theano-users/0dCxdDSKFjA

dimatura:
@f0k Ah yeah, I did rebase but you're right, I probably  didn't do it the right way. I'll try the cherry picking. 
@benanne I am using cudnn R2 (it's the only one I've installed), and it has been somewhat flaky lately. I've been keeping up to date with theano master and recent commits have given me trouble that seems related to a series of commits for doing shape inference for some convolution configurations, but it's a different error than you get. I have a K40 and 340.29, for what it's worth.

dimatura:
Ok, sorry for the delay. I think the cherry-picked branch looks sane, without the duplicate commits. Thanks @f0k for the git wisdom :)

benanne:
Thanks! Sorry for the trouble :)

It looks like this contains both dnn.py and cudnn.py though, I think cudnn.py can just be deleted.

dimatura:
Ah curses! Fixed.

benanne:
Alright, looks good to me. I'll wait for Travis and then merge. At some point we'll want to add some tests for this, but it makes more sense to write tests for the regular `Conv2DLayer` first since a lot of them will be reusable for all the variants.

dimatura:
I agree, I saw that there's no conv2d tests. I've had various problems with certain combinations of strides, pad sizes and so on - probably more Theano-related than Lasagne-related. Sometimes due to upstream problems, sometimes myself.

benanne:
Looks like it's still trying to run the example test, which causes the Travis failure. We'll need to modify test_examples.py to skip it if cudnn is not available. Feel free to do that if you want, otherwise I'll probably have a look at it tomorrow.

dimatura:
Added an exception for the cudnn error.

benanne:
Sweet! Merging in 3... 2... 1...

PR #94, opened 2015-01-08
Add ReshapeLayer

f0k:
This adds `layers.ReshapeLayer`.

In addition to supporting fixed shapes and shapes with `-1` for one dimension to be inferred automatically, it also supports `None` for dimensions to be copied from the input shape. This should be handy especially for the batch size, so you can keep the batch size constant even if you don't know its value and still have the `-1` available for automatically inferring any of the other dimensions. (Only works if the batch size is the first dimension, though.)

If you're fine with the proposed interface and functionality, I'll add tests. If you have any ingenious idea on how to improve the "copy dimensions" feature, let me know.

### DIFF
--- a/lasagne/layers/shape.py
+++ b/lasagne/layers/shape.py
@@ -10,6 +10,7 @@ from .base import Layer
 __all__ = [
     "FlattenLayer",
     "flatten",
+    "ReshapeLayer",
     "PadLayer",
     "pad",
 ]
@@ -24,6 +25,78 @@ class FlattenLayer(Layer):
         return input.flatten(2)
 
 flatten = FlattenLayer # shortcut
+
+
+class ReshapeLayer(Layer):
+    """
+    A layer reshaping its input tensor to another tensor of the same total
+    number of elements.
+
+    :parameters:
+        - incoming : a :class:`Layer` instance or a tuple
+            the layer feeding into this layer, or the expected input shape
+
+        - shape : tuple
+            The target shape. Any of its elements can be `None`, denoting to
+            retain the size of the input shape for this dimension. At most one
+            element can be `-1`, denoting to infer the size for this dimension
+            to match the total number of elements of the input shape. Any
+            remaining elements must be positive integers.
+
+    :usage:
+        >>> from lasagne.layers import InputLayer, ReshapeLayer
+        >>> l_in = InputLayer((100, 20))
+        >>> l1 = ReshapeLayer(l_in, (None, 2, 10))
+        >>> l1.get_output_shape()
+        (100, 2, 10)
+        >>> l2 = ReshapeLayer(l_in, (None, 1, 2, 5, -1))
+        >>> l2.get_output_shape()
+        (100, 1, 2, 5, 2)
+
+    :note:
+        The tensor elements will be fetched and placed in C-like order. That
+        is, reshaping `[1,2,3,4,5,6]` to shape `(2,3)` will result in a matrix
+        `[[1,2,3],[4,5,6]]`, not in `[[1,3,5],[2,4,6]]` (Fortran-like order),
+        regardless of the memory layout of the input tensor. For C-contiguous
+        input, reshaping is cheap, for others it may require copying the data.
+    """
+
+    def __init__(self, incoming, shape):
+        super(ReshapeLayer, self).__init__(incoming)
+        shape = tuple(shape)
+        if not all(s is None or isinstance(s, int) for s in shape):
+            raise ValueError("`shape` must be a tuple of int and/or None")
+        if any(s is not None and (s == 0 or s < -1) for s in shape):
+            raise ValueError("`shape` integers must be positive or -1")
+        if sum(s == -1 for s in shape) > 1:
+            raise ValueError("`shape` cannot contain multiple -1")
+        self.shape = shape
+
+    def get_output_shape_for(self, input_shape, *args, **kwargs):
+        # First, replace all `None` with the corresponding input dimension
+        output_shape = list(self.shape)
+        for dim, o in enumerate(output_shape):
+            if o is None:
+                output_shape[dim] = input_shape[dim]
+        # Secondly, infer value for -1 if needed
+        if -1 in output_shape:
+            dim = output_shape.index(-1)
+            output_shape[dim] = np.prod(input_shape) // -np.prod(output_shape)
+        # Sanity check
+        if np.prod(input_shape) != np.prod(output_shape):
+            raise ValueError("%s cannot be reshaped to specification %s. "
+                             "The total size mismatches." %
+                             (input_shape, self.shape))
+        return tuple(output_shape)
+
+    def get_output_for(self, input, *args, **kwargs):
+        # Replace all `None` with the corresponding input dimension
+        output_shape = list(self.shape)
+        for dim, o in enumerate(output_shape):
+            if o is None:
+                output_shape[dim] = input.shape[dim]
+        # Everything else is handled by Theano
+        return input.reshape(tuple(output_shape))
 
 
 class PadLayer(Layer):
--- a/lasagne/layers/shape.py
+++ b/lasagne/layers/shape.py
@@ -11,6 +11,7 @@ __all__ = [
     "FlattenLayer",
     "flatten",
     "ReshapeLayer",
+    "reshape",
     "PadLayer",
     "pad",
 ]
@@ -133,6 +134,8 @@ class ReshapeLayer(Layer):
                 output_shape[dim] = input.shape[o[0]]
         # Everything else is handled by Theano
         return input.reshape(tuple(output_shape))
+
+reshape = ReshapeLayer # shortcut
 
 
 class PadLayer(Layer):
--- a/lasagne/layers/shape.py
+++ b/lasagne/layers/shape.py
@@ -37,21 +37,23 @@ class ReshapeLayer(Layer):
             the layer feeding into this layer, or the expected input shape
 
         - shape : tuple
-            The target shape. Any of its elements can be `None`, denoting to
-            retain the size of the input shape for this dimension. At most one
-            element can be `-1`, denoting to infer the size for this dimension
-            to match the total number of elements of the input shape. Any
-            remaining elements must be positive integers.
+            The target shape specification. Any of its elements can be `[i]`,
+            a single-element list of int, denoting to use the size of the ith
+            input dimension. At most one element can be `-1`, denoting to
+            infer the size for this dimension to match the total number of
+            elements of the input tensor. Any remaining elements must be
+            positive integers directly giving the size of the corresponding
+            dimension.
 
     :usage:
         >>> from lasagne.layers import InputLayer, ReshapeLayer
-        >>> l_in = InputLayer((100, 20))
-        >>> l1 = ReshapeLayer(l_in, (None, 2, 10))
+        >>> l_in = InputLayer((None, 100, 20))
+        >>> l1 = ReshapeLayer(l_in, ([0], [1], 2, 10))
         >>> l1.get_output_shape()
-        (100, 2, 10)
-        >>> l2 = ReshapeLayer(l_in, (None, 1, 2, 5, -1))
+        (None, 100, 2, 10)
+        >>> l2 = ReshapeLayer(l_in, ([0], 1, 2, 5, -1))
         >>> l2.get_output_shape()
-        (100, 1, 2, 5, 2)
+        (None, 1, 2, 5, 200)
 
     :note:
         The tensor elements will be fetched and placed in C-like order. That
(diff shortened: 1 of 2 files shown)


f0k:
> If you have any ingenious idea on how to improve the "copy dimensions" feature, let me know.

Thinking about it, one *could* use a syntax like `shape=([0], 10, 5, -1)` to denote: First output dimension size should be the same as the first input dimension size, second dimension should be 10, third dimension should be 5, and fourth dimension should be whatever is left. That would be more expressive than using `None`. Would that be nice to have or too complex? At any rate, I think auto-inferring the batch size is quite useful.

craffel:
FWIW I also made a `ReshapeLayer` while implementing recurrent stuff.
https://github.com/craffel/nntools/blob/master/lasagne/layers/recurrent.py#L167
I think yours is better though.

f0k:
> I think yours is better though.

Yep, yours might be a little limited on functionality and error checks :) But whatever we have should be backwards-compatible to yours, as that's the minimum people will expect.

lasagne/layers/shape.py
```diff
@@ -26,6 +27,78 @@ def get_output_for(self, input, *args, **kwargs):
...
+                output_shape[dim] = input_shape[dim]
+        # Secondly, infer value for -1 if needed
+        if -1 in output_shape:
+            dim = output_shape.index(-1)
+            output_shape[dim] = np.prod(input_shape) // -np.prod(output_shape)
+        # Sanity check
+        if np.prod(input_shape) != np.prod(output_shape):
+            raise ValueError("%s cannot be reshaped to specification %s. "
+                             "The total size mismatches." %
+                             (input_shape, self.shape))
```
f0k: Reading again, we will have to skip this check if any `input_shape` dimension is `None`.

lasagne/layers/shape.py
```diff
@@ -26,6 +27,78 @@ def get_output_for(self, input, *args, **kwargs):
...
+    def get_output_shape_for(self, input_shape, *args, **kwargs):
+        # First, replace all `None` with the corresponding input dimension
+        output_shape = list(self.shape)
+        for dim, o in enumerate(output_shape):
+            if o is None:
+                output_shape[dim] = input_shape[dim]
+        # Secondly, infer value for -1 if needed
+        if -1 in output_shape:
+            dim = output_shape.index(-1)
+            output_shape[dim] = np.prod(input_shape) // -np.prod(output_shape)
```
f0k: If any `input_shape` dimension is `None`, this will fail. If all such dimensions are copied over to `output_shape`, we could still correctly infer the value for `-1`, though.
Example: for `input_shape = (None, 100)`, `self.shape = (None, -1, 20)`, we can infer `output_shape = (None, 5, 20)`.

benanne:
Looks great! The ability to use this with shape tuples containing `None` is valuable, I think.

f0k:
> The ability to use this with shape tuples containing `None` is valuable, I think.

Yep, but as I said, this won't work for copying over the batch size if it's the last dimension. Say the input shape is `(32, 5, 20, batchsize)`, and we want the output shape to be `(-1, 20, batchsize)` (concatenating all channels, for whatever reason, without knowing the number of channels, for whatever reason) -- this cannot be done with the current proposal of using `None` to copy over a dimension. It would work if we used `[x]` to copy over the `x`th dimension: `(-1, 20, [3])` or even `(-1, 20, [-1])`. Would that be better? We should decide before merging, we won't want to support both `None` and `[x]`.

benanne:
hmm, good point! Is there a reason we wouldn't want to support both? Although I guess avoiding a proliferation of options and interfaces is a good enough reason by itself. And it would make the code more complex.

I like the [] trick. It's a bit of an abuse of Python's list literal syntax I suppose, but it's intuitive because [] is also used for indexing (it could be seen as indexing the input shape). I can't really think of anything cleaner either.

craffel:
Are you using `[]` to denote "use this dimension's shape"?  I.e. it should always be a single-element list, and you're just using the list syntax to differentiate it from an `int` which normally denotes size?  If so, I don't like it, because that's not what `[]` means.  What's wrong with `shape=(layer.get_output_shape()[0], 10, 5, -1)` instead of `shape=([0], 10, 5, -1)`?  It's a little more verbose but also more readable (assuming I'm actually understanding what's going on).  If I'm misunderstanding or in the minority on this, feel free to ignore.

f0k:
> I.e. it should always be a single-element list, and you're just using the list syntax to differentiate it from an int which normally denotes size?

Yep. It's an abuse of syntax, sure, but it's quite readable.

> What's wrong with `shape=(layer.get_output_shape()[0], 10, 5, -1)` instead of `shape=([0], 10, 5, -1)`?

This won't work if `layer.get_output_shape()[0]` is `None`, i.e., it's a network compiled for variable batch size. The second form would work in `get_output_shape_for`, because we can infer that all unknown input shape dimensions are copied over to the output shape, so we can perform the `-1` inference just using the remaining dimensions and be sure it's correct; and it would also work in `get_output_for`, because we can use `input.shape[0]` for the first dimension.

I'm not sure exactly how useful this would be either... after all, even the `-1` inference is just a convenience feature, you can usually find ways around that. But it seems it could save some trouble sometime.

craffel:
>This won't work if layer.get_output_shape()[0] is None, i.e., it's a network compiled for variable batch size.

I'm having trouble seeing how it would be different than the `[0]` form, but I would probably understand better once you code it up.  

> I'm not sure exactly how useful this would be either...
> it would make the documentation more complicated and I don't think it would be any more intuitive

Given these two comments and my own hangups about the syntax, I'd hold off on it until it seems like there's an explicit need.

f0k:
>> This won't work if layer.get_output_shape()[0] is None, i.e., it's a network compiled for variable batch size.

> I'm having trouble seeing how it would be different than the `[0]` form, but I would probably understand better once you code it up.

Without `[x]`:
```python
def get_output(self, input=None, *args, **kwargs):
    return input.reshape(self.shape)
```

With `[x]`:
```python
def get_output(self, input=None, *args, **kwargs):
    shape = [input.shape[s[0]] if isinstance(s, list) else s for s in self.shape]
    return input.reshape(tuple(shape))
```

Now if you have a layer `foo` that has a variable batch size (such as `foo = InputLayer((None, 20, 6))`), you can stack a `ReshapeLayer(foo, ([0], 120))` on top of it, but not a `ReshapeLayer(foo, (foo.get_output_shape()[0], 120))` (`get_output()` would break in the second case).
And a similar point can be made for `-1` inference.

> Given these two comments and my own hangups about the syntax, I'd hold off on it until it seems like there's an explicit need.

So we shouldn't support `-1` inference for partially undefined input shapes for now?

benanne:
I think we definitely need `-1` not in the least because people will just expect it to work (see #102, not entirely the same thing I guess, but still).

As for supporting `None` or `[idx]` in the shapes, we probably need at least one of those. Supporting `None` is the most straightforward option, since we already support that for `InputLayers`. But obviously `[idx]` is a little more flexible.

The main use case for the latter seems to be layers with `c01b` arrangement (i.e. the cuda-convnet based layers with `dimshuffle=False`. I'm not sure how essential that use case is though, since `dimshuffle=True` tends to be just as fast for my experiments (has anybody had a different experience?). Are there any other use cases we can think of?

f0k:
> Are there any other use cases we can think of?

@craffel and me discussed this a little via email. First I noted that there are use cases for multiple variable input dimensions (i.e., more than just the batch size being `None` in `input_shape`):

> When doing predictions on overlapping spectrogram excerpts of an audio recording, it pays off applying the convolutional layers to the full audio recording at once and chopping it up into excerpts afterwards (before applying the dense layers), to avoid redundant computations. If audio recordings are not all the same length, then this means the image size will be variable in one dimension.
Via the same argument, applying networks to photographs might require a variable image size in two dimensions. 

Then I brought up the following:

>If we want to support the case of multiple variable input dimensions, we need a way to tell which output shape dimension is which. [x] provides such a way. Is there a better one? Relying on the order of Nones, maybe?

>Assume input_shape = (128, 3, None, None, 10).
>Merging the last two: shape = ([0], [1], [2], -1)
>Merging the unknowns: shape = ([0], [1], -1, [4])
>Splitting the last one: shape = ([0], [1], [2], [3], 2, -1)

>If we strip away "convenience usage":
>Merging the last two: shape = (128, 3, [2], -1)
>Merging the unknowns: shape = (128, 3, -1, 10)
>Splitting the last one: shape = (128, 3, [2], [3], 2, -1)

>Can you do all this with a None-based syntax? Would it be easier? (Not sure if these use cases cover everything, but it's a start.)

benanne:
I think I prefer the 'indexing' syntax. It's a bit 'clever', which isn't always a good thing in software development, but in this case I think it's actually clearer than the alternative, and probably more flexible.  I fully understand @craffel's gripes though. Maybe we should get some more input for this decision?

benanne:
Looks like there isn't any more input coming... how should we proceed?

craffel:
> Looks like there isn't any more input coming... how should we proceed?

I can't endorse the `[0]` syntax but I can't think of another way, so I'm just going to shut up about this.  Let's get this merged!

f0k:
Rebased, changed from `None` syntax to `[i]` syntax, properly implemented shape inference for all possible cases, and added tests covering the full class.

benanne:
Sweet! very nice work. It's a bit unfortunate that we're not able to reach a consensus about this shape specification syntax, but I believe this is the best we can do to get this level of flexibility (and we do need this kind of flexibility, I think). I'll leave this open for comments and merge it this weekend (I'll be traveling so it may take a bit longer than usual).

f0k:
> It's a bit unfortunate that we're not able to reach a consensus about this shape specification syntax, but I believe this is the best we can do to get this level of flexibility (and we do need this kind of flexibility, I think).

At least there is no way to reach the same flexibility with a `None` syntax:

* Possibility 1: `None` syntax as in my first implementation, i.e., copy over the corresponding input dimension: Cannot reshape `(100, None)` to `(2, 50, None)`.
* Possibility 2: `None` syntax relying on the order of `None`s in the input shape: Breaks as soon as that input shape becomes known. `ReshapeLayer.get_output_for()` should be independent of whether particular construction-time input dimensions are known or not.
* Possibility 3: `None` syntax relying on the order of `None`s in the output shape: Maybe one could design an algorithm that infers how to interpret reshaping `(10, 30, 50)` to `(2, 5, None, -1)`, but that would be complicated to implement, difficult to understand, and all the ones I could think of would break if the input shape changes to `(None, 30, None)`.

So I think there's no way around explicitly indexing the input shape in some way, and for that `[i]` is much shorter than creating a special object for that, and easier on the eye than `(i,)`. The only similarly brief syntax I can think of would be imaginary numbers, but that's even more difficult to check for correctness :)

craffel:
The only thing I'm not seeing is a check that indexed dimensions are within the range of the input shape.  I.e., if `shape` is provided as `(10, 100, [5])` and `len(input_shape) == 3` in `get_output_shape_for` then you'll get index out of bound errors, for example, here: https://github.com/f0k/Lasagne/blob/e16b4677acc744fbe35bd50a45ef0721f3454bd6/lasagne/layers/shape.py#L94
It'd be nice to give helpful error messages in this case.

f0k:
> The only thing I'm not seeing is a check that indexed dimensions are within the range of the input shape.

Good idea! Added a check, error message and test.

benanne:
One more thing I thought of: it would be good to add a shortcut for this layer, because this operation might not typically be seen as constituting a layer in a neural net. We already have `dropout = DropoutLayer`, `concat = ConcatLayer` and `pad = PadLayer`, so I think it'd be useful to have `reshape = ReshapeLayer` as well. What does everyone think?

craffel:
> What does everyone think?

Sounds good to me.

f0k:
All right, added a shortcut. I was feeling that we're a bit too generous about introducing shortcuts, but the argument that "this operation might not typically be seen as constituting a layer" is reasonable.

benanne:
Alright, merging!

benanne:
I guess I shouldn't have merged this and the Py3k one simultaneously... apparently `reduce` is no longer a built-in in Python 3! The docs say:

> Removed reduce(). Use functools.reduce() if you really need it; however, 99 percent of the time an explicit for loop is more readable.

I'll leave it up to @f0k to decide what the most readable solution is :)
```from functools import reduce``` should fix this at any rate (it also exists as an alias in Python 2.7).

**Comment by f0k** [comment]
Booh, I found that part pretty readable:
```python
noneprod = lambda x, y: None if x is None or y is None else x * y
input_size = reduce(noneprod, masked_input_shape)
output_size = reduce(noneprod, masked_output_shape)
```
The explicit for loop would be:
```python
def noneprod(shape):
    p = 1
    for s in shape:
        if s is None:
            return None
        p *= s
    return p
input_size = noneprod(masked_input_shape)
output_size = noneprod(masked_output_shape)
```
That's cumbersome. Another way would be:
```python
input_size = (None if any(x is None for x in masked_input_shape)
               else np.prod(masked_input_shape))
output_size = (None if any(x is None for x in masked_output_shape)
               else np.prod(masked_output_shape))
```
That's just possible because we're dependent on numpy anyway. I'll use that one.

PR #105, opened 2015-01-21
Split up layer tests into several modules

dnouri:
It's awkward, but here it is.


benanne:
Good call. Why is it awkward?

f0k:
+1 (can you upvote things on github or was that just gitlab? Anyway, thanks, that makes it a lot more convenient to handle.)

benanne:
I'll go ahead and merge this but I'm still curious about why you think it's awkward :p

dnouri:
Never mind, I guess it's not awkward (petty?) if you guys think it's useful.  :-)

PR #113, opened 2015-01-24
Negative-log-likelihood objective function.

Britefury:
Implemented negative-log-likelihood objective function in objectives module.
Modified MNIST example to use it.
Also (accidentally) included a modification to the MNIST example that prevents it from re-downloading the dataset each time it is run.

### DIFF
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -4,6 +4,7 @@ import cPickle as pickle
 import gzip
 import itertools
 import urllib
+import os
 
 import numpy as np
 import lasagne
@@ -22,7 +23,9 @@ MOMENTUM = 0.9
 
 
 def _load_data(url=DATA_URL, filename=DATA_FILENAME):
-    urllib.urlretrieve(url, filename)
+    if not os.path.exists(filename):
+        urllib.urlretrieve(url, filename)
+
     with gzip.open(filename, 'rb') as f:
         data = pickle.load(f)
     return data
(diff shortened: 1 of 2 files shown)


benanne:
Cool, thanks for contributing! I think it's useful to have a function for this, but I'm not sure if it should be called `negative_log_likelihood`. This is the negative log likelihood of a model with a softmax output, but for a regression problem for example, the MSE would be the negative log likelihood. So this may lead to misunderstandings.

I don't really know what would be a better name for this - `multinomial_nll` or something perhaps? Maybe someone else has a better idea.

I believe `urllib.urlretrieve` is already supposed to prevent redownloading if the target file already exists. Was this not working for you? I don't think it should be necessary to check manually whether the path exists.

Britefury:
Its now called `multinomial_nll`.
The file check is now gone.

benanne:
Thanks! Looks good to me, will leave it open for a bit longer in case anyone else has comments.

dimatura:
How about using `T.nnet.categorical_crossentropy` for this functionality? It is functionally equivalent, but I have found that in some cases it makes it easier for `theano` to figure out the appropriate transfromations for numerical stability.

f0k:
Cool, didn't know `T.nnet.categorial_crossentropy` covers both the case of `true_dist` being binary one-hot vectors and `true_dist` being integers. Maybe we should use this instead, but maybe it's clearer to keep the two variants separately and also to spell out the implementation here. I tend towards the latter, unless @dimatura is right and there are some optimizations that only work for `T.nnet.categorial_crossentropy`?
The docstring is good, but not in the format we use elsewhere. For consistency, please see [layers/base.py](https://github.com/benanne/Lasagne/blob/master/lasagne/layers/base.py) for some examples and adapt to this format! (And feel free to fix the other docstrings in this file while you're at it, that would be very helpful.)

dimatura:
Actually I didn't know about the two label cases either ;) Regarding the optimization, it's more about numerical stability. I just have my own anecdotal experience: at some point I was using my own log-loss function that was slightly more complex and while it worked at first, after a few iterations it ran into `NaN`s and such. I figured out it was something similar to what is described [here](http://deeplearning.net/tutorial/code/rbm.py), in the comments for `get_reconstruction_error`; basically `theano` didn't see `log( softmax(x) )` but `log( f(x) )` where `f` was something that was functionally equivalent to the `softmax`, but `theano` couldn't figure this out. I tried the same fix from `rbm.py`, but it still failed with the `NaN` problems. Then I tried `nnet.categorical_crossentropy` (feeding it the output of `nnet.softmax`) and it worked great.

benanne:
Excellent, thank you! Merging.

PR #114, opened 2015-01-25
Add a few conv2d layer tests

dnouri:
test_conv.py tests three conv 2d layer implementations using mostly
default parameters.  

The tests run convolutions with small kernels on small inputs.  (Which
is why it doesn't work with cuda-convnet.)  The convolution math was
done by hand, to allow checking actual outputs against expected
outputs.

The tests so far did turn out a bit hairy, maybe hard to follow.
Also, there's no way a few generic tests like this one could cover all
the special cases between the different implementations.  But it does
maybe cover a lot of common ground between them.  Interested to hear
what people think.

A part of this pull request changes the way that modules with missing
dependencies are dealt with, particurly lasagne.layers.dnn and
lasagne.layers.cuda_convnet.

Previously, these two modules failed at runtime, when we were running
their code.  This is changed herein; these modules will now fail with
ImportError when using from-imports, like these:

  - from lasagne.layers import Conv2DDNNLayer
  - from lasagne.layers.dnn import Conv2DDNNLayer

Unforunately, accessing the module contents using attribute-style
module traversal, like in 'Layer = lasagne.layers.Conv2DDNNLayer',
will result in AttributeError, without an error message.  (This might
be controversial.)

The upshot is that with this we now have a consistent way to know
whether or not a particular module has unsatisfied dependencies,
without having to run the actual code.  This is something I really
want to be able to do to conditionally to run tests against different
implementations based on availability.  Not sure there's a better way
to deal with this.

### DIFF
--- a/examples/mnist_conv_dnn.py
+++ b/examples/mnist_conv_dnn.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 
 import lasagne
+from lasagne.layers import dnn  # fails early if not available
 import theano
 import theano.tensor as T
 
--- a/lasagne/conftest.py
+++ b/lasagne/conftest.py
@@ -1,0 +1,12 @@
+ignore_test_paths = [
+    "*/layers/corrmm.py",
+    "*/layers/cuda_convnet.py",
+    "*/layers/dnn.py",
+    ]
+
+
+def pytest_ignore_collect(path, config):
+    """Ignore paths that would otherwise be collceted by the doctest
+    plugin and lead to ImportError due to missing dependencies.
+    """
+    return any(path.fnmatch(ignore) for ignore in ignore_test_paths)
--- a/lasagne/layers/__init__.py
+++ b/lasagne/layers/__init__.py
@@ -7,6 +7,14 @@ from .conv import *
 from .pool import *
 from .shape import *
 from .merge import *
-from .cuda_convnet import *
 from .corrmm import *
-from .dnn import *
+
+try:
+    from .cuda_convnet import *
+except ImportError:
+    pass
+
+try:
+    from .dnn import *
+except ImportError:
+    pass
--- a/lasagne/layers/__init__.py
+++ b/lasagne/layers/__init__.py
@@ -7,7 +7,11 @@ from .conv import *
 from .pool import *
 from .shape import *
 from .merge import *
-from .corrmm import *
+
+try:
+    from .corrmm import *
+except ImportError:
+    pass
 
 try:
     from .cuda_convnet import *
--- a/lasagne/layers/corrmm.py
+++ b/lasagne/layers/corrmm.py
@@ -21,6 +21,10 @@ __all__ = [
 ]
 
 
+if not theano.config.device.startswith("gpu"):
+    raise ImportError("requires a GPU to work")
+
+
 # base class for all layers that rely on GpuCorrMM directly
 class MMLayer(Layer):
     pass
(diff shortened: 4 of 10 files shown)


benanne:
Cool! I think it's preferable to fail early, so that's good, but getting AttributeErrors without error messages doesn't sound very good. People will expect it to be there because it's in the docs (... or it will be, at least).

How do Python libraries typically deal with missing optional dependencies? (Because that's basically what this is, except the dependency is hardware :)) Maybe we could draw inspiration for a better solution from that.

Other than that this looks great. One possible issue is that `Conv2DDNNLayer` and `Conv2DCCLayer` both compute correlations by default, not convolutions (i.e. the filters are not flipped), so they might fail that test. They have a `flip_filters` argument that is `False` by default.

f0k:
> Unforunately, accessing the module contents using attribute-style
module traversal, like in 'Layer = lasagne.layers.Conv2DDNNLayer',
will result in AttributeError, without an error message.

One solution would be not to auto-import these into the `lasagne.layers` namespace, so users always have to import them explicitly, and in turn users will always be greeted with a meaningful `ImportError` at the right place if they don't have the dependencies available.

> The tests so far did turn out a bit hairy, maybe hard to follow.
Also, there's no way a few generic tests like this one could cover all
the special cases between the different implementations. But it does
maybe cover a lot of common ground between them. Interested to hear
what people think.

I think we shouldn't try to replicate what Theano tests already (the correctness of the convolution algorithms). We should instead compare the output of a layer with the output of an equivalent `theano.nnet.conv2d()` call (also for 1D convolution), for a few relevant shape configurations. For inputs, I would use random matrices then, rather than hand-written matrices of ones and zeros. Directly testing the convolutions against known good results isn't a bad move, of course, but it's too much work to reasonably cover the space of possible shape configurations.

dimatura:
Regarding the flipping filter behavior, I just sort of guessed what would the right thing would be for the `Conv2DDNN` layer, but it could go either way. By default the `conv2d` Theano op does convolution, for what it's worth.

dnouri:
On 01/25/2015 03:40 PM, Jan Schlüter wrote:
>     Unforunately, accessing the module contents using attribute-style
>     module traversal, like in 'Layer = lasagne.layers.Conv2DDNNLayer',
>     will result in AttributeError, without an error message.
> 
> One solution would be not to auto-import these into the |lasagne.layers|
> namespace, so users always have to import them explicitly, and in turn
> users will always be greeted with a meaningful |ImportError| at the
> right place if they don't have the dependencies available.

Yes, we could do that (and I like).  But someone will surely argue that
it's inconsistent that they can access some layers inside the
lasagne.layers namespace, and others not.

>     The tests so far did turn out a bit hairy, maybe hard to follow.
>     Also, there's no way a few generic tests like this one could cover all
>     the special cases between the different implementations. But it does
>     maybe cover a lot of common ground between them. Interested to hear
>     what people think.
> 
> I think we shouldn't try to replicate what Theano tests already (the
> correctness of the convolution algorithms). We should instead compare
> the output of a layer with the output of an equivalent
> |theano.nnet.conv2d()| call (also for 1D convolution), for a few
> relevant shape configurations. For inputs, I would use random matrices
> then, rather than hand-written matrices of ones and zeros. Directly
> testing the convolutions against known good results isn't a bad move, of
> course, but it's too much work to reasonably cover the space of possible
> shape configurations.

Good point.  I'll change the fixtures to use conv2d to calculate the
expected value.  That's going to be easier to deal with in the general case.

benanne:
I will merge this and create a separate issue to discuss how we want to handle the gpu-dependent modules. Thanks!

Issue #115, opened 2015-01-25
Naming layers


dimatura:
This arose in the context of issue #111. One issue I've had with `lasagne` is easily identifying layers and their respective (hyper)parameters. This comes up when configuring layers, but also when serializing them, analyzing them after training, and performing 'surgery' on nets for various purposes, such as discriminative pretraining. 

Currently I rely on creating my own dictionaries with name keys and used that to manage them, but it gets messy sometimes to correlate my own names with what e.g., `get_all_param_values` returns.  I feel like simply having a `name` member for the base layer class would make this somewhat easier. I don't necessarily think  they should have a functional role; they can be mostly for debugging/usabilty, and defaulting to `None`, similar to the ways names work for `theano` variables.

benanne:
Personally I haven't had much use for this, but I think this feature has plenty of use cases that are worth supporting. It's easy enough to add support for it to the base class. Unfortunately we will have to update all existing layer classes to pass the optional 'name' argument to the superclass constructor, which is a bit unfortunate. (Luckily users won't have to update their custom layer classes if they don't want to at least - the names of their layer instances will just be `None`). I guess it should be the second argument on the `Layer` constructor, and preferably used as a keyword argument.

It does add a tiny bit of overhead to creating your own `Layer` subclass, but I suppose it's not excessive (and it can safely be ignored, as long as you're not planning to send a pull request).

Maybe we should consider having all `Layer` classes pass `*args` and `**kwargs` to the base class instead? Just as a way of future-proofing the code. We currently already use this scheme with `get_output()` and `get_output_for()`, to support flags and optional arguments that only affect certain layers in a network (for example, `deterministic=True` to turn off noise generating layers). I can't really think of any other use cases for this at the moment (besides this one), but maybe that's kind of the point :)

Anyway, I'm fine with adding a `name` argument. What does everyone else think? @f0k usually has insightful stuff to say about things like this ;)

craffel:
I have definitely thought this too. I think layers should take an optional name argument, and their parameters should have names based on this name too. E.g., a dense layer with name "blah" would have its weight matrix shared named "blah_W" or something.

dimatura:
Yes, definitely a keyword argument.

@craffel Funny you say that, I use that same convention and I was going to propose it maybe later down the line ;)

[...]

PR #117, opened 2015-01-28
naming layers

benanne:
See #115 for discussion.

### DIFF
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -201,9 +201,11 @@ class Layer(object):
                 name = "%s.%s" % (self.name, name)
 
         if isinstance(param, theano.compile.SharedVariable):
-            # cannot check shape here, the shared variable might not be
-            # initialized correctly yet. Note that we cannot assign
-            # a name here.
+            # We cannot check the shape here, the shared variable might not be
+            # initialized correctly yet. We can check the dimensionality though.
+            # Note that we cannot assign a name here.
+            if param.ndim != len(shape):
+                raise RuntimeError("shared variable has %d dimensions, should be %d" % (param.ndim, len(shape)))
             return param
 
         elif isinstance(param, np.ndarray):
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -164,8 +164,7 @@ class Layer(object):
         """
         raise NotImplementedError
 
-    @staticmethod
-    def create_param(param, shape):
+    def create_param(self, param, shape, name=None):
         """
         Helper method to create Theano shared variables for layer parameters
         and to initialize them.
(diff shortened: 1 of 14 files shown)


lasagne/layers/base.py
```diff
@@ -23,8 +23,10 @@ class Layer(object):
     network's output :class:`Layer` instance doubles as a handle to the full
     network.
     """
-    def __init__(self, input_layer):
-        self.input_layer = input_layer
+    def __init__(self, input_layer, name=None):
+        if input_layer is not None:
+            self.input_layer = input_layer
```
f0k: `self.input_layer` should always exist as a property, shouldn't it? Also changing this will create a merge conflict in #52, so I'd appreciate leaving this line unchanged :)

lasagne/layers/base.py
```diff
@@ -189,28 +190,33 @@ def create_param(param, shape):
...
             existing Theano shared variables, and callables for generating
             initial parameter values.
         """
-        if isinstance(param, np.ndarray):
-            if param.shape != shape:
-                raise RuntimeError("parameter array has shape %s, should be %s" % (param.shape, shape))
-            return theano.shared(param)
+        if name is not None:
+            if self.name is not None:
+                name = "%s.%s" % (self.name, name)
```
f0k: If this isn't a static method any more, I need to change back some of the tests in #52. This will remind me of it.

lasagne/layers/base.py
```diff
@@ -189,28 +190,33 @@ def create_param(param, shape):
...
 
-        elif isinstance(param, theano.compile.SharedVariable):
+        if isinstance(param, theano.compile.SharedVariable):
             # cannot check shape here, the shared variable might not be
-            # initialized correctly yet.
+            # initialized correctly yet. Note that we cannot assign
+            # a name here.
             return param
```
f0k: While you're at it, you might want to add a check for `param.ndim == len(shape)`. While we cannot check the shape, we can at least check the dimensionality.

f0k:
Looks good to me except for one minor comment (the first one) and one possible addition (the last one)!

lasagne/layers/base.py
benanne: No, currently the presence of `self.input_layer` or `self.input_layers` is used by `lasagne.layers.get_all_layers` to traverse the graph: https://github.com/benanne/Lasagne/blob/master/lasagne/layers/helper.py#L57
So `InputLayer` is not supposed to have it.

Of course we could make it so this property exists for `InputLayer`, but is set to `None`. Then we just have to update `get_all_layers` to check for `None`. Either approach is fine for me.

lasagne/layers/base.py
benanne: Good call.

benanne:
I went ahead and modified the code so that all layer instances now have either `input_layer` or `input_layers`. For `InputLayer`, `input_layer` is set to `None`. `lasagne.layers.get_all_layers` has been updated to check for this case.

lasagne/layers/base.py
f0k: Ah, I see. It's because `InputLayer` didn't have to call the super constructor before. But then we'll have the same problem with `MultipleInputsLayer`, which shouldn't have a `self.input_layer` member either. Maybe `InputLayer` and `MultipleInputsLayer` shouldn't call the super constructor, but take care of `name` by themselves? Or we need a common base class for `Layer`, `MultipleInputsLayer` and `InputLayer` that takes care of this?

The current solution is a bit dangerous because it means whenever I pass `None` as the first argument of a `DropoutLayer`, for example, it will end up without an `input_layer` property; instead this should be caught and lead to an exception.

benanne: Duplicating the logic for `name` in `InputLayer` kind of smells imo, which is why I modified it so it would call the constructor now. But clearly that isn't an ideal situation either.

Either way catching `input_layer is None` and raising an exception is not something we can do in the `Layer` base class, because the `InputLayer` either doesn't have this property, or it should be `None` (depending on which solution we go for).

But we can't really go around and put that check in every `Layer` subclass either. An intermediate subclass for layers with an input_layer is a possibility, but that complicates things quite a bit and I'm not sure if we want that.

f0k:
OK, reading the code again, I'd suggest to not have `InputLayer` call the super constructor and take care of the name itself, similar to how `MultipleInputsLayer` does it. I'd also suggest the InputLayer to name the input variable `name.input_var` if it was not provided by the user. A common base class for the three classes (`Layer`, `MultipleInputsLayer` and `InputLayer`, as per [my earlier comment](https://github.com/benanne/Lasagne/pull/117#discussion_r23697732)) would be cleaner, but this can still be refactored if we find we change the constructor more often. After this change, `InputLayer` wouldn't have an `input_layer` member any more, similar to how a `MultipleInputsLayer` doesn't have an `input_layer` member. The tree traversal code could still check for `input_layer = None` if that's what we want for #52.

benanne:
I got rid of the last commit and modified `InputLayer` to handle the name attribute itself. I still think it's a bit smelly but all things considered it's probably the most pragmatic solution. And it's only one line of code anyway. If the code for handling this attribute gets any more complicated we will have to refactor this though.

I also changed the input_var name to include the layer name if it exists.

f0k:
Nice! +1 for merging.

benanne:
I'll merge this tomorrow if nobody else has comments.

dimatura:
Looks good as far as I can tell, this will be useful :)

Issue #23, opened 2014-09-12
Which versions of Python / numpy / Theano should be supported?


benanne:
Which versions of Python, numpy and Theano should we support? It's probably okay to rely on Theano 0.6+, but what versions of Python and numpy are in common use at the moment? I would be in favor of not burdening the project with too much 'backward compatibility', so requiring recent Python and numpy versions is okay for me.

Also, what about Python 3? Should we support this? Afaik the scientific community is still relying mostly on Python 2 (and that's what I use), but Python 3 support is probably desirable in the interest of future-proofing the library.

jseabold:
FWIW, many libraries either rely on, or ship with a subset of, the [six](https://pypi.python.org/pypi/six) library. Using it you can target Python 2 and Python 3 with a single codebase.

benanne:
Thanks, that sounds like something we should look into.

ebattenberg:
I've personally been attempting a switchover to python 3.  I've run into a few snags with Theano (you can't use it in "setup.py develop" mode, have to "setup.py install" so the 2to3 script is run). 

For now, should we attempt something similar with Lasagne, i.e., run the 2to3 script when installing for python3?  Then we could at least see if the current python2 code is forward compatible?

ebattenberg:
@jseabold Which popular libraries already rely on six?

ebattenberg:
From the python 2 to 3 porting guide: https://docs.python.org/3/howto/pyporting.html

They suggest using "modernize" and "futurize".  

Modernize (http://python-modernize.readthedocs.org/en/latest/) is a simple script that inserts all the best py2/3 compability practices for existing python 2 code.  

Futurize (http://python-future.org/automatic_conversion.html) goes a bit further by adding some imports for even stronger py3 compatibility with the option of using the extra "future" module for near complete py3 compatibility.  

Futurize seems a bit like the "six" library that @jseabold suggested.

dnouri:
So this write up by Armin Ronacher (the Flask guy) looks like a very useful resource with lots of advice on how to do the porting to Python 3: http://lucumr.pocoo.org/2013/5/21/porting-to-python-3-redux/

I guess it wouldn't be a mistake to try and improve test coverage before we port to Python 3.

dnouri:
Oh I'm seeing that @f0k wrote similar things about a compatibility module back in September.

jseabold:
[...]

PR #119, opened 2015-01-28
python-modernize (python3 compatibility)

ebattenberg:
This PR has:
1. add dtype argument to shared_empty
2. don't re-download mnist if already present
3. ran "python-modernize --no-six"

Basically modernize just added `from __future__ import absolute_import` to everything and fixed the implicit relative imports by making them explicit relative imports.  Looks like the 2/3 compatibility is already pretty darn good.  I can remove the `from __future__ import absolute_import` if you think it just adds clutter.  Probably a good thing to add early though.


benanne:
Cool! Someone told me `urllib.urlretrieve` already checks if the target exists. I actually told someone to remove a fix for this same issue from a pull request a few days ago because of that . But I actually tried it out yesterday and I can't reproduce it (it redownloads every time). So we should probably fix it.

I'd just check with `os.path.exists` though, instead of trying to open it. I know the mantra is "ask for forgiveness, not permission" but in this case I think checking if the path exists before trying to open it is a lot cleaner. That's probably a matter of personal taste though :)

It's good to know that things run fine under Python 3, even though I've never really paid attention to that. 

What does `from __future__ import absolute_import` do exactly? The documentation isn't too clear about this imo. Won't `from . import foo` work just fine without this?

dnouri:
On 01/29/2015 12:02 AM, Sander Dieleman wrote:
> Cool! Someone told me |urllib.urlretrieve| already checks if the target
> exists. I actually told someone to remove a fix for this same issue from
> a pull request a few days ago because of that . But I actually tried it
> out yesterday and I can't reproduce it (it redownloads every time). So
> we should probably fix it.

That was me.  I read the docs and I thought that a previously downloaded
file at that location would be a "valid cached copy", but it's not.

> What does |from __future__ import absolute_import| do exactly? The
> documentation isn't too clear about this imo. Won't |from . import foo|
> work just fine without this?

In Python 2, the default is that "import sys" will first look inside the
package you're in.  So if you have a sys.py file next to the module that
has the import, it will load that instead of the global sys module.
This behaviour has changed with Python 3, where you'll have to use
relative imports ("from . import sys") to get the local sys.py.

dnouri:
@ebattenberg I just made a pull request for your pull request: https://github.com/ebattenberg/Lasagne/pull/1

ebattenberg:
@dnouri, great, I'm a Travis noob.  

@benanne, I haven't run any tests in python3.  I just tried to import it and then fixed the syntax errors.  We'll find out if it works when CI finishes.

ebattenberg:
@benanne  And I agree that try/except isn't as concise.  I'll change it even though it's less pythonic.  ;)

ebattenberg:
> @benanne And I agree that try/except isn't as concise. I'll change it even though it's less pythonic. ;)

Actually, we might want to catch the exception that occurs when the file is only partially downloaded/corrupt.

ebattenberg:
It looks like IOError catches the CRC check failure for incomplete .gzip.  And EOFError catches an empty file.  

BTW, the build failed for 3.3/3.4. :(

dnouri:
It's a total waste building numpy and scipy from scratch every time, which is what Travis is doing now.  I'll look into using miniconda instead of this tomorrw.  Conda has numpy and scipy binaries and so it's much faster with installing.

ebattenberg:
I moved the commits unrelated to py2/3 compat to #120 .

ebattenberg:
@dnouri BTW, how are numpy/scipy obtained for the current python2 tests?  conda?

ebattenberg:
It looks like this will be a little messier than I originally thought, but it should all still be able to be done semi-automatically using modernize.  The main drawback is reliance on the "six" module that @jseabold mentioned.  This provides things like a portable "range", portable "zip", cPickle==pickle, etc.

dnouri:
On 01/29/2015 01:54 AM, Eric Battenberg wrote:
> @dnouri <https://github.com/dnouri> BTW, how are numpy/scipy obtained
> for the current python2 tests? conda?

@ebattenberg You must have missed this comment:
https://github.com/benanne/Lasagne/pull/119#issuecomment-71945841

Or else, maybe by "current" you mean the master branch.  In that case,
they're installed with "apt-get", see
https://github.com/benanne/Lasagne/blob/ab7c590987ac6d75aad14fba6fdda5707999a439/.travis.yml#L7

dnouri:
@ebattenberg Care to merge my latest pull request?  Makes it easier to stop with Travis where there's errors with Python 3.

ebattenberg:
@dnouri 
>Or else, maybe by "current" you mean the master branch.  In that case,
they're installed with "apt-get", see
https://github.com/benanne/Lasagne/blob/ab7c590987ac6d75aad14fba6fdda5707999a439/.travis.yml#L7

Yeah, that's what I meant.  So why can't we use apt-get for python3-numpy/scipy?  That's the package I use on Ubuntu 14.04.

dnouri:
@ebattenberg See https://github.com/travis-ci/travis-ci/issues/2219#issuecomment-41804942

ebattenberg:
> @ebattenberg See travis-ci/travis-ci#2219 (comment)
@dnouri Well that's annoying.  Thanks for adding the workaround.

ebattenberg:
I glanced at the travis failure.  It looks like just a couple errors with importing cPickle (renamed to "pickle" in 3), which can be fixed by either using *six* or using our own compatibility module.  

I'd prefer just using six.  We could also just rip out the internals of six for our own compat module.  How do other feel about this?  @benanne?

dnouri:
Let's just add a `compat.py` module and add stuff we need on a case by case basis, shall we?  I think that's much better than including a module wholesale of which we're not gonna use 80%.

Also, the [Porting to Python 3 Redux](http://lucumr.pocoo.org/2013/5/21/porting-to-python-3-redux/) guide that I linked to earlier discourages the use of six.  Probably a good idea to take advice from people like Armin who maintain some of the most popular Python packages out there.

In the case of `pickle` versus `cPickle`: why not just use `pickle` in Python 2 as well.  This is in examples, so speed won't matter so much.

dnouri:
Also, the `from __future__ import absolute_import` seems superfluous.  I'd say let's try to be compatible but introduce as little noise as possible.

benanne:
To be honest Python 3 is not a big concern for me right now. It'd be great if we can support it, as long as it doesn't get in the way. So a custom `compat.py` module is probably the way to go.

I would also prefer to see `from __future__ import absolute_import` removed, It doesn't seem to be doing anything useful.

ebattenberg:
I can see the argument for keeping the code as simple and dependency-free as possible.  And I do think that the flask guy gives some good advice on 2/3 compatibility.  However, his situation is quite a bit different from ours.  In the case of a popular package like flask, the effort makes sense.  But I think there's a reason that things like six exist, and by reimplementing something that has already been implemented in order to avoid dependencies, we don't benefitting from the work that others have already done for us.  Being able to just run python-modernize on new code and then use whatever has already been written in "six" is compelling.  As @jseabold said, many scientific package already use six.  

At this point it would be really easy to just implement a portable pickle and range, but how much of six would we end up reimplementing as we go forward?  I guess we can pull the relevant portions from the six code into the compat.py.  

@benanne Thanks for being honest about your interest in python3.  I personally don't use python3 really at all, but I've been trying to for a while now, and the thing that keeps me from finally making the jump is modules like this.  All of the main scientific packages seem to support 3 now, but things like Theano aren't quite there yet.  I wonder when the community will actually make the jump.  Or if it's even worth it.  Better unicode support FTW?

For now, I agree we should just implement some minimal compat.py and remove the unnecessary absolute path stuff.

benanne:
I'm still not clear on what `from __future__ import absolute_import` does. Why is this important? Do we have any imports in the library whose semantics change when this import is added? If not, then I see no point in adding it everywhere. It's definitely a bit of an eyesore :)

ebattenberg:
It doesn't *do* anything.  It just disables implicit relative imports, you you can't `import layers`, for example.  You'd have to `from . import layers`.

ebattenberg:
And I agree that it can be removed.  I'll work on this PR more in a bit.

benanne:
I see :) I think we're disciplined enough to remember that. Sounds good!

ebattenberg:
Ok, I added a very basic _compat.py module.  For now, it's only used for the pickle module.  

I still think that using six is the easiest solution for this.  Pylearn2 uses it and Theano will most likely end up using it.  I mean, all users of Lasagne use numpy, scipy, matplotlib, and don't those all have a dependency on six anyway?  That or they ship with it.  I'm not sure what the advantage is in trying to avoid having it as a dependency when all you're doing is making more work for yourself in order to have slightly cleaner code.  At the very least, we should just copy the current version of six into Lasagne.  No reason to duplicate the work of other, more qualified developers.

ebattenberg:
One more thing.  I'm having trouble running the tests on the example code from my local machine.  I'm not very familiar with py.test, but when I try to run it, it always skips the example code tests.  I have cudnn, pylearn2, etc. installed and working.  What am I missing?

ebattenberg:
What do you guys think of just shipping six.py along with and then having a compat.py to make the six imports look cleaner?

ebattenberg:
Annnd there seem to be pickle incompatibilities due to text encoding between python 2 and 3.  

http://stackoverflow.com/questions/11305790/pickle-incompatability-of-numpy-arrays-between-python-2-and-3

dnouri:
A few things:
1) Pickles don't have to be compatible between Python 2 and 3; not sure why you would be worried about that.
2) I mentioned it before but: the compatibility issue with pickle is best solved by using "pickle" (instead of "cPickle") under both Python 2 and 3.
3) numpy has a six dependency since when?  AFAICT, it only has [this small module](https://github.com/numpy/numpy/blob/master/numpy/compat/py3k.py); we should aim for something similar.
4) Adding a 800 line module of which we only use a few lines is just sloppy; it's not a matter of "slightly cleaner code".

You have to use the `--runslow` switch (look at how it's configured for Travis), to run the example tests.  These are normally skipped because they're too slow.

ebattenberg:
>You have to use the --runslow switch (look at how it's configured for Travis), to run the example tests. These are normally skipped because they're too slow.

Thanks.  

>1) Pickles don't have to be compatible between Python 2 and 3; not sure why you would be worried about that.

I wasn't worried.  I don't expect serialized objects to be portable across python versions.   I was just sad it didn't just work.  ;)

>2) I mentioned it before but: the compatibility issue with pickle is best solved by using "pickle" (instead of "cPickle") under both Python 2 and 3.

Yes, I'm sure you know this, but the only option in Python 3 is "pickle" which is equivalent to "cPickle" in Python 2.  Using "pickle" under python 3 is what produced the ASCII decode error.  

>3) numpy has a six dependency since when? AFAICT, it only has this small module; we should aim for something similar.

My knowledge of six is limited to our conversations here and at #23 [e.g.](https://github.com/benanne/Lasagne/issues/23#issuecomment-71942985)


>4) Adding a 800 line module of which we only use a few lines is just sloppy; it's not a matter of "slightly cleaner code".

Maybe overkill, not sloppy.  I personally like code reuse, and the file is only a few kB; I think most people can spare the disk space.  And as we've established, I'm not a Python 2/3 portability expert.  @dnouri , I'd be happy to pass the baton to you.  :)

dnouri:
On 02/02/2015 10:46 PM, Eric Battenberg wrote:
>     1) Pickles don't have to be compatible between Python 2 and 3; not
>     sure why you would be worried about that.
> 
> I wasn't worried. I don't expect serialized objects to be portable
> across python versions. I was just sad it didn't just work. ;)
> 
>     2) I mentioned it before but: the compatibility issue with pickle is
>     best solved by using "pickle" (instead of "cPickle") under both
>     Python 2 and 3.
> 
> Yes, I'm sure you know this, but the only option in Python 3 is "pickle"
> which is equivalent to "cPickle" in Python 2. Using "pickle" under
> python 3 is what produced the ASCII decode error.

I just saw this SO question (and maybe you did too) about the unpickling
error:
http://stackoverflow.com/questions/11305790/pickle-incompatability-of-numpy-arrays-between-python-2-and-3

And now it makes sense why you brought up the compatibility issue
between pickles in 2 and 3.  So the mnist pickle doesn't seem to work
with Python 3, that's pretty crappy.  Maybe we can find it somewhere
else in some other format to download?  Probably preferable to adding a
workaround in the code.

> [...]

>     4) Adding a 800 line module of which we only use a few lines is just
>     sloppy; it's not a matter of "slightly cleaner code".
> 
> Maybe overkill, not sloppy. I personally like code reuse, and the file
> is only a few kB; I think most people can spare the disk space. And as
> we've established, I'm not a Python 2/3 portability expert. @dnouri,
> I'd be happy to pass the baton to you. :)

You're right, overkill is the word.  So let's stick to copying stuff out
of six when we need it, and put it in a compat.py module, agree?

Hope I wasn't too much of a party pooper here.  If you wanna continue
with this work go ahead.  I will probably not have time for it until
next week, but I'll need it eventually for one project that I'm working on.

ebattenberg:
>I just saw this SO question (and maybe you did too) about the unpickling
error:
http://stackoverflow.com/questions/11305790/pickle-incompatability-of-numpy-arrays-between-python-2-and-3

Yeah, I linked to it [above](https://github.com/benanne/Lasagne/pull/119#issuecomment-72415695).  Now I feel like you don't even read my posts. :(  ;)    Yeah, it'd be nice to find a portable version of mnist.  Maybe a .npz file somewhere?  


>You're right, overkill is the word.  So let's stick to copying stuff out
of six when we need it, and put it in a compat.py module, agree?

That's a fair compromise.  But why not just copy it all now?  Copying individual lines means understanding what each line does and what they depend on.  It's not immediately obvious to me.  Here's the source:  https://bitbucket.org/gutworth/six/src/78315705481c3f5e491a3816e41aa66c54c8f965/six.py?at=default

 I think stealing from an existing compat module that's simpler would be easier to do if we want to pick and choose lines.

f0k:
> Yeah, it'd be nice to find a portable version of mnist. Maybe a .npz file somewhere?

To chime in here, we noticed the same with the examples in cudamat. My suggestion was to use the original MNIST files from Yann's website instead. @pallegro gave [a code snipped to read them](https://github.com/cudamat/cudamat/pull/43#issuecomment-70135249):

> ```python
with gzip.GzipFile('train-images-idx3-ubyte.gz', 'rb') as fo:
    magic,nimg,nrow,ncol = np.fromstring(fo.read(4*4), '>i4', 4)
    data = np.fromstring(fo.read(), uint8).reshape(nimg,nrow*ncol))
```

We'd just need to check whether they are the same orientation, and we'd need to download multiple files, but it's not that bad as everything is encapsulated in the `mnist.py` file.

benanne:
Since most of this is superseded by #130, is there anything from this we want to cherry-pick (e.g. the `dtype` argument for `shared_empty`)?

ebattenberg:
>Since most of this is superseded by #130, is there anything from this we want to cherry-pick (e.g. the dtype argument for shared_empty)?

I don't think so.  I cherry-picked the `dtype` from `shared_empty` and mnist redownload into #120, which were already merged.  I think we can close this one.  :)

PR #120, opened 2015-01-29
Mnist redownload check, dtype for shared_empty

ebattenberg:
Moving these unrelated parts out of the py2/3 effort (#119)


benanne:
Looks good to me!

PR #123, opened 2015-01-30
removed automatic import of modules with GPU-related dependencies.

benanne:
also updated convolution implementation tests to match. @dnouri does this look okay? You wrote the tests originally so I guess you should have a look.


dnouri:
Looks good to me!

benanne:
Alright then, not gonna wait a day with this one :)

PR #124, opened 2015-01-31
Preventing None from appearing in get_all_layers' returned list

craffel:
This should fix the issue I mentioned here:
https://github.com/benanne/Lasagne/pull/52#issuecomment-72264767
I.e., now that a layer can have an input attribute which is `None`, `get_all_layers` can return a list which has entries which are `None`.  This causes problems, e.g. causes `get_all_params` to fail because 
```Python
    layers = get_all_layers(layer)
    params = sum([l.get_params() for l in layers], [])
```
I'm not 100% sure this is the best way/only place to fix this, but I think it is.

### DIFF
--- a/lasagne/layers/helper.py
+++ b/lasagne/layers/helper.py
@@ -60,7 +60,9 @@ def get_all_layers(layer):
             children = [current_layer.input_layer]
 
         # filter the layers that have already been visited.
-        children = [child for child in children if child not in layers]
+        children = [child for child in children
+                    if child not in layers
+                    and child is not None]
         layers_to_expand.extend(children)
         layers.extend(children)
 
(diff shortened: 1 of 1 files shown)


benanne:
I believe it is. Maybe the comment should also be updated to mention the `None` case. Just in case we're both missing something it'd be good if someone else also reviews this issue.

craffel:
Which comment?

benanne:
It says `# filter the layers that have already been visited.` - but now `None` is also being filtered from the list.

f0k:
> I'm not 100% sure this is the best way/only place to fix this, but I think it is.

Seems the best way. This also correctly handles the case where a `MultipleInputLayer` is constructed partly from shapes and partly from `Layer` instances.

> Which comment?

Probably the one [directly before your change](https://github.com/craffel/nntools/blob/none-input/lasagne/layers/helper.py#L62). You could change it to:
```python
# filter the layers that have already been visited, and filter
# `None` elements (for layers without incoming layers)
```
or something like that.

craffel:
Ah, ok, I was thinking an issue comment.  Comment added to the code.

benanne:
Looks good, gonna merge it straight away as this is a pretty serious bug.

Issue #15, opened 2014-09-12
Objective functions


benanne:
There is currently a submodule `nntools.objectives`, but like the regularization one this was a bit of an afterthought.

I would be great if the code in `nntools.objectives` is defined in terms of Theano expressions (not `Layer` instances), so that it can be used in isolation.

But sometimes it might be convenient to have an object that represents the objective function, which defines a Theano variable for the target labels (in the case of a supervised training objective), so that you can just do something like:
```
l_out = ... # output layer of the network
obj = nntools.objectives.Objective(l_out, loss_function=nntools.objectives.mse)
loss = obj.get_loss()
```
I don't know if this warrants the extra class though. It doesn't really do that much. What does everyone think?

ebattenberg:
I wanted to revive this discussion with something I came across.  In my work, I sometimes use a weighted loss function where one might weight individual examples differently or weight individual output labels or classes differently.  For this, I would need to add additional arguments to the loss function besides `x` and `t` (the predicted and target outputs).  

So it would be nice to be able to pass additional parameters to the loss function I define.  Right now, the `Objective` constructor passes all additional args to `get_output`.  

I guess I could do this using a function closure as well.  Or I could implement it separately since the `Objective` class is more of a convenience function.

benanne:
Have a look at the discussion on #127 :)

benanne:
The `lasagne.objectives` module has been slightly reworked in the meantime so that the various objective functions (mse etc.) no longer compute the average across the batch. This is taken care of by the `Objective` class instead, which has allowed us to implement a `MaskedObjective`.

The question was raised whether we really need `Objective` objects - they only really define one method after all (`get_loss()`). Maybe their job could be done by simple Python functions instead.

The reason I like `Objective` as a class is because it allows for a default target variable (`target_var`) to be defined automatically, so I don't need to worry about this. This is nicely symmetric with the input variable (`input_var`) that is automatically created by the `InputLayer`. So I think I'm in favour of keeping things like they are for this reason. But we agreed during the dev meeting that this should be discussed, so: discuss :)

f0k:
> The question was raised whether we really need `Objective` objects - they only really define one method after all (`get_loss()`). Maybe their job could be done by simple Python functions instead.

[...]

PR #127, opened 2015-02-03
Objectives now have an optional mask.

Britefury:
Masked objectives allow specified rows to be ignored when computing the loss value. For classification problems, this can be useful if training or test set is not an exact multiple of the mini-batch size; on the last mini-batch you provide a mask with zeros for the entries in the mini-batch for which you do not have data. For image segmentation problems the mask can be used to define a region of interest.

### DIFF
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -106,11 +106,11 @@ def create_iter_functions(dataset, output_layer,
     batch_slice = slice(
         batch_index * batch_size, (batch_index + 1) * batch_size)
 
-    objective = lasagne.objectives.Objective(
-        output_layer, loss_function=lasagne.objectives.multinomial_nll)
+    objective = lasagne.objectives.Objective(output_layer,
+        loss_function=lasagne.objectives.categorical_crossentropy)
 
     loss_train = objective.get_loss(X_batch, target=y_batch)
-    loss_eval = objective.get_loss(X_batch, target=y_batch, deterministic=True)
+    loss_eval = objective.get_loss(X_batch, target=y_batch,deterministic=True)
 
     pred = T.argmax(
         output_layer.get_output(X_batch, deterministic=True), axis=1)
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -110,7 +110,8 @@ def create_iter_functions(dataset, output_layer,
         loss_function=lasagne.objectives.categorical_crossentropy)
 
     loss_train = objective.get_loss(X_batch, target=y_batch)
-    loss_eval = objective.get_loss(X_batch, target=y_batch,deterministic=True)
+    loss_eval = objective.get_loss(X_batch, target=y_batch,
+                                   deterministic=True)
 
     pred = T.argmax(
         output_layer.get_output(X_batch, deterministic=True), axis=1)
--- a/lasagne/objectives.py
+++ b/lasagne/objectives.py
@@ -94,7 +94,7 @@ class Objective(object):
 
 
 class MaskedObjective(object):
-    _valid_mask_normalization = {None, 'none', 'mean', 'sum'}
+    _valid_aggregation = {None, 'sum', 'mean', 'normalized_sum'}
 
     """
     Masked training objective
@@ -103,7 +103,7 @@ class MaskedObjective(object):
     training with a gradient descent approach, with masking applied to weight
     the contribution of samples to the final loss.
     """
-    def __init__(self, input_layer, loss_function=mse, mask_normalization='mean'):
+    def __init__(self, input_layer, loss_function=mse, aggregation='mean'):
         """
         Constructor
 
@@ -113,25 +113,26 @@ class MaskedObjective(object):
             - loss_function : a loss function of the form `f(x, t, m)` that
                 returns a scalar loss given tensors that represent the
                 predicted values, true values and mask as arguments.
-            - mask_normalization : either:
-                - `'mean'` : the elements of the loss will be multiplied by
-                the mask and the mean returned
+            - aggregation : either:
+                - `None` or `'mean'` : the elements of the loss will be
+                multiplied by the mask and the mean returned
                 - `'sum'` : the elements of the loss will be multiplied by
-                the mask, summed and divided by the sum of the mask
-                - `None` or `'none'` : the elements of the loss will be
-                multiplied by the mask and the sum returned
+                the mask and the sum returned
+                - `'normalized_sum'` : the elements of the loss will be
+                multiplied by the mask, summed and divided by the sum of
+                the mask
         """
         self.input_layer = input_layer
         self.loss_function = loss_function
         self.target_var = T.matrix("target")
         self.mask_var = T.matrix("mask")
-        if mask_normalization not in self._valid_mask_normalization:
-            raise ValueError('mask_normalization must be \'mean\', \'sum\', ' \
-                '\'none\' or None, not {0}'.format(mask_normalization))
-        self.mask_normalization = mask_normalization
+        if aggregation not in self._valid_aggregation:
+            raise ValueError('aggregation must be \'mean\', \'sum\', ' \
+                '\'normalized_sum\' or None, not {0}'.format(aggregation))
+        self.aggregation = aggregation
 
     def get_loss(self, input=None, target=None, mask=None,
-                 mask_normalization=None, *args, **kwargs):
+                 aggregation=None, *args, **kwargs):
         """
         Get loss scalar expression
 
@@ -145,7 +146,7 @@ class MaskedObjective(object):
                 as - or broadcast-able to the shape of - the result of
                 applying the loss function. It selects/weights the
                 contributions of the resulting loss values
-            - mask_normalization : None to use the value passed to the
+            - aggregation : None to use the value passed to the
                 constructor or a value to override it
             - args : additional arguments passed to `input_layer`'s
                 `get_output` method
@@ -161,21 +162,21 @@ class MaskedObjective(object):
         if mask is None:
             mask = self.mask_var
 
-        if mask_normalization not in self._valid_mask_normalization:
-            raise ValueError('mask_normalization must be \'mean\', \'sum\', ' \
-                '\'none\' or None, not {0}'.format(mask_normalization))
+        if aggregation not in self._valid_aggregation:
+            raise ValueError('aggregation must be \'mean\', \'sum\', ' \
+                '\'normalized_sum\' or None, not {0}'.format(aggregation))
 
-        # Get mask_normalization value passed to constructor if None
-        if mask_normalization is None:
-            mask_normalization = self.mask_normalization
+        # Get aggregation value passed to constructor if None
+        if aggregation is None:
+            aggregation = self.aggregation
 
         masked_losses = self.loss_function(network_output, target) * mask
 
-        if mask_normalization is None  or  mask_normalization == 'none':
-            return masked_losses.sum()
-        elif mask_normalization == 'sum':
-            return masked_losses.sum() / mask.sum()
-        elif mask_normalization == 'mean':
+        if aggregation is None  or  aggregation == 'mean':
             return masked_losses.mean()
+        elif aggregation == 'sum':
+            return masked_losses.sum()
+        elif aggregation == 'normalized_sum':
+            return masked_losses.sum() / mask.sum()
         else:
             raise RuntimeError('This should have been caught earlier')
--- a/lasagne/objectives.py
+++ b/lasagne/objectives.py
@@ -181,6 +181,6 @@ class MaskedObjective(object):
             normalize_mask = self.normalize_mask
 
         if normalize_mask:
-            mask = mask / T.cast(1.0 / T.sum(mask), theano.config.floatX)
+            mask = mask / T.sum(mask)
 
         return self.loss_function(network_output, target, mask)
--- a/lasagne/objectives.py
+++ b/lasagne/objectives.py
@@ -1,35 +1,50 @@
+import theano
 import theano.tensor as T
 
 
-def mse(x, t):
+def mse(x, t, m):
     """Calculates the MSE mean across all dimensions, i.e. feature
      dimension AND minibatch dimension.
 
     :parameters:
         - x : predicted values
         - t : target values
+        - m : mask; None for no mask, or a 1-dimensional array, that
+            selects which rows in `x` and `t` should count towards the
+            result
 
     :returns:
         - output : the mean square error across all dimensions
     """
-    return T.mean((x - t) ** 2)
+    if m is None:
+        return T.mean((x - t) ** 2)
+    else:
+        recip_mask_sum = T.cast(1.0 / T.sum(m), theano.config.floatX)
+        return T.sum(m * (x - t) ** 2) * recip_mask_sum
 
 
-def crossentropy(x, t):
+def crossentropy(x, t, m):
     """Calculates the binary crossentropy mean across all dimentions,
     i.e.  feature dimension AND minibatch dimension.
 
     :parameters:
         - x : predicted values
         - t : target values
+        - m : mask; None for no mask, or a 1-dimensional array, that
+            selects which rows in `x` and `t` should count towards the
+            result
 
     :returns:
         - output : the mean binary cross entropy across all dimensions
     """
-    return T.mean(T.nnet.binary_crossentropy(x, t))
+    if m is None:
+        return T.mean(T.nnet.binary_crossentropy(x, t))
+    else:
+        recip_mask_sum = T.cast(1.0 / T.sum(m), theano.config.floatX)
+        return T.sum(T.nnet.binary_crossentropy(x, t)) * recip_mask_sum
 
 
-def multinomial_nll(x, t):
+def multinomial_nll(x, t, m):
     """Calculates the mean multinomial negative-log-loss
 
     :parameters:
@@ -41,11 +56,21 @@ def multinomial_nll(x, t):
             a 1D integer array that gives the class index of each sample
             (the position of the 1 in the row in a 1-of-N encoding, or
             1-hot encoding),
+        - m : mask; None for no mask, or a 1-dimensional array, that
+            selects which rows in `x` and `t` should count towards the
+            result
 
     :returns:
         - output : the mean multinomial negative log loss
     """
-    return T.mean(T.nnet.categorical_crossentropy(x, t))
+    if m is None:
+        return -T.mean(T.log(x)[T.arange(t.shape[0]), t])
+    else:
+        recip_mask_sum = T.cast(1.0 / T.sum(m), theano.config.floatX)
+        masked_nll_sum = T.sum(T.log(x)[T.arange(t.shape[0]), t] * m)
+        return -masked_nll_sum * recip_mask_sum
+
+
 
 
 class Objective(object):
@@ -69,8 +94,9 @@ class Objective(object):
         self.input_layer = input_layer
         self.loss_function = loss_function
         self.target_var = T.matrix("target")
+        self.mask_var = None
 
-    def get_loss(self, input=None, target=None, *args, **kwargs):
+    def get_loss(self, input=None, target=None, mask=None, *args, **kwargs):
         """
         Get loss scalar expression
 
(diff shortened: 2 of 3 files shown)


lasagne/objectives.py
```diff
@@ -1,35 +1,50 @@
+import theano
 import theano.tensor as T
 
 
-def mse(x, t):
+def mse(x, t, m):
```
f0k: That's not exactly an optional mask. (Hint: Should be `m=None` instead.)

lasagne/objectives.py
```diff
@@ -1,35 +1,50 @@
...
     """Calculates the MSE mean across all dimensions, i.e. feature
      dimension AND minibatch dimension.
 
     :parameters:
         - x : predicted values
         - t : target values
+        - m : mask; None for no mask, or a 1-dimensional array, that
+            selects which rows in `x` and `t` should count towards the
+            result
```
f0k: Does it have to be a 1-dimensional array? I'd prefer being able to use it with a mask the same shape as `t`.

lasagne/objectives.py
```diff
@@ -1,35 +1,50 @@
...
 
     :returns:
         - output : the mean square error across all dimensions
     """
-    return T.mean((x - t) ** 2)
+    if m is None:
+        return T.mean((x - t) ** 2)
+    else:
+        recip_mask_sum = T.cast(1.0 / T.sum(m), theano.config.floatX)
+        return T.sum(m * (x - t) ** 2) * recip_mask_sum
```
f0k: I'm not sure about the renormalization. When you have some unmasked minibatches and then some half-masked minibatches, each datapoint in a half-masked minibatch will contribute twice as strongly to the weight updates as a datapoint in an unmasked minibatch. If at all, the renormalization should be optional.

lasagne/objectives.py
f0k: Checking the code, it correctly handles masks that are the same shape as `x` and `t`, and it correctly handles non-binary masks (something I use a lot). The docstring should be updated to reflect that.

lasagne/objectives.py
```diff
@@ -41,11 +56,21 @@ def multinomial_nll(x, t):
...
+        - m : mask; None for no mask, or a 1-dimensional array, that
+            selects which rows in `x` and `t` should count towards the
+            result
 
     :returns:
         - output : the mean multinomial negative log loss
     """
-    return T.mean(T.nnet.categorical_crossentropy(x, t))
+    if m is None:
+        return -T.mean(T.log(x)[T.arange(t.shape[0]), t])
```
f0k: That's not the same as before. `t` can explicitly be either an integer array (the case you assumed here) or a 2D tensor of targets in one-hot encoding. If it's too difficult to treat both in one function, we should think about explicitly splitting it up.

lasagne/objectives.py
```diff
@@ -70,8 +95,9 @@ def __init__(self, input_layer, loss_function=mse):
         self.input_layer = input_layer
         self.loss_function = loss_function
```
f0k: The docstring for the `loss_function` argument needs to be updated to account for the new argument.

lasagne/objectives.py
```diff
@@ -70,8 +95,9 @@ def __init__(self, input_layer, loss_function=mse):
...
         self.target_var = T.matrix("target")
+        self.mask_var = None
```
f0k: It should be possible to set this from the constructor, otherwise it should be omitted.

f0k:
Thanks for your contribution! I'm using masks myself (but not binary ones, so I usually call them "weights"), so I think it could be a useful addition, but on the other hand, I've done quite fine without using `lasagne.objectives` for that. I'm not sure whether we want to force all loss functions to support the `mask` parameter, though -- on the one hand, it would allow people to use it everywhere, but on the other hand, it would make it more difficult to implement new loss functions. We could change `Objective.get_loss` to only pass `mask` to `self.loss_function` if `mask is not None`, or we could devise a `MaskedObjective` class that only works with 3-argument loss functions, or we could decide not to support this in Lasagne and leave it up to the user.
What do others think? I'm slightly leaning towards `MaskedObjective` and making sure that it allows non-binary masks / target weights / soft masks as well.

benanne:
To be honest I think this is a case where the use cases are not common enough to warrant a modification of the code of this extent. With all the mask handling code added, the loss function implementations suddenly take effort to understand, which they did not before.

If we want to provide this masking/weighting functionality I agree that it should preferably be added somewhere else, either in a separate set of loss function implementations, or with some other method.

The current implementation of `objectives` is very rough around the edges (I didn't spend much thought on it and I think it hasn't changed since the start). To support these use cases more easily it may actually be beneficial to take the mean operation over the batch axis out of the individual loss function implementations. Then it would be trivial to have `Objective` and `MaskedObjective` and the masking logic does not need to be duplicated in every new loss function.

One possible concern is that it could sometimes impact performance (maybe calling `T.mean` only once to average over both the batch axis and the other axes is cheaper than doing it in two steps? Or does Theano optimize this?) But it would definitely yield cleaner code. Sometimes you just want to compute the loss for each individual example anyway, and this is currently impossible with the provided code.

f0k:
> With all the mask handling code added, the loss function implementations suddenly take effort to understand, which they did not before.

On the other hand, without the mask handling code, the loss functions do not provide any benefit over the existing `T.nnet.categorical_crossentropy` etc., and the original, easy-to-understand implementations are preserved in the `mask=None` case. If you don't use masks, they are as easy to understand as before, and if you use masks, there is no way around additional complexity.

> To support these use cases more easily it may actually be beneficial to take the mean operation over the batch axis out of the individual loss function implementations. Then it would be trivial to have `Objective` and `MaskedObjective` and the masking logic does not need to be duplicated in every new loss function.

Hmm, good point. But is it possible to have all conceivable loss functions return an element-wise loss, or would that exclude some loss functions? And what about masks that are not per data point, but per target (for tasks of multiple independent binary targets)? That would be impossible if you average over the other axes before. The renormalization step could easily be taken out into a `MaskedObjective` class, though, which would get rid of half of the code duplication and easily allow the renormalization to become optional.

So I think extending some of the provided loss functions with an optional `mask` parameter and adding a `MaskedObjective` that makes use of it would be feasible.

benanne:
> Note that we're not compatible to the loss functions in Theano now because we include a `T.mean()`.

Yep, probably a good idea to change that regardless.

> That's what came to my mind as well, but you can just reshape the minibatch to have half the number of examples with twice the number of features before shoving it into your cost function.

The version I was thinking of actually matches every point in a minibatch to every other point in the minibatch, so that wouldn't work :) (It's a bit hackish to implement it this way but it's worked for me in the past.)

> Hmm, that's not a good term, yes... I was referring to computing the mean over the unmasked datapoints instead of computing the mean over everything: https://github.com/benanne/Lasagne/pull/127#discussion_r24007830 (Essentially, dividing the sum by another factor than before.)

Ah, I see.

> If we had the loss functions not do any mean and let `Objective` and `MaskedObjective` cope with that, we could support both unmasked, per-element-masked and per-target-masked objectives, and we wouldn't run into the multiple `T.mean()` case either (which could really affect performance or numeric stability).

True. I suppose there are loss functions that don't take the mean across outputs as well though (and those are probably more common than those that don't take the mean across the batch dimension). But maybe we could set it up so that the `Objective` classes are agnostic to the output dimensionality of the loss function (except that the first dimension should always be the batch dimension). Then we can cover all those cases in one go.

f0k:
> But maybe we could set it up so that the `Objective` classes are agnostic to the output dimensionality of the loss function

Well, with the few types of loss we currently support there are already three different cases of loss functions:

1. binary cross-entropy, squared difference: 2D input, 2D target -> 2D loss
2. multinomial cross-entropy (1-hot targets): 2D input, 2D target -> 1D loss
3. categorical cross-entropy (integer targets): 2D input, 1D target -> 1D loss

For `Objective`, it's easy, it just needs to take the overall mean (that's what every loss function does right now). For `MaskedObjective`, it would need a multiplication with the mask before taking the overall mean, or before taking the sum and dividing it by `mask.sum()`. It would depend on the loss function whether the mask must be 1D or can be 2D, but a 1D mask would always work -- except that it must be dimshuffled into a column vector for a 2D loss function. That's something we could do automatically. If a user wants to mask out particular features rather than particular data points, she or he would need to properly dimshuffle the mask herself or himself into a 2D row vector.

The only possible problem is that existing code may use `lasagne.objectives.something` without the `Objective` class already, but I'd say before we release anything, we're free to break the interface.

benanne:
I agree that we can break the interface for now. We can announce breaking changes on the mailing list to make sure everybody's aware (we should also do this when we move `get_output()` outside of the `Layer` class).

Britefury:
The loss functions `m` argument now defaults to None. Tests implemented and pass.
Removed masking functionality from `Objective` class into new `MaskedObjective` class. Mask normalisation is handled by `MaskedObjective` and is optional, defaulting to off.

benanne:
That's definitely an improvement, but as we discussed on this PR before, I feel we need some more fundamental changes to handle this cleanly.

We should modify the interface of the objective functions so they don't compute the mean across examples (at the very least), then we can implement the masking once and use it in conjunction with all of them, instead of having to add masking logic to every objective function separately.

ebattenberg:
> Thanks for your contribution! I'm using masks myself (but not binary ones, so I usually call them "weights")

If we are going to add this functionality, I personally think it should be called `WeightedObjective` rather than `MaskedObjective` because this points to the fact that the values can be non-binary.  But I would be interested to hear how others typically refer to this kind of thing.  @Britefury's initial use case for this is masking, but the generalization of this I would call weighting.  

> To be honest I think this is a case where the use cases are not common enough to warrant increased complexity of the code of this extent.

I agree with this.  I feel like whenever I start working on a new project (unless it's MNIST or some other sanitized dataset), I always write a custom objective function whether it has weights per training example, or per output, or per class.  To address all of these use cases at once, can we just add an optional `loss_kwargs` dictionary that is passed to the custom loss function we implement?  That's kind of the functionality I was pondering in #15.  

In the meantime, I've solved this by using a function closure for my custom loss function.

f0k:
> @Britefury's initial use case for this is masking, but the generalization of this I would call weighting.

I also call it weighting, but I'd be quite fine to refer to it as "soft masks" to avoid confusion with network weights. And then `MaskedObjective` is okay for a name.

> In the meantime, I've solved this by using a function closure for my custom loss function.

I've solved it by not using `Objective` at all, it hardly provides any functionality so far. It could if we change the loss functions to not take any mean and then let `Objective` and `MaskedObjective` deal with that.

ebattenberg:
> I've solved it by not using Objective at all, it hardly provides any functionality so far. It could if we change the loss functions to not take any mean and then let Objective and MaskedObjective deal with that.

That makes more sense than what I did with the loss function closure.  I guess I was trying to adhere to using the current `Objective` interface so I could benefit from it in the future if anything worth using was added.  With that in mind, what could `Objective` and `MaskedObjective` classes add that just implementing specialized loss functions couldn't?  It seems like just calling `get_output` with whatever arguments you want to give it and then calling your particular loss function with its own arguments isn't that hard.  

What are we trying to abstract away here?

benanne:
I like `Objective` because it creates the target variable for me (just like I like to use the input variable that `InputLayer` creates for me). Other than that it currently doesn't have any real benefits. But as we discussed here earlier, I think averaging over the batch (and possibly other dimensions) should be moved into the `Objective` class.

Britefury:
The mean/sum and masking functionality has been moved into `Objective`/`MaskedObjective`. The removal of the mean/sum functionality means that the loss function tests now need to use `Objective`/`MaskedObjective` in order to get the same results; there are now some convenience methods to ease this by applying `Objective`/`MaskedObjective` . The masked objective test has been commented out since `Mock` objects do not support multiplication and the functionality that it tests is now covered by the loss function tests.

benanne:
Nice work! I think the `crossentropy` and `multinomial_nll` functions can be removed since they're just aliases for existing Theano functions now.

One thing that may lead to confusion and unexpected behaviour is that the `MaskedObjective` sums across all dimensions, whereas `Objective` takes the mean. If there is only one dimension this is okay because the weights can be normalized, but if there are more (as with the MSE objective for example) this means the objective is scaled differently (which means it requires different learning rates, and so on).

Does it make sense to use `mean` in `MaskedObjective` as well? I guess it's just a question of normalizing the weights differently, right?

Britefury:
`crossentropy` and `multinomial_nll` have been replaced with imports.
As for using `mean` in `MaskedObjective`, I think that would complicate how you must normalize the weights. For a typical use case, when using `sum` the weights are normalized with `weights / weights.sum()`. When using `mean`, you would have to use `weights * np.prod(weights.shape) / weights.sum()`.

f0k:
> For a typical use case, when using `sum` the weights are normalized with `weights / weights.sum()`. When using `mean`, you would have to use `weights * np.prod(weights.shape) / weights.sum()`.

I would use `mean` (or `sum` with `weights / T.prod(weights.shape)`) by default, and `sum` with `weights / weights.sum()` if you choose the different normalization mode (`normalize_mask = True`).

Britefury:
Oops, I didn't see that; that was a maths fail on my part! :(

Okay. The `normalize_mask` has been replaced by `mask_normalization` which allows you to choose between mean, sum and normalized sum.
I have added a similar `aggregation` parameter to Objective to allow you to choose between mean and sum.
My use case is as follows:
During training, you would want to use MaskedObjective with mean, so that - as you said - parameters with 1.0 masks don't get extra weight. This is fine.
During testing, your last minibatch is incomplete. In this case, you want to use sum for regular `Objective` on all your previous complete minibatches and `MaskedObjective` with no normalization on your last minibatch and divide by the test set size at the end.
Is this okay?

f0k:
> Okay. The `normalize_mask` has been replaced by `mask_normalization` which allows you to choose between mean, sum and normalized sum.
I have added a similar `aggregation` parameter to Objective to allow you to choose between mean and sum.

Sounds good to me! But for consistency, I would simply rename `mask_normalization` to `aggregation`. It's the same just with an additional possibility.

> My use case is as follows: [...] Is this okay?

Sounds reasonable, yes.

My problem with the normalization was as follows: Assume your dataset has the data points A, B, and C, and your batchsize is 2. Assume that you include incomplete minibatches in training via a MaskedObjective (I usually just drop incomplete batches because I shuffle the datapoints in each epoch anyway). Assume that your first epoch has the minibatches (A,B) and (C). That is, the first weight update will rely on both A and B, and the second only on C. If you use a MaskedObjective with normalized sum, C will affect your network twice as strongly as either of A or B just because it ended up in an incomplete minibatch. If you use mean, the second weight update will be half as strong as the first to account for the fact that it used half the number of datapoints. I'd usually prefer the latter.

Britefury:
Alright, `mask_normalization` is now `aggregation`. The values for `aggregation` are now consistent with those of `Objective`.

examples/mnist.py
```diff
@@ -97,10 +97,11 @@ def create_iter_functions(dataset, output_layer,
     batch_slice = slice(
         batch_index * batch_size, (batch_index + 1) * batch_size)
 
-    objective = lasagne.objectives.Objective(output_layer, loss_function=lasagne.objectives.multinomial_nll)
+    objective = lasagne.objectives.Objective(output_layer,
+        loss_function=lasagne.objectives.categorical_crossentropy)
 
     loss_train = objective.get_loss(X_batch, target=y_batch)
-    loss_eval = objective.get_loss(X_batch, target=y_batch, deterministic=True)
+    loss_eval = objective.get_loss(X_batch, target=y_batch,deterministic=True)
```
f0k: Did you actually remove the blank there or is that because it's based off of an old `master`?

lasagne/tests/test_objectives.py
```diff
@@ -1,65 +1,228 @@
...
+    #     # the `input` argument we provided, plus the extra positional and
+    #     # keyword arguments.
+    #     input_layer.get_output.assert_called_with(input, arg1, kwarg1=kwarg1)
+    #     network_output = input_layer.get_output.return_value
+    #
+    #     # The `network_output` and `target` are fed into the loss
+    #     # function:
+    #     loss_function.assert_called_with(network_output, target, mask)
+    #     loss_function.return_value.__mul__.assert_called_with(mask)
+    #     assert result == loss_function.return_value.__mul__.return_value.sum.return_value
```
f0k: I'd prefer this to be removed, we don't want to start collecting dead code as Theano does :)

f0k:
Thanks for updating! Looks good to me in principle. If it's an easy merge anyway, it would be helpful if you could rebase onto master so Travis runs. In git, that would be:
```bash
git fetch upstream
git rebase upstream/master
# resolve conflicts if needed, use `git add` to mark them resolved, then `git rebase --continue`
git push --force origin
```

Britefury:
Alright, done.

lasagne/objectives.py
```diff
@@ -12,74 +14,136 @@ def mse(x, t):
...
-            a 2D tensor that gives the class probabilities in its rows, OR
-            a 1D integer array that gives the class index of each sample
-            (the position of the 1 in the row in a 1-of-N encoding, or
-            1-hot encoding),
+        :parameters:
+            - input : (default `None`) an expression that results in the
+                input data that is passed to the network
+            - target : (default `None`) an expression that results in the
+                desired output that the network is being trained to generate
+                given the input
```
f0k: The `aggregation` parameter needs to be documented (can be copied from the `MaskedObjective` below)

lasagne/objectives.py
```diff
@@ -12,74 +14,136 @@ def mse(x, t):
...
-    :parameters:
-        - x : predicted values
-        - t : target values
+    """
+    Training objective
 
-    :returns:
-        - output : the mean binary cross entropy across all dimensions
+    The  `get_loss` method returns an expression that can be used for
+    training with a gradient descent approach.
```
f0k: That's a little misleading, it's not limited to gradient descent. How about: "The `get_loss` method returns a cost expression useful for training or evaluating a network." ?

f0k:
> Alright, done.

Thank you! I only have three minor comments -- you can either address them yourself or Sander can fix them while merging.
@benanne: I'd be fine with merging this. We can still think about refactoring this (to not use classes) in a separate PR.

benanne:
Agreed, this is great work :) Thanks for the extensive collection of tests as well! Let me know if you want to address f0k's remarks, if you don't I'll try to take care of it this weekend or early next week at the latest.

I will also make sure to make a post on the mailing list about this when I merge this, since it involves changes to an existing interface which may break some code.

Britefury:
I have made the doc-string improvements suggested by f0k. Thanks for your compliments. :)

examples/mnist.py
f0k: This one is still open (reinserting the blank). Sander can take care of this while merging.
@dnouri: I'm not a pep8 expert, but shouldn't this have been caught by Travis with your new setup?

dnouri: @f0k Looks like `examples/` aren't PEP8-checked right now.

benanne: Do they need to be explicitly included in a config file somewhere? It would be great if all *.py files were checked.

dnouri: The problem with the examples is that some of them fail with `ImportError` when included for pytest.  So adding the pep8 check would likely come at the expense of making the examples less comprehensible (with a weird `pytest.skiptest` thingy or something).  I'd see we use our eyeballs for the examples; looks like we have some pretty good ones on board. ;-)

benanne:
Great! I will merge this tomorrow if there are no further comments.

f0k:
+1 for merging!

lasagne/tests/test_objectives.py
```diff
@@ -1,65 +1,210 @@
...
+                                                output, target_2d, mask_1hot,
+                                                aggregation='normalized_sum')
+        assert result_with_mask.eval() == 5.0/2.0
+
+    def test_objective(self):
+        from lasagne.objectives import Objective
+
+        input_layer = mock.Mock()
+        loss_function = mock.Mock()
+        input, target, arg1, kwarg1 = (object(),) * 4
```
dnouri: This will give you four identical objects, which isn't the intention here, and makes the tests below that `assert_called_with` much weaker.

Britefury: Fixed.

benanne:
Merging, thanks again!

PR #129, opened 2015-02-07
Flatten pixels layer

Britefury:
`FlattenPixelsLayer` reshapes a (batch,channel,height,width) array into (batch_height_width,channel) array for per pixel softmax. Its main use is for segmentation in a block-by-block basis. 

Segmentation can be performed without the use of `FlattenPixelsLayer` in a sliding window fashion by extracting a window surrounding each output pixel. This however is very slow, since the entire net must be evaluated once for each output pixel, with some parallelism achieved through the use of mini-batches. Significant performance improvements can be realised by segmenting complete images or blocks of images. `FlattenPixelsLayer` has one downside, in that the batch, height and width dimensions are collapsed into one, meaning that you have to reshape the target values at training time and reshape the network output when in use. Other than that, its still quite usable. :)

### DIFF
--- a/lasagne/layers/shape.py
+++ b/lasagne/layers/shape.py
@@ -9,6 +9,7 @@ from .base import Layer
 
 __all__ = [
     "FlattenLayer",
+    "FlattenPixelsLayer",
     "flatten",
     "PadLayer",
     "pad",
@@ -24,6 +25,14 @@ class FlattenLayer(Layer):
         return input.flatten(2)
 
 flatten = FlattenLayer # shortcut
+
+
+class FlattenPixelsLayer(Layer):
+    def get_output_shape_for(self, input_shape):
+        return (input_shape[0] * int(np.prod(input_shape[2:])), input_shape[1])
+
+    def get_output_for(self, input, *args, **kwargs):
+        return input.dimshuffle(1,0,2,3).flatten(2).dimshuffle(1,0)
 
 
 class PadLayer(Layer):


dnouri:
I can't say anything about the usefulness of this.  Nevertheless, I think the `FlattenPixelsLayer` class would benefit from a docstring that explains the motivation.

benanne:
This does seem like a very specific use case for it to be included in the main library...
Its functionality is not covered by the upcoming `ReshapeLayer` (#94) because it involves dimshuffling, but can we think of any other use cases for this?

Is there a more general solution we could provide as part of the library that covers a larger number of use cases and is not too cumbersome to use? Since we're adding `ReshapeLayer`, we may as well add a `DimshuffleLayer` while we're at it, which would cover this - although it would require a sequence of 3 layer instances (dimshuffle -> reshape -> dimshuffle) so that is definitely a lot more cumbersome. Then again, maybe this is not an extremely important use case to cover (we should make the common use cases easy and the uncommon ones possible).

craffel:
> Since we're adding ReshapeLayer, we may as well add a DimshuffleLayer while we're at it, which would cover this

Huh, I thought we already had one.  But yeah, I think we should have a `ReshapeLayer` and a `DimshuffleLayer`.  Then, defining a macro for creating "dimshuffle -> reshape -> dimshuffle" (which is what this layer does) I think would be totally reasonable to leave up to the user.

Britefury:
I would like to request that this PR be dropped; I have implemented `DimShuffleLayer` within [PR134](https://github.com/benanne/Lasagne/pull/134). I can add flatten pixels as a macro/function if you like.

benanne:
Sure, that would be great :)

PR #130, opened 2015-02-07
Py3k

dnouri:
Intended to supersede #119, this pull request attempts to add Python 3 compatibility with fewer changes to the code.

I think it's fair to say with this, that we have experimental Python 3 support.  It's maybe not bug-free, given that we don't have 100% coverage, but once Python 3 users start using this (hopefully I will soon), we might be able to declare this stable.

I haven't included a `_compat.py` module for now. The `mnist.py` example uses its own little `if PY2` block to do the right imports.  Why?  Because I think it might be common that people look at the examples and copy code as a starting point for their own experiments.  For those users, we wouldn't want them to have imports from a private Lasagne compatibility module (which I think `_compat.py` should be; we don't want to support a compat API).  That's of course once we start having a compatibility module; we don't have with this pull request.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
+# Virtualenv
 bin/
 include/
 src/
+
+# Dataset used in examples
 mnist.pkl.gz
 
 # Byte-compiled / optimized / DLL files
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,8 @@
+bin/
+include/
+src/
+mnist.pkl.gz
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -1,10 +1,20 @@
 from __future__ import print_function
 
-import cPickle as pickle
 import gzip
 import itertools
-import urllib
+import pickle
 import os
+import sys
+
+PY2 = sys.version_info[0] == 2
+
+if PY2:
+    from urllib import urlretrieve
+    pickle_load = lambda f, encoding: pickle.load(f)
+else:
+    from urllib.request import urlretrieve
+    pickle_load = lambda f, encoding: pickle.load(f, encoding=encoding)
+
 
 import numpy as np
 import lasagne
@@ -23,15 +33,12 @@ MOMENTUM = 0.9
 
 
 def _load_data(url=DATA_URL, filename=DATA_FILENAME):
-    try:
-        with gzip.open(filename, 'rb') as f:
-            data = pickle.load(f)
-    except (IOError, EOFError):
+    if not os.path.exists(filename):
         print("Downloading MNIST")
-        urllib.urlretrieve(url, filename)
-        with gzip.open(filename, 'rb') as f:
-            data = pickle.load(f)
-    return data
+        urlretrieve(url, filename)
+
+    with gzip.open(filename, 'rb') as f:
+        return pickle_load(f, encoding='latin-1')
 
 
 def load_data():
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 python:
   - "2.7"
+  - "3.4"
 before_install:
   - sudo apt-get update -qq
   - sudo apt-get install -qq libatlas-dev libatlas-base-dev liblapack-dev gfortran
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,17 @@
 language: python
 python:
   - "2.7"
-virtualenv:
-  system_site_packages: true
 before_install:
-  - sudo apt-get install -qq python-numpy python-scipy libatlas3gf-base libatlas-dev liblapack-dev gfortran
+  - sudo apt-get update -qq
+  - sudo apt-get install -qq libatlas-dev libatlas-base-dev liblapack-dev gfortran
+  - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
+  - chmod +x miniconda.sh
+  - ./miniconda.sh -b
+  - export PATH=/home/travis/miniconda/bin:$PATH
+  - conda update --yes conda
 install:
-  - pip install -r requirements.txt
-  - python setup.py dev
+  - travis_retry conda install --yes python=$TRAVIS_PYTHON_VERSION pip numpy scipy
+  - travis_retry pip install -r requirements.txt
+  - travis_retry python setup.py dev
 script: travis_wait py.test --runslow
 cache: apt
--- a/lasagne/__init__.py
+++ b/lasagne/__init__.py
@@ -2,10 +2,10 @@
 Tools to train neural nets in Theano
 """
 
-import nonlinearities
-import init
-import layers
-import objectives
-import regularization
-import updates
-import utils
+from . import nonlinearities
+from . import init
+from . import layers
+from . import objectives
+from . import regularization
+from . import updates
+from . import utils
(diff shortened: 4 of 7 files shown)


benanne:
I agree, any modifications required for Python 2/3 compatiblity of the example code should be handled in those files themselves. Does the 'encoding' stuff solve the Python 3 pickle incompatibility issue? If so, that's great :)

One thing I'm not sure about is the changes to the Travis config - do these add Python 3 as a target? I skimmed over the discussion in the other PR, but I'm guessing the miniconda stuff is necessary to support both versions? Is there any risk of this breaking Travis? (I'm assuming you were not able to test locally whether the config is correct - but maybe you've used this/seen this used for other repositories).

dnouri:
On 02/07/2015 03:02 PM, Sander Dieleman wrote:
> I agree, any modifications required for Python 2/3 compatiblity of the
> example code should be handled in those files themselves. Does the
> 'encoding' stuff solve the Python 3 pickle incompatibility issue? If so,
> that's great :)

Yes, that mnist pickle issue is fixed with the extra kwarg (which only
exists for Python 3).

> One thing I'm not sure about is the changes to the Travis config - do
> these add Python 3 as a target? I skimmed over the discussion in the
> other PR, but I'm guessing the miniconda stuff is necessary to support
> both versions? Is there any risk of this breaking Travis? (I'm assuming
> you were not able to test locally whether the config is correct - but
> maybe you've used this/seen this used for other repositories).

Yes, mini/conda is necessary to be able to install numpy and scipy
reasonably fast for Python 3.4.  These two aren't available with apt-get
on Travis with Python 3.4.

Travis already passed OK with these changes, see here:
https://travis-ci.org/benanne/Lasagne/builds/49854761

examples/mnist.py
```diff
@@ -23,15 +33,12 @@
 
 
 def _load_data(url=DATA_URL, filename=DATA_FILENAME):
-    try:
-        with gzip.open(filename, 'rb') as f:
-            data = pickle.load(f)
-    except (IOError, EOFError):
+    if not os.path.exists(filename):
```
f0k: The `try..except` block had the advantage that it also caught an incomplete download.

.gitignore
```diff
@@ -1,3 +1,8 @@
+bin/
+include/
+src/
+mnist.pkl.gz
+
```
f0k: The latter should probably come with a comment `# Dataset used in examples`. But what are the first three for?

examples/mnist.py
```diff
@@ -1,10 +1,20 @@
...
+
+PY2 = sys.version_info[0] == 2
+
+if PY2:
+    from urllib import urlretrieve
+    pickle_load = lambda f, encoding: pickle.load(f)
+else:
+    from urllib.request import urlretrieve
+    pickle_load = lambda f, encoding: pickle.load(f, encoding=encoding)
+
```
f0k: If we're doing the `PY2` conditional anyway, can't we `import cPickle as pickle` in Python 2? That's a little more efficient, so it's a little more environmentally friendly.

f0k:
Looks good to me! It's basically #119 just without the `six` dependency, which feels a lot easier.

examples/mnist.py
dnouri: I kinda tended towards fewer lines.  This little benchmark suggests that the bottleneck with loading isn't with cPickle versus pickle anyway:
```
>>> from timeit import timeit
>>> timeit("pickle.load(gzip.open('mnist.pkl.gz', 'rb'))", setup="import cPickle as pickle; import gzip", number=10)
5.9051759243011475
>>> timeit("pickle.load(gzip.open('mnist.pkl.gz', 'rb'))", setup="import pickle; import gzip", number=10)
5.526721000671387
```
(Don't ask me!)

examples/mnist.py
dnouri: With incomplete downloads, the user will now be greeted with something like:
``IOError: CRC check failed 0x15161f5c != 0x9ad4c9afL``

Seems like a friendly enough error to me.

.gitignore
dnouri: Done!

examples/mnist.py
ebattenberg: Hmm.  I'm not sure I understand the advantage of throwing an error (which most users won't easily understand) over just redownloading.  It seems to me that making the example as seamless as possible is desired.  Especially since I ran into that issue myself after killing the script because I thought the download was taking too long.  I agree that catching EOFError can be remove because encountering an empty file is unlikely to ever happen.  I guess it depends on if the goal is shorter code (3 fewer lines) or an example script that is less likely to break.  Also, try/except is more pythonic (apparently). ;)

examples/mnist.py
ebattenberg: Benchmarking results can be influenced by the order of the routines when both routines access the same memory or disk area.  This is usually called "cache warming".

dnouri: Not a case of disk caching here.  Try for yourself, run the lines in reverse order, you'll get consistent results.

ebattenberg:
BTW.  Nice job fixing the encoding problem with the mnist pickle.  :)

examples/mnist.py
dnouri: Not sure "most users" won't understand what a checksum error is and a traceback that points right at the line that opens the file in question.  I guess we might be underestimating our users.  :-)

Anyway, my goal is an example that's concise; fine if it doesn't cover every conceivable edge case.  And since you added that `print("Downloading MNIST")` line I do hope that users are less inclined now to kill the process in the middle.

examples/mnist.py
ebattenberg: You're right that `pickle` is faster after the cache has been warmed.  But you can understand my suspicion due to the results from these first two (un-warmed) lines:

```
>>> from timeit import timeit
>>> timeit("pickle.load(gzip.open('mnist.pkl.gz', 'rb'))", setup="import pickle; import gzip", number=10)
5.902479887008667
>>> timeit("pickle.load(gzip.open('mnist.pkl.gz', 'rb'))", setup="import cPickle as pickle; import gzip", number=10)
5.870376825332642
>>> timeit("pickle.load(gzip.open('mnist.pkl.gz', 'rb'))", setup="import pickle; import gzip", number=10)
5.449158191680908
>>> timeit("pickle.load(gzip.open('mnist.pkl.gz', 'rb'))", setup="import cPickle as pickle; import gzip", number=10)
5.793623924255371
```

However, I guess that warming shouldn't even be an issue when I and [others](http://justinfx.com/2012/07/25/python-2-7-3-serializer-speed-comparisons/) have observed orders of magnitude speed differences between the two.  I wonder what's going on here.

examples/mnist.py
ebattenberg: >Not sure "most users" won't understand what a checksum error is and a traceback that points right at the line that opens the file in question. I guess we might be underestimating our users. :-)

Just don't be surprised when it pops up on the mailing list or SO. ;)

> And since you added that print("Downloading MNIST") line I do hope that users are less inclined now to kill the process in the middle.

That's what I was hoping.  That's why I killed the process in the first place! :)

benanne:
Nice work all of you :) Will merge this tomorrow if there are no further comments.

benanne:
Merging!

examples/mnist.py
f0k: > I wonder what's going on here.

Must be because of this pickle's content; it's mostly numpy arrays that probably won't make much of a difference between `pickle` and `cPickle`. It's still a bit of a mystery, but now that `pickle` is faster for this example, it's more environmentally friendly and I'm fine with keeping it!

PR #131, opened 2015-02-08
ReshapeLayer: ensure Python 3 compatibility

f0k:
This removes the usage of `reduce` from the `ReshapeLayer` implementation (noted here: https://github.com/benanne/Lasagne/pull/94#issuecomment-73416833) so it's compatible with Python 3.


benanne:
Thanks for the quick fix :)

PR #132, opened 2015-02-09
Cross-channel Local response normalization

dimatura:
This is a first stab at a 2D cross-channel LRN layer. Lines 52 through 61 are pretty much verbatim from `pylearn2`, except for some variable renaming. I added some comments and the `lasagne` wrapping. It gets the job done but perhaps it could be generalized to handle `cb01` as well as `bc01`, though in the former case `cuda-convnet` could be preferable. Maybe a 1D version would also be desirable, though I don't know if anyone does this for 1D data.

### DIFF
--- a/lasagne/__init__.py
+++ b/lasagne/__init__.py
@@ -9,4 +9,3 @@ from . import objectives
 from . import regularization
 from . import updates
 from . import utils
-from . import normalization
--- a/lasagne/__init__.py
+++ b/lasagne/__init__.py
@@ -9,3 +9,4 @@ from . import objectives
 from . import regularization
 from . import updates
 from . import utils
+from . import normalization
--- a/lasagne/layers/__init__.py
+++ b/lasagne/layers/__init__.py
@@ -7,3 +7,4 @@ from .conv import *
 from .pool import *
 from .shape import *
 from .merge import *
+from .normalization import *
--- a/lasagne/layers/normalization.py
+++ b/lasagne/layers/normalization.py
@@ -1,6 +1,10 @@
 # -*- coding: utf-8 -*-
 
 """
+This file contains code from pylearn2, which is convered by the following
+license:
+
+
 Copyright (c) 2011--2014, Université de Montréal
 All rights reserved.
 
--- a/lasagne/layers/normalization.py
+++ b/lasagne/layers/normalization.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 """
+
 This file contains code from pylearn2, which is convered by the following
 license:
 
@@ -19,8 +20,8 @@ modification, are permitted provided that the following conditions are met:
    and/or other materials provided with the distribution.
 
 3. Neither the name of the copyright holder nor the names of its contributors
-   may be used to endorse or promote products derived from this software without
-   specific prior written permission.
+   may be used to endorse or promote products derived from this software
+   without specific prior written permission.
 
 THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
 ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
@@ -48,6 +49,7 @@ from .base import Layer
 __all__ = [
     "LocalResponseNormalization2DLayer",
 ]
+
 
 class LocalResponseNormalization2DLayer(Layer):
     """
@@ -79,7 +81,8 @@ class LocalResponseNormalization2DLayer(Layer):
             - beta: see equation above
             - n: number of adjacent channels to normalize over.
         """
-        super(LocalResponseNormalization2DLayer, self).__init__(incoming, **kwargs)
+        super(LocalResponseNormalization2DLayer, self).__init__(incoming,
+                                                                **kwargs)
         self.alpha = alpha
         self.k = k
         self.beta = beta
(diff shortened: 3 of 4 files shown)


benanne:
Thanks! A few remarks:

* I don't think we need to support c01b anywhere except for the `cuda_convnet` wrappers, and only because there is a potential performance advantage there (although in my experience it is negligible). Let's stick to `bc01` unless there is a good reason not to :)

* Since some of this code is copied from pylearn2, we will need to sort out licensing before we merge this. What's the best way to handle this?

* It looks like the docstring is not in our (current) standard format, could you take a look at this?

* A couple of tests would be a great addition :)

f0k:
> Since some of this code is copied from pylearn2, we will need to sort out licensing before we merge this. What's the best way to handle this?

As [previously suggested](https://github.com/benanne/Lasagne/issues/121#issuecomment-72841799), we should include pylearn2's license text with the source code. I'd suggest to add it to the source file as a block comment in the beginning (along with a statement that this file contains code of pylearn2 covered by the following license), and change the note in the class docstring to state that this code is ported from pylearn2, covered by BSD, and "for detailed license information, please see the source file".

benanne:
Are there any consequences of this that we should consider? Are people less likely to use this library if some parts of the code have a different license than others? Is pylearn2's version of the BSD license more or less restrictive than MIT? If it is more restrictive, does this have any inadvertent effects on the rest of the codebase?

dimatura:
Thanks for the feedback, I agree with the comments. Re: license, yes, we could just paste the `pylearn2` license on this one file. It's a little bit annoying that just a handful lines could cause this problem, and I feel somewhat tempted to just change the implementation, but there's not much wiggle room. 

I can't say really regarding the effects on usage, I've never worried much about licensing.

f0k:
> If it is more restrictive, does this have any inadvertent effects on the rest of the codebase?

It's about as free as MIT, i.e., you can still use the code for commercial purposes and everything. Have a look yourself, [I've linked both licenses here](https://github.com/benanne/Lasagne/issues/121#issuecomment-72841799) and they're both pretty short. It also doesn't magically affect the rest of the codebase.

benanne:
ok, then I think it's safe to do it this way and I agree we should include the license in the source code.

run2:
I tried this out with nolearn. Couple of observations
1) Can you rename your incoming argument to input_layer (that is how the other lasagne layers have it)
2) Probably the  input_shape if None can be got as input_shape = self.input_layer.get_output_shape() - that will make it complaint with nolearn too
3) The normalization (after first and second conv layers) is doubling my epoch time from 177 seconds to 378 seconds. Note I am not using cudacovnet (as I have non square images and kernels)

benanne:
1) actually, we recently renamed it to `incoming` everywhere, because the input argument can now be both a `Layer` instance and a shape tuple. So `input_layer` didn't really make sense anymore :)

2) Not sure what you mean by this - one thing I noticed is that this layer's `get_output_shape_for()` just returns the `input_shape` because this doesn't change the shape. This implementation is unnecessary, it's what the `Layer` class does by default. So this method does not need to be overridden (and it should probably be removed for consistency).

3) It's no surprise that this implementation of response normalization is going to be slow, it involves a lot of fairly small operations that don't have a lot of inherent parallelism (there's a for loop in there!). I don't think there's an easy way to make this faster in Theano though, apart from implementing a custom op.

run2:
1) ok. I am still using a previous version of lasagne and nolearn hence could not get it working without that change.
2) Sorry - point taken back. The problem was because I am using a nolearn NeuralNetwork where I pass None as the first param of my 4D input shape. This causes a problem with the b as b becomes None. To get over this I am currently sending a hard coded batch size - that solves it.
3) Ok - got it. Thanks again.

benanne:
2) is that a problem with this particular layer? Because we do want to support `None` in shapes wherever it's possible (especially for the batch size). If the code can easily be modified to support this, we should do it.

run2:
Yes - it is a problem with this layer. I think it can check and get the shape from input - like it is done for DropoutLayer (like 34 , 35). I can check and confirm or if you are sure that's right give me a shout

benanne:
Ah, I see it. I guess we just need to copy these two lines from `DropoutLayer`: https://github.com/benanne/Lasagne/blob/master/lasagne/layers/noise.py#L35

run2:
Yes - bang on!

dimatura:
Ah! Sorry I haven't responded, a bit swamped right now. Thanks for pointing out these problems, I'll add that fix tomorrow.

dimatura:
Ok, added license and did `shape` thing. @run2 Regarding computation time, if that's an issue consider leaving the LRN out. It's still not understood when it really helps; OverFeat doesn't use and it works fine.

benanne:
Looks like the é in Montréal is making Travis fail for Python 2 :p

f0k:
Then we need a `# -*- coding: utf-8 -*-` line in the beginning. We could also swap the é for an e, but then we are technically violating the license :)

benanne:
I'll try to finish this up and merge it this weekend or early next week.

dimatura:
Hi, sorry for going MIA. The paper deadline was yesterday (with Lasagne! But I need the DOI to cite properly ;) so I can do it this weekend if you haven't already.

benanne:
Oh, no worries :) Be my guest! If you'd rather I take care of it just let me know.

benanne:
Looks good, I will merge this tomorrow if there are no further comments.

dimatura:
The test module is more of a placeholder right now, I'm adding a bit more.

dimatura:
Okay, added the test from pylearn2, also with the license for that file. First time I use pytest so it's nothing too fancy.

benanne:
Thanks for the tests as well! I'll wait a while and merge if there are no further comments.

StevenReitsma:
There's a small spelling error in `lasagne/layers/normalization.py`: `convered` instead of `covered`.

dimatura:
fixed!

benanne:
Awesome, time to merge at last :)

Issue #26, opened 2014-09-12
Documentation


benanne:
There is no documentation at the moment - we should probably get started with this as soon as possible. We talked about this previously and I believe we were thinking of using [Sphinx](http://sphinx-doc.org/). Are there any alternatives we should consider? Does anybody have any experience setting up Sphinx?

EDIT: relevant reading: http://www.jeffknupp.com/blog/2013/08/16/open-sourcing-a-python-project-the-right-way/

benanne:
I just added some more docs (input layer and dense layer) and changed the documentation theme because I thought the previous one looked very messy. I stole the one that librosa uses (sphinx-bootstrap-theme).

I actually meant to do this on a branch + pull request, because I believe the theme is an additional dependency. But apparently I forgot to check out the branch or something... so if you can't build the docs right now, that would be my fault. Sorry!

f0k:
`<offtopic>`
> I actually meant to do this on a branch + pull request, because I believe the theme is an additional dependency. But apparently I forgot to check out the branch or something...

It's too late now, but if stuff like that happens, you can always do something like
```bash
git checkout -b wrong-commit  # just to have a backup
git checkout master
git reset --hard HEAD^  # or the SHA of the commit you want to revert to
git push --force origin master
```
to undo what you pushed to the server. If nobody pulled in between, nobody will notice. Maybe you know this anyway, but don't be too shy about force-pushing.
`</offtopic>`

benanne:
I did not know that :) Thanks for the hint!

benanne:
The current layout on http://lasagne.readthedocs.org/ looks a little messy, so we should probably experiment with this a bit (maybe reduce the number of levels of indentation by making the subtitles like "usage" and "parameters" be headers instead.

As a side note it looks like Brian McFee is actually converting the librosa docs to the numpy format: https://github.com/bmcfee/librosa/issues/145

JackKelly:
Really excited about Lasagne!

One small tweak: it might be worth linking to http://lasagne.readthedocs.org from Lasagne's front page on github.  For example, link to http://lasagne.readthedocs.org from the "Website" attribute on the project's github front page.

benanne:
Good call. Done :)

JackKelly:
perfect, thank you ;)

benanne:
[...]

PR #134, opened 2015-02-09
DimShuffle layer

Britefury:
Implemented DimShuffleLayer in lasagne.layers.shape. Implemented tests.

### DIFF
--- a/lasagne/layers/shape.py
+++ b/lasagne/layers/shape.py
@@ -139,6 +139,48 @@ class ReshapeLayer(Layer):
 reshape = ReshapeLayer # shortcut
 
 
+class DimShuffleLayer(Layer):
+    def __init__(self, incoming, pattern):
+        super(DimShuffleLayer, self).__init__(incoming)
+        self.pattern = pattern
+
+    def get_output_shape_for(self, input_shape):
+        # Build output shape while keeping track of the dimensions that we are
+        # attempting to collapse, so we can ensure that they are broadcastable
+        output_shape = []
+        dims_used = [False] * len(input_shape)
+        for p in self.pattern:
+            if isinstance(p, (int, long)):
+                if p < 0  or  p >= len(input_shape):
+                    raise ValueError, "pattern contains {0}, but input shape has " \
+                        "{1} dimensions only".format(p, len(input_shape))
+                # Dimension p
+                if dims_used[p]:
+                    raise ValueError, "pattern contains dimension {0} more "\
+                            "than once".format(p)
+                o = input_shape[p]
+                dims_used[p] = True
+            elif p == 'x':
+                # Broadcast; will be of size 1
+                o = 1
+            else:
+                raise ValueError, "pattern should only contain dimension" \
+                                  "indices or 'x', not {0}".format(p)
+            output_shape.append(o)
+
+        for i, (dim_size, used) in enumerate(zip(input_shape, dims_used)):
+            if not used and dim_size != 1:
+                raise ValueError, "pattern attempted to collapse dimension " \
+                    "{0} of size {1}; dimensions with size != 1 are not" \
+                    "broadcastable and cannot be " \
+                    "collapsed".format(i, dim_size)
+
+        return tuple(output_shape)
+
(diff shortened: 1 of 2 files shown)


f0k:
Thanks, looks good superficially (didn't check the implementation in detail)! It's just missing a docstring.

lasagne/layers/shape.py
```diff
@@ -139,6 +139,62 @@ def get_output_for(self, input, *args, **kwargs):
...
+            elif p == 'x':
+                # Broadcast; will be of size 1
+                o = 1
+            else:
+                raise RuntimeError("invalid pattern entry, should have "
+                    "caught in the constructor")
+            output_shape.append(o)
+
+        for i, (dim_size, used) in enumerate(zip(input_shape, dims_used)):
+            if not used and dim_size != 1:
```
f0k: `dim_size` could also be `None`. In this case, we cannot know whether it's okay to collapse it and should just assume that it is.

benanne:
Looks great! The tests look good as well. I have just a few minor comments:

* Since the Theano function this wraps is called `dimshuffle` and not `dim_shuffle`, I think it would make more sense to call this layer `DimshuffleLayer` (lowercase s). Seems a bit more consistent that way. I don't know, I guess I have API-OCD :p 

* To support layer naming (and possibly additional features in the future), make sure to add `**kwargs` to `__init__()` and pass it to the superclass's `__init__()` method.

f0k:
> To support layer naming (and possibly additional features in the future), [...]

Ah, that's missing in `ReshapeLayer` as well... Sander, could you fix this directly on master, please?

benanne:
will do!

Britefury:
Should all be done now (`**kwargs`, `DimshuffleLayer` and `None` in input shape).

benanne:
Awesome! Will merge this tomorrow if there are no further comments.

f0k:
It's still missing the docstring! It should have a class docstring explaining what the purpose of the layer is and what the constructor arguments mean. (Although I'm not sure if the latter shouldn't be in the constructor docstring instead? I've seen both here in Lasagne.)

benanne:
Whoops, good catch! Yeah, we need to sort that out. What's the preferred way to do it? Are there any advantages/disadvantages to putting it in the class / constructor docstring?

Also, this is probably another `Layer` that deserves a shortcut, because people typically don't see a shuffling operation as a separate layer. Since we already have `lasagne.layers.reshape` and such, could you add the following line below the class definition to create a shortcut for it?

```
dimshuffle = DimshuffleLayer
```

Thanks!

craffel:
My impression/convention is that the class docstring should document the class and its members, and the `__init__` docstring should document the init method and its parameters.

benanne:
But in practice, where do you draw the line? For example, sometimes it's hard to describe what a layer does without referring to some of its parameters by name. I don't think it's always that clear-cut what should go where.

craffel:
> But in practice, where do you draw the line? For example, sometimes it's hard to describe what a layer does without referring to some of its parameters by name.

The line is members (i.e. `foo.b`) vs. initialization parameters (i.e. `foo = Foo(b)`).  For example, the class docstring for `lasagne.layers.Layer` should descibe what `input_layer` and `input_shape` are, while the docstring for its initializer should describe what `incoming` is.  Or, in a `DenseLayer`, the class docstring for the member `W` should describe what a weight matrix is, while the initializer docstring for `W` should describe the suitable methods for initializing it.  See https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#documenting-classes for some more info.

Britefury:
The doc-string and shortcut are done. I also added `DimshuffleLayer` and the shortcut to `__all__`, something else I omitted! :)

benanne:
@craffel: that makes sense, thanks. We should keep this in mind for the future (gonna tag #26 here so this shows up there).

@Britefury: looks great, thanks! Second attempt:I'll merge this tomorrow if there are no further comments.

benanne:
Merging, thanks!

Issue #84, opened 2015-01-02
Constrained Optimization as a Regularizer


cancan101:
Consider adding constrained optimization as a regularizer.

From: https://blog.safaribooksonline.com/2014/02/13/constrained-optimization-pylearn2/ one of the regularizers that pyearn2 supports is `col_norm`.

See also: http://deeplearning.net/software/pylearn2/library/models.html?highlight=linear#pylearn2.models.mlp.Linear.

benanne:
Good idea. I've never had much luck with this approach in the past, but I might end up implementing it myself anyway, eventually. Pull requests are welcome, at any rate :)

ebattenberg:
I've used weight norm constraints in the past.  I think they can be useful if you have some filters (or rows/cols) in the lower layers that tend to die with weight decay.  

I have some Theano code that does this, and I'd be willing to add it in a PR.  

Since just setting a maximum value for the norm by hand can be a bit tricky, I tend to use a multiple of the initial norm of the layer as the argument that ends up setting the constraint.  For example, if the incoming weights are initialized with standard deviation `std`, then the expected incoming 2-norm of an incoming weight vector is `std*sqrt(fan_in)`.  The argument I use is a multiple of this value (something between 1x and ~5x) depending on the nonlinearity used and how close to the output layer we are.  

Of course then the weight norm constraint would need information about the initialization values, but I guess you could pass the initialization type information through.  

Does that sound a reasonable way to hyperparametrize this?  I can also allow just the exact constraint as well, but I don't know if that is as meaningful to the user.  I hate having to recompute the constraints when layer sizes change.

f0k:
> Does that sound like a reasonable way to hyperparametrize this?

Sounds interesting.

> Of course then the weight norm constraint would need information about the initialization values, but I guess you could pass the initialization type information through.

If the regularization is applied directly to shared variables rather than more complex expressions, it could just check the values of said variables (e.g., their initial column norms) on construction. For more complex expressions, that would still be possible via an `eval()` call. Seems easier and more flexible to me than passing on the weight initializer.

> I can also allow just the exact constraint as well

It should probably allow both, maybe as `max_col_norm=somevalue` and `max_col_norm='auto'`.

We should probably rework the regularization code first, though (see #14).

ebattenberg:
> It should probably allow both, maybe as `max_col_norm=somevalue` and `max_col_norm='auto'.`

[...]

PR #135, opened 2015-02-13
Weight Norm Constraints

ebattenberg:
Here's an attempt at weight norm constraints as we discussed in #84.  I also did a little PEP8 cleanup in update.py (mostly line lengths stuff).

Unless I totally overlooked an easy assumption, it seems like we have to handle all the different types of parameter dimensionalities differently (not quite as clean as in the default Uniform initializer).  So, it makes assumptions about the mean of the dimensions in a 2D and 4D parameter array.  

Let me know what you guys think.  I'll add better docs later.

### DIFF
--- a/lasagne/updates.py
+++ b/lasagne/updates.py
@@ -140,37 +140,20 @@ def adadelta(loss, all_params, learning_rate=1.0, rho=0.95, epsilon=1e-6):
     return updates
 
 
-def norm_constraint(tensor_var, param=None, abs_max=None, rel_max=None,
-                    norm_axes=None, epsilon=1e-7):
+def norm_constraint(tensor_var, max_norm, norm_axes=None, epsilon=1e-7):
     '''
     Max weight norm constraints and gradient clipping
 
     This takes a TensorVariable and rescales it so that incoming weight
-    norms are below a specified constraint value.  The constraint value is
-    the lesser of `abs_max` (if provided) and `rel_max` (if provided) times
-    the average norm of the values originally stored in `param`.
-
-    Vectors violating the constraint are rescaled so that they are
-    within the allowed range.
-
+    norms are below a specified constraint value. Vectors violating the
+    constraint are rescaled so that they are within the allowed range.
 
     :parameters:
         - tensor_var : TensorVariable
             Theano expression for update, gradient, or other quantity.
-        - param : TheanoSharedVariable
-            Shared variable containing initial parameter values.
-            (Required if `rel_max` is used)
-        - abs_max : scalar
-            This value sets the absolute maximum value allowed for the
-            incoming weight vectors after the update.
-            (Optional)
-        - rel_max : scalar
-            This value sets the relative maximum value allowed for the
-            incoming weight vectors after the update.  The relative maximum
-            is computed by multiplying this value by the average incoming
-            weight vector norm for the values originally stored in the
-            `param` shared variable.
-            (Optional)
+        - max_norm : scalar
+            This value sets the maximum allowed value of any norm in
+            `tensor_var`.
         - norm_axes : sequence (list or tuple)
             The axes over which to compute the norm.  This overrides the
             default norm axes defined for the number of dimensions
@@ -187,22 +170,17 @@ def norm_constraint(tensor_var, param=None, abs_max=None, rel_max=None,
             that violate the specified constraints.
 
     :usage:
-        >>> # Weight norm constraints
-        >>> from collections import OrderedDict
-        >>> updates = nesterov_momentum(loss, all_params, learning_rate)
-        >>> updates = OrderedDict(updates)
-        >>> # Absolute constraint of 5.0 for `param1`
-        >>> updates[param1] = norm_constraint(updates[param1], abs_max=5.0)
-        >>> # Relative constraint of 1.0 for `param2`
-        >>> updates[param2] = norm_constraint(updates[param2], rel_max=1.0)
-
-        >>> # Gradient clipping
-        >>> all_grads = theano.grad(loss, all_params)
-        >>> # Clip all gradients to 5.0
-        >>> updates = []
-        >>> for param, grad in zip(all_params, all_grads):
-        >>>     clipped_grad = norm_constraint(grad, abs_max=5.0)
-        >>>     updates.append((param, param - learning_rate * clipped_grad))
+        >>> param = theano.shared(
+        ...     np.random.randn(100, 200).astype(theano.config.floatX))
+        >>> update = param + 100
+        >>> update = norm_constraint(update, 10)
+        >>> func = theano.function([], [], updates=[(param, update)])
+        >>> # Apply constrained update
+        >>> _ = func()
+        >>> from lasagne.utils import compute_norms
+        >>> norms = compute_norms(param.get_value())
+        >>> np.isclose(np.max(norms), 10)
+        True
 
     :note:
         Right now this has predefined norm ops for:
@@ -212,23 +190,6 @@ def norm_constraint(tensor_var, param=None, abs_max=None, rel_max=None,
         For other uses, you can use the `norm_axes` argument.
     '''
 
-    constraint = None
-
-    if rel_max is not None:
-
-        if param is None:
-            raise ValueError("`rel_max` requires that `param` is provided")
-
-        # Compute average norm of `param`
-        vals = param.get_value()
-        avg_norm = np.mean(compute_norms(vals))
-        constraint = rel_max * avg_norm
-
-    if abs_max is not None:
-        constraint = abs_max if constraint is None else min(abs_max, constraint)
-
-    if constraint is None:
-        return tensor_var
 
     ndim = tensor_var.ndim
 
@@ -246,45 +207,10 @@ def norm_constraint(tensor_var, param=None, abs_max=None, rel_max=None,
 
     dtype = np.dtype(theano.config.floatX).type
     norms = T.sqrt(T.sum(T.sqr(tensor_var), axis=sum_over, keepdims=True))
-    target_norms = T.clip(norms, 0, dtype(constraint))
+    target_norms = T.clip(norms, 0, dtype(max_norm))
     constrained_output = \
         (tensor_var * (target_norms / (dtype(epsilon) + norms)))
 
     return constrained_output
 
 
-def compute_norms(array, norm_axes=None):
-    '''
-    Compute incoming weight vector norms.
-
-    :parameters:
-        - array : ndarray
-            Weight array
-        - norm_axes : sequence (list or tuple)
-            The axes over which to compute the norm.  This overrides the
-            default norm axes defined for the number of dimensions
-            in array
-            (Optional)
-
-    :returns:
-        - norms : 1D array
-            1D array of incoming weight vector norms.
-    '''
-
-    ndim = array.ndim
-
-    if norm_axes is not None:
-        sum_over = tuple(norm_axes)
-    elif ndim == 2:  # DenseLayer
-        sum_over = (0,)
-    elif ndim in [3, 4, 5]:  # Conv{1,2,3}DLayer
-        sum_over = tuple(range(1, ndim))
-    else:
-        raise ValueError(
-            "Unsupported tensor dimensionality {}."
-            "Must specify `norm_axes`".format(array.ndim)
-        )
-
-    norms = np.sqrt(np.sum(array**2, axis=sum_over))
-
-    return norms
--- a/lasagne/updates.py
+++ b/lasagne/updates.py
@@ -150,6 +150,33 @@ def norm_constraint(orig_update, param=None, abs_max=None, rel_max=None,
     the lesser of `abs_max` (if provided) and `rel_max` (if provided) times
     the average norm of the values originally stored in `param`.
 
+    Weight vectors violating the constraint are rescaled so that they are
+    within the allowed range.
+
+
+    :parameters:
+        - orig_update : TensorVariable
+            Theano expression for original update
+        - param : TheanoSharedVariable
+            Shared variable containing initial parameter values.
+            (Required if `rel_max` is used)
+        - abs_max : scalar
+            This value sets the absolute maximum value allowed for the
+            incoming weight vectors after the update.
+            (Optional)
+        - rel_max : scalar
+            This value sets the relative maximum value allowed for the
+            incoming weight vectors after the update.  The relative maximum
+            is computed by multiplying this value by the average incoming
+            weight vector norm for the values originally stored in the
+            `param` shared variable.
+
+    :returns:
+        - update : TensorVariable
+            Original update with rescaling applied to weight vectors
+            that violate the specified constraints.
+
+
     Right now this supports:
         * 2D param matrices with shape (input_dim, output_dim)
         * {3,4,5}D param tensors with shape
--- a/lasagne/updates.py
+++ b/lasagne/updates.py
@@ -6,7 +6,6 @@ import numpy as np
 
 import theano
 import theano.tensor as T
-from theano.config import floatX
 
 
 def sgd(loss, all_params, learning_rate):
@@ -25,7 +24,7 @@ def momentum(loss, all_params, learning_rate, momentum=0.9):
 
     for param_i, grad_i in zip(all_params, all_grads):
         mparam_i = theano.shared(np.zeros(param_i.get_value().shape,
-                                          dtype=floatX))
+                                          dtype=theano.config.floatX))
         v = momentum * mparam_i - learning_rate * grad_i
         updates.append((mparam_i, v))
         updates.append((param_i, param_i + v))
@@ -42,7 +41,7 @@ def nesterov_momentum(loss, all_params, learning_rate, momentum=0.9):
 
     for param_i, grad_i in zip(all_params, all_grads):
         mparam_i = theano.shared(np.zeros(param_i.get_value().shape,
-                                          dtype=floatX))
+                                          dtype=theano.config.floatX))
         v = momentum * mparam_i - learning_rate * grad_i  # new momemtum
         w = param_i + momentum * v - learning_rate * grad_i  # new param values
         updates.append((mparam_i, v))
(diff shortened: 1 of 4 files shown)


benanne:
Cool! Maybe we should move it elsewhere because all the other things in this module generate updates, and this processes existing updates. I don't know what would be a better place for it though.

Also I think it would be good to make the `1e-7` configurable (as `self.epsilon`).

The docstring is not in our default format, but then again, so are none of the others in that module, so I don't blame you :p Still it would be great if you could update it so we don't have to modify that one later.

It's unfortunate that we need a check on the dimensionality to determine what to do, but I don't have a better idea.

For some reason `from theano.config import floatX` is making Travis fail. It might be cleaner to write it in full where it's used anyway, what's the motivation for changing this?

ebattenberg:
> Also I think it would be good to make the 1e-7 configurable (as self.epsilon).

Do you mean to make it a parameter of the function?  



> The docstring is not in our default format, but then again, so are none of the others in that module, so I don't blame you :p Still it would be great if you could update it so we don't have to modify that one later.

Yeah, what I meant by "I'll add better docs later" is that I'll update the docstring once we've agreed on the implementation and interface.  :)



> It's unfortunate that we need a check on the dimensionality to determine what to do, but I don't have a better idea.

Me neither.  Are there any other cases we want it to work with here?  A 1D convolution is a tensor3?  I guess there could be element-wise constraints on a vector of biases. 

What about adding min constraints?


> For some reason from theano.config import floatX is making Travis fail. It might be cleaner to write it in full where it's used anyway, what's the motivation for changing this?

I did it for brevity/readability and for making PEP8 line-widthing easier.  Not sure why that's breaking things.  I'll fix it.

benanne:
> Do you mean to make it a parameter of the function?
Yes, sorry for the confusion. This is a function, not a class, my bad :)

> Yeah, what I meant by "I'll add better docs later" is that I'll update the docstring once we've agreed on the implementation and interface. :)
Looks like I missed that, sorry again!

> Me neither. Are there any other cases we want it to work with here? A 1D convolution is a tensor3? I guess there could be element-wise constraints on a vector of biases.
It should be doable to support Conv1D, 2D and 3D in one go because the dimensionality is always going to be bc012... so you can assume that axis 2 and beyond are going to be spatial dimensions.

> What about adding min constraints?
Are min constraints useful? Have they been used in literature?

ebattenberg:
> It should be doable to support Conv1D, 2D and 3D in one go because the dimensionality is always going to be bc012... so you can assume that axis 2 and beyond are going to be spatial dimensions.

Good idea.



> Are min constraints useful? Have they been used in literature?

I'm not sure.  Maybe if someone wanted to use weight decay but prevent certain filters from going to zero?  I was just asking because @f0k suggested it in #84 at the end of [this comment](https://github.com/benanne/Lasagne/issues/84#issuecomment-73436424).

benanne:
I see. If it doesn't change anything fundamental in the code we could support that as well, otherwise I'd be wary of making up use cases that we don't even know are actual use cases :)

ebattenberg:
> I see. If it doesn't change anything fundamental in the code we could support that as well, otherwise I'd be wary of making up use cases that we don't even know are actual use cases :)

Yeah, I agree that we should stick with just max constraints unless someone else has anything to add.  

I just fixed that float error, added some tests, and simplified/generalized the support for {3,4,5}-d arrays for Conv{3,4,5}DLayer).  

I'll do the doc soon.

ebattenberg:
Added docs.

skaae:
Hi, 
In RNN's you often constrain the norm of the weight updates. If W is all your parameters you scale norm(delta_W) to be below some constant. You rescale all updates "globally".  If I understand your code correctly you constrain the norm of the incoming weights to each hidden unit?

Could the RNN use case be merged to into your code? otherwise i can write the RNN norm constraint function.

Can you add a usage example to the docs where it is used with one of the functions for computing updates?

ebattenberg:
> Could the RNN use case be merged to into your code? otherwise i can write the RNN norm constraint function.

This should be possible and is a good use case for this function.  Basically all you would need to do is hand your gradient to this function to have it clipped.  So, you would need to use it in the function that computes your updates. Something like:

```
grad = theano.grad(rnn_loss, param)
clipped_grad = norm_constraint(grad, abs_max=max_grad_norm)
updates = [(param, param - learning_rate * clipped_grad)]
```

> Can you add a usage example to the docs where it is used with one of the functions for computing updates?

That's a good idea.  I'll keep that in mind as we decide where this type of function should actually live.  But real quick, to clip the updated value of `param1`, you can do for example:

```
from collections import OrderedDict
updates = OrderedDict(nesterov_momentum(loss, all_params, learning_rate))

# for abs_max constraint
updates[param1] = norm_constraint(updates[param1], abs_max=ABS_MAX)

# OR

# for rel_max constraint
updates[param1] = norm_constraint(updates[param1], param=param1, rel_max=REL_MAX)

```

lasagne/updates.py
```diff
@@ -113,3 +139,116 @@ def adadelta(loss, all_params, learning_rate=1.0, rho=0.95, epsilon=1e-6):
...
+    :parameters:
+        - orig_update : TensorVariable
+            Theano expression for original update
+        - param : TheanoSharedVariable
+            Shared variable containing initial parameter values.
+            (Required if `rel_max` is used)
+        - abs_max : scalar
+            This value sets the absolute maximum value allowed for the
+            incoming weight vectors after the update.
+            (Optional)
```
f0k: Minor note: If you write `(Optional)` here, you should do so for `rel_max` as well.

f0k:
> So, it makes assumptions about the mean of the dimensions in a 2D and 4D parameter array.

I haven't thought that through, but what about allowing users to specify the axes that make up the incoming weights?

ebattenberg:
> I haven't thought that through, but what about allowing users to specify the axes that make up the incoming weights?

That might make it a little harder to use if the axes are required; however, we could add it as an optional argument that overrides the default for a specific number of dimensions.

lasagne/updates.py
ebattenberg: Fixed.

ebattenberg:
> That might make it a little harder to use if the axes are required; however, we could add it as an optional argument that overrides the default for a specific number of dimensions.

Did that in https://github.com/ebattenberg/Lasagne/commit/7b84cf34c5e90a39ad498cc08f2e6ae799df42ce

f0k:
Oops, I meant to comment on the PR diff, not on the commit. Anyways. Except for my comment, this looks very very good! It's well-documented and it does everything one could ever want. Good job!

ebattenberg:
@f0k  Thanks for the suggestion.  I initially was trying for a nice clean generator expression, but couldn't think of one quickly.  Iterator within iterator is much cleaner.  Thanks!

benanne:
Awesome work! Is there a consensus now that we want to keep this in `lasagne.updates`? I'm okay with it. We may want to split up `updates.py` at a later date, like we did with `layers`, but it's still good to import everything into the `lasagne.updates` namespace for convenience I suppose.

ebattenberg:
> Awesome work! Is there a consensus now that we want to keep this in lasagne.updates?

For now, it seems like the right place since this would mainly be used to clip the updates or clip the gradients which are both part of the update computation flow.  

I'm going to write a few more tests.  I'll let you know when it's ready for merge.

benanne:
Alright, thanks!

f0k:
> @f0k Thanks for the suggestion.

Unfortunately, Travis fails for Python 3 now: `AttributeError: 'range_iterator' object has no attribute 'next'`
Turns out the way to do it in Python 2 and 3 is `next(count)` instead of `count.next()`.

ebattenberg:
> Turns out the way to do it in Python 2 and 3 is next(count) instead of count.next().

Ahh, yeah, I knew that.  Fixed.  

Also, I added some final tests.  Ready for merge if Travis passes.

benanne:
Looks great! Very complete and well tested. I will merge this tomorrow if there are no more comments.

Just one minor thing, I wonder if we can avoid the `broadcast` stuff altogether by using `keepdims=True` argument of `theano.tensor.sum`. Maybe that would be a bit easier to understand.

f0k:
> Just one minor thing, I wonder if we can avoid the `broadcast` stuff altogether by using `keepdims=True` argument of `theano.tensor.sum`.

D'oh. Good to know that library :) Yes, this seems to do exactly what we need, in a very similar way:
https://github.com/Theano/Theano/blob/master/theano/tensor/basic.py#L2629
https://github.com/Theano/Theano/blob/master/theano/tensor/basic.py#L1449

ebattenberg:
> Just one minor thing, I wonder if we can avoid the broadcast stuff altogether by using keepdims=True argument of theano.tensor.sum. Maybe that would be a bit easier to understand.

:+1:  Cool.  That'll save some lines.

dnouri:
Nice work.  Citing from an earlier comment:

> Can you add a usage example to the docs where it is used with one of the functions for computing updates?

It might be worthwhile to have a little `:usage:` block in the docstring for such an example.

ebattenberg:
> It might be worthwhile to have a little :usage: block in the docstring for such an example.

Will do.

ebattenberg:
Ok, I added some usage examples.  But it looks like they're getting run as doctests by py.test.  Not sure how to disable that.  I don't really want to write a complete working example in the docs.

f0k:
I'd generally say that doctests would be a good thing, but here I agree that writing down a complete working example wouldn't help anybody, and stripping it down to something that's easily doc-testable would be too far from a typical usage pattern.
Researching a bit, it seems the only way to skip a doctest is to add `# doctest: +SKIP` after every line. Another way is wrapping it in a function definition so the code gets parsed, but not executed. Both is ugly.
Maybe turn it into a working example in the end? After all, we only need to add something like that in the beginning:
```python
# assume some neural network parameters and loss (faked here)
import theano  # (is that needed at all?)
param1, param2 = all_params = theano.matrices('a', 'b')
loss = param2.sum()
```
And then omit `learning_rate` to save another unknown variable, and supply `param2` for the relative constraint (that was missing). Ah, `param2` needs to be a shared variable. So if we want that, it should be something like:
```python
param1, param2 = all_params = map(theano.shared,
        [np.zeros((10,20)), np.ones((30,40))])
```

ebattenberg:
Thanks @f0k.  Another option would be to just skip the `>>>` and write the example in the doc string under `:usage:` but not as an executable doctest.  But does adding the `>>>` make it nice and syntax highlighted when creating html docs with Sphinx?

dnouri:
You should be able to add a `.. code-block:: python` to get back any syntax highlighting should it gets lost by removing the >>> prompt.  I'm thinking runnable would be nice, too.

f0k:
> But does adding the `>>>` make it nice and syntax highlighted when creating html docs with Sphinx?

I assumed so... who knows that? @benanne? One could probably also use `>>` instead of `>>>` and make Sphinx render it the same... but we shouldn't waste too much time on this, I'd like to use the feature :)

benanne:
I say we make it executable, it's always nice to be able to copy and paste examples and run them yourself :)

ebattenberg:
So it seems like you guys prefer to have a running doctest.  However as @f0k pointed out above, this means it won't really illustrate a real-world use case.  For example, it will be difficult to show examples of both weight norm constraints and gradient clipping.

benanne:
I suppose some comments could clarify what `param1` and `param2` represent? Later on we should add a bunch of extensive examples to the documentation that cover all the features anyway (including more obscure things like weight sharing and such). But let's get all the docstrings sorted first so the API is at least documented :)

ebattenberg:
It feels like getting a fake example of this running in the doctest isn't all that useful to the user.  It could be that it's more confusing for someone to look at something like:

```
# assume some neural network parameters and loss (faked here)
import theano  # (is that needed at all?)
param1, param2 = all_params = theano.matrices('a', 'b')
loss = param2.sum()

#OR

param1, param2 = all_params = map(theano.shared,
        [np.zeros((10,20)), np.ones((30,40))])
```

And be confused about what this function is even used for since those are so far from something anyone would actually do.  Maybe it would be better to demonstrate usage in an example script, like one of the MNIST ones.  For a runnable doctest, I'm starting to favor the option of making a large working example that is more illustrative of actual usage.  Maybe an MLP with one hidden layer or a logistic regression?

ebattenberg:
Sorry, maybe I'm overthinking this.  Would someone be willing to PR my PR or merge the PR and then edit the doctest?  I think it's ready to be merged, so we shouldn't dwell on this doctest thing.  :)

f0k:
> Sorry, maybe I'm overthinking this.

It's good to think about this as it will also apply to other parts of our documentation. But maybe we should get this merged (in some way that doesn't make Travis fail) and sort out the documentation issue separately? @benanne could remove the `>>>` and add `.. code-block:: python` as part of the merge.

---

> Maybe it would be better to demonstrate usage in an example script, like one of the MNIST ones.

It'd be good to have some tutorial in the documentation, and then just link to the respective part of the tutorial from the docstring.

> For a runnable doctest, I'm starting to favor the option of making a large working example that is more illustrative of actual usage. Maybe an MLP with one hidden layer or a logistic regression?

If we want that, we could define the working example somewhere centrally (but where?) and import it in whichever doctest needs it. But this discussion should probably be moved to the Documentation issue (#26).

benanne:
I checked out this pull request locally and I am unable to make `.. code-block:: python` work. It just doesn't show up in the html docs if I do that.

Sphinx seems to be smart enough to remove the `# doctest: +SKIP` in the html docs, so I think that's the best solution for now.

One of the examples seems to be wrong:
```
updates[param2] = norm_constraint(updates[param2], rel_max=1.0)
```
If you specify `rel_max`, the `param` argument is required, but it is not specified here. I guess this should be:

```
updates[param2] = norm_constraint(updates[param2], param2, rel_max=1.0)
```


An unrelated, more general remark about this pull request: while appreciate the practicality of being able to specify a relative maximum based on the initial values of a shared variable, I feel that maybe the `norm_constraint` function is trying to do a bit too much.

Maybe it's better to always expect an absolute maximum, and if the user wants to specify a relative maximum they should convert it to an absolute value first. We can of course provide a convenience function for this. This would considerably simplify the signature of `norm_constraint` and make the code easier to read and understand.

So instead of:
```
updates[param2] = norm_constraint(updates[param2], param2, rel_max=1.0)
```

We'd have something like:
```
updates[param2] = norm_constraint(updates[param2], compute_avg_norm(param2) * 1.0)
```

or something like that.

f0k:
> One of the examples seems to be wrong:

Yep, probably should have told so more prominently.

> An unrelated, more general remark about this pull request: while appreciate the practicality of being able to specify a relative maximum based on the initial values of a shared variable, I feel that maybe the `norm_constraint` function is trying to do a bit too much.

Hmm, maybe you're right.
```python
updates[param2] = norm_constraint(updates[param2], compute_avg_norm(param2) * 1.0)
```
looks easy enough. However, when you consider the case of custom axes, you'd get:
```python
updates[param2] = norm_constraint(updates[param2], compute_avg_norm(param2, norm_axes=(0,1) * 1.0), norm_axes=(0,1))
```

I don't have a strong opinion in either direction. But in any case, `compute_avg_norm()` probably shouldn't be part of `lasagne.updates`, but `lasagne.utils`?

ebattenberg:
> One of the examples seems to be wrong:

And that's a strong argument in favor of working doctests. :)  What I think I'll do for now is add a doctest or two to show very simple low-level usage of the function, and then maybe add some non-doctest examples in the docstring to illustrate an actual use case.  We can then move that non-doctest example to an example script later.


> An unrelated, more general remark about this pull request: while appreciate the practicality of being able to specify a relative maximum based on the initial values of a shared variable, I feel that maybe the norm_constraint function is trying to do a bit too much.

Since we're drilling down so deep into this function, I'll bring up that I'm slightly unsatisfied with the implementation of the `rel_max` feature as it is now.  In my original code (not in Lasagne), I use the expected value of the norm given the initialization distribution to calibrate the `rel_max`.  In order to keep this function as standalone as possible, I settled for the sample average norm as an approximation to this.  However, this causes the `rel_max` constraint to be non-deterministic which I don't like.  I can't think of an easy way to deal with this though.

ebattenberg:
> Since we're drilling down so deep into this function, I'll bring up that I'm slightly unsatisfied with the implementation of the rel_max feature as it is now. In my original code (not in Lasagne), I use the expected value of the norm given the initialization distribution to calibrate the rel_max. In order to keep this function as standalone as possible, I settled for the sample average norm as an approximation to this. However, this causes the rel_max constraint to be non-deterministic which I don't like. I can't think of an easy way to deal with this though.

So maybe we should actually just remove the `rel_max` feature for now since it doesn't do what I actually want it to do, and then we can implement a function that comes up with specific constraint values at a later date.

benanne:
I think that's a good idea. Sorry for keeping this PR going for ages :p

ebattenberg:
Ok, removed `rel_max` feature.  Moved `compute_norms` to utils.  Doctests are working.

ebattenberg:
For posterity, the `rel_max` feature  can be found in my "rel-max" branch.

f0k:
Looks good to me! Not sure how useful the usage example is now, but that's something we can address when including a proper tutorial. One very last issue is that the default value for `norm_axes` is not documented. You could add: `(0 for 2D, all but 0 for 3D to 5D)` to both `norm_constraint` and `compute_norms`. Sorry, I promise I will not request anything more for this PR! On the positive side, this will instantly get you to the third place of the [contributor's list](https://github.com/benanne/Lasagne/graphs/contributors) :)

ebattenberg:
> Sorry, I promise I will not request anything more for this PR! On the positive side, this will instantly get you to the third place of the contributor's list :)

I realize you guys are just trying to help and get the code real nice and stuff, but I'm officially done with this PR for now.  It's such a small function, and there's so much to be desired with the rest of the library in terms of documentation.  I'm not sure why we're stressing out so hard about this one.  

Feel free to merge and make any further changes in separate PR's if desired.

benanne:
Don't worry about it, I'll take care of it when merging. This has gone on for long enough :)

benanne:
Done!

PR #142, opened 2015-02-17
set dnn border_mode to None when using pad

ebenolson:


benanne:
Good catch! I assume this wasn't working before? Just to make sure, could you clarify the problem you ran into and why this fix solves it? The dnn convolutions unfortunately don't have tests yet, we should sort that out ASAP.

ebenolson:
yes, I was getting an AttributeError on line 122 in get_output_for:
`border_mode = self.border_mode if (self.border_mode is not None) else self.pad`
an alternative would be to use `hasattr` there instead, but this seemed like the simpler fix.

benanne:
Yep, looks like it's the correct fix to me too. Thanks for the clarification!

PR #147, opened 2015-02-21
Add note to install instructions about lack of PyPI release

dnouri:
Tiny PR that addresses today's "pip install does'nt work?" mail on the mailing list.


benanne:
Good call, thanks!

Issue #21, opened 2014-09-12
PEP8


benanne:
What is everyone's stance on enforcing [PEP8](http://legacy.python.org/dev/peps/pep-0008/)? Personally I think it's good to enforce some coding standards, but the 80 character line length limit really annoys me sometimes. @avdnoord feels the same way. Maybe there are good arguments in favor of it though?

Are there any good tools (perhaps with GitHub integration) that make enforcing this easier?

ebattenberg:
I think python mode in vim is really handy.    And +1 for sticking with PEP8.

ebattenberg:
And while python mode in vim shows you all the violations from installed checkers in-line, you can always run those programs separately to get the feedback.  Those that are supported include:

https://pypi.python.org/pypi/pep8
https://pypi.python.org/pypi/pyflakes
https://pypi.python.org/pypi/pylint
https://pypi.python.org/pypi/mccabe
https://pypi.python.org/pypi/pep257

craffel:
Sander, in #149, writes

>  But the PEP8 issue (#21) never came to a conclusion - we should take care of that first.

I think the conclusion reached here is "Some people think 80 characters is too short but we should follow PEP-8 anyways".  As I mentioned, it's very easy to force yourself to follow PEP-8 using a plug-in in your favorite editor, and it's much easier to point to a (universally accepted) style guide then to make your own up when you are soliciting pull requests.

benanne:
Let's do it. As I said before I can live with it if necessary. I'll adapt :) If anyone has any other concerns, please comment as soon as possible.

We should probably do a sweep of the entire codebase (preferably in a limited number of commits) to ensure that all of the current code follows PEP8. That will make it much easier to enforce for future pull requests.

craffel:
> We should probably do a sweep of the entire codebase (preferably in a limited number of commits) to ensure that all of the current code follows PEP8.

I can do this.  I usually advocate for one commit per file - makes it easier to roll back changes if something breaks, although if you'd prefer a single commit that's fine too.

benanne:
Good point, I guess one commit per file is best.

ebattenberg:
I prefer exactly two commits per file.  ;)

I just realized that I've been using `'''` in my docstrings all along while PEP 257 suggests always using `"""` for consistency.  I've been doing it wrong all along!

benanne:
I changed them to `"""` when I merged #135 :)

dnouri:
[...]

PR #149, opened 2015-02-25
Fix examples

diogo149:
Changes:
- fix dnn imports in the examples
- cuda-convnet example didn't return output_layer
- the accuracy returned was float64, which caused problems when the warn_float64 theano flag is set

### DIFF
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -59,7 +59,7 @@ def load_data():
         num_examples_test=X_test.shape[0],
         input_dim=X_train.shape[1],
         output_dim=10,
-        )
+    )
 
 
 def build_model(input_dim, output_dim,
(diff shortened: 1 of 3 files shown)


benanne:
Thanks for this! We should have caught this sooner. In my defence, my laptop has a GT 540M so I can't run the dnn example on that :p 

Since the changes are fairly small and Travis passes, I'll go ahead and merge this.

dnouri:
I wish this PR and others included fewer random changes to indentation.

craffel:
> I wish this PR and others included fewer random changes to indentation.

This is why we need to enforce a style guide (starting, at least, with PEP-8).  Not to be pedantic, but all of these aesthetic code changes will disrupt the blame pretty heavily.

benanne:
Agreed, this is something I haven't been paying enough attention to. But the PEP8 issue (#21) never came to a conclusion - we should take care of that first.

dnouri:
Agree too.  Consider though that the indentation changes that went into this PR weren't even a PEP8 issue.  I'd say collaborators need to stop configuring their editor to randomly move parenthesis around, if that's what's happening.

PR #150, opened 2015-02-27
PEP-8/Pyflakes cleanup

craffel:
Here's a sweep through the codebase fixing all PEP-8 issues, and also pyflakes for good measure (primarily for things that get imported/defined but are unused).  Before:
```
> pep8 | wc -l
    565
```
After:
```
> pep8 | wc -l
    0
```

There's one commit per file, generally stating the issue being fixed in the file (or misc/various if there's lots of pep-8's being fixed).  This was, in fact, useful in the course of preparing this PR at various times.  Atomic version control FTW.


ebattenberg:
Hmm. rebase -i -> squash all? I don't understand the benefit of having a
separate commit per file.

craffel:
Because if one of these commits broke something, it'll be easier to revert it if they're separated  As I said earlier, this was useful in the course of creating this PR; I made some errors here and there which were much easier to fix because the commits were separate.  If I'm in the minority in my preference for atomic commits, I can squash.

benanne:
I have to say I'm not a big fan of some of the new PEP8 rules I'm discovering through this. I agree that "comments should start with a space", but I've made a habit of using this to distinguish actual comments from commented code (where the space would be missing). I figured a lot of people do this. But I guess we can't :p 

Also, although imports like `theano`, `theano.tensor as T`, `numpy as np`, ... etc. aren't generally used, I put them in almost every file because they will be more than likely be necessary at some point, and making the "preamble" of each file somewhat consistent is probably a good idea. We now run the risk of someone needing numpy in one of those files and adding `import numpy as N` or something like that (contrived example). We'll have to pay extra attention to this in the future.

In https://github.com/craffel/nntools/commit/3cb8cb548d48230978a5cdd99bacab87e3045c87 , the comment should probably be above the line starting with `update_i`, and not "in between" :)

In https://github.com/craffel/nntools/commit/4c72088af741486dbf24e6846afa8ffa4567bbce : I'm not a huge fan of reusing variable names like that (`output_length`), it makes later editing more error-prone. Although I guess this is a fairly benign case of that. At least the usage is consistent in the file.

In https://github.com/craffel/nntools/commit/bff033b95022840ef406aaa9a88d45141ea4259d , why was the mention of the default nonlinearity being rectification removed from the docstring? I suppose it's also in the method signature but it doesn't hurt to mention this explicitly in the text as well. Of course I realise some people want to change this, but for now it's still there.

In https://github.com/craffel/nntools/commit/53521380b1b2a57719c9cb2dc0d9e8bea87047c8 , why were the spaces removed around the `/` sign? This seems inconsistent, maybe I don't know all the rules for whitespace around operators.

All in all these are fairly minor comments, so even if there are a few things I don't like, we should either follow the entire standard or not follow it at all. So I guess the former is the least bad option :)

Since we are planning to enforce this henceforth, is there anything like Travis for automatic PEP8 checking of commits / PRs? Or can we make Travis do it by adding a test for it?

dnouri:
+1 for squashing.

> Since we are planning to enforce this henceforth, is there anything like Travis for automatic PEP8 checking of commits / PRs? Or can we make Travis do it by adding a test for it?

I've added automatic pep8 checks to pytest in this PR: https://github.com/craffel/nntools/pull/14

I'm ignoring some of the most annoying PEP8 warnings (some of which I believe aren't even enforced in the Python PEP8 standard, but were added by the anal PEP8 tool developers themselves).  We can adjust this list to our liking, in `setup.cfg`:

```
[pytest]
# ...
# http://pep8.readthedocs.org/en/latest/intro.html#error-codes
pep8ignore = E122 E123 E125 E128
```

And finally, @craffel's branch (still?) has a few "E226 missing whitespace around arithmetic operator" errors for me.

benanne:
So are we actually going to follow PEP8 selectively in the end? I thought the idea was to follow the standard to avoid any confusion and make contributing easier.

Apparently E123 is not part of the official standard (and not enforced by default), but the others should be. What's the motivation for excluding the others? If it's just because they are annoying, I could say the same thing about the 80 character line limit :p

benanne:
We can squash this when it's ready if necessary, but for now I think the separate commits are actually pretty handy for the review.

dnouri:
Re: 'handy for review'.  Why wouldn't you switch to the "Files changed" tab and just comment inline?

I've updated my PR-PR to not exclude any of the rules.  Which adds back some default ignores, so @craffel's PR is again without errors.  (Confused already?)

benanne:
Good point.

And yes, this is confusing. But I thought the point of following a coding standard is to actually follow it :p if we're going to make exceptions, we should at least discuss this and come to a consensus. Also, if some rules are ignored by default (because they aren ot part of the standard), then we should ignore them too. My point was mainly that it seems silly to pick and choose a few others that we decide to ignore, which kind of defeats the point imo.

dnouri:
@benanne Understood!

f0k:
+1 for squashing from me as well.

> Because if one of these commits broke something, it'll be easier to revert it if they're separated

You can always reset a single file to an earlier state (`git checkout <commithash> <filename>`). We're only in trouble if we want to undo the pep8 commits on a specific file several PRs down the road, but I don't see why we would want that. If the changes break anything, this should be caught by our tests, otherwise it's not broken by definition :)

craffel:
> Also, although imports like theano, theano.tensor as T, numpy as np, ... etc. aren't generally used, I put them in almost every file because they will be more than likely be necessary at some point, and making the "preamble" of each file somewhat consistent is probably a good idea. 

This isn't actually a PEP-8 issue, it's a pyflakes issue, which is another useful syntax checker. The idea being that if you import or define something but then don't use it, it might mean you forgot to do something with the import/new variable - a form of code smell.  If you do want a consistent preamble, then we need to ensure that 
```
import theano
import theano.tensor as T
import numpy as np
```
appears at the top of every file.

> In craffel@3cb8cb5 , the comment should probably be above the line starting with update_i, and not "in between" :)

The idea was that because it was an end-of-line comment, it was referring to `T.sqrt(acc_i_new + epsilon))`, not the line above it.  But I can change this.

> In craffel@4c72088 : I'm not a huge fan of reusing variable names like that (output_length), it makes later editing more error-prone. Although I guess this is a fairly benign case of that. At least the usage is consistent in the file.

I did this in many places.  I can try to go through and change these to parentheses instead.

> In craffel@bff033b , why was the mention of the default nonlinearity being rectification removed from the docstring? 

Because it'll already show up in the docs that that's the default.  Also, for consistency - should we add a "default xyz" to the docstring for every parameter with a default value.

lasagne/layers/dense.py
```diff
@@ -129,17 +132,21 @@ def get_output_shape_for(self, input_shape):
...
         if self.b is None:
             activation = out
         else:
             if self.untie_biases:
-                remaining_dims_biases = range(1, input.ndim - 1) # no broadcast
+                # no broadcast
+                remaining_dims_biases = range(1, input.ndim - 1)
             else:
-                remaining_dims_biases = ['x'] * (input.ndim - 2) # broadcast
+                remaining_dims_biases = ['x'] * (input.ndim - 2)  # broadcast
```
f0k: `# broadcast` should be moved to a separate line before, because you (or pep8 or pyflakes) did the same with `# no broadcast`.

lasagne/layers/pool.py
```diff
@@ -18,18 +17,18 @@
...
         if self.ignore_border:
-            output_shape[2] = int(np.floor(float(output_shape[2]) / self.ds[0]))
-            output_shape[3] = int(np.floor(float(output_shape[3]) / self.ds[1]))
+            output_shape[2] = int(np.floor(float(output_shape[2])/self.ds[0]))
+            output_shape[3] = int(np.floor(float(output_shape[3])/self.ds[1]))
         else:
-            output_shape[2] = int(np.ceil(float(output_shape[2]) / self.ds[0]))
-            output_shape[3] = int(np.ceil(float(output_shape[3]) / self.ds[1]))
+            output_shape[2] = int(np.ceil(float(output_shape[2])/self.ds[0]))
+            output_shape[3] = int(np.ceil(float(output_shape[3])/self.ds[1]))
```
f0k: @benanne: Sorry, this is unrelated to the PR, but why isn't that `output_shape[2] //= self.ds[0]` instead of a long float -> true div -> ceil -> int chain? Should we change that?

benanne: Because we need to round up, not down. I guess the first part of the if clause (with floor) could be replaced by integer division, but I think I did it this way for consistency.

lasagne/layers/shape.py
```diff
@@ -98,8 +95,8 @@ def get_output_shape_for(self, input_shape, *args, **kwargs):
                                      (o[0], len(input_shape)))
                 output_shape[dim] = input_shape[o[0]]
                 masked_output_shape[dim] = input_shape[o[0]]
-                if ((input_shape[o[0]] is None)
-                    and (masked_input_shape[o[0]] is None)):
+                if (input_shape[o[0]] is None) \
+                   and (masked_input_shape[o[0]] is None):
```
f0k: I'd prefer brackets over line continuation. Which tool did complain here? Should we change that?

craffel:
> /edit: And with respect to the PR, why was the whitespace removed? It's required by pep8. 

No, it's not; see my comment above:

> It's a little ambiguous: https://www.python.org/dev/peps/pep-0008/#other-recommendations Basically, you can have one or no spaces, and you're supposed to put spaces around operators with less priority.

If it was required, pep8 wouldn't pass without complaining!

lasagne/nonlinearities.py
```diff
@@ -14,13 +11,15 @@
 # tanh
 from theano.tensor import tanh
 
+
 # rectify
 # The following is faster than lambda x: T.maximum(0, x)
 # Thanks to @SnippyHolloW for pointing this out.
-# See: https://github.com/SnippyHolloW/abnet/blob/807aeb98e767eb4e295c6d7d60ff5c9006955e0d/layers.py#L15
+# See: https://github.com/SnippyHolloW/abnet/blob/807aeb9/layers.py#L15
```
f0k: (Cool, didn't know that was possible.)

benanne:
Regardless of whether it's required, I really don't see why the space should be removed there. If anything it makes it less readable to me.

lasagne/layers/shape.py
craffel: Me too, but it's because this is bad:
```
if (something == foo and
    something_else == bar):
    this_line = has_the_same_indent_level
```
This, I think, is the only place I use line continuation.

lasagne/layers/pool.py
dnouri: Yes, removing that whitespace isn't PEP8.  The PEP8 tool has [E226](http://pep8.readthedocs.org/en/latest/intro.html#error-codes), which requires whitespace around arithmetic operators, but it's turned off by default.  There's certainly no rule that says that this whitespace should be removed, so this change arguably isn't in the scope of the PR.

lasagne/layers/shape.py
dnouri: Not sure I agree to this change and other changes that add backslashes.  PEP8 says:

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. **These should be used in preference to using a backslash for line continuation.**

A misunderstanding?

lasagne/layers/pool.py
f0k: > Because we need to round up, not down.

Oops, didn't see the difference between the two cases at first glance. Haven't slept enough. Consistency is good then!

lasagne/layers/dense.py
craffel: Yeah, that was because `# no broadcast` is too long and `# broadcast` isn't.  I can change this.

lasagne/layers/shape.py
craffel: Yes, absolutely I'm using parentheses preferentially over line continuation, but see my example above -- when there's an `if` statement with parentheses, it causes a different PEP-8 issue, which is the only place I use line continuation.

f0k: Hmm, good point. I usually don't do visual indent there, but indent twice, but I guess pep8 wouldn't like that. I'm fine with line continuation then. The indent level is still pretty close to the line following the if statement now, though... what about a double indent? Is that allowed when we're doing line continuation anyway, and don't have to match visual indent for brackets?

craffel: > I usually don't do visual indent there, but indent twice, but I guess pep8 wouldn't like that.

Yes, two indents gives the same error, `E129 visually indented line with same indent as next logical line`.  We can double indent when using line continuation, would you prefer that?

f0k: > Yes, two indents gives the same error, `E129 visually indented line with same indent as next logical line`.

Really? Shouldn't it say `indented too much for visual indent` (whatever the wording of that is)?

> We can double indent when using line continuation, would you prefer that?

I'd prefer that, but I don't know what the majority thinks. @benanne, @dnouri: What do you prefer for long if clauses? There's an overview of alternatives here, for example: http://stackoverflow.com/a/181557

craffel: Frankly, I think they're all ugly.  I switched to double indent with line continuation, if you all decide you prefer a different technique, feel free to change it (will push soon).

craffel:
OK, I changed some of the requests mentioned above and squashed the commits.  The idea behind this PR was to get the code to pass pep8, if there are additional aesthetic changes you all want to make please feel free to do so but I need to start spending time on other things!

f0k:
All right. @benanne can take care of the whitespace in `MaxPool2D` while merging, I guess?
What's troubling me is that my commit about `lasagne.utils.concatenate` shows up here -- we don't want that to end up twice in Lasagne's history. Probably rebasing this PR onto `master` will fix this. I don't know if @benanne can do that.

craffel:
> @benanne can take care of the whitespace in MaxPool2D while merging, I guess?

Sorry, just changed this.

> What's troubling me is that my commit about lasagne.utils.concatenate shows up here -- we don't want that to end up twice in Lasagne's history.

This showed up because I rebased my `pep8` branch onto `master`.  I'm guessing it will disappear when this PR is merged, right?

benanne:
I'll merge this tomorrow if there are no further comments. We can think about aesthetics later :)

benanne:
merging in 3... 2... 1...

Issue #151, opened 2015-02-27
remove `lasagne.utils.concatenate`, use `theano.tensor.concatenate` instead


benanne:
Since Theano's own `concatenate` now has a proper GPU gradient (see https://github.com/benanne/Lasagne/issues/102), we should start using it and get rid of our custom implementation.

PR #152, opened 2015-02-27
Deprecate lasagne.utils.concatenate and do not use it within library

f0k:
Addresses #151 (but does not close it; we still need to remove the function for the first release).


benanne:
Thanks!

PR #153, opened 2015-03-03
Misc diagnostics/cleanup to examples

craffel:
I was running the examples recently to compare the speed of Theano's convolutions (FFT and non-FFT), cudnn, and cuda_convnet, and in the process I made a few changes to the example files.  This PR reflects those changes, which are:

- Switched string syntax from `%` to `format`, because `%` are ["obsolete and may go away in future versions of Python"](https://docs.python.org/release/3.1.5/library/stdtypes.html#old-string-formatting-operations)
- Making it so that issuing ctrl-C while training just exits the script cleanly without a traceback (just an aesthetic thing when you don't want it to train for 500 epochs)
- Added print statements for loading data and building the models
- Added the amount of time each epoch took to the print statement

I can split these up if you want, or remove one of these changes if you want, but none of them really change the functionality in any big way.

### DIFF
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -190,15 +190,19 @@ def main(num_epochs=NUM_EPOCHS):
     iter_funcs = create_iter_functions(dataset, output_layer)
 
     print("Starting training...")
-    for epoch in train(iter_funcs, dataset):
-        print("Epoch {} of {}".format(epoch['number'], num_epochs))
-        print("  training loss:\t\t{:.6f}".format(epoch['train_loss']))
-        print("  validation loss:\t\t{:.6f}".format(epoch['valid_loss']))
-        print("  validation accuracy:\t\t{:.2f} %%".format(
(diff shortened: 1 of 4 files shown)


benanne:
These all seem to be reasonable changes, so I'll merge this tomorrow if there are no comments.

I'm a little bit wary about making the examples too 'nice' and 'clean', because the main use case for these examples is reading and understanding the code. So too much 'cleanliness' might actually overcomplicate the code to the point where it is less useful as an example.

I'm definitely not saying that that is the case here, these are all simple additions / modifications that are easy to understand. Just something to keep in mind when modifying the example code.

craffel:
> I'm a little bit wary about making the examples too 'nice' and 'clean', because the main use case for these examples is reading and understanding the code. So too much 'cleanliness' might actually overcomplicate the code to the point where it is less useful as an example.

Very true.

PR #156, opened 2015-03-04
change Conv2DLayer.get_output_shape to match Theano for border_mode='ful...

### LINKED ISSUE #154: error in output shape calculation for strided Conv2DLayer with border_mode='full'
Simple example:

        l_in = lasagne.layers.InputLayer(
            shape=(1, 1, 10, 10),
        )

        l_conv = lasagne.layers.Conv2DLayer(
            l_in,
            num_filters=1,
            filter_size=(3, 3),
            strides=(2,2),
            border_mode='full',
        )

`l_conv.get_output_shape()` gives (1, 1, 5, 5)
But `l_conv.get_output(x).shape` yields (1, 1, 6, 6)

ebenolson:
To fix #154, change the calculation of output shape in Conv2DLayer with `border_mode=='full'` to match [Theano](https://github.com/Theano/Theano/blob/master/theano/tensor/nnet/conv.py#L252])

### DIFF
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -13,6 +13,24 @@ __all__ = [
 ]
 
 
+def conv_output_length(input_length, filter_length,
+                       stride, border_mode):
+    if border_mode == 'valid':
+        output_length = input_length - filter_length + 1
+    elif border_mode == 'full':
+        output_length = input_length + filter_length - 1
+    elif border_mode == 'same':
+        output_length = input_length
+    else:
+        raise RuntimeError('Invalid border mode: {0}'.format(border_mode))
+
+    # This is the integer arithmetic equivalent to
+    # np.ceil(output_length / stride)
+    output_length = (output_length + stride - 1) // stride
+
+    return output_length
+
+
 class Conv1DLayer(Layer):
     def __init__(self, incoming, num_filters, filter_length, stride=1,
                  border_mode="valid", untie_biases=False, W=init.Uniform(),
@@ -52,16 +70,10 @@ class Conv1DLayer(Layer):
         return [self.b] if self.b is not None else []
 
     def get_output_shape_for(self, input_shape):
-        if self.border_mode == 'valid':
-            output_length = ((input_shape[2] - self.filter_length) //
-                             self.stride + 1)
-        elif self.border_mode == 'full':
-            output_length = ((input_shape[2] + self.filter_length) //
-                             self.stride - 1)
-        elif self.border_mode == 'same':
-            output_length = input_shape[2] // self.stride
-        else:
-            raise RuntimeError("Invalid border mode: '%s'" % self.border_mode)
+        output_length = conv_output_length(input_shape[2],
+                                           self.filter_length,
+                                           self.stride,
+                                           self.border_mode)
 
         return (input_shape[0], self.num_filters, output_length)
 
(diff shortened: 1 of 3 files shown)


benanne:
Thanks!

As  @f0k mentioned in #154, it's probably cleaner to use the 'nice' formula that's mentioned in the comments in the `ConvOp` Theano code that I linked. We have no real reason to resort to integer arithmetic here. Or if you'd prefer to keep the formulas as is, there should also be a comment explaining where the formulas come from, like in the Theano code. That way we'll still know what's going on here in six months :)

ebenolson:
I think it's cleaner to use the same formula as in Theano, but I'll add some comments on the derivation. 

Correct me if I'm wrong, but I think there is a similar error for `border_mode=='same'`, it should be `(i + d - 1) // d`?

Conv1DLayer probably needs the same fix as well, should I do that here or make a separate PR?

benanne:
Feel free to fix all of them :) Some tests for this would also be a welcome addition, but we can take care of this later if necessary.

f0k:
> I think it's cleaner to use the same formula as in Theano

As you already found out, we have an additional border mode here. I think having three formulas of the tricky kind isn't too helpful. Wouldn't it be much easier to do something like (this is untested):
```python
filter_size = np.array(self.filter_size)
input_shape = np.array(input_shape)
strides = np.array(self.strides).astype(np.float)
if self.border_mode = 'full':
    padding = filter_size
elif self.border_mode = 'same':  # should have been 'half', in my opinion
    padding = filter_size // 2
elif self.border_mode = 'valid':
    padding = 0
output_size = np.ceil((input_shape[2:] + padding - filter_size) / strides).astype(np.int)
return (input_shape[0], self.num_filters) + tuple(output_size)
```
This should work for 1D, 2D, and 3D, so it could become a central helper function. And it would be easy to extend for the case of custom padding that will be added to Theano anyway (where `border_mode` is either an integer or a tuple of integers).

> Some tests for this would also be a welcome addition, but we can take care of this later if necessary.

+1! This would make sure the function does the right thing.

ebenolson:
That seems to have a couple of bugs in it, but a helper function is probably the way to go. 

I think it's only tricky because of the integer round-up, but with a comment to explain that it seems cleaner to me than multiple casts and using numpy.

f0k:
> That seems to have a couple of bugs in it

Yes, I guess so. I just wanted to illustrate the idea of using `border_mode` to set the padding and then use a common formula, rather than having a separate formula for each border mode (and then another one for custom padding when it's there). And the idea of having a single equation for each dimension, so the code can be shared between 1D, 2D, and 3D (whether it's using numpy, a list comprehension or a for loop is not crucial).

> I think it's only tricky because of the integer round-up, but with a comment to explain that it seems cleaner to me than multiple casts and using numpy.

Okay, I don't have a strong opinion on that, I'd welcome any solution that fixes the bug and is comprehensible :) Do you want to continue on that in this PR? Things left to do would be:
* [ ] also fix border_mode='same'
* [ ] fix the 1D and 3D case
* [ ] comment on the formula
* [ ] add tests (we can do that ourselves if needed)

ebenolson:
Sure, I'm a little busy the next few days, but I should be able to do it by the weekend.

benanne:
Awesome, thanks for contributing :)

ebenolson:
Unfortunately, the tests I added show some new issues.

Currently the Theano-backed conv layer implements "same" by trimming a strided "full" convolution.
To get the proper output, instead it should first trim the convolution, then perform striding. 

The cuda-convnet version also has issues, but I'm less sure what's going on there.

benanne:
```
instead it should first trim the convolution, then perform striding.
```

How does that work though? If I understood correctly that means you would compute a convolution with stride=1 and then throw away most of the result, which seems very wasteful.

The reason it's currently implemented the way it is, is to be able to do same convolutions with as little overhead as possible. Luckily the Theano convolution implementations are going to be cleaned up sometime soon ( https://github.com/Theano/Theano/issues/2505 ) so in the near future this 'hack' should no longer be necessary.

ebenolson:
Yeah, I think raising `NotImplementedError

benanne:
I see you always raise now for `border_mode='same'`, I was thinking it should only raise when `border_mode == 'same' and stride > 1`. But I guess there's something to be said for not implementing it at all, since the implementation is suboptimal. Anyway, so as not to break too many people's code, it would be great to have it working when the stride is 1, for now.

Once the Theano convolution interface has been cleaned up this will not be an issue anymore anyway, so I think that's the best solution in the meantime :)

ebenolson:
Oops, fixed.

f0k:
@benanne: I think you need to retrigger Travis, the build has stalled.

benanne:
If we want to rename this let's create a separate issue / pull request for that, so as not to draw out this one :) I've restarted Travis.

benanne:
This looks good to me, and Travis passes :) So unless there are any further comments I will merge this soon. It may take a few days because I'm currently abroad.

benanne:
Merged, thanks again!

PR #160, opened 2015-03-09
Document variable `InputLayer` shape dimensions

f0k:
It wasn't documented that the `shape` of `InputLayer` may include `None`s, nor what that would be useful for.


benanne:
Thanks!

PR #165, opened 2015-03-12
Added ReluNormal initializer from PReLU paper

dimatura:
Hi, this is the initialization scheme they use in the parametric ReLU paper. If there's interest I could add more tests. I wasn't sure about the name, I chose this once since the idea is that it's for convolutional layers with ReLU units and it's a normal.

### DIFF
--- a/lasagne/init.py
+++ b/lasagne/init.py
@@ -109,3 +109,25 @@ class Orthogonal(Initializer):
         q = u if u.shape == flat_shape else v
         q = q.reshape(shape)
         return floatX(self.gain * q)
+
+
+class ReluNormal(Initializer):
+    """
+    Initializer based on He et al, "Delving Deep into Rectifiers:
+        Surpassing Human-Level Performance on Imagenet Classification".
+    """
+
+    def __init__(self):
+        pass
+
+    def sample(self, shape):
+        if len(shape) < 2:
+            raise RuntimeError("Only shapes of length 2 or more are "
+                               "supported.")
+        # Same assumptions about meaning of different dimensions as Uniform.
+        # eg, for a Conv2D layer with spatial 5x5 dimensions this would be 25.
+        receptive_field_size = np.prod(shape[2:])
+        input_channels = shape[1]
+        n_l = input_channels*receptive_field_size
+        std = np.sqrt(2.0/(n_l))
+        return floatX(np.random.normal(0, std, size=shape))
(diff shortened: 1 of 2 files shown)


benanne:
Thanks! We were actually planning to refactor the current `Uniform` initializer to have two versions: one where you directly set the range to sample from, and one where the range is derived from fanin / fanout and you specify a gain. We should probably do the same for the `Normal` initializer. Once we've done that, the `ReLUNormal` initializer just becomes a special case of this, and we probably don't need a separate class for it.

We'd probably also need an option to switch between estimation based on fanin, fanout, or both as well. 

What does everyone think? I'd like to keep this simple and avoid adding more classes for every flavour of initialization strategy, because there is a lot of overlap there.

ebenolson:
I think there's nothing wrong with convenience classes, but they should simply subclass a shared base class and set parameters. Probably `Uniform` and `Normal` can also share code by having a param specifying the distribution.

How important is incorporating fanout in practice? It seems like there will be issues trying to do it in the current way - the initializer is called as soon as the layer is created, so fanout information is not available.

Perhaps it would be better to factor that into a helper function that works on the net as a whole. A similar approach could be used to implement stuff like http://arxiv.org/abs/1412.6558

f0k:
> How important is incorporating fanout in practice? It seems like there will be issues trying to do it in the current way - the initializer is called as soon as the layer is created, so fanout information is not available.

Often it can be deducted from the weight shape, but if a pooling layer follows, then you get it wrong.

> Perhaps it would be better to factor that into a helper function that works on the net as a whole.

Agreed. It would be cool to have a function that does a pass over a network and reinitializes the weights by calling an initializer with the correct fan-in and fan-out factors as well as the correct nonlinearity (where needed). This shouldn't be the only way to initialize weights, but an additional way for perfectionists. It might be useful if `get_all_layers()` returned the layers in topographic order for this. (hint, hint ;) )

To accomplish this it would be nice to have a generic `Uniform` and `Normal` that you call with a range or mean and stddev, and then subclasses of that such as `Glorot` and `He` which you call with a `fan_in`, `fan_out` (optionally left blank to derive the factors from the weight shape) and `gain` (optionally set to a nonlinearity name as in the `Orthogonal` initializer).

dimatura:
Ok, makes sense. On having convenience classes, I agree with @f0k that having too many kwargs is hard to remember (and type). As @ebenolson says, perhaps having a few base classes with all the functionality and a few convenience classes for the common cases seems the best.

benanne:
Maybe we should have a separate issue to come to a consensus about a good interface? @f0k and @ebenolson's remarks make a lot of sense, but this needs some fleshing out before we start implementing every possible variant, I reckon :)

benanne:
I'll close this since it's being superseded by #191.

PR #169, opened 2015-03-16
Make Conv layers support None in input shape

ebenolson:
Adds test coverage of `None` in the b01 dimensions of Conv layers, and updates those that needed fixes to support it.

This partially resolves https://github.com/benanne/Lasagne/issues/166, although it needs to be documented somewhere that this is the proper way to handle variable input size, since until https://github.com/Theano/Theano/issues/2620 is fixed, it will still be possible to get results of the wrong shape without an error.

### DIFF
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -15,6 +15,8 @@ __all__ = [
 
 def conv_output_length(input_length, filter_length,
                        stride, border_mode, pad=0):
+    if input_length is None:
+        return None
     if border_mode == 'valid':
         output_length = input_length - filter_length + 1
     elif border_mode == 'full':
@@ -103,7 +105,7 @@ class Conv1DLayer(Layer):
                                       filter_shape=filter_shape,
                                       border_mode='full')
             shift = (self.filter_length - 1) // 2
-            conved = conved[:, :, shift:input_shape[2] + shift]
+            conved = conved[:, :, shift:input.shape[2] + shift]
         else:
             raise RuntimeError("Invalid border mode: '%s'" % self.border_mode)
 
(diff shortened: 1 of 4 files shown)


f0k:
Looks good to me! There should probably be a test for Conv1D as well, but there is none yet, so that falls into our general test-coverage duties and is outside the scope of this PR.

> it needs to be documented somewhere that this is the proper way to handle variable input size

It's documented in `InputLayer`, which is where you specify the input shape (this has only been added recently, though, in case you missed it). It will also be mentioned in the tutorial when there is one. Would you document this anywhere else?

ebenolson:
> It's documented in InputLayer, which is where you specify the input shape (this has only been added recently, though, in case you missed it). It will also be mentioned in the tutorial when there is one. Would you document this anywhere else?

Ah ok, that seems fine. I would maybe add a note in the `InputLayer` docs that some Layers require the input shape to be known, the errors can be a bit cryptic.

f0k:
> some Layers require the input shape to be known, the errors can be a bit cryptic.

Good point. That should probably also be courtesy of these layers (e.g., the `DenseLayer` should raise a meaningful error if any dimension besides the batch dimension is unspecified). But as the shapes are ultimately specified in the `InputLayer`, I agree an extra remark there wouldn't hurt.

benanne:
Thanks for this! Do you want to add the extra documentation in this PR? Otherwise I'll merge this soon if there are no further comments.

f0k:
> Do you want to add the extra documentation in this PR?

Hmm, I don't think this belongs here. We'll have to go through all layer implementations and add notes which (if any) shapes have to be specified, plus raise an exception if they aren't.

benanne:
Oh, I thought we were talking about the extra remark in the `InputLayer` docs. But maybe that's outside of the scope of this PR as well.

benanne:
Merging!

PR #170, opened 2015-03-18
Remove *args from get_output() and get_output_for()

f0k:
This removes `*args` from `get_output()` and `get_output_for()`, leaving only `**kwargs` for customizations.

@benanne stated that the `*args` [were there intentionally](https://github.com/benanne/Lasagne/issues/115#issuecomment-71394148), but without an existing use case. Later in the same thread we established that we should not include `*args` [for the layer constructor](https://github.com/benanne/Lasagne/issues/115#issuecomment-71466435) to avoid people writing code that can break in future (or, put differently, to allow us to add additional arguments without breaking people's existing code).

In any case, please see this as a proposal to be discussed rather than a request to be merged, it just took 5 minutes.

### DIFF
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -91,7 +91,7 @@ class Layer(object):
         """
         return self.get_output_shape_for(self.input_shape)
 
-    def get_output(self, input=None, *args, **kwargs):
+    def get_output(self, input=None, **kwargs):
         """
         Computes the output of the network at this layer. Optionally, you can
         define an input to propagate through the network instead of using the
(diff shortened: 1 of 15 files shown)


benanne:
You're probably right. If any important use cases for this resurface we could always re-add it later, in the meantime removing it might save us a headache or two. I'm in favour.

benanne:
Anybody else want to chime in on this? If not I'll merge it tomorrow.

benanne:
and it's gone!

f0k:
> and it's gone!

and it's gone

Issue #166, opened 2015-03-14
Conv2DLayer gives the wrong (?) shape when mode='same'


craffel:
I ran across this when showing @bmcfee some lasagne stuff.  Basically, when a Conv2DLayer has `mode='same'` and you pass it an input where the second-to-last (or last, I think, but I'm not testing that here) dimension is different than expected, it gives you an output whose shape is what it had expected.  An example is probably easier to understand:

```Python
import numpy as np
import theano
import lasagne
import lasagne.layers.dnn
import lasagne.layers.cuda_convnet
import itertools

# Create network, starting with input layer
l_in = lasagne.layers.InputLayer(shape=(None, 16, 100, 20))
x_same = np.random.randn(1, 16, 100, 20).astype(theano.config.floatX)
x_diff = np.random.randn(1, 16, 200, 20).astype(theano.config.floatX)

for layer, mode in itertools.product([lasagne.layers.Conv2DLayer,
                                      lasagne.layers.cuda_convnet.Conv2DCCLayer,
                                      lasagne.layers.dnn.Conv2DDNNLayer],
                                     ['valid', 'full', 'same']):
    l_1 = layer(l_in, num_filters=16, filter_size=(5, 5), border_mode=mode)
    out = theano.function([l_in.input_var], l_1.get_output())
    for x, shape in zip([x_same, x_diff], ['same', 'diff']):
        print "{}, {}, {}, {}".format(layer, mode, shape, out(x).shape)
```

yields

```
<class 'lasagne.layers.conv.Conv2DLayer'>, valid, same, (1, 16, 96, 16)
<class 'lasagne.layers.conv.Conv2DLayer'>, valid, diff, (1, 16, 196, 16)
<class 'lasagne.layers.conv.Conv2DLayer'>, full, same, (1, 16, 104, 24)
<class 'lasagne.layers.conv.Conv2DLayer'>, full, diff, (1, 16, 204, 24)
<class 'lasagne.layers.conv.Conv2DLayer'>, same, same, (1, 16, 100, 20)
<class 'lasagne.layers.conv.Conv2DLayer'>, same, diff, (1, 16, 100, 20)
<class 'lasagne.layers.cuda_convnet.Conv2DCCLayer'>, valid, same, (1, 16, 96, 16)
<class 'lasagne.layers.cuda_convnet.Conv2DCCLayer'>, valid, diff, (1, 16, 196, 16)
[...]

PR #172, opened 2015-03-18
added None propagation to pooling operators

bmcfee:
I haven't tested thoroughly (not sure how to?) but it appears to push out data of the correct shape, following Colin's example in #166 with a few modifications.

There are a few bits of code that look like they code be simplified:
* forced integer division inside `pool_output_length` (copied from `get_output_shape_for` previously)
* the `ignore_border` check

I'm happy to fix those too prior to merging.

### DIFF
--- a/lasagne/layers/pool.py
+++ b/lasagne/layers/pool.py
@@ -14,6 +14,28 @@ __all__ = [
 ]
 
 
+def pool_output_length(input_length, pool_length):
+    '''Compute the output length of a pooling operator.
+
+    Parameters
+    ----------
+    input_length 
+    pool_length
+
+    Returns
+    -------
+    output_length
+        * None if either input is None
+        * `input_length / pool_length` otherwise
+    '''
+
+    if input_length is None or pool_length is None:
+        return None
+
+    # Should this just be "input_length // pool_length" ?
+    return int(np.floor(float(input_length) / pool_length))
+
+
 class MaxPool2DLayer(Layer):
     def __init__(self, incoming, ds, ignore_border=False, **kwargs):
         super(MaxPool2DLayer, self).__init__(incoming, **kwargs)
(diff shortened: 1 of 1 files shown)


craffel:
I think `pool_output_length` needs to take in a parameter `ignore_border` to decide whether to `floor` or `ceil`.  Can you add this to all of the other pooling classes (`MaxPool2DCCLayer`, `MaxPool2DDNNLayer`) too?  I think this could also cause pukes with `None` dimensions:
https://github.com/benanne/Lasagne/blob/master/lasagne/layers/pool.py#L66

bmcfee:
> I think pool_output_length needs to take in a parameter ignore_border to decide whether to floor or ceil.

Derp, I missed that difference.  Easily fixed though.


> Can you add this to all of the other pooling classes (MaxPool2DCCLayer, MaxPool2DDNNLayer) too? I think this could also cause pukes with None dimensions:

I could, but i have no way to test it.  (DNN doesn't run on my older gpu machine.)

lasagne/layers/pool.py
```diff
@@ -14,6 +14,39 @@
...
+    if input_length is None or pool_length is None:
+        return None
+
+    if ceil:
+        rounder = np.ceil
+    else:
+        rounder = np.floor
+
+    # Should this just be "input_length // pool_length" ?
+    return int(rounder(float(input_length) / pool_length))
```
f0k: What about:
```python
if input_length is None or pool_length is None:
    return None
if ignore_border:
    # round down
    return input_length // pool_length
else:
    # round up
    return (input_length - 1) // pool_length + 1
```
(At least I'd suggest to change `ceil=False` to `ignore_border=False`, with inverted meaning, to match the `MaxPool2DLayer` constructor. I'm probably picky here, but `ceil` refers to the implementation of the shape computation, not to the operation performed by a pooling layer.)

lasagne/layers/pool.py
```diff
@@ -24,15 +57,13 @@ def get_output_shape_for(self, input_shape):
...
+            output_shape[3] = pool_output_length(input_shape[3], self.ds[1])
         else:
-            output_shape[2] = int(np.ceil(float(output_shape[2]) /
-                                          self.ds[0]))
-            output_shape[3] = int(np.ceil(float(output_shape[3]) /
-                                          self.ds[1]))
+            output_shape[2] = pool_output_length(input_shape[2], self.ds[0],
+                                                 ceil=True)
+            output_shape[3] = pool_output_length(input_shape[3], self.ds[1],
+                                                 ceil=True)
```
f0k: That bit could then be turned into:
```
output_shape[2] = pool_output_length(input_shape[2], self.ds[0], ignore_border=self.ignore_border)
output_shape[3] = pool_output_length(input_shape[3], self.ds[1], ignore_border=self.ignore_border)
```

bmcfee:
@f0k yup, those all make sense. Thanks!

f0k:
Thanks, looks good to me! Tests and adaption of the other pooling layers would be welcome, but that can also become another PR, we'll take what we get :)

benanne:
Looks good. To apply it to the other pooling layers, we'll need to take into account the case where the strides differ from the pool window size as well, but I guess that can be handled in a later PR. I'll merge this tomorrow if there are no further comments.

Issue #104, opened 2015-01-21
Layer outputs being computed multiple times


benanne:
Lasagne `Layers` compute their outputs by calling `get_output()` recursively on their parents. When a network has a tree structure, which is usually the case, this works fine. But when a network has a directed acyclic graph (DAG) structure, i.e. some layers get their input from the same layer, this leads to `get_output()` being called multiple times on this layer. As a result the output expression of this shared input layer (and all layers further down) is constructed multiple times.

In terms of performance this is not an issue since all these computations are symbolic anyway. We count on Theano's smarts to notice that these sub-expressions are the same, so it only computes them once at execution time. This works great in most cases, but unfortunately there are exceptions:

* it fails when scan is used (see https://github.com/benanne/Lasagne/issues/17#issuecomment-70317085 and onwards, also https://github.com/benanne/Lasagne/issues/97#issuecomment-70718707). ~~This is because Theano does not seem to merge two scan operations that do the same computations.~~
EDIT: this is wrong, see @craffel's comment below. Theano does merge the scan operations, but it doesn't merge their *gradients*, which is also a problem.

* it fails when layer outputs are nondeterministic (because of e.g. dropout), see #97. Random variates are re-sampled for each call to `get_output()`, so Theano rightly does not merge the sub-expressions here. But of course this isn't what we want.

* there may be other cases that have not come up yet or have not been discovered yet.

If we are able to ensure that `get_output()` is called only once on each layer in a DAG, we can solve this issue. There are a couple of ways to do this:

* memoization: make `get_output_for()` remember the output associated with a given input after it's been computed for the first time. The second time it just returns the remembered output (which will be the same Theano expression). Doing this library-wide may result in some gotchas (people expecting code to be executed that will no longer be executed, etc.), so I don't know if this is a good idea. Making it a per-layer thing that should be enabled or disabled seems like it wouldn't solve the problem adequately (and is hard to implement as well), and also adds a cognitive burden for the user.

* making `get_output()` iterative instead of recursive: currently, `get_output()` calls `get_output()` on the parents and then feeds the result to `get_output_for()`. Instead of recursively computing output expressions like this, we could make this function iterative. We would still have to recurse to grab all the 'ancestors' of the layer (we can use `get_all_layers()` for this), but after that we can ensure that `get_output_for()` is only called once for each of them, even when the network structure is a DAG.

* there may be other ways to solve this, ideas are welcome!

[...]

PR #176, opened 2015-03-23
Topological order for get_all_layers()

f0k:
This renames the current implementation of `lasagne.layers.get_all_layers()` to `lasagne.layers.get_all_layers_old()` and installs a new implementation `lasagne.layers.get_all_layers()` which is guaranteed (and tested) to return layers in topological order. This is a requirement for #104, among others.

It currently gives a warning whenever it is called, along with a note on how to disable that. The warning serves to notify all users of the transition and `get_all_layers_old()` eases it, but both should be removed for the first release of Lasagne.

### DIFF
--- a/lasagne/layers/helper.py
+++ b/lasagne/layers/helper.py
@@ -1,3 +1,5 @@
+from collections import deque
+
 import numpy as np
 
 from .. import utils
@@ -5,6 +7,7 @@ from .. import utils
 
 __all__ = [
     "get_all_layers",
+    "get_all_layers_old",
     "get_all_params",
     "get_all_bias_params",
     "get_all_non_bias_params",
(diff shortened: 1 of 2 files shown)


benanne:
Beautiful :) I really appreciate all the comments, it's very easy to understand this way. Tests look great too.

Since this is a drastic change and we will have to remove the warning / old implementation in about a month if we want to release on schedule, I will merge this ASAP. If there are no further comments I will merge it tomorrow afternoon.

craffel:
Looks cool! +1 on nice comments

f0k:
Thanks :) The only improvement I can think of would be swapping the `elif layer in seen` and `else` cases so they come in chronological order (when thinking of what happens at runtime). Just to make the story easier to read.

f0k:
It's tomorrow afternoon! :) I can cope with the "[interface change]" announcement on lasagne-users if needed.

benanne:
That would be great :) go ahead! Merging now.

f0k:
Announced here: https://groups.google.com/forum/#!topic/lasagne-users/mkCXB7sdWIc

To document it on github as well, a useful code snippet to read or convert old models is the following:
```python
def permute_param_values(layer, values):
    params_new = lasagne.layers.get_all_params(layer)
    params_old = sum((l.get_params()
                     for l in lasagne.layers.get_all_layers_old(layer)), [])
    return [values[params_old.index(param)] for param in params_new]
```
It permutes a parameter value list of an old model to match the new order, so you can use it both for reading old models:
`set_all_param_values(layer, permute_param_values(layer, values))`
and for converting old models:
`your_write_method(filename, layer, permute_param_values(your_read_method(filename, layer)))`
Where `layer` is your network's output layer or list of output layers.

PR #177, opened 2015-03-25
Make set_all_param_values() check the number of provided values

f0k:
I just got tricked by Lasagne staying silent when `set_all_param_values()` was called with an empty `values` list. This PR throws an exception if there are not enough or too many values.


benanne:
Good call, thanks. Since this is a small fix and Travis passes I'll just go ahead and merge it.

PR #178, opened 2015-03-26
Remove lasagne.utils.concatenate

f0k:
`lasagne.utils.concatenate` has been deprecated for a month, we should remove it before we forget.


benanne:
baleeted!

PR #179, opened 2015-03-27
change requirements spec to depend on the new Theano release (0.7)

benanne:
Doing this as a PR so Travis can do its thing :)


benanne:
seems to work :)

Issue #87, opened 2015-01-05
Saxe style gain for Uniform Glorot Initialization


cancan101:
Currently it exists for [Orthogonal](https://github.com/benanne/Lasagne/blob/6f23af2cd948b00a30c1428798a73ce0ab123dfb/lasagne/init.py#L95) but not `Uniform`.

benanne:
Good point. It might actually be cleaner to split up the current `Uniform` initializer into two separate classes, one that takes a range as input, and one that does the Glorot-style initialization (with an optional gain argument).

cancan101:
I agree. I think composition is cleaner than having a lot of extra arguments to something a simple as `Uniform`.

benanne:
This has been addressed in #180. You can specify a gain when using the `GlorotUniform` or `GlorotNormal` initializers.

Issue #49, opened 2014-10-21
default initialization for `Conv2DCCLayer` does not make sense when dimshuffle=False


benanne:
The default initialization for the cuda-convnet based 2D convolution (`Conv2DCCLayer`) is `init.Uniform` without any parameters, which defaults to the "Glorot"-style uniform initialization based on fan in and fan out.

If `Conv2DCCLayer` is used with `dimshuffle=False`, it uses `c01b` ordering for the input dimensions instead of the default of `bc01`. This causes the initializer to compute the fan in and fan out incorrectly: it computes the receptive field size as the product of the last two dimensions, which is correct for `bc01` but not for `c01b`.

The easiest solution is probably to define a custom `Uniform_c01b` initializer and use that as the default instead when `dimshuffle=False`. I'll try to sort that out soon, unless someone has a better idea for this.

f0k:
> I'll try to sort that out soon, unless someone has a better idea for this.

In #87, we more or less decided to do a separate Glorot initializer. In addition to an optional `gain` argument, it could also have optional `fan_in` and `fan_out` arguments overriding the default computation of those from the shape. `Conv2DCCLayer` could compute the fan in and fan out and pass that to the `init.Glorot` then. This won't work with the current way of creating the default initializer directly in the function declaration, though... if we want to keep this, deriving a `Glorot_c01b` initializer from `Glorot` will be the easiest solution.
While we're at it, what about 1D and 3D convolution? Is everything correct there? If not, we might want to go for the `fan_in` and `fan_out` arguments after all, instead of creating a bunch of specialized `Glorot` subclasses.

benanne:
Everything should be fine for 1D and 3D convolution for now, the fan in is just computed as the product of all trailing dimensions beyond the 1st in the bc01 case, so that works out okay.

Are you proposing to add `fan_in` and `fan_out` as optional arguments for initializers? Because for the Orthogonal initializer, I don't think this is sufficient, you actually need to pass in the set of axes that constitute the fan-in in that case.

f0k:
> Are you proposing to add `fan_in` and `fan_out` as optional arguments for initializers? Because for the Orthogonal initializer, I don't think this is sufficient, you actually need to pass in the set of axes that constitute the fan-in in that case.

[...]

PR #180, opened 2015-03-28
Revamp lasagne.init

benanne:
This PR cleans up the `lasagne.init` module. It introduces separate classes for "Glorot-style" initialization, `GlorotUniform` and `GlorotNormal`.

This means `Uniform` no longer does Glorot-style init by default. Instead, it will use a default range of (-0.01, 0.01). I've added a warning to alert people of this, in case they're using `Uniform` in this way. Their training runs might suddenly take a lot longer to converge!

Accordingly, I've also changed the default initialization to `GlorotUniform` everywhere in the library.

I've added separate `GlorotUniform_c01b` and `GlorotNormal_c01b` initializers for the c01b case (useful for cuda-convnet-based layers). We discussed a fancier approach using `fan_in` and `fan_out` parameters in #49, but for now I don't think this will be worth it. Having separate classes for the `c01b` case seemed like a simpler approach, although it does lead to a little bit of code duplication.

This PR also addresses the bug reported in #49: the default initializer for `W` in `lasagne.layers.cuda_convnet.Conv2DCCLayer` is now `GlorotUniform` or `GlorotUniform_c01b`, depending on whether `dimshuffle` is `True` or `False`.

In the `mnist_conv_cc.py` example, I've removed all explicit specifications of initializers for `W`, because this should now depend on whether `dimshuffle` is `True` or `False`. I believe they were only specified explicitly because the default was broken to begin with.

This also addresses #87: the Glorot-style initializers have a `gain` parameter, like `Orthogonal` already had before.

Glorot-style initialization means that the standard deviation of the initialization distribution is proportional to `2/(fan_in + fan_out)`. Some people prefer to use `1/fan_in` instead, but this PR does not address that. Does anyone think this is worth adding as well? I think this was basically the idea of PR #165.

Feedback is welcome!

### DIFF
--- a/lasagne/init.py
+++ b/lasagne/init.py
@@ -18,46 +18,16 @@ class Initializer(object):
 
 
 class Normal(Initializer):
-    def __init__(self, std=0.01, avg=0.0):
+    def __init__(self, std=0.01, mean=0.0):
         self.std = std
-        self.avg = avg
+        self.mean = mean
 
     def sample(self, shape):
-        return floatX(np.random.normal(self.avg, self.std, size=shape))
-
-
-class GlorotNormal(Normal):
-    def __init__(self, gain=1.0):
-        if gain == 'relu':
-            gain = np.sqrt(2)
-
-        self.gain = gain
-
-    def sample(self, shape):
-        # This code makes some assumptions about the meanings of
-        # the different dimensions, which hold for
-        # layers.DenseLayer and layers.Conv*DLayer, but not
-        # necessarily for other layer types.
-        n1, n2 = shape[:2]
-        receptive_field_size = np.prod(shape[2:])
-        std = self.gain * np.sqrt(2.0 / ((n1 + n2) * receptive_field_size))
-        return floatX(np.random.normal(0.0, std, size=shape))
-
-
-class GlorotNormal_c01b(GlorotNormal):
-    def sample(self, shape):
-        if len(shape) != 4:
-            raise RuntimeError(
-                "This initializer only works with shapes of length 4")
-
-        n1, n2 = shape[0], shape[3]
-        receptive_field_size = shape[1] * shape[2]
-        std = self.gain * np.sqrt(2.0 / ((n1 + n2) * receptive_field_size))
-        return floatX(np.random.normal(0.0, std, size=shape))
+        return floatX(np.random.normal(self.mean, self.std, size=shape))
 
 
 class Uniform(Initializer):
-    def __init__(self, range=0.01):
+    def __init__(self, range=0.01, std=None, mean=0.0):
         import warnings
         warnings.warn("The uniform initializer no longer uses Glorot et al.'s "
                       "approach to determine the bounds, but defaults to the "
@@ -65,7 +35,11 @@ class Uniform(Initializer):
                       "GlorotUniform initializer to get the old behavior. "
                       "GlorotUniform is now the default for all layers.")
 
-        if isinstance(range, Number):
+        if std is not None:
+            a = mean - np.sqrt(3) * std
+            b = mean + np.sqrt(3) * std
+            self.range = (a, b)
+        elif isinstance(range, Number):
             self.range = (-range, range)
         else:
             self.range = range
@@ -75,25 +49,32 @@ class Uniform(Initializer):
             low=self.range[0], high=self.range[1], size=shape))
 
 
-class GlorotUniform(Uniform):
-    def __init__(self, gain=1.0):
+class Glorot(Initializer):
+    def __init__(self, initializer=Normal, gain=1.0):
         if gain == 'relu':
             gain = np.sqrt(2)
 
+        self.initializer = initializer
         self.gain = gain
 
     def sample(self, shape):
-        # This code makes some assumptions about the meanings of
-        # the different dimensions, which hold for
-        # layers.DenseLayer and layers.Conv*DLayer, but not
-        # necessarily for other layer types.
         n1, n2 = shape[:2]
         receptive_field_size = np.prod(shape[2:])
-        m = self.gain * np.sqrt(6.0 / ((n1 + n2) * receptive_field_size))
-        return floatX(np.random.uniform(low=-m, high=m, size=shape))
+        std = self.gain * np.sqrt(2.0 / ((n1 + n2) * receptive_field_size))
+        return self.initializer(std=std).sample(shape)
 
 
-class GlorotUniform_c01b(GlorotUniform):
+class GlorotNormal(Glorot):
+    def __init__(self, gain=1.0):
+        super(GlorotNormal, self).__init__(Normal, gain)
+
+
+class GlorotUniform(Glorot):
+    def __init__(self, gain=1.0):
+        super(GlorotUniform, self).__init__(Uniform, gain)
+
+
+class Glorot_c01b(Glorot):
     def sample(self, shape):
         if len(shape) != 4:
             raise RuntimeError(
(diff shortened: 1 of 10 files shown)


craffel:
> Some people prefer to use 1/fan_in instead, but this PR does not address that. Does anyone think this is worth adding as well?

I have used initialization schemes like this before; I think when it only depends on `fan_in` or `fan_out`, it's easy enough to do yourself and is probably not worth having a special initializer.

ebenolson:
off topic, but what would you think about renaming this module to something like initializers.py? 

it's a bit strange to have an ``init.py`` that's not ``__init.py__``.

benanne:
I don't really see a huge problem with having __init__ and init, the distinction is pretty clear imo.  Also I don't like typing :) Has it confused you in the past?

ebenolson:
It's not a big deal but I was surprised the first time I opened that file.

f0k:
+1 for cleaning this up!

In your proposal, it doesn't really make sense to derive `GlorotUniform` from `Uniform`, though, as you don't have a single super call in there and have zero overlap in the attributes (same for `GlorotNormal` and `Normal`). Also, as you noted, there's still quite some code duplication. Maybe this could be decoupled a bit?

We currently have two initializers that fill a weight matrix with iid. random samples following a given mean and standard deviation (although the `Uniform` initializer doesn't provide that interface for now), and we have one scheme that suggests a particular mean and standard deviation depending on the weight matrix shape and nonlinearity (Glorot). Another such scheme has been proposed in [Delving Deep into Rectifiers](http://arxiv.org/abs/1502.01852). In principle, these could be freely combined. Should we try to find a solution that does not need six classes with duplicated code (Uniform, GlorotUniform, HeUniform, Normal, GlorotNormal, HeNormal)? I.e., either a solution with six classes and minimal code duplication, or a solution with four classes to be combined by the user?

benanne:
> Should we try to find a solution that does not need six classes with duplicated code

This was my original goal when I started working on it, which is why `GlorotUniform` derives from `Uniform`. But the problem with this is that for `GlorotUniform`, you don't know the mean and variance until `sample()` is called, because `shape` is unknown when the initializer is instantiated. So this makes it a bit more challenging to do it cleanly. I suppose `GlorotUniform` could simply create a new `Uniform` instance every time `sample()` is called, with the right parameters, i.e. composition instead of inheritance.

I would keep 6 separate initializers though, the code is easy enough to understand so people can always write their own versions if they want to. The additional cognitive overhead of having to combine these initializer 'traits' into a valid initializer somehow is probably a bit too much for something this simple (and for such a small reduction in code).

f0k:
> I suppose `GlorotUniform` could simply create a new `Uniform` instance every time `sample()` is called, with the right parameters

Sounds reasonable. What about a solution like this:
```python
class Uniform:
    def __init__(self, a=-0.01, b=0.01, mean=None, std=None):
        if mean is not None and std is not None:
            a = ...
            b = -a
class Glorot:
    def __init__(self, initializer, gain=1):
        ...
    def sample(self, shape):
        std = ...
        return self.initializer(mean=0, std=std).sample(shape)
class GlorotUniform(Glorot):
    def __init__(self, gain=1):
        super(GlorotUniform, self).__init__(Uniform, gain)
class GlorotNormal(Glorot):
    def __init__(self, gain=1):
        super(GlorotNormal, self).__init__(Normal, gain)
```

> I would keep 6 separate initializers though, the code is easy enough to understand so people can always write their own versions if they want to.

The proposal above would allow both using a pre-made class or combining traits, and make it very obvious how to do your own version.

benanne:
looks good to me. The mean and std arguments overriding a and b if they are defined may raise some eyebrows, but I like how it makes it really simple to create new versions with your own distribution.

f0k:
> The mean and std arguments overriding a and b if they are defined may raise some eyebrows, but I like how it makes it really simple to create new versions with your own distribution.

A cleaner solution would be a `Uniform` subclass that takes `mean` and `std` instead of `a` and `b` and just issues the correct super call in the initializer. I'm just not sure what this subclass should be called then.

benanne:
There we go, things should now be a bit more modular. I made `mean` default to 0.0 everyhwere for convenience.

I haven't added classes for He-style init (i.e. only taking into account the fan-in and not the fan-out), but we can do that later if necessary. A minor annoyance with this is that the fan-in of a `DenseLayer` is `shape[0]`, whereas the fan-in of a `Conv*Layer` is `np.prod(shape[1:])`. So there is no one-size-fits-all code for this.

lasagne/init.py
```diff
@@ -18,12 +18,82 @@ def sample(self, shape):
...
+                      "GlorotUniform is now the default for all layers.")
+
+        if std is not None:
+            a = mean - np.sqrt(3) * std
+            b = mean + np.sqrt(3) * std
+            self.range = (a, b)
+        elif isinstance(range, Number):
+            self.range = (-range, range)
+        else:
+            self.range = range
```
f0k: That `isinstance` check looks a little brittle to me. For example, `isinstance(np.float32(10), Number)` returns `False`. What about:
```python
if std is not None:
    a = ...
    b = ...
else:
    try:
        a, b = range
    except (TypeError, ValueError):
        a, b = -range, range
self.range = (a, b)
```

lasagne/tests/test_init.py
```diff
@@ -15,55 +15,137 @@ def test_normal():
...
-    from lasagne.init import Constant
+def test_uniform_range_as_number():
+    from lasagne.init import Uniform
 
-    sample = Constant(1.0).sample((10, 20))
-    assert (sample == 1.0).all()
+    sample = Uniform(1.0).sample((300, 400))
+    assert sample.shape == (300, 400)
+    assert -1.1 < sample.min() < -0.9
+    assert 0.9 < sample.max() < 1.1
```
f0k: I'd do `assert -1.0 <= sample.min() < -0.9` and `assert 0.9 < sample.max() <= 1.0`. We can check the bounds more strictly.

lasagne/tests/test_init.py
```diff
@@ -15,55 +15,137 @@ def test_normal():
...
+def test_uniform_range_as_range():
+    from lasagne.init import Uniform
 
-    sample = Sparse(sparsity=0.5).sample((10, 20))
-    assert (sample == 0.0).sum() == (sample != 0.0).sum()
-    assert (sample == 0.0).sum() == (10 * 20) / 2
+    sample = Uniform((0.0, 1.0)).sample((300, 400))
+    assert sample.shape == (300, 400)
+    assert -0.1 < sample.min() < 0.1
+    assert 0.9 < sample.max() < 1.1
```
f0k: Same here, we can do a more strict check of the lower and upper bound.

lasagne/tests/test_init.py
```diff
@@ -15,55 +15,137 @@ def test_normal():
...
+    sample = Constant(1.0).sample((10, 20))
+    assert (sample == 1.0).all()
+
+
+def test_sparse():
+    from lasagne.init import Sparse
+
+    sample = Sparse(sparsity=0.5).sample((10, 20))
+    assert (sample == 0.0).sum() == (sample != 0.0).sum()
+    assert (sample == 0.0).sum() == (10 * 20) / 2
```
f0k: Hmm, what if someone changes `sparsity` to behave like `1 - sparsity`? That would go unnoticed. I'd prefer:
```python
sample = Sparse(sparsity=0.1).sample((10, 20))
assert (sample != 0.0).sum() == (10 * 20) * 0.1
```
(There's no need for a second assert unless you want to ensure `sample` has 10*20 items.)

f0k:
I like it! (Except for a few minor comments.)

benanne:
I've changed the number vs. tuple check and made the tests for uniform distributions more strict. Changing the sparse initializer test is sort of out of the scope of this PR but I did it anyway :)

We can add the He-style initializers in a separate PR.

f0k:
Great! +1 for merging!

> Changing the sparse initializer test is sort of out of the scope of this PR

Sorry, but you started it! :) You added the test in b237390.

> We can add the He-style initializers in a separate PR.

Sure, that's not vital for the first release anyway as it won't break compatibility. It's also a low-hanging fruit, which can be good to attract new contributors. (We should add a tag for that.)

benanne:
> Sorry, but you started it! :) You added the test in b237390.
Nope, that's just github's diffing getting confused because I moved a lot of stuff around to match the order of the classes in `init.py` :)

Will merge tomorrow if there are no further comments.

PR #182, opened 2015-03-30
turn get_output() into a global function

f0k:
To address #104, this moves `get_output()` out of the layer class hierarchy and turns it into a helper function `lasagne.layers.get_output()`. This is work in progress.
TODO:
* [x] move `get_output()`
* [x] rewrite tests for `get_output()`
* [x] move `get_output_shape()`
* [x] rewrite tests for `get_output_shape()`
* [x] modify examples

The current tests don't run through because they rely on everything-goes `Mock()` instances. That is, they always have a `input_layers` attribute, which tricks `get_all_layers()`. Changing `get_all_layers()` to first check for the `input_layer` attribute isn't better either, because every `Mock()` instance also has a `input_layer` attribute which is a `Mock()` instance with an `input_layer` attribute. Changing the `Mock()` instances to only have the attributes they should makes the tests fail in other ways because they assume `get_output()` to be called where the new code calls `get_output_for()` or returns `input_var`. Long story short, the tests need to be rewritten, with the current test suite we cannot check whether the refactoring did any harm.

### DIFF
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -93,40 +93,14 @@ class Layer(object):
 
     def get_output(self, input=None, **kwargs):
         """
-        Computes the output of the network at this layer. Optionally, you can
-        define an input to propagate through the network instead of using the
-        input variables associated with the network's input layers.
-
-        :parameters:
-            - input : None, Theano expression, numpy array, or dict
-                If None, uses the inputs of the :class:`InputLayer` instances.
-                If a Theano expression, this will replace the inputs of all
-                :class:`InputLayer` instances (useful if your network has a
-                single input layer).
-                If a numpy array, this will be wrapped as a Theano constant
-                and used just like a Theano expression.
-                If a dictionary, any :class:`Layer` instance (including the
-                input layers) can be mapped to a Theano expression or numpy
-                array to use instead of its regular output.
-
-        :returns:
-            - output : Theano expression
-                the output of this layer given the input to the network
-
-        :note:
-            When implementing a new :class:`Layer` class, you will usually
-            keep this unchanged and just override `get_output_for()`.
+        Deprecated. Use `lasagne.layers.get_output(layer, input, **kwargs)`.
         """
-        if isinstance(input, dict) and (self in input):
-            # this layer is mapped to an expression or numpy array
-            return utils.as_theano_expression(input[self])
-        elif self.input_layer is None:
-            raise RuntimeError("get_output() called on a free-floating layer; "
-                               "there isn't anything to get its input from. "
-                               "Did you mean get_output_for()?")
-        else:  # in all other cases, just pass the input on to the next layer.
-            layer_input = self.input_layer.get_output(input, **kwargs)
-            return self.get_output_for(layer_input, **kwargs)
+        import warnings
+        warnings.warn("layer.get_output(...) is deprecated and will be "
+                      "removed for the first release of Lasagne. Please use "
+                      "lasagne.layers.get_output(layer, ...) instead.")
+        from .helper import get_output
+        return get_output(self, input, **kwargs)
 
     def get_output_shape_for(self, input_shape):
         """
@@ -271,20 +245,6 @@ class MultipleInputsLayer(Layer):
 
     def get_output_shape(self):
         return self.get_output_shape_for(self.input_shapes)
-
-    def get_output(self, input=None, **kwargs):
-        if isinstance(input, dict) and (self in input):
-            # this layer is mapped to an expression or numpy array
-            return utils.as_theano_expression(input[self])
-        elif any(input_layer is None for input_layer in self.input_layers):
-            raise RuntimeError("get_output() called on a free-floating layer; "
-                               "there isn't anything to get its inputs from. "
-                               "Did you mean get_output_for()?")
-        # In all other cases, just pass the network input on to the next layers
-        else:
-            layer_inputs = [input_layer.get_output(input, **kwargs) for
-                            input_layer in self.input_layers]
-            return self.get_output_for(layer_inputs, **kwargs)
 
     def get_output_shape_for(self, input_shapes):
         """
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -36,10 +36,13 @@ class Layer(object):
             self.input_shape = incoming
             self.input_layer = None
         else:
-            from .helper import get_output_shape
-            self.input_shape = get_output_shape(incoming)
+            self.input_shape = incoming.output_shape
             self.input_layer = incoming
         self.name = name
+
+    @property
+    def output_shape(self):
+        return self.get_output_shape_for(self.input_shape)
 
     def get_params(self):
         """
@@ -78,14 +81,13 @@ class Layer(object):
 
     def get_output_shape(self):
         """
-        Deprecated. Use `lasagne.layers.get_output_shape(layer)`.
+        Deprecated. Use `layer.output_shape`.
         """
         import warnings
         warnings.warn("layer.get_output_shape() is deprecated and will be "
                       "removed for the first release of Lasagne. Please use "
-                      "lasagne.layers.get_output_shape(layer) instead.")
-        from .helper import get_output_shape
-        return get_output_shape(self)
+                      "layer.output_shape instead.")
+        return self.output_shape
 
     def get_output(self, input=None, **kwargs):
         """
@@ -231,14 +233,17 @@ class MultipleInputsLayer(Layer):
             - name : a string or None
                 an optional name to attach to this layer
         """
-        from .helper import get_output_shape
         self.input_shapes = [incoming if isinstance(incoming, tuple)
-                             else get_output_shape(incoming)
+                             else incoming.output_shape
                              for incoming in incomings]
         self.input_layers = [None if isinstance(incoming, tuple)
                              else incoming
                              for incoming in incomings]
         self.name = name
+
+    @Layer.output_shape.getter
+    def output_shape(self):
+        return self.get_output_shape_for(self.input_shapes)
 
     def get_output_shape_for(self, input_shapes):
         """
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -36,7 +36,8 @@ class Layer(object):
             self.input_shape = incoming
             self.input_layer = None
         else:
-            self.input_shape = incoming.get_output_shape()
+            from .helper import get_output_shape
+            self.input_shape = get_output_shape(incoming)
             self.input_layer = incoming
         self.name = name
 
@@ -77,19 +78,14 @@ class Layer(object):
 
     def get_output_shape(self):
         """
-        Computes the output shape of the network at this layer.
-
-        :returns:
-            - output shape: tuple
-                a tuple that represents the output shape of this layer. The
-                tuple has as many elements as there are output dimensions, and
-                the elements of the tuple are either integers or `None`.
-
-        :note:
-            When implementing a new :class:`Layer` class, you will usually
-            keep this unchanged and just override `get_output_shape_for()`.
+        Deprecated. Use `lasagne.layers.get_output_shape(layer)`.
         """
-        return self.get_output_shape_for(self.input_shape)
+        import warnings
+        warnings.warn("layer.get_output_shape() is deprecated and will be "
+                      "removed for the first release of Lasagne. Please use "
+                      "lasagne.layers.get_output_shape(layer) instead.")
+        from .helper import get_output_shape
+        return get_output_shape(self)
 
     def get_output(self, input=None, **kwargs):
         """
@@ -235,16 +231,14 @@ class MultipleInputsLayer(Layer):
             - name : a string or None
                 an optional name to attach to this layer
         """
+        from .helper import get_output_shape
         self.input_shapes = [incoming if isinstance(incoming, tuple)
-                             else incoming.get_output_shape()
+                             else get_output_shape(incoming)
                              for incoming in incomings]
         self.input_layers = [None if isinstance(incoming, tuple)
                              else incoming
                              for incoming in incomings]
         self.name = name
-
-    def get_output_shape(self):
-        return self.get_output_shape_for(self.input_shapes)
 
     def get_output_shape_for(self, input_shapes):
         """
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -131,7 +131,8 @@ def create_iter_functions(dataset, output_layer,
                                    deterministic=True)
 
     pred = T.argmax(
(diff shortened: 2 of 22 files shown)


benanne:
Looks great so far! That's a real bummer about the tests, I don't think I fully understand the problem yet though. Basically `get_all_layers()` thinks everything is a `MultipleInputsLayer`? I didn't realize `Mock` instances always return `True` when `hasattr` is called on them. Of course that does make sense.

The docs say you can force it to not have certain attributes simply by deleting them:

> You may want a mock object to return False to a hasattr() call, or raise an AttributeError when an attribute is fetched. You can do this by providing an object as a spec for a mock, but that isn’t always convenient.

> You “block” attributes by deleting them. Once deleted, accessing an attribute will raise an AttributeError.

(from https://docs.python.org/3/library/unittest.mock.html#deleting-attributes )

Would that be sufficient to solve the issue?

f0k:
> Would that be sufficient to solve the issue?

Thanks, but no. I already went the "providing an object as a spec for a mock" route, but then the tests fail because they check whether `get_output()` was called on the Layer's `input_layer`, while we're now calling `get_output_for()` instead. A simple "does the network produce the correct output" test would have passed seamlessly after the change, but our "does the implementation do what we implemented" test doesn't. That's not really a problem, though, the tests have to be rewritten anyway to not use `Layer.get_output()` any more, and we have the MNIST test for "does the network produce the correct output". I'll proceed with the PR the next time I'm waiting for some experiments to finish.

benanne:
Oh I see, the tests are just testing the wrong thing :p I guess there isn't a clear cut solution for that, other than to rewrite them.

f0k:
> I guess there isn't a clear cut solution for that, other than to rewrite them.

Yep. I was just hoping that before rewriting them to reflect the API change, I could use them to check whether the deprecated `Layer.get_output()` does exactly the same as before. Turns out I can't. That's all I meant to convey :)
I'm not sure if that's a problem with the tests, i.e., if they know too much about the implementation, but on the other hand they are probably still better than a "record-and-replay" approach that just tests a few specific input-output pairs.

benanne:
Yeah, I don't think it's necessarily a bad thing or a problem that the tests were a bit implementation-specific before. This is a pretty radical change anyway.

f0k:
Hmm, small problem: When you have a network like `l1 -> l2 -> l3 -> l4` and you call `get_output()` with a mapping `{l2: ...}`, it will still start propagating data from `l1`. This (a) wastes some effort and (b) fails if `l1` does not have an associated input variable or input layer although it shouldn't. The most obvious solution would be to extend `get_all_layers` such that you can supply a list of layers blocking graph traversal, i.e., a list of layers that you don't need the input layers for. `get_output()` would call `get_all_layers()` with `inputs.keys()` as this blocking list if `inputs` is a dictionary. Is that okay with everyone? :)

benanne:
It only adds a few lines to `get_all_layers()` so I don't have any problems with it. The additional complexity is limited and warranted in this case, I think.

f0k:
Did a few changes (in separate commits to make reviewing easier), added [what I suggested above](https://github.com/benanne/Lasagne/pull/182#issuecomment-88528372), and rewrote all the tests. Phew. Next up: Do the same for `get_output_shape()`. And add some test proving that this actually solves #104.

benanne:
Looks good from a quick skim, I'll try to have a closer look at it soon (the tests in particular). I'm not sure about the name `treat_as_input`, maybe there are other use cases where you would want to prevent the inputs to certain layers to show up in the list? Then that name may be a bit too specific. I'm not sure what else we could call it that's short enough, though.

f0k:
> maybe there are other use cases where you would want to prevent the inputs to certain layers to show up in the list?

Would they require any different handling than treating these layers as input layers? I mean, is there a use case for which `treat_as_input` would be confusing?
Theano's `theano.grad()` has `consider_constant=`, and `theano.toolbox.graph.inputs` has `blockers=`. Doesn't really help here.
Another name would be `not_traverse` or something, but that would imply something about the algorithm that the name `get_all_layers` does not imply.
Anyway, I'm open to suggestions!

> I'll try to have a closer look at it soon (the tests in particular)

I'm thinking that maybe we should document the tests a bit. I just followed the existing style (i.e., no comments at all), but it's not always obvious to see what's tested and why. I'll add comments everywhere along with the next set of tests.

benanne:
Tests look good, I agree that documenting them a little might be helpful. So far the name of the test method has served as a description of what's going on, but I guess that isn't always sufficient.

f0k:
Next step: Moved `get_output_shape()` into a global function as well. For now it doesn't allow propagating alternative shapes, but I'll add that for symmetry to `get_output()`.

Another question that popped up now: Since every layer already has an `input_shape` or `input_shapes` property, `get_output_shape(some_layer)` is not at all symmetric to `get_output(some_layer)`. All it needs to return is `some_layer.get_output_shape_for(some_layer.input_shape)`, or `some_layer.get_output_shape_for(some_layer.input_shapes)`, or `some_layer.shape`, depending on the class of `some_layer` (`Layer`, `MultipleInputsLayer` or `InputLayer`). Moving `get_output()` into a global function makes it a bit inconvenient in the case I have a layer at hands that knows its input shape and just has to call `get_output_shape_for()` on it once. This happens during construction of a layer, when we set the new layer's input shape to the incoming layer's output shape, and it happens during construction of convolutional layers for `untie_biases=True`, when the new layer has to query its own output shape.

For symmetry with `input_shape`, it would be convenient if each layer got either:
a) an `output_shape` property that just passes `self.input_shape` or `self.input_shapes` to `self.get_output_shape_for()`
b) an `output_shape` attribute that is initialized in the constructor

If we do that, a lot of `get_output_shape(layer)` calls in the code could be replaced by accessing `layer.output_shape`, and the global `get_output_shape()` helper function could focus on the task of propagating alternative input shapes through the network.
A side question is whether `InputLayer` should get this property/attribute as well. It would be logical and convenient, but also a bit awkward because it would provide its shape twice.

benanne:
I had a look at the new changes, the comments in the tests do help a lot to make sense of them, so that was a great idea.

I think adding `output_shape` in addition to `input_shape` is a good idea.

Is there any advantage to making it a property, so it is computed on the fly? As you mentioned, we might as well make it an attribute and set it in the constructor, just like `input_shape`.

I think it's okay for `InputLayer` to get `output_shape` as well. If anything, it should only have `output_shape` and not `input_shape`, since it does not have an incoming layer itself. But I guess that would potentially cause problems as well, so lets just live with this tidbit of awkwardness :)

f0k:
> Is there any advantage to making it a property, so it is computed on the fly? As you mentioned, we might as well make it an attribute and set it in the constructor, just like `input_shape`.

The advantage is that you can't set it from the outside, so it's guaranteed to always be consistent with `input_shape`. Not many layers would be happy about somebody changing `input_shape` after construction, though, so it doesn't really matter. Also it would be the only property in the library, which would raise the question why we don't encapsulate other attributes as well. So I'm fine with just setting an attribute.

> I think it's okay for `InputLayer` to get `output_shape` as well. If anything, it should only have `output_shape` and not `input_shape`, since it does not have an incoming layer itself.

It doesn't have `input_shape` anyway, it only has `shape` and always needs special treatment for now. Afterwards, it would have `shape` and `output_shape`, unless we decide to drop `shape` (which we can iff we make `output_shape` an attribute and not a property).

benanne:
Another reason to make it an attribute then. Let's do that :)

f0k:
> Another reason to make it an attribute then. Let's do that :)

Bonus advantage: Things like `ReshapeLayer` will fail on construction rather than on the first `get_output_shape_for()` call triggered by creating a layer on top of them.
Bonus disadvantage: Every attribute required for `get_output_shape_for()` to work has to be set before calling the super constructor, because this will call `get_output_shape_for(self.input_shape)`. If setting up the attributes already requires `self.input_shape` to be present, we're out of luck. (Haven't checked whether this is the case anywhere.)

If we make it a property instead, we lose the bonus advantage and the bonus disadvantage. I'm undecided. The attribute would be nice, but it's a bit weird to not have the super constructor call be the first statement.

benanne:
Hmm, good point about the disadvantage... I wish I had your level of insight into things like this!

 `get_output_shape_for()` will almost always depend on a bunch of stuff that's set in the constructor (`self.num_units` in `DenseLayer` is an example). Bummer. I was hoping to avoid introducing a property since it's less common and slightly harder to read than just having it be an attribute. But it sounds like it may actually be the best approach.

Both alternatives `a` and `b` seem pretty inconvenient. And I agree that not having the superconstructor call be the first thing in the constructor is weird and unintuitive, so we should probably avoid that.

f0k:
> Hmm, good point about the disadvantage... I wish I had your level of insight into things like this!

Granted, this time the insight came only after I had turned `output_shape` into an attribute set in the constructor and briefly wondered why suddenly all tests failed along the lines of `'Conv2DMMLayer' object has no attribute 'filter_size'` :)

> Both alternatives `a` and `b` seem pretty inconvenient.

Alternative b) is the current state of this PR. It doesn't happen too often in the code anyway, but I agree it would be nice to be able to ask any network's output layer for its shape directly, similar to how `layer.get_output_shape()` allows it. Property it is, then.

benanne:
Looks like it! I was hoping we'd be able to avoid fancier Python features like this one (as well as decorators etc.) but it looks like this is really the cleanest solution. It's preferable to the inconsistency of `layer.input_shape` but `layer.get_output_shape()` at any rate.

f0k:
> It's preferable to the inconsistency of `layer.input_shape` but `layer.get_output_shape()` at any rate.

It's also the inconsistency of `layer.get_output_shape()` vs. `lasagne.layers.get_output()` and `lasagne.layers.get_output_shape()`.


I've implemented the property now and manually added the "bonus advantage" mentioned above to the `ReshapeLayer` and `DimshuffleLayer` constructors by calling `get_output_shape()` manually at the end of the constructor. It would be a lot cooler to have this happen for all layers, but that would mean either a) setting the base class constructor to something like: `self.input_shape = ...; self.init(); self.output_shape = ...` and then having subclasses override `init()` instead of `__init__()` or b) requiring each constructor to set `self.output_shape` somewhere.

benanne:
What's the current status of this? Is the TODO list above still accurate? If so, is there anything I can help with (`get_output_shape()` tests maybe)? There are a bunch of open PRs waiting for this one, not to mention the recurrence work by @craffel and @skaae, so it would be great to get it done soon :)

f0k:
The TODO list is still accurate, I'm just busy with experiments and writing... and I think this might need rebasing in the meantime. The next step will be extending `get_output_shape()` to support propagating alternative input shapes, then copying all the `get_output()` tests to turn them into `get_output_shape()` tests. I'll try to work on that in the afternoon.

benanne:
Considering that it seems to need a rebase in the meantime, would you be okay with having #204 and maybe #201 (less urgent) merged first? Since both of those are currently waiting for this one, which was basically to avoid a rebase in the first place if I'm not mistaken :)

f0k:
> Since both of those are currently waiting for this one, which was basically to avoid a rebase in the first place if I'm not mistaken :)

No, it's also about making the rebase easier than starting from scratch. #201 would give a lot of additional conflicts because I've touched a lot of code around the base classes (including `MultipleInputsLayer`). #204 will be less painful, except that it places the additional burden on me to convert the docstrings in layers/base.py and layers/helper.py and layers/input.py that I've touched or added.

Sorry for being so slow with this, but if #204 and #201 are not too urgent, it would still be nice to have them wait a bit...

benanne:
Okay, fine by me. This is a big overhaul so it's probably best to keep it as straightforward as possible :)

benanne:
I'm having a look at this currently, do you have any place where I can read about doing a PR of a PR? I'm not sure how to go about it. I know how to check out a PR, but I don't know how to then send a PR on top of that (which should go to the correct fork I guess, i.e. f0k/Lasagne and not Lasagne/Lasagne).

f0k:
> do you have any place where I can read about doing a PR of a PR?

After checking out my PR, commit your changes on top and push them into a new branch, then use the web interface on github.com/f0k/Lasagne to send a pull request. When opening a pull request, you can freely define the source and target repository as well as the source and target branch.

> but I realised that `get_output_shape()` still needs to be extended to support custom input shapes. I think I'd best leave that to you to make sure it matches what you have in mind.

It should just be perfectly symmetric to `get_output()`, i.e., allowing to pass in a shape or a shape dictionary, or just leave it as is. (I actually wrote that function already, but it seems I lost it when I had to shutdown the machine for power maintenance...) I won't be able to do this before the ISMIR deadline (Tuesday 11:00 GMT), and I'll be abroad from Wednesday for which I have to prepare a presentation on Tuesday afternoon. So... feel free to have a go, or otherwise we can merge it as is and leave `get_output_shape()` open for later. Just check if the examples still work.

benanne:
Thanks! I don't think we should merge this before it's completely done. I'll see what I can do.

benanne:
I've added `get_output_shape()` as you described it in a PR to this PR: https://github.com/f0k/Lasagne/pull/3

I'll see if I can update the tests and the examples as well so we can finally wrap this up.

f0k:
Okay, merged in Sanders work (thanks again!) and fixed everything that needed fixing. Ready for a final review! If there are no comments, I'll merge it tomorrow or the day after.

benanne:
I'll have another look over all the changes tonight or tomorrow morning.

lasagne/layers/base.py
```diff
@@ -262,30 +234,17 @@ def __init__(self, incomings, name=None):
...
-                             else incoming.get_output_shape()
+                             else incoming.output_shape
                              for incoming in incomings]
         self.input_layers = [None if isinstance(incoming, tuple)
                              else incoming
                              for incoming in incomings]
         self.name = name
 
-    def get_output_shape(self):
+    @Layer.output_shape.getter
```
benanne: Just curious: why is this necessary? Can't properties be overridden or something? (I would have expected `@property`.)

benanne:
Alright, looks okay to me. Feel free to merge whenever :) Do you want to post to the mailing list about the interface change or should I do it?

lasagne/layers/base.py
f0k: With `@property` you create a new property replacing the inherited one, with what I used here you overwrite the getter of the inherited property. Both works in this case: as the base class only defines a getter *for now*, we could also just redefine the property as a whole. I guess the question generally boils down to what you want to happen in the subclasses if the base class ever introduces a setter (or deleter): Should they inherit it or not? We'll arguably never introduce a setter for the output shape, so it still doesn't matter which way we choose. I chose this one because it explicitly tells that we're overriding a property that's also present in the base class.

benanne: Fair enough, thanks for the clarification :)

PR #185, opened 2015-03-31
add MaxPool1DLayer + tests

benanne:
This adds a `MaxPool1DLayer` that just uses `max_pool_2d` with an extra dimension of size 1. It also adds rudimentary tests for `MaxPool1DLayer` and `MaxPool2DLayer`.

### DIFF
--- a/lasagne/layers/pool.py
+++ b/lasagne/layers/pool.py
@@ -7,6 +7,7 @@ from theano.tensor.signal import downsample
 
 
 __all__ = [
+    "MaxPool1DLayer",
     "MaxPool2DLayer",
     "FeaturePoolLayer",
     "FeatureWTALayer",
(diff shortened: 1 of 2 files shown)


f0k:
Looks good to me. Maybe the tests should also check if `get_output_for` gives correct values, not only correct shapes. Either by comparing it to a solution in numpy or by computing a simple manually constructed example. Pooling in numpy along the first axis with `ignore_border=True` would be:
```python
def maxpool(data, factor):
        return np.max(data[:len(data)//factor*factor].reshape((-1, factor) + data.shape[1:]), axis=1)
```

benanne:
I added value tests for the ignore_border=False case, since that's the easiest to do. Not sure if it's worth adding the ignore_border=True case as well. Kinda feels like we're testing Theano here, not Lasagne.

f0k:
> I added value tests for the ignore_border=False case, since that's the easiest to do.

No, you added tests for the `ignore_border=True` case, since that's the easiest to do :)

> Kinda feels like we're testing Theano here, not Lasagne.

Yeah, I know... I commented along the same lines when it came to adding tests for the convolutions. The alternative would be creating an equivalent Theano expression and checking that for equivalence either symbolically or by evaluation. But let's leave it at that for now! +1 for merging!

benanne:
> No, you added tests for the ignore_border=True case, since that's the easiest to do :)

doh. I'll amend the commit msg tomorrow, and also merge it unless there are further comments.

PR #187, opened 2015-03-31
Fix possibly incorrect output shape calculation in Conv2DCCLayer.

JeffreyDF:


f0k:
Thanks, good catch. Seems we should focus on improving our test coverage!

benanne:
The convolutions definitely need the most work in that department. Since this is a bugfix and Travis passes, I will merge it now. Thanks!

PR #188, opened 2015-04-01
Improved MNIST example; added link to GitHub in the documentation; added...

MartinThoma:
... link to the documentation in the README as a badge (to make it easier to directly find the docs, as this is commonly used in GitHub)


MartinThoma:
Eventually `example/mnist.py` should be changed, as it gives a Lasagne warning:

```
/usr/local/lib/python2.7/dist-packages/Lasagne-0.1dev-py2.7.egg/lasagne/layers/helper.py:52: UserWarning: get_all_layers() has been changed to return layers in topological order. The former implementation is still available as get_all_layers_old(), but will be removed before the first release of Lasagne. To ignore this warning, use `  warnings.filterwarnings('ignore', '.*topo.*')`.
+    warnings.warn("get_all_layers() has been changed to return layers in "
```

But I don't know how to fix this (and it is not introduced by my commit).

benanne:
Thanks for this! I'll have a closer look at it later today. Don't worry about the warning, this will go away when we release the first version of the library. Pretty much all Lasagne code should raise this warning right now, that's intentional :)

docs/user/tutorial.rst
```diff
@@ -27,9 +26,47 @@ Now run the ``mnist.py`` example:
...
 
-XXX Explain output
+You will get the following output:
+
+.. code-block::text
+
+  Loading data...
+  Building model and compiling functions...
+/usr/local/lib/python2.7/dist-packages/Lasagne-0.1dev-py2.7.egg/lasagne/layers/helper.py:52: UserWarning: get_all_layers() has been changed to return layers in topological order. The former implementation is still available as get_all_layers_old(), but will be removed before the first release of Lasagne. To ignore this warning, use `  warnings.filterwarnings('ignore', '.*topo.*')`.
+    warnings.warn("get_all_layers() has been changed to return layers in "
```
f0k: I think this warning shouldn't be replicated in the tutorial, as it will go away in about 3-4 weeks anyway.

docs/user/tutorial.rst
```diff
@@ -27,9 +26,47 @@ Now run the ``mnist.py`` example:
...
+validation set in line 206. It continues with automatically building a model
+with two hidden layers. It automatically detects the necessary number of input
+neurons (= number of features, 28*28=784 pixels in case of MNIST) and the
+necessary number of output neurons (= number of classes, 10 in case of MNIST).
+
+Then it automatically trains and evaluates the feed forward neural network
+with mini-batch gradient descent.
+
+A detailed explanation of how the training of feed forward neural networks
+works can be found in "Machine Learning" by  Tom Mitchell.
```
f0k: I think I'd prefer a web reference here, maybe linking to the deeplearning.net tutorial (which in turn gives suitable textbook references).

examples/mnist.py
```diff
@@ -66,7 +72,11 @@ def load_data():
 
 def build_model(input_dim, output_dim,
                 batch_size=BATCH_SIZE, num_hidden_units=NUM_HIDDEN_UNITS):
-
+    """Create a symbolic representation of a neural network with
+    `intput_dim` input nodes, `output_dim` output nodes, a training function
+    with a mini-batch size of `batch_size` and `num_hidden_units` per hidden
+    layer.
```
f0k: +1 for documenting, but I'd change that a bit. It sounds as if the training function had a particular batch size and a particular number of hidden units now, but in fact both apply to the network, and the training function is not created in `build_model` at all.

examples/mnist.py
```diff
@@ -100,29 +110,35 @@ def create_iter_functions(dataset, output_layer,
...
         loss_function=lasagne.objectives.categorical_crossentropy)
 
-    loss_train = objective.get_loss(X_batch, target=y_batch)
-    loss_eval = objective.get_loss(X_batch, target=y_batch,
-                                   deterministic=True)
+    loss = {'train': objective.get_loss(X_batch, target=y_batch),
+            'eval': objective.get_loss(X_batch,
+                                       target=y_batch,
+                                       deterministic=True)
+           }
```
f0k: What's the advantage of changing that?

MartinThoma: Personal preference :-)  Is there a disadvantage?

I thought I could generate the iter-functions in a more compact way like

```python
iters = {}
for modes in ['train', 'valid']:
    iters[mode] = theano.function(
        [batch_index], loss[mode],
        updates=updates,
        givens={
            X_batch: dataset['X_train'][batch_slice],
            y_batch: dataset['y_train'][batch_slice],
        },
    )
```

as the code is very similar, but then I noticed that it is `loss[mode]` for train and `[loss[mode], accuracy]` for the others.

I like it this way, because it groups two different types of loss functions in one object, now two as it was before.

Should I undo it?

docs/user/tutorial.rst
MartinThoma: I've just added a link to the deeplearning tutorial start page at the beginning (by the way: thanks for all the constructive input :-) )

examples/mnist.py
MartinThoma: I've updated the docstring.

Isn't it the case that the training function has to have a batch size of `batch_size`? If not, why do we pass this parameter? What would happen if we passed `batch_size=8` to `build_model` and `batch_size=9`  to `train`?

examples/mnist.py
benanne: I think it would be better to keep it as is, as you said it's not really possible to easily 'generalize' the creation of theano functions across these two cases, so using a dictionary here seems superfluous and it just makes the code a tiny bit harder to understand.

benanne:
Looks great! I don't have much to add in addition to @f0k's comments. There are a few sentences I would formulate slightly differently, but those are details and we can always take care of cleaning up grammar and style at a later date. Thanks a lot for helping out!

examples/mnist.py
f0k: I think so, too, it's easier to understand without the dictionaries in between (same for the similar change further below). Keep in mind that many people come to Lasagne via Daniel's or Sander's blog posts and may not have used Python before, so we should try to keep it simple. (Maybe even simpler than now, as most users copy and paste the example as their starting point -- but that's outside the scope of this PR.)

docs/user/development.rst
```diff
@@ -0,0 +1,95 @@
+Development
+===========
+
+The ``Lasagne`` project is developed by Sander Dieleman and contributers via
```
f0k: *contributors

docs/user/development.rst
```diff
@@ -0,0 +1,95 @@
...
+* http://lasagne.readthedocs.org/ for hosting the documentation
+
+
+Code coverage can be tested with
+
+.. code:: bash
+
+    $ py.test
+
+Testing takes over 5 minutes.
```
f0k: Testing will take over 5 minutes for the first run, but less than a minute for subsequent runs when Theano can reuse compiled code.

docs/user/development.rst
```diff
@@ -0,0 +1,95 @@
...
+
+
+Theano
+------
+
+Lasagne makes use of Theano. For this reason, you should read Colin Craffels
+[Theano Tutorial](http://nbviewer.ipython.org/github/craffel/theano-tutorial/blob/master/Theano%20Tutorial.ipynb)
+if you haven't worked with Theano before.
```
f0k: Hmm, not sure about this... it starts off well, but then it defines a `Layer` class, which is confusing because it's not the same as the `Layer` class in Lasagne. I think we should link to [the official Theano tutorial](http://deeplearning.net/software/theano/tutorial/) instead, even if this looks a bit overwhelming.
Secondly, knowledge about Theano is not only required to develop Lasagne, but also to use it. Lasagne just helps with creating the symbolic graph, you still need to know Theano to work with it. So probably this should be mentioned in the Lasagne user tutorial instead as soon as there is one.

MartinThoma:
@f0k I've already linked it in `user/tutorial.rst`. Probably I should only remove the link at this position?
@craffel could this tutorial eventually be merged into Lasagne?

docs/user/tutorial.rst
f0k: I still feel weird about giving a single text book reference here. Why should we promote this one and not another?

MartinThoma: Because it is a (the?) standard text book about machine learning. It goes very detailed through how the training works.

Do you think I should remove it or add another text book? Which text book would you add?

f0k:
> Probably I should only remove the link at this position?

You can just remove the "Theano" section, I'd say.

> could this tutorial eventually be merged into Lasagne?

As far as I know, Colin will do a new tutorial including Lasagne soon. We can add a reference to that later. I'm not sure if we should try to include a Theano crash course as part of the Lasagne documentation. It would surely be nice, but it would also be outside the scope of Lasagne.

craffel:
> @craffel could this tutorial eventually be merged into Lasagne?

Thanks for the shoutout, but I think it needs to be kept separate - as Sander said, I think teaching Theano is out of the scope of Lasagne.  It's worth making clear, though, that Lasagne is best enjoyed with a background knowledge of Theano.

> As far as I know, Colin will do a new tutorial including Lasagne soon. 

Yes, it'll have to be done by the end of the month.

docs/user/tutorial.rst
f0k: Reading the full page again, I think it's misplaced here. In the introduction, you already state that "If you want to get detailed information on how neural networks work [...] read [...]". I think it's a good idea to list such prerequisites in the beginning. We don't aim for a tutorial explaining neural networks here, we just want to explain how Lasagne can help to build and train neural networks compared to using Theano directly.

Regarding the text book reference, it can be considered one of the standard text books, yes. I'm not sure if it's *the* standard text book, that's why I'm hesitating to refer to a single book.

What about removing it here and amending the sentence in the introduction:
[...] then you should read the Deeplearning Tutorial, possibly along with an online course for more theoretical background (["Neural Networks and Deep Learning" by Michael Nielsen](http://neuralnetworksanddeeplearning.com/), ["Convolutional Neural Networks for Visual Recognition" by Andrej Karpathy et al.](http://cs231n.github.io/)) or a standard text book such as "Machine Learning" by Tom Mitchell.

MartinThoma: I don't know either of the two links but  they look fine to me (I trust you at this point; I only took a quick glance at the content).

I've just updated it as you suggested (and put both links on my reading list :-) )

f0k:
Thanks! I'd be fine with merging now. We can still reformulate/rearrange things later, it's a good start for now.

benanne:
lgtm! Will merge tomorrow if there are no further comments.

benanne:
Merging, thanks!

PR #191, opened 2015-04-01
He-style initializers

benanne:
I figured I might as well go ahead and add them.

This should supersede #165: `lasagne.init.HeNormal('relu')` should be equivalent to `lasagne.init.ReluNormal()` introduced in that PR.

For some reason I botched my pep8/flake8/pytest install on this PC, so I'm counting on Travis here :)

### DIFF
--- a/lasagne/init.py
+++ b/lasagne/init.py
@@ -62,7 +62,7 @@ class Glorot(Initializer):
         if self.c01b:
             if len(shape) != 4:
                 raise RuntimeError(
-                    "If c01b is True, only shapes of lenght 4 are accepted")
+                    "If c01b is True, only shapes of length 4 are accepted")
 
             n1, n2 = shape[0], shape[3]
             receptive_field_size = shape[1] * shape[2]
@@ -97,7 +97,7 @@ class He(Initializer):
         if self.c01b:
             if len(shape) != 4:
                 raise RuntimeError(
-                    "If c01b is True, only shapes of lenght 4 are accepted")
+                    "If c01b is True, only shapes of length 4 are accepted")
 
             fan_in = np.prod(shape[:3])
         else:
--- a/lasagne/init.py
+++ b/lasagne/init.py
@@ -50,101 +50,77 @@ class Uniform(Initializer):
 
 
 class Glorot(Initializer):
-    def __init__(self, initializer=Normal, gain=1.0):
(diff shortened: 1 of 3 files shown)


f0k:
Looks good! I just wonder if `c01b` should be a boolean constructor argument to save on classes. I don't have a strong opinion on that, though.

benanne:
Yeah, could be a little cleaner that way.

benanne:
I've changed `c01b` to be a keyword argument for both the `Glorot*` and the `He*` classes.

lasagne/init.py
```diff
@@ -50,50 +50,77 @@ def sample(self, shape):
...
         self.gain = gain
+        self.c01b = c01b
 
     def sample(self, shape):
-        n1, n2 = shape[:2]
-        receptive_field_size = np.prod(shape[2:])
+        if self.c01b:
+            if len(shape) != 4:
+                raise RuntimeError(
+                    "If c01b is True, only shapes of lenght 4 are accepted")
```
f0k: *length

lasagne/init.py
```diff
@@ -50,50 +50,77 @@ def sample(self, shape):
...
         self.initializer = initializer
         self.gain = gain
+        self.c01b = c01b
 
     def sample(self, shape):
-        n1, n2 = shape[:2]
-        receptive_field_size = np.prod(shape[2:])
+        if self.c01b:
+            if len(shape) != 4:
+                raise RuntimeError(
```
f0k: Hmm, I though that would be a ValueError. `shape` is of correct type, but inappropriate value. Or `c01b` is of correct type, but inappropriate value. But actually it's just the combination thats inappropriate, so the [descriptions](https://docs.python.org/2/library/exceptions.html) don't fit exactly. What would numpy do, or Theano?

f0k:
Looks good! I'm just not sure about the exception type. Maybe we should find a consensus on that some time. It's not critical for now.

benanne:
I've never really spent time figuring out what types there are, I usually just go with RuntimeError :p But I guess it wouldn't be a bad idea to specialise them a bit across the library (not just in this PR).

benanne:
Okay, let's handle a possible exception cleanup in a separate PR for convenience. I'll merge this tomorrow unless there are further comments.

PR #194, opened 2015-04-03
Modular momentum updates plus docstrings

f0k:
This pulls out `apply_momentum` and `apply_nesterov_momentum` as I suggested in https://github.com/benanne/Lasagne/issues/190#issuecomment-88960154. Also it changes the update functions to accept either a cost expression or a list of gradients, and to return OrderedDicts instead of `(param, update)` tuple lists. Finally, it adds numpy-style docstrings to `sgd` and the four `momentum` functions, and converts the `norm_constraint` docstring to numpy format.

### DIFF
--- a/lasagne/updates.py
+++ b/lasagne/updates.py
@@ -2,78 +2,286 @@
 Functions to generate Theano update dictionaries for training.
 """
 
+from collections import OrderedDict
+
 import numpy as np
 
 import theano
 import theano.tensor as T
 
 
-def sgd(loss, all_params, learning_rate):
-    all_grads = theano.grad(loss, all_params)
-    updates = []
+__all__ = [
+    "sgd",
+    "apply_momentum",
+    "momentum",
+    "apply_nesterov_momentum",
+    "nesterov_momentum",
+    "adagrad",
+    "rmsprop",
+    "adadelta",
+    "norm_constraint",
+]
 
-    for param_i, grad_i in zip(all_params, all_grads):
-        updates.append((param_i, param_i - learning_rate * grad_i))
+
+def get_or_compute_grads(loss_or_grads, params):
+    """Helper function returning a list of gradients.
+
+    Parameters
+    ----------
+    loss_or_grads : symbolic expression or list of expressions
+        A scalar loss expression, or a list of gradient expressions
+    params : list of shared variables
+        The variables to return the gradients for
+
+    Returns
+    -------
+    list of expressions
+        If `loss_or_grads` is a list, it is assumed to be a list of
+        gradients and returned as is, unless it does not match the length
+        of `params`, in which case a `ValueError` is raised.
+        Otherwise, `loss_or_grads` is assumed to be a cost expression and
+        the function returns `theano.grad(loss_or_grads, params)`.
+    """
+    if isinstance(loss_or_grads, list):
+        if not len(loss_or_grads) == len(params):
+            raise ValueError("Got %d gradient expressions for %d parameters" %
+                             (len(loss_or_grads), len(params)))
+        return loss_or_grads
+    else:
+        return theano.grad(loss_or_grads, params)
+
+
+def sgd(loss_or_grads, params, learning_rate):
+    """Stochastic Gradient Descent (SGD) updates.
+
+    Generates update expressions of the form:
+    * ``param := param - learning_rate * gradient``
+
+    Parameters
+    ----------
+    loss_or_grads : symbolic expression or list of expressions
+        A scalar loss expression, or a list of gradient expressions
+    params : list of shared variables
+        The variables to generate update expressions for
+    learning_rate : float or symbolic scalar
+        The learning rate controlling the size of update steps
+
+    Returns
+    -------
+    OrderedDict
+        A dictionary mapping each parameter to its update expression
+    """
+    grads = get_or_compute_grads(loss_or_grads, params)
+    updates = OrderedDict()
+
+    for param, grad in zip(params, grads):
+        updates[param] = param - learning_rate * grad
 
     return updates
 
 
-def momentum(loss, all_params, learning_rate, momentum=0.9):
-    all_grads = theano.grad(loss, all_params)
-    updates = []
+def apply_momentum(updates, params=None, momentum=0.9):
+    """Returns a modified update dictionary that includes momentum terms.
 
-    for param_i, grad_i in zip(all_params, all_grads):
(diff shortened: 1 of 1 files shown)


benanne:
Looks good to me! I spent a few minutes verifying whether the new momentum / nesterov momentum formulas are equivalent to the old ones, and this seems to be the case. The `OrderedDict`s actually make the code a bit more readable too, which is nice.

skaae:
In line line 309  I think `loss` should be `loss_or_grad` and `all_params` should be `params`

f0k:
Good catch @skaae, I've amended the commit. We definitely need tests, I'm just not sure what they should look like. Perform an update step in a fully controlled situation and compare the result with a pre-recorded one? Generate an update dictionary and compare it with manually-constructed expressions?

benanne:
Let's handle tests in a separate PR so this can be merged ASAP :)

f0k:
> Let's handle tests in a separate PR so this can be merged ASAP :)

We should probably still make sure the refactoring doesn't break anything. I'll try to do that tomorrow.

f0k:
OK, did the following test (with `lasagne.updates_old` being the current master `updates.py`, and `lasagne.updates` being the one from this PR):
```python
import numpy as np
import theano
import lasagne
import lasagne.updates_old
np.random.seed(10)
x, y, z = theano.tensor.matrices(*'xyz')
a, b, c, p = [np.random.randn(10,10).astype(np.float32) for _ in range(4)]
p1 = theano.shared(p, 'p')
p2 = theano.shared(p, 'p')
for func in lasagne.updates.__all__[:-1]:
    if hasattr(lasagne.updates_old, func):
        f1 = getattr(lasagne.updates, func)
        f2 = getattr(lasagne.updates_old, func)
        loss1 = theano.tensor.mean((theano.tensor.dot(x, y)*p1 - z)**2)
        loss2 = theano.tensor.mean((theano.tensor.dot(x, y)*p2 - z)**2)
        up1 = f1(loss1, [p1], learning_rate=0.1)
        up2 = f2(loss2, [p2], learning_rate=0.1)
        fup1 = theano.function([x,y,z], updates=up1)
        fup2 = theano.function([x,y,z], updates=up2)
        for r in range(10):
            if not np.allclose(p1.get_value(), p2.get_value()):
                print "Function %s: %d. update does not match" % (
                        func, r+1)
            fup1(a, b, c)
            fup2(a, b, c)
```

I get:
```
Using gpu device 0: GeForce GT 640
Function nesterov_momentum: 6. update does not match
Function adagrad: 2. update does not match
Function adagrad: 4. update does not match
Function adagrad: 7. update does not match
Function rmsprop: 3. update does not match
Function rmsprop: 6. update does not match
Function rmsprop: 7. update does not match
Function rmsprop: 9. update does not match
```

There's no easy way to do a more direct comparison (in terms of `theano.toolbox.graph.is_same_graph`, for example), because the two instances create their own shared variables, and momentum/nesterov_momentum create slightly different graphs (some operands are in a different order).
This test shows that there are slight rounding differences in between that eventually match up again. Should we investigate and try to match the original versions more closely or leave it at that?

benanne:
Interesting! How big of a difference are we talking about in relative and absolute terms? Honestly I don't think it matters, you will probably see the same difference when running identical code on multiple different GPUs. The default tolerance for `numpy.allclose` is too low for single precision numbers anyway.

It's interesting that adagrad and rmsprop are affected more than the others, but that's probably because of the squaring and square roots in those algorithms.

f0k:
> It's interesting that adagrad and rmsprop are affected more than the others, but that's probably because of the squaring and square roots in those algorithms.

Err... nope. It's because `p1` and `p2` were initialized outside of the loop so the differences from the momentum updates were still in there. At least you didn't see it either :)
When I move the initialization inside the loop, and add another test like this:
```python
print "Function %s: Perfect match of code?" % func, (theano.printing.debugprint(fup1, file='str') == theano.printing.debugprint(fup2, file='str'))
```
Then I get:
```
Function sgd: Perfect match of code? True
Function momentum: Perfect match of code? False
Function nesterov_momentum: Perfect match of code? False
Function adagrad: Perfect match of code? True
Function rmsprop: Perfect match of code? True
Function adadelta: Perfect match of code? True
```
No noticeable differences, at least for `np.allclose` and at least for 10 iterations. With 20 iterations there are slight differences for momentum (updates 13, 19, 20). I'll see if there's a way to reorder things to get a perfect match of code.

benanne:
Hmmm okay :p 
Do you think getting it to produce the exact same output is worth your time though? I would doubt it :)

f0k:
> I'll see if there's a way to reorder things to get a perfect match of code.

I can't find one. It's a bit annoying. The former momentum updates were constructed like this:
```python
updates[velocity] = momentum * velocity - learning_rate * grad
updates[param] = param + updates[velocity]
```
This gets compiled into a single `GpuElemwise{Composite{...}}` for the velocity, and a `GpuElemwise{Add}` for the param.

The currently proposed momentum updates are constructed like this:
```python
# note: updates[param] == param - learning_rate * grad
x = momentum * velocity + updates[param]
updates[velocity] = x - param
updates[param] = x
```
Now as `x` is used in both updates, this prevents the first part to be compiled into a single giant composite expression. Instead, it's compiled like this:
```python
tmp1 = momentum * velocity
tmp2 = learning_rate * grad
updates[velocity] = tmp1 - tmp2
updates[param] = tmp1 + (p - tmp2)
```
Theano doesn't see that `tmp1 - tmp2` could be merged. This results in two more CUDA kernel calls and in more temporary memory being used. That's a bit suboptimal.

When I change the updates to variations of:
```python
# note: updates[param] == param - learning_rate * grad
step = updates[param] - param
updates[velocity] = momentum * velocity + step
updates[param] = param + updates[velocity]
```
I still get the same.

The only way to get exactly the same code (and performance) as before the change is to do do:
```python
step = updates[param].owner.inputs[1]
updates[velocity] = momentum * velocity - step
updates[param] = param + updates[velocity]
```
But we'd need to closely examine `updates[param]` before to verify it is in the `param - step` form, and that should probably be solved by a subsequent graph optimization in Theano, not by us on constructing the graph.

This is all a bit unfortunate... let's merge the PR as it is for now, we can still optimize things later if we want to. By the way, for nesterov momentum, while the resulting updates are not exactly the same, the performance is not affected.

benanne:
Agree, if there are no performance regressions it shouldn't be an issue. I'll go ahead and merge it.

Issue #38, opened 2014-09-19
Move this repository to an organization?


benanne:
I created this repository on my personal GitHub account, but since this is a joint project, maybe it's better to create an organization and move this repository to it. Added benefits would be more granular access control (commit rights etc.).

Also I would just be able to submit pull requests like everyone else, instead of committing to the repo directly. Currently this is convenient because a lot of the 'base' library is still missing, but I probably don't want to make a habit out of this in the long term.

What do you guys think?

If we decide to do this, it would be a good time to change the name as well (if we're going to do that, see #3), so we only have to update configurations once.

benanne:
I'm reopening this, as @avdnoord mentioned in #74 it might be a good idea to move this repository to an organization before we put out the first release. What are everyone's current thoughts on this?

craffel:
I think either is fine; making an organization makes it seem slightly more official in a superficial way.

f0k:
> What are everyone's current thoughts on this?

My thoughts are the same as before:
> It's mainly a question of how you want to present it to the outside: as a project of yours with a couple of contributions from others (= personal account), or as a joint project with you being one of the core maintainers (= organization account)?

benanne:
The latter is a bit more accurate :) Even if it doesn't make much of a difference in practice, it does look a bit more "professional" I suppose.

If nobody has any objections to this, I'll look into it.

benanne:
So, let's do this. I know @f0k was joking about the organization name vote earlier, but it turns out 'Lasagne' is already taken :p So what should we use instead?

craffel:
lasagne_chefs? pasta_chefs?

avdnoord:
Deep_Learning_Jedis_from_Napoli

ebattenberg:
"Layered" is available.  As is "nntools".  I like "Layered" as a kind of abstract name.

Layered/Lasagne
nntools/Lasagne

@benanne Are there other projects that could eventually fit under this organization?  If so, what sort of common theme would be good to express in the organization name?

benanne:
No idea. I guess we can go with something generic to take that into account. I like 'pasta_chefs', or something similarly clever :p

bmcfee:
ChefBoyardeep?

I'll show myself out.

f0k:
[...]

PR #197, opened 2015-04-04
Various documentation changes

MartinThoma:


benanne:
Travis fails because you removed a blank line before `LeakyRectify`. I'll have a proper look at the changes tomorrow!

f0k:
You have two merge commits in there that shouldn't be part of the PR. Also it's good practice to do your PRs from another branch than `master` so it's easier to work on multiple things in parallel or clean up git accidents like this one.
What I would do now is:
```bash
git stash
git checkout -b master-backup
git checkout master
git reset --hard HEAD^^^^
git fetch upstream
git merge --ff-only upstream/master
git cherry-pick 91513b2
git cherry-pick 02fa284
git push --force origin master
git stash pop
```

benanne:
Apologies for the delay - busy week! Will have a look soon.

MartinThoma:
No problem. As long as you take a look at it and I get such helpful feedback, I'm happy :-)

docs/modules/nonlinearities.rst
```diff
@@ -3,5 +3,20 @@
 
 .. automodule:: lasagne.nonlinearities
 
+.. autosummary::
+
+   sigmoid
```
benanne: These autosummaries look nice, but since `sigmoid`, `softmax` and `tanh` are directly imported from Theano, it seems that the docstrings from Theano are inserted here. Here's what I see:
```
sigmoid	Generalizes a scalar op to tensors.
softmax	WRITEME
tanh	hyperbolic tangent of a
```

not particularly useful ;) Is there any way to override these without wrapping them in a custom function? If not I guess we should just do that instead, i.e.
```
def sigmoid(x):
    """our custom docstring"""
    return theano.tensor.nnet.sigmoid(x)
```

Would be nice to avoid it though.

lasagne/nonlinearities.py
```diff
@@ -22,10 +34,30 @@ def rectify(x):
...
+        Calculate the neurons output by applying this leaky rectifier to the
+        activation `x`.
 
+    References
+    ----------
+    The leaky rectifier is described in [1]_.
+
+    .. [1] Maas et al, "Rectifier Nonlinearities Improve Neural Network
+       Acoustic Models",
+       http://web.stanford.edu/~awni/papers/relu_hybrid_icml2013_final.pdf
```
benanne: For some reason this reference ends up being called [R2], which is kind of strange. Any idea what might be going on here? Doesn't matter too much though, just an oddity I noticed.

docs/modules/nonlinearities.rst
MartinThoma: I don't know any way to avoid this without wrapping it in its own function.

However, it seems not too bad to do so. It is very easy to read and only one line of "overhead".

I could also adjust the docstring in Theano. That would probably be the best solution. However, I am not too sure what to write for softmax there.

lasagne/nonlinearities.py
```diff
@@ -22,10 +34,30 @@ def rectify(x):
 
 
 # leaky rectify
-# Maas et al: Rectifier Nonlinearities Improve Neural Network Acoustic Models
-# http://web.stanford.edu/~awni/papers/relu_hybrid_icml2013_final.pdf
 class LeakyRectify(object):
+    """Implementation of a leaky rectifier.
```
benanne: Since we have formulas for `rectify` and `identity`, it might be nice to add one here as well for consistency.

lasagne/nonlinearities.py
```diff
@@ -46,6 +78,18 @@ def __call__(self, x):
...
+
+    Parameters
+    ----------
+    x : float32
+        The activation (the summed, weigthed input of a neuron).
+
+    Returns
+    -------
+    float32
+        The output of the rectify function applied to the activation.
```
benanne: copy-paste error here I guess, this should read 'linear function' instead of 'rectify function'.

lasagne/nonlinearities.py
MartinThoma: I get two references instead of one here. I tried various things to fix this / get a work-around, but everything failed. My hope was that it was a problem of my system and would not happen on readthedocs.

setup.py
```diff
@@ -35,8 +35,13 @@
     long_description="\n\n".join([README, CHANGES]),
     classifiers=[
         "Development Status :: 3 - Alpha",
+        "Intended Audience :: Developers",
+        "Intended Audience :: Science/Research",
+        "License :: OSI Approved :: MIT License",
         "Programming Language :: Python :: 2.7",
+        "Programming Language :: Python :: 3",
         "Programming Language :: Python :: 3.4",
+        "Topic :: Scientific/Engineering :: Artificial Intelligence",
```
benanne: These additions are great! Thanks :)

docs/modules/nonlinearities.rst
benanne: I can imagine there are situations where we would want our docstring to differ from the one in Theano though, so maybe the wrapping approach is the safest bet :)

If you can't come up with a good docstring feel free to just put 'WRITEME' so we can handle it in a different PR.

lasagne/nonlinearities.py
benanne: Right... maybe someone else can figure out what's going on here. In the meantime I don't think this is a showstopper so let's ignore it for now.

benanne:
Looking good, just a few minor comments :)

MartinThoma:
Ok, I think I fixed all issues you mentioned :-)

But I also get

```
$ py.test
[...]
lasagne/theano_extensions/__init__.py SKIPPED
lasagne/theano_extensions/conv.py SKIPPED
lasagne/theano_extensions/padding.py SKIPPED

=============================================================================== FAILURES ===============================================================================
_____________________________________________________________ TestReshapeLayer.test_reference_out_of_range _____________________________________________________________

self = <test_shape.TestReshapeLayer object at 0x7f12ec0baf60>, layerclass = <class 'lasagne.layers.shape.ReshapeLayer'>
two_unknown = (<lasagne.layers.input.InputLayer object at 0x7f12ed23b240>, <TensorType(float64, 5D)>)

    def test_reference_out_of_range(self, layerclass, two_unknown):
        inputlayer, inputdata = two_unknown
        layer = layerclass(inputlayer, (16, 3, 5, 7, [5]))
        with pytest.raises(ValueError):
>           layer.get_output_for(inputdata)

lasagne/tests/layers/test_shape.py:85: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
lasagne/layers/shape.py:137: in get_output_for
    return input.reshape(tuple(output_shape))
/usr/local/lib/python3.4/dist-packages/theano/tensor/var.py:297: in reshape
    return theano.tensor.basic.reshape(self, shape, ndim=ndim)
/usr/local/lib/python3.4/dist-packages/theano/tensor/basic.py:3957: in reshape
    rval = op(x, newshape)
/usr/local/lib/python3.4/dist-packages/theano/gof/op.py:399: in __call__
    node = self.make_node(*inputs, **kwargs)
/usr/local/lib/python3.4/dist-packages/theano/tensor/basic.py:3809: in make_node
    y.get_scalar_constant_value() == 1)
/usr/local/lib/python3.4/dist-packages/theano/tensor/var.py:545: in get_scalar_constant_value
    return theano.tensor.basic.get_scalar_constant_value(self)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

v = Subtensor{5}.0

    def get_scalar_constant_value(v):
        """return the constant scalar(0-D) value underlying variable `v`
    
        If v is the output of dimshuffles, fills, allocs, rebroadcasts, cast
        this function digs through them.
    
        If `v` is not some view of constant scalar data, then raise a
        NotScalarConstantError.
    
        :note: There may be another function similar to this one in the
            code, but I'm not sure where it is.
        """
    
        if v is None:
            # None is not a scalar (and many uses of this function seem to depend
            # on passing it None)
            raise NotScalarConstantError()
    
        if isinstance(v, (numpy.integer, int, float)):
            return numpy.asarray(v)
    
        def numpy_scalar(data):
            """ Return a scalar stored in a numpy ndarray, or raise
            NotScalarConstantError if the numpy ndarray is not a scalar
            """
    
            # handle case where data is numpy.array([])
            if data.ndim > 0 and  (len(data.shape) == 0 or
                __builtins__['max'](data.shape) == 0):
                assert numpy.all(numpy.array([]) == data)
                raise EmptyConstantError()
            try:
                numpy.complex(data)  # works for all numeric scalars
                return data
            except Exception:
                raise NotScalarConstantError(
                    'v.data is non-numeric, non-scalar, or has more than one'
                    ' unique value', data)
    
        if isinstance(v, numpy.ndarray):
            return numpy_scalar(v)
    
        if isinstance(v, Constant):
            if getattr(v.tag, 'unique_value', None) is not None:
                data = v.tag.unique_value
            else:
                data = v.data
            return numpy_scalar(data)
    
        if v.owner:
            if isinstance(v.owner.op, (Alloc, DimShuffle, Rebroadcast,
                                       compile.ops.OutputGuard,
                                       compile.DeepCopyOp)):
                return get_scalar_constant_value(v.owner.inputs[0])
            if isinstance(v.owner.op, Elemwise) and \
                    isinstance(v.owner.op.scalar_op, scal.Second):
                shape, val = v.owner.inputs
                return get_scalar_constant_value(val)
            if isinstance(v.owner.op, scal.Second):
                x, y = v.owner.inputs
                return get_scalar_constant_value(y)
            # Don't act as the constant_folding optimization here as this
            # fct is used too early in the optimization phase.  This would
            # mess with the stabilization optimization.
            if (isinstance(v.owner.op, Elemwise) and isinstance(
                v.owner.op.scalar_op, scal.Cast)) or \
                isinstance(v.owner.op, scal.Cast):
                const = get_scalar_constant_value(v.owner.inputs[0])
                ret = [[None]]
                v.owner.op.perform(v.owner, [const], ret)
                return ret[0][0]
            if isinstance(v.owner.op, theano.tensor.subtensor.Subtensor) and v.ndim == 0:
                # This condition depends on Subtensor always embedding constant
                # indices in the Op rather than making them inputs to the Apply
                # node.
                if isinstance(v.owner.inputs[0], TensorConstant) and \
                    len(v.owner.inputs) == 1:
                    try:
                        return v.owner.inputs[0].data.__getitem__(
                        tuple(v.owner.op.idx_list))
                    except IndexError:
                        raise IndexError(
                                str(tuple(v.owner.op.idx_list)) +
                                " is not a valid index into " +
                                str(v.owner.inputs[0].data))
    
                # The index list 'idx_list' should have length the same
                # shape as the input.
                # TODO: implement the case where we take a scalar in a matrix
                assert len(v.owner.op.idx_list) == v.owner.inputs[0].ndim
    
                # Needed to make better graph in this test in theano/tensor/tests:
                # test_sharedvar.py:test_shared_options.test_specify_shape_partial
                if (v.owner.inputs[0].owner and
                    isinstance(v.owner.inputs[0].owner.op, Join) and
                    # Ensure the Join is joining only scalar variables (so that
                    # the constant value can be found at the same index as the one
                    # used in the sub-tensor).
                    python_all(var.ndim == 0 for var in
                               v.owner.inputs[0].owner.inputs) and
                    len(v.owner.op.idx_list) == 1):
    
                    # Note the '+ 1' is because the first argument to Join is the
                    # axis.
                    ret = v.owner.inputs[0].owner.inputs[
                        v.owner.op.idx_list[0] + 1]
                    ret = get_scalar_constant_value(ret)
                    # join can cast implicitly its input in some case.
                    return theano._asarray(ret, dtype=v.type.dtype)
    
                if (v.owner.inputs[0].owner and
                    isinstance(v.owner.inputs[0].owner.op,
                               theano.tensor.opt.MakeVector) and
                    # MakeVector normally accept only scalar as input.
                    # We put this check in case there is change in the future
                    python_all(var.ndim == 0 for var in
                               v.owner.inputs[0].owner.inputs) and
                    len(v.owner.op.idx_list) == 1 and
                    #idx_list can contain Scalar Type object.
                    isinstance(v.owner.op.idx_list[0], (int,
                                                        numpy.integer))):
    
                    # Python 2.4 does not support indexing with numpy.integer
                    # So we cast it.
                    idx = int(v.owner.op.idx_list[0])
                    ret = v.owner.inputs[0].owner.inputs[idx]
                    ret = get_scalar_constant_value(ret)
                    # MakeVector can cast implicitly its input in some case.
                    return theano._asarray(ret, dtype=v.type.dtype)
    
                # This is needed when we take the grad as the Shape op
                # are not already changed into MakeVector
                owner = v.owner
                leftmost_parent = owner.inputs[0]
                if (leftmost_parent.owner and
                    isinstance(leftmost_parent.owner.op,
                               theano.tensor.Shape)):
                    op = owner.op
                    idx_list = op.idx_list
                    idx = idx_list[0]
                    grandparent = leftmost_parent.owner.inputs[0]
                    gp_broadcastable = grandparent.type.broadcastable
                    ndim = grandparent.type.ndim
    
                    assert ndim == len(gp_broadcastable)
    
                    if not (idx < len(gp_broadcastable)):
                        msg = "get_scalar_constant_value detected " + \
                                "deterministic IndexError: x.shape[%d] " + \
>                               "when x.ndim=%d." % (ndim, idx)
E                       TypeError: not all arguments converted during string formatting

/usr/local/lib/python3.4/dist-packages/theano/tensor/basic.py:657: TypeError
------------------------------------------------------------ coverage: platform linux, python 3.4.2-final-0 ------------------------------------------------------------
```

Is this another issue or did I cause it?

benanne:
This seems to be an error in the Theano code due to operator precedence. The string interpolation operator `%` is executed before string concatenation `+`, so it tries to insert `ndim` and `idx` into the string `"when x.ndim=%d."`, which only contains one placeholder. Adding parens around the concatenation would solve this.

What version are you using? If it's from git and recent you may want to report this to them. The error does not occur on Travis so this code must not have been present in the 0.7 release.

MartinThoma:
I used Theano 0.6.0. I've just updated it to 0.7.0. That fixed it :-)

benanne:
Seems like Travis has disappeared from this particular PR, probably due to the repository move. It was passing before the last commit, so I'm pretty sure it would still pass now :)

There are a few stylistic things I would do differently, but I don't want to nitpick at this point. I'm okay with merging. I'll go ahead and do that tomorrow unless there are further comments.

MartinThoma:
I don't have a problem with fixing those stylistic things if I can learn from it (I'm less interested in single typo or similar). What would you do different?

benanne:
Nah, don't worry about it. It's hard to tell sometimes where personal preference begins and ends :) But this is great, and we should try to improve our documentation coverage before we start finetuning the vocabulary and formulations used. We can do that in later PRs and discuss these things then.

lasagne/nonlinearities.py
```diff
@@ -1,19 +1,79 @@
...
+
+
 # sigmoid
-from theano.tensor.nnet import sigmoid
+def sigmoid(x):
+    """Sigmoid activation function :math:`\\varphi(x) = \\frac{1}{1 + e^{-x}}`
+
+    Parameters
+    ----------
+    x : float32
```
f0k: Not sure about this one. I think this should be `x : scalar or tensor`, as it can take any kind of Theano variable (and even Python scalars or ndarrays). But we need to agree on what to call this in the documentation. In any case, the data type (float32) is not important.

lasagne/nonlinearities.py
```diff
@@ -1,19 +1,79 @@
...
+        The activation (the summed, weigthed input of a neuron).
+
+    Returns
+    -------
+    float32 in [0, 1]
+        The output of the sigmoid function applied to the activation.
```
f0k: The range is useful extra information, but [according to the specifications](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#id6), that first line should only hold the type (`scalar or tensor`). I would include the range in the description instead.

lasagne/nonlinearities.py
```diff
@@ -1,19 +1,79 @@
...
+
 # sigmoid
-from theano.tensor.nnet import sigmoid
+def sigmoid(x):
+    """Sigmoid activation function :math:`\\varphi(x) = \\frac{1}{1 + e^{-x}}`
+
+    Parameters
+    ----------
+    x : float32
+        The activation (the summed, weigthed input of a neuron).
```
f0k: *weighted

lasagne/nonlinearities.py
```diff
@@ -22,10 +82,31 @@ def rectify(x):
...
 # leaky rectify
-# Maas et al: Rectifier Nonlinearities Improve Neural Network Acoustic Models
-# http://web.stanford.edu/~awni/papers/relu_hybrid_icml2013_final.pdf
 class LeakyRectify(object):
+    """Implementation of a leaky rectifier
+    :math:`\\varphi(x) = \max(leakiness \cdot x, x)`.
+
+    Parameters
+    ----------
+    leakiness : float in [0, 1]
```
f0k: Hmm, here the range is really useful to have as part of the type. Maybe just ignore my comment on the return type further above. Also `float` is probably okay for a type even if it can be a Theano scalar instead.

lasagne/nonlinearities.py
```diff
@@ -22,10 +82,31 @@ def rectify(x):
...
+        A leakiness of 0 will lead to the standard rectifier,
+        a leakiness of 1 will lead to a linear activation function.
+
+    Methods
+    -------
+    __call__(x)
+        Calculate the neurons output by applying this leaky rectifier to the
```
f0k: *neuron's

lasagne/nonlinearities.py
f0k: Thinking again, the range is not needed here. Values outside of that range work fine. I'd just prepend a sentence to the description below: "Slope of the function for :math:`x < 0`."

benanne:
Looks like I missed a couple of things! We haven't really standardized on a 'type system' for our docs yet, so for now I wouldn't worry about this too much. We can fix this and work on consistency later. Right now quantity is a little more important than quality imo :)

benanne:
I will go ahead and merge this. We have a bunch more documentation to write and we need to get a move on :) We can do minor fixes and make the notation consistent everywhere in later PRs.

MartinThoma:
Sorry, I didn't have much time the last days to think about the `float32` <-> `scalar or tensor` issue. I hope I can do so in about 10 hours.

The new documentation seems not to be online (see http://lasagne.readthedocs.org/en/latest/modules/nonlinearities.html)

Do I have to wait a little bit longer or did something break with issue #38? (By the way, how do you make sure that there is the latest doc at readthedocs.org? Does readthedocs do that automatically or do you have to push it / notify readthedocs to pull the latest version from pip/github?)

benanne:
Yeah, it's related to #38. Needed to update some settings to point to the new repository. I've manually triggered a build now :) In the future it should be triggered automatically because readthedocs integrates with GitHub.

benanne:
Looks like it's online now :) I had to add `numpydoc` to the `docs/requirements.txt` file to make it work.

PR #199, opened 2015-04-08
Add support for striding and padding to pooling layers

ebenolson:
Work in progress on https://github.com/benanne/Lasagne/issues/186.

I think it's functional but I need to figure out some tests for the `ignore_border = False` mode, and update the documentation.

- [X] Updated tests for `ignore_border = True`
- [X] Tests for `ignore_border = False`
- [X] MaxPool1DLayer stride and pad support
- [X] MaxPool2DLayer stride and pad support
- [X] Documentation

### DIFF
--- a/lasagne/layers/pool.py
+++ b/lasagne/layers/pool.py
@@ -31,6 +31,7 @@ def pool_output_length(input_length, ds, st, ignore_border=True, pad=0):
         The number of elements to be added to the input on each side.
     ignore_border: bool
         If True, partial pooling regions will be ignored.
+        Must be True if pad != 0.
 
     Returns
     -------
@@ -40,13 +41,12 @@ def pool_output_length(input_length, ds, st, ignore_border=True, pad=0):
 
     Notes
     -----
-    When ignore_border = True, this is given by the number of full
+    When `ignore_border == True`, this is given by the number of full
     pooling regions that fit in the padded input length,
     divided by the stride (rounding down).
 
-    When ignore_border = False, partial pooling regions are also
-    allowed, as long as each pooling region contains at least one element
-    not present in another pooling region.
+    If `ignore_border == False`, a single partial pooling region is
+    appended if at least one input element would be left uncovered otherwise.
     """
     if input_length is None or ds is None:
         return None
@@ -58,6 +58,8 @@ def pool_output_length(input_length, ds, st, ignore_border=True, pad=0):
     # output length calculation taken from:
     # https://github.com/Theano/Theano/blob/master/theano/tensor/signal/downsample.py
     else:
+        assert pad == 0
+
         if st >= ds:
             output_length = (input_length + st - 1) // st
         else:
--- a/lasagne/layers/pool.py
+++ b/lasagne/layers/pool.py
@@ -14,76 +14,115 @@ __all__ = [
 ]
 
 
-def pool_output_length(input_length, ds, stride, ignore_border=True, pad=0):
-    '''Compute the output length of a pooling operator
+def pool_output_length(input_length, ds, st, ignore_border=True, pad=0):
+    """
+    Compute the output length of a pooling operator
     along a single dimension.
 
     Parameters
     ----------
-    input_length
-    ds
-    stride
-    pad
-        Length of the input and parameters of the pooling operator.
-
-    ignore_border:
-        if True, the output length is rounded down.
-        if False, it is rounded up.
-        must be True if pad is nonzero.
+    input_length : integer
+        The length of the input in the pooling dimension
+    ds : integer
+        The length of the pooling region
+    st : integer
+        The stride between successive pooling regions
+    pad : integer
+        The number of elements to be added to the input on each side.
+    ignore_border: bool
+        If True, partial pooling regions will be ignored.
 
     Returns
     -------
     output_length
-        * None if either input is None
-        * Computed length of the pooling operator otherwise
-    '''
+        * None if either input is None.
+        * Computed length of the pooling operator otherwise.
 
+    Notes
+    -----
+    When ignore_border = True, this is given by the number of full
+    pooling regions that fit in the padded input length,
+    divided by the stride (rounding down).
+
+    When ignore_border = False, partial pooling regions are also
+    allowed, as long as each pooling region contains at least one element
+    not present in another pooling region.
+    """
     if input_length is None or ds is None:
         return None
 
     if ignore_border:
         output_length = input_length + 2 * pad - ds + 1
-        output_length = (output_length + stride - 1) // stride
+        output_length = (output_length + st - 1) // st
 
     # output length calculation taken from:
     # https://github.com/Theano/Theano/blob/master/theano/tensor/signal/downsample.py
     else:
-        if stride >= ds:
-            output_length = (input_length + stride - 1) // stride
+        if st >= ds:
+            output_length = (input_length + st - 1) // st
         else:
             output_length = max(
-                0, (input_length - ds + stride - 1) // stride) + 1
+                0, (input_length - ds + st - 1) // st) + 1
 
     return output_length
 
 
 class MaxPool1DLayer(Layer):
+    """
+    This layer performs max pooling over the final dimension
+    of a 3D tensor.
+    """
+    def __init__(self, incoming, ds, st=None, pad=0, ignore_border=False,
+                 **kwargs):
+        """
+        Instantiates the layer.
 
-    def __init__(self, incoming, ds, stride=None,
-                 ignore_border=False, pad=0, **kwargs):
+        Parameters
+        ----------
+        incoming : a :class:`Layer` instance or tuple
+            The layer feeding into this layer, or the expected input shape.
+        ds : integer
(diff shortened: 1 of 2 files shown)


lasagne/layers/pool.py
```diff
@@ -15,81 +14,119 @@
...
+                                        ignore_border=self.ignore_border,
+                                        padding=(self.pad, 0),
+                                        )
         return pooled[:, :, :, 0]
 
 
 class MaxPool2DLayer(Layer):
-    def __init__(self, incoming, ds, ignore_border=False, **kwargs):
+
+    def __init__(self, incoming, ds, stride=None,
```
benanne: In `MaxPool2DDNNLayer` the argument to specify the strides is called `strides` (because it is a tuple here). It's probably a good idea to make these interfaces consistent, either by changing it to `strides` here or changing it to `stride` in `MaxPool2DDNNLayer`. Personally I think `strides` makes more sense, as there are two of them in the 2D case.

benanne:
I had a look at the code and the tests so far, and only had one minor comment. Nice work! I don't think `ignore_border=False` tests are super crucial to have (although it is the default case), but it would of course be nice since it's the default. But if that turns out to be more work than anticipated I would be okay with moving that to a separate PR so we can merge this sooner. Up to you!

ebenolson:
Good point. I agree `strides` is better.

What do you think of changing `ds` to something more descriptive, like `kernel` or `pool_shape`?
I assume it's short for 'downsample', but that isn't exactly accurate if you have a different stride.

benanne:
I borrowed `ds` from Theano since that is what it uses. Personally I think that's a good enough reason to use that name (also it's short, which is nice), but I'm not opposed to giving it a better name. I just hope we can keep this from devolving into another drawn out naming discussion as we are wont to do :P

ebenolson:
I suppose as long as it's documented it doesn't matter, and `ds` is nice and short. 

Although for consistency maybe we ought to go with `st` instead of `strides` then.

ebenolson:
I should have time to finish the tests and docs this weekend, I think merging can wait till after that.

benanne:
It seems Theano is very inconsistent with these names (for `dnn_pool` they are called `ws` and `stride`, but for `max_pool_2d` they are `ds` and `st` respectively). So maybe following Theano isn't as great an idea as I thought :p 

So if we decide to change any names it should probably be to make them nicer and easier to understand, not to make them more consistent with Theano since it is itself inconsistent, apparently!

> I should have time to finish the tests and docs this weekend, I think merging can wait till after that.

Sure, take your time!

lasagne/layers/pool.py
f0k: What about the 1D case? The argument should probably have the same name for 1D, 2D and 3D implementations to make it easier to write generic code applying to all of them.

benanne: I don't know, I feel that it's more important to provide an intuitive interface for `Conv1DLayer`. I don't think forcing people to specify 1-element tuples for `ds` and `strides` for the 1D case would be worth it.

If we ever end up with a lot of code that applies to all dimensionalities, we could always write a `max_pool` wrapper function that does the right thing based on an additional `dimensionality` argument, or something.

f0k: > I don't think forcing people to specify 1-element tuples for  ds  and  strides  for the 1D case would be worth it.

No, that shouldn't be necessary, of course. I just think calling it `stride` in one class and `strides` in another could make some things more complicated. I'd call it the same in all classes and have it accept both tuples of the correct length or a single integer (that gets promoted to a tuple automatically). That's more of a feature request than a comment to this PR, though.

ebenolson:
Ok, I think this is ready for review.

I changed `stride` to `st` to be more consistent with `ds`, and because it can work for both 1D and 2D, but   if you prefer `stride` and `strides` I can change it to that.

lasagne/layers/pool.py
```diff
@@ -15,81 +14,212 @@
...
+
+    Notes
+    -----
+    When ignore_border = True, this is given by the number of full
+    pooling regions that fit in the padded input length,
+    divided by the stride (rounding down).
+
+    When ignore_border = False, partial pooling regions are also
+    allowed, as long as each pooling region contains at least one element
+    not present in another pooling region.
```
f0k: I'm a bit confused about that definition. I think it's because it speaks of partial pooling regions (in plural). What about:
```
For `ignore_border=False`, a single partial pooling region is appended if at least one input element would be left uncovered otherwise.
```

lasagne/layers/pool.py
```diff
@@ -15,81 +14,212 @@
...
+        output_length = (output_length + st - 1) // st
+
+    # output length calculation taken from:
+    # https://github.com/Theano/Theano/blob/master/theano/tensor/signal/downsample.py
+    else:
+        if st >= ds:
+            output_length = (input_length + st - 1) // st
+        else:
+            output_length = max(
+                0, (input_length - ds + st - 1) // st) + 1
```
f0k: I think you're missing the padding here. [In theano.tensor.signal.downsample](https://github.com/Theano/Theano/blob/master/theano/tensor/signal/downsample.py#L143-144), they pad unconditionally before checking for `ignore_border`. If I'm right, there should be a test for this scenario, otherwise a comment explaining it would be great.

lasagne/layers/pool.py
ebenolson: Yeah, that's a lot better. I think I was thinking of the 2D case but that's not relevant in this function.

lasagne/layers/pool.py
f0k: Ah, I see, this case is forbidden in the constructor of the pooling classes (hadn't got to that bit yet). A comment or even an assertion would be nice then.

lasagne/layers/pool.py
```diff
@@ -15,81 +14,212 @@
...
+        -----
+        The value used to pad the input is chosen to be less than
+        the minimum of the input, so that the output of each pooling region
+        always corresponds to some element in the unpadded input region.
+        """
         super(MaxPool1DLayer, self).__init__(incoming, **kwargs)
         self.ds = ds  # an integer
+        self.st = ds if st is None else st
+        self.pad = pad
         self.ignore_border = ignore_border
```
f0k: I thought that the `ignore_border`/`pad` incompatibility needs to be checked somewhere, but then I had a look at the Theano implementation to see that [it raises a `NotImplementedError`](https://github.com/Theano/Theano/blob/master/theano/tensor/signal/downsample.py#L214-216). I think it's good to leave it up to Theano along with all other input checking then.

lasagne/tests/layers/test_pool.py
```diff
@@ -1,90 +1,215 @@
...
+        for st in [1, 2, 3, 4]:
+            for pad in range(ds):
+                yield (ds, st, pad)
 
-    def test_get_output_shape_for(self, layer):
+
+class TestMaxPool1DLayer:
+
+    def input_layer(self, output_shape):
+        return Mock(get_output_shape=lambda: output_shape)
```
f0k: This will break with #182. I can deal with that, we just need to remember.

f0k:
Looks good to me. Personally I don't like either of `ds` or `st` (and other short names in Theano such as `bs`, `ishp`, `kshp`), but I like that it's the same name for all implementations now and I don't want to start another name discussion :)

benanne:
> I changed `stride` to `st` to be more consistent with `ds`, and because it can work for both 1D and 2D, but   if you prefer `stride` and `strides` I can change it to that.

Unfortunately this makes it inconsistent with the `Conv*DLayer`s, because they use `stride` / `strides`. I think we should at least try to be consistent within the library, so if we change it here, we will have to change it there as well.

lasagne/layers/pool.py
benanne: Agreed.

f0k:
Before we shorten the names of the convolution as well, I'd prefer to expand the names here to `stride` (and `strides`, if it must be so) for `st` and `pool_size` for `ds`. `pool_size` nicely fits `filter_size` for the convolution.

Checking again, I see that Conv1D has `stride` and `filter_length`, the others have `strides` and `filter_size`. I'd suggest `stride` and `filter_size` regardless of dimensionality for convolution, and `stride` and `pool_size` regardless of dimensionality for pooling. Unifying the convolution names is out of scope for this PR, but we can plan for that already.

lasagne/layers/pool.py
f0k: Can you change that bit then? With this and the assertion below it would be good for merging! If you don't have the time, let us know and we'll take care of it as part of the merge.

ebenolson:
ok, added the assert and fixed the docs. leaving `st` as is for now.

f0k:
Great, thanks a lot! Will wait for Travis to pass and merge tomorrow (unless Sander beats me to it).

benanne:
I guess I'm beating you to it :) Let's indeed keep the naming stuff for a separate PR so we don't drag things out. But we should nevertheless handle it soon, I reckon.

@ebenolson thanks a lot for this!

PR #210, opened 2015-04-17
Unify keyword argument names for convolution and pooling

### LINKED ISSUE #207: Unify keyword argument names for convolution and pooling
The keyword argument names for 1D and {2,3}D convolution and pooling are currently not the same.

* 1D-convolution has `filter_length` and `stride`
* {2,3}D-convolution has `filter_size` and `strides`
* {1,2,3}D-pooling has `ds` and `st` in #199

I'd suggest to use `filter_size` and `stride` for all convolutions, and `pool_size` and `stride` for all poolings.
For a really uniform interface, it would be nice if all of `filter_size`, `pool_size` and `stride` either accepted a single integer or a tuple of integers matching the dimensionality.

Arguments for or against it?

ebenolson:
This is to close #207 

I've made changes to use `filter_size`, `pool_size` and `stride` throughout all conv and pooling layers. 

Added a helper function `as_tuple` in `utils.py` which takes an input and a length and returns a tuple of that length, and use it in the 2D layers to accept either a number or tuple for `pool_size` and `stride`. 

I also added some tests for the cuda_convnet and dnn pooling layers, and updated the examples to use `pool_size`.

### DIFF
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -13,18 +13,18 @@ __all__ = [
 ]
 
 
-def conv_output_length(input_length, filter_length,
+def conv_output_length(input_length, filter_size,
                        stride, border_mode, pad=0):
     if input_length is None:
         return None
     if border_mode == 'valid':
-        output_length = input_length - filter_length + 1
+        output_length = input_length - filter_size + 1
     elif border_mode == 'full':
-        output_length = input_length + filter_length - 1
+        output_length = input_length + filter_size - 1
     elif border_mode == 'same':
         output_length = input_length
     elif border_mode == 'pad':
-        output_length = input_length + 2 * pad - filter_length + 1
+        output_length = input_length + 2 * pad - filter_size + 1
     else:
         raise RuntimeError('Invalid border mode: {0}'.format(border_mode))
 
@@ -36,7 +36,7 @@ def conv_output_length(input_length, filter_length,
 
 
 class Conv1DLayer(Layer):
-    def __init__(self, incoming, num_filters, filter_length, stride=1,
+    def __init__(self, incoming, num_filters, filter_size, stride=1,
                  border_mode="valid", untie_biases=False,
                  W=init.GlorotUniform(), b=init.Constant(0.),
                  nonlinearity=nonlinearities.rectify,
@@ -48,7 +48,7 @@ class Conv1DLayer(Layer):
             self.nonlinearity = nonlinearity
 
         self.num_filters = num_filters
-        self.filter_length = filter_length
+        self.filter_size = filter_size
         self.stride = stride
         self.border_mode = border_mode
         self.untie_biases = untie_biases
@@ -66,7 +66,7 @@ class Conv1DLayer(Layer):
 
     def get_W_shape(self):
         num_input_channels = self.input_shape[1]
-        return (self.num_filters, num_input_channels, self.filter_length)
+        return (self.num_filters, num_input_channels, self.filter_size)
 
     def get_params(self):
         return [self.W] + self.get_bias_params()
@@ -76,7 +76,7 @@ class Conv1DLayer(Layer):
 
     def get_output_shape_for(self, input_shape):
         output_length = conv_output_length(input_shape[2],
-                                           self.filter_length,
+                                           self.filter_size,
                                            self.stride,
                                            self.border_mode)
 
@@ -105,7 +105,7 @@ class Conv1DLayer(Layer):
                                       image_shape=input_shape,
                                       filter_shape=filter_shape,
                                       border_mode='full')
-            shift = (self.filter_length - 1) // 2
+            shift = (self.filter_size - 1) // 2
             conved = conved[:, :, shift:input.shape[2] + shift]
         else:
             raise RuntimeError("Invalid border mode: '%s'" % self.border_mode)
@@ -121,7 +121,7 @@ class Conv1DLayer(Layer):
 
 
 class Conv2DLayer(Layer):
-    def __init__(self, incoming, num_filters, filter_size, strides=(1, 1),
+    def __init__(self, incoming, num_filters, filter_size, stride=(1, 1),
                  border_mode="valid", untie_biases=False,
                  W=init.GlorotUniform(), b=init.Constant(0.),
                  nonlinearity=nonlinearities.rectify,
@@ -134,7 +134,7 @@ class Conv2DLayer(Layer):
 
         self.num_filters = num_filters
         self.filter_size = filter_size
-        self.strides = strides
+        self.stride = stride
         self.border_mode = border_mode
         self.untie_biases = untie_biases
         self.convolution = convolution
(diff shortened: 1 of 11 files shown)


lasagne/layers/conv.py
```diff
@@ -133,8 +134,8 @@ def __init__(self, incoming, num_filters, filter_size, strides=(1, 1),
             self.nonlinearity = nonlinearity
 
         self.num_filters = num_filters
-        self.filter_size = filter_size
-        self.strides = strides
+        self.filter_size = as_tuple(filter_size, 2)
+        self.stride = as_tuple(stride, 2)
```
f0k: Nice. That looks very clean.

lasagne/layers/cuda_convnet.py
```diff
@@ -198,19 +202,33 @@ def get_output_for(self, input, **kwargs):
...
 
 class MaxPool2DCCLayer(CCLayer):
-    def __init__(self, incoming, ds, ignore_border=False, strides=None,
+    def __init__(self, incoming, pool_size, ignore_border=False, stride=None,
                  dimshuffle=True, **kwargs):
         from pylearn2.sandbox.cuda_convnet.pool import MaxPool
 
+        if 'pad' in kwargs:
+            raise NotImplementedError("MaxPool2DCCLayer does not "
+                                      "support padding")
```
f0k: Good idea, that's more descriptive than letting the base class constructor raise something about an unknown kwarg.

ebenolson:
just realized my tests are not testing very much :/ 

need to think a bit more about how to do this cleanly.

lasagne/layers/conv.py
```diff
@@ -48,7 +49,7 @@ def __init__(self, incoming, num_filters, filter_length, stride=1,
             self.nonlinearity = nonlinearity
 
         self.num_filters = num_filters
-        self.filter_length = filter_length
+        self.filter_size = filter_size
```
benanne: For consistency with the 2D (and possibly in the future 3D) case, I think it would be nice to do `as_tuple(filter_size, 1)` here, so that the `filter_size` argument can also accept single element tuples. This will also require some other modifications to the code, as it is currently assumed to be an integer (and it would become a single element tuple).

benanne:
Looks great so far :) Thanks a lot! I will have a more detailed look at the code when you're happy with the tests.

lasagne/layers/conv.py
ebenolson: If we're going to use tuples everywhere, what do you think about standardizing on `filter_shape` and `pool_shape` instead? Seems like it would be more consistent with `input_shape` and `output_shape`.

benanne: That's a very good point, although it makes it slightly more awkward when specifying it as an integer. `pool_shape=3` looks a bit weird to me because I've been conditioned to expect a shape to look like a tuple, and I think this goes for most numpy / Theano users :)

But of course there is a lot to be said for consistency. Would love to hear @f0k's thoughts on this (and well, anyone else's)!

f0k: Wohoo, another name discussion! ;) `filter_shape` would also be consistent with Theano, but we don't necessarily need that. As a counterexample, `scipy.ndimage.median_filter` has a `size` keyword that can be a scalar or tuple. And I agree that `pool_shape=3` looks weird, as if we would enumerate possible shapes (1: square, 2: triangle, 3: circle).

We could just make it a rule that things that *must* be tuples are `shape`, and things that can also be scalars are `size`. If we follow that everywhere, this would still count as consistent.

Regarding the original comment, I thought about that, too, but I think `as_tuple(filter_size, 1)` would make the rest of the code a bit harder to read. It could be helpful for refactoring the code, though (there's a lot of duplication between the convolution classes).

benanne: I think if we're going to accept integers and tuples for the 2D and 3D case, there is no reason to only accept integers for the 1D case. It seems more consistent to use `as_tuple` everywhere imo.

> We could just make it a rule that things that must be tuples are `shape`, and things that can also be scalars are `size`. If we follow that everywhere, this would still count as consistent.

Works for me. But using `shape` everywhere is also fine really. It's not that crucial.

benanne:
Looks like there are 3 things left to do on this one
- [ ] use as_tuple in the 1D case as well
- [ ] (optionally) rename the arguments from `_size` to `_shape` (do whatever you prefer here, I'm fine either way)
- [ ] fix tests

If there's any of these I can take off your hands let me know. I believe it should be possible to send PRs for PRs ;)

ebenolson:
ok, I'll try to get this wrapped up later today.

ebenolson:
Reworked `test_pool.py`, due to all the restrictions on `cuda_convnet` it seemed simpler to make separate tests for each implementation. I also added some tests for `Conv1DLayer`, at some point I want to clean up the conv tests more but for now it's better than nothing.

Changed the 1D conv and pool layers to use `as_tuple` as well.

I added the `pad` argument to `MaxPool2DDNNLayer` since it seems to work now (I guess v2 wasn't out when that was written?)

I decided to stick with `_size`, I think @f0k's suggestion to standardize on `_size` for things that may be either tuples or scalar makes sense.

lasagne/layers/cuda_convnet.py
```diff
@@ -198,34 +202,51 @@ def get_output_for(self, input, **kwargs):
...
+            raise NotImplementedError("MaxPool2DCCLayer only supports "
+                                      "stride <= pool_size.")
 
         # ignore_border argument is for compatibility with MaxPool2DLayer.
         # it is not supported. Borders are never ignored.
         if ignore_border is not False:
-            raise RuntimeError("MaxPool2DCCLayer does not support "
-                               "ignore_border.")
+            raise NotImplementedError("MaxPool2DCCLayer does not support "
+                                      "ignore_border.")
```
f0k: Makes sense. It's even something that one *could* add here if anybody ever needs it.

lasagne/layers/pool.py
```diff
@@ -236,38 +224,45 @@ class FeaturePoolLayer(Layer):
...
 
-    def __init__(self, incoming, ds, axis=1, pool_function=T.max, **kwargs):
+    def __init__(self, incoming, pool_size, axis=1, pool_function=T.max,
+                 **kwargs):
         """
-        ds: the number of feature maps to be pooled together
-        axis: the axis along which to pool. The default value of 1 works
-        for DenseLayer and Conv*DLayers
-        pool_function: the pooling function to use
+        Instrideantiates the layer.
```
f0k: Haha, nice neologism! Freudian slip? :D

lasagne/layers/pool.py
```diff
@@ -145,11 +151,11 @@ def __init__(self, incoming, ds, st=None,
         ----------
         incoming : a :class:`Layer` instance or tuple
             The layer feeding into this layer, or the expected input shape.
-        ds : integer or iterable
+        pool_size : integer or iterable
             The length of the pooling region in each dimension
```
f0k: That's quite elegant, actually. The text fits both the integer and iterable case.

lasagne/utils.py
```diff
@@ -66,6 +66,28 @@ def unique(l):
...
 
+def as_tuple(x, N):
+    """
+    Coerce a value to a tuple of length N.
+
+    Parameters:
+    -----------
+    x : value or iterable
+    N : integer
+        length of the desired tuple
```
f0k:
```
Returns:
--------
tuple
    ``tuple(x)`` if `x` is iterable, ``(x,) * N`` otherwise.
```

f0k:
Thanks, looks very clean and mergeable to me, except for two minor comments!

ebenolson:
oops, thought I'd caught all those. thanks for reviewing.

benanne:
Looking good! Will merge tomorrow then :)

benanne:
Merging! I'll make a post on the mailing list about the changed interface.

PR #211, opened 2015-04-18
updates.py docs

skaae:
Add docs to 

1.  Adadelta
2. Adagrad
3. RMSprop

Should i include usage examples?

### DIFF
--- a/lasagne/updates.py
+++ b/lasagne/updates.py
@@ -1,11 +1,11 @@
 """
 Functions to generate Theano update dictionaries for training.
 
-The update functions implements different methods to control the learning
+The update functions implement different methods to control the learning
 rate for use with stochastic gradient descent.
 
 
-In vanilla SGD generates update expressions of the form:
+Vanilla SGD generates update expressions of the form:
 
     * ``param := param - learning_rate * gradient``
 
@@ -14,12 +14,13 @@ With momentum the generated updates are of the form:
     * ``velocity := momentum * velocity - learning_rate * gradient``
     * ``param := param + velocity``
 
-For description of more advanced methods refer to the function descriptions.
+For description of more advanced methods please refer to the function
+descriptions.
 
 Update functions take a loss function or an ordered dictionary of updates as
 input and return an ordered dict of updates.
 
-The loss function must return a single scalar value.
+The loss function expression must return a single scalar value.
 
 Usage
 --------
@@ -31,14 +32,14 @@ Usage
 >>> from lasagne.updates import sgd, apply_momentum
 >>> l_in = InputLayer((100, 20))
 >>> l1 = DenseLayer(l_in, num_units=3, nonlinearity=softmax)
->>> sym_x = T.matrix('sym_x')  # shp: num_batch x num_features
->>> sym_y = T.ivector('sym_y') # shp: num_batch
->>> l_out = l1.get_output(sym_x)
+>>> x = T.matrix('x')  # shp: num_batch x num_features
+>>> y = T.ivector('y') # shp: num_batch
+>>> l_out = l1.get_output(x)
 >>> params = lasagne.layers.get_all_params(l1)
->>> loss = T.mean(T.nnet.categorical_crossentropy(l_out, sym_y))
+>>> loss = T.mean(T.nnet.categorical_crossentropy(l_out, y))
 >>> updates_sgd = sgd(loss, params, learning_rate=0.0001)
 >>> updates = apply_momentum(updates_sgd, params, momentum=0.9)
->>> train_function = theano.function([sym_x, sym_y], updates=updates)
+>>> train_function = theano.function([x, y], updates=updates)
 """
 
 from collections import OrderedDict
--- a/lasagne/updates.py
+++ b/lasagne/updates.py
@@ -1,5 +1,44 @@
 """
 Functions to generate Theano update dictionaries for training.
+
+The update functions implements different methods to control the learning
+rate for use with stochastic gradient descent.
+
+
+In vanilla SGD generates update expressions of the form:
+
+    * ``param := param - learning_rate * gradient``
+
+With momentum the generated updates are of the form:
+
+    * ``velocity := momentum * velocity - learning_rate * gradient``
+    * ``param := param + velocity``
+
+For description of more advanced methods refer to the function descriptions.
+
+Update functions take a loss function or an ordered dictionary of updates as
+input and return an ordered dict of updates.
+
+The loss function must return a single scalar value.
+
+Usage
+--------
+>>> import lasagne
+>>> import theano.tensor as T
+>>> import theano
+>>> from lasagne.nonlinearities import softmax
+>>> from lasagne.layers import InputLayer, DenseLayer
+>>> from lasagne.updates import sgd, apply_momentum
+>>> l_in = InputLayer((100, 20))
+>>> l1 = DenseLayer(l_in, num_units=3, nonlinearity=softmax)
+>>> sym_x = T.matrix('sym_x')  # shp: num_batch x num_features
+>>> sym_y = T.ivector('sym_y') # shp: num_batch
+>>> l_out = l1.get_output(sym_x)
+>>> params = lasagne.layers.get_all_params(l1)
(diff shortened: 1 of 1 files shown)


MartinThoma:
I love usage examples. However, I am not too sure if explained examples how to apply adagrad (for example) would probably fit better in the tutorial section of the documentation.

benanne:
Nice! It seems that some of the stuff that was in the docstrings before (like the references) would be better suited as part of the 'notes' section. What do you think?

Usage examples for the different update functions will look largely the same I imagine, so I agree that it's probably better to add this to the user guide / a tutorial instead.

skaae:
I'm not sure what you mean :) Is there a references section for all py files or?

I agree that usage examples are better put in the tutorial.

benanne:
For example:

```
Epsilon is not included in the typical formula,
+    See "Notes on AdaGrad" by Chris Dyer for more info [2].
```

is the first thing in the docstring, before the epsilon is even listed (because the parameters list becomes afterwards). Maybe it would be better to move this to 'notes' instead.

skaae:
yes. I'll clean that up.

benanne:
Looks like Travis fails on Python 2 and Python 3, but for different reasons. On Python 2 it fails because there's a unicode character in a docstring somewhere. Python 2 requires the encoding to be declared in this case (see http://stackoverflow.com/questions/6289474/working-with-utf-8-encoding-in-python-source ).

On Python 3 it fails because there's a line in one of the docstrings that is 80 characters, whereas PEP8 requires lines in docstrings to have at most 79 characters (no idea why).

MartinThoma:
We should probably add the line `# -*- coding: utf-8 -*-` to all files.

> whereas PEP8 requires lines in docstrings to have at most 79 characters (no idea why).

Quote from Guido van Rossum:

> Because an 80 char wide Emacs window starts wrapping at 79.

Source: http://www.reddit.com/r/learnpython/comments/1h2eug/pep_8_why_is_the_character_limit_79_and_not_80/

benanne:
Oh, I thought it was only in docstrings, but I guess it's everywhere then. Good to know.

Agree, we should probably add it to all files. Although we might want to wait with doing that until we've sorted out all the planned interface changes, to avoid excessive rebasing :)

skaae:
Do you know how i find the unicode character?

MartinThoma:
You could try http://stackoverflow.com/a/9395552/562769

skaae:
ok, the last name should have been non ascii :)

benanne:
It's in the Traceback on Travis: https://travis-ci.org/Lasagne/Lasagne/jobs/59040844

skaae:
2.7 passes now, but 3.4 throws an error about downloading mongo db?

benanne:
Looks like the server was down for a bit or something. I've restarted the build.

f0k:
Regarding line length, [the numpydoc guide](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#docstring-standard) recommends docstring lines to not exceed 75 characters, but I'm not sure if that excludes the indentation (ending at the same 79 characters for most things) or includes it (so docstring lines should end with the 75th column; that's what I did for the first couple of docstrings in `updates.py`).

Documentation looks good, thank you! I'll just add one question to the diff in a minute.

lasagne/updates.py
```diff
@@ -54,9 +54,6 @@ def get_or_compute_grads(loss_or_grads, params):
 def sgd(loss_or_grads, params, learning_rate):
     """Stochastic Gradient Descent (SGD) updates.
 
-    Generates update expressions of the form:
-    * ``param := param - learning_rate * gradient``
-
```
f0k: Why did you remove these explanations? I thought they were useful to compare momentum and nesterov momentum (and to be explicit about the formulation we use, as there is another momentum formulation that scales the learning rate by `1 - momentum`). I'm not sure if that was the best place to put them, but I don't think they should just be dropped. Are they confusing/unneeded?

lasagne/updates.py
```diff
@@ -262,10 +251,52 @@ def nesterov_momentum(loss_or_grads, params, learning_rate, momentum=0.9):
...
+
+    Parameters
+    ----------
+    loss_or_grads : symbolic expression or list of expressions
+        A scalar loss expression, or a list of gradient expressions
+    params : list of shared variables
+        The variables to generate update expressions for
+    learning_rate : float or symbolic scalar
+        The learning rate controlling the size of update steps
+    epsilon : float or symbolic scalar,
```
f0k: There shouldn't be a comma here (same thing further below), but that's probably not really critical.

lasagne/updates.py
```diff
@@ -262,10 +251,52 @@ def nesterov_momentum(loss_or_grads, params, learning_rate, momentum=0.9):
 
 
 def adagrad(loss_or_grads, params, learning_rate=1.0, epsilon=1e-6):
+    """Adagrad updates [1]
```
f0k: From [the guide](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections) (down at "11. References"), I think the syntax should be `[1]_` to reference something, and `[1]` to define the reference. Haven't tried it myself, though.

lasagne/updates.py
skaae: no, I'll remove the comma.

lasagne/updates.py
skaae: I removed them to make the documentation consistent accross different update types, but can definetly put them back in. Is it possible to add comments that apply to all update functions?

f0k: > Is it possible to add comments that apply to all update functions?

Such comments could be put in the module docstring, but I don't know what that will render like in Sphinx.

benanne:
I think there's one or two more remarks that have not been addressed. If you think you won't have time to do that within the next day or 2-3 I can do it when merging, just let me know. I'd like to keep the "documentation writing momentum" up so I'd like to merge it soon :)

skaae:
It's fine. Let me know if I need to change anything.

benanne:
Just the last few inline remarks by @f0k that haven't been addressed (at least I think so, else they would show up as "f0k commented on an outdated diff" right?)

skaae:
whops... i'll remove all the pycharm files...

skaae:
I think the last thing that needs to be adressed is @f0k's about line 57-59. Any suggestions?

benanne:
If we want to be certain it shows up in the docs, I guess the best idea would be to add it directly to `docs/modules/updates.rst`.

skaae:
I looked at the docs at readthedocs and it seems that we can just modify the docstring at line 1-3 

```
Functions to generate Theano update dictionaries for training.

- some general description of updates.....
```

benanne:
Great! It's definitely nicer to be able to keep all the API docs in docstrings.

skaae:
I'll add the general description later today.

skaae:
I added a suggestion for a overall description and a simple usage example.

feedback is welcome

benanne:
Looking good! Just a few minor spelling remarks.

skaae:
Yes. I have included your corrections. I'll change sym_x and sym_y to x and y. I use prefixes in my
code to distinquish shared vars, symbolic variables etc.

benanne:
Alright, looks good to me! Will merge tomorrow if there are no further comments.

benanne:
I guess I'll wait until tomorrow again then, just in case :)

f0k:
I'm happy!

benanne:
Merging!

lasagne/updates.py
```diff
@@ -222,8 +267,8 @@ def nesterov_momentum(loss_or_grads, params, learning_rate, momentum=0.9):
     """Stochastic Gradient Descent (SGD) updates with Nesterov momentum.
 
     Generates update expressions of the form:
-    * ``velocity := momentum * velocity - learning_rate * gradient``
-    * ``param := param + momentum * velocity - learning_rate * gradient``
+    * ``velocity := momentum * velocity + updates[param] - param``
+    * ``param := param + momentum * velocity + updates[param] - param``
```
f0k: This change was wrong, see #527.

PR #212, opened 2015-04-18
add doc to noise.py

skaae:
adds documentation to the GuassianNoiseLayer and the DropoutLayer.

I'm not sure what is the correct way to document `get_output_for` with `deterministic`, `rescale` etc keywords?


benanne:
I think what you did looks fine. Not sure what `use full at test time` means though, did you mean `false`?

The use of keyword arguments that 'propagate' through the network is a feature that needs to be documented in the user guide.

skaae:
I updated it to see notes? I wanted to make people aware that they should use deterministic differently 
during training and testing.

benanne:
Looks good to me :)

benanne:
Travis is saying "no newline at end of file", could you add it back?

benanne:
Another PEP8 error at the dropout shortcut, because you removed a space before a comment there. Could you put it back?

skaae:
yes. Do i install pyflakes to check for pep8 before i upload?

benanne:
if you run the tests with pytest-pep8 installed it should check. Your editor may also have the option to check on the fly (I use sublime with sublime linter for this).

If you're comfortable enough with git it might be useful to squash those last few commits as well. If not then don't worry about it.

skaae:
Not really that part of git.

MartinThoma:
We should eventually add detailed instructions how to run all tests (especially PEP8) with a git pre-commit hook to http://lasagne.readthedocs.org/en/latest/user/development.html

@benanne Do you have a pre-commit hook for PEP8? Do you know how to install it?If you don't know how it is done / don't have time for writing it, but think its a good idea I would add it to the development page. According to http://codeinthehole.com/writing/tips-for-using-a-git-pre-commit-hook/ it seems to be quite simple.

benanne:
I've never done anything with pre-commit hooks. My git-fu is limited, @f0k is the master around here :) I usually just run the tests before I create a PR, which includes PEP8 checks.

But I agree it's a good idea. Not a huge priority at the moment though :)

lasagne/layers/noise.py
```diff
@@ -40,11 +88,47 @@ def get_output_for(self, input, deterministic=False, **kwargs):
 
 
 class GaussianNoiseLayer(Layer):
-    def __init__(self, incoming, sigma=0.1, **kwargs):
-        super(GaussianNoiseLayer, self).__init__(incoming, **kwargs)
+    """Gaussian noise layer [1]
+
+    Add zero Guassian noise with mean 0 and std. sigma to the input
```
benanne: Should be `Gaussian` :) The period after 'std' also looks a bit out of place.

lasagne/layers/noise.py
```diff
@@ -40,11 +88,47 @@ def get_output_for(self, input, deterministic=False, **kwargs):
...
+    Parameters
+    ----------
+    input_layer : `Layer` instance
+        The layer from which this layer will obtain its input
+    sigma : float or tensor scalar
+        Std. of added gaussian noise
+
+    Notes
+    ----------
+    The Guassian noise layer is a regularizer. During training you should set
```
benanne: Same here

skaae:
@benanne: Thaks for the corrections, don't know what happend to my spelling of Gaussian.

Do you know a good page where i can learn to squash all these small commits?

f0k:
> Do you know a good page where i can learn to squash all these small commits?

To merge the last four commits into one, with a new description:
```bash
git reset --soft HEAD~4
git commit -m "some good description"
git push --force origin updates_noise
```

lasagne/layers/noise.py
```diff
@@ -15,12 +17,58 @@
...
+
+    Sets values to zero with probability p. See notes for disabling dropout
+    during testing.
+
+    Parameters
+    ----------
+    input_layer : `Layer` instance
+        The layer from which this layer will obtain its input
+    p : float or scalar tensor
+        The probability of setting a value to zero
```
f0k: `input_layer` should still be `incoming`, until we decide to not allow creating layers from shape tuples any longer. You can copy the description from the base class. `rescale` is missing in the parameter list.

lasagne/layers/noise.py
```diff
@@ -15,12 +17,58 @@
...
+
+    Notes
+    ----------
+    The dropout layer is a regularizer that randomly sets input values to
+    zero, see [1,2] for a why this might improve generalization.
+    During training you should set deterministic to false and during
+    testing you should set deterministic to true.
```
f0k: If I understand the [numpydoc guide](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) correctly, I think the correct markup would be:
```
During training you should set `deterministic` to ``False`` and during testing you should set `deterministic` to ``False``.
```
I.e., single backticks for referencing argument names, and double backticks for other things to be set in a monospace font.

lasagne/layers/noise.py
```diff
@@ -15,12 +17,58 @@
...
+    Improving neural networks by preventing co-adaptation of feature detectors.
+    arXiv Preprint, 1207.0580(Hinton, Geoffrey E., et al.
+
+    [2] Srivastava Nitish, Hinton, G., Krizhevsky, A., Sutskever,
+    I., & Salakhutdinov, R. R. (2014).
+    Dropout: A Simple Way to Prevent Neural Networks from Overfitting.
+    Journal of Machine Learning Research, 5(Jun)(2), 1929-1958.
+    """
+    def __init__(self, input_layer, p=0.5, rescale=True):
+        super(DropoutLayer, self).__init__(input_layer)
```
f0k: Hmm, while you're at it (and renaming `input_layer` back to `incoming`), could you please add `**kwargs` as the last argument in both lines (the `__init__` signature and the super class constructor call)? This is needed to be able to pass `name=...` upwards the class hierarchy, but the tests currently don't check that for every single layer class.

skaae:
yes. I misplaced the rescale description in get_output_for.

lasagne/layers/noise.py
```diff
@@ -15,12 +17,58 @@
...
         self.p = p
         self.rescale = rescale
 
-    def get_output_for(self, input, deterministic=False, **kwargs):
+    def get_output_for(self, input, deterministic=False, *args, **kwargs):
```
f0k: The `*args` have been removed some time ago. Are you working from the latest `master`?

lasagne/layers/noise.py
```diff
@@ -29,7 +77,7 @@ def get_output_for(self, input, deterministic=False, **kwargs):
                 input /= retain_prob
 
             # use nonsymbolic shape for dropout mask if possible
-            input_shape = self.input_shape
+            input_shape = self.input_layer.get_output_shape()
```
f0k: This also seems like you didn't start from the latest master.

skaae:
I thought i was, but it doesn't seem so. Isn't i easier to just close this PR and create new one with the correct master?

f0k:
> I thought i was, but it doesn't seem so. Isn't i easier to just close this PR and create new one with the correct master?

That's one option, but you can also overwrite your `updates_noise` branch with whatever needs to be in there. Something like:
```bash
git branch -m updates_noise backup
git checkout master
# make sure it's the latest one, as detailed above
git checkout -b updates_noise
# copy back your docstrings from somewhere, manually, and commit
git push --force origin updates_noise  # that's the key, using --force you can overwrite anything
```

skaae:
I'll close this and make a new PR

f0k:
> I'll close this and make a new PR

All right! Checking your commit history, you *did* start from the latest master, you just seem to have copied in an older version of `noise.py` then. Thanks for fixing this up!

skaae:
Ok, I used the github app, and it seems to dublicate the branches every time i make a commit?? I don't know whats going on in the app :)

skaae:
Created new PR in https://github.com/Lasagne/Lasagne/pull/214

PR #213, opened 2015-04-18
Githook and Lasagne organization url

MartinThoma:
I've just 

* added a pre-commit githook which makes sure that `py.test` still passes after the commit
* and changed the urls from https://github.com/benanne/Lasagne -> https://github.com/Lasagne/Lasagne

The pre-commit hook worked for this commit.


benanne:
Looks good, but I don't know how practical this is to be honest... running the test takes several minutes (even longer when they haven't run before), I wouldn't want to wait that long every time I commit. I know it's optional since you have to install the hook manually, but I think a lot of people would get tired of it pretty quickly :p

MartinThoma:
Well, even if you have it installed you can always `git commit -n` / `git commit --no-verify`. This will commit without executing the hook. You can also adjust the hook (is there a way to run only PEP8 / fast tests?)

benanne:
Fair enough, I'm still not sure whether it's worth adding to the repository though. I think I'll defer to people who have more git experience about the practicality of this.

I think you can tell py.test to run only the pep8 checks but I haven't bothered to find out how yet :)

ebenolson:
For some of my other projects, I use [jig](https://pythonhosted.org/jig/index.html) to run pep8 and pyflakes pre-commit. 

As far as I can tell, it's not possible to run single tests unless you remove the `setup.cfg` file, would like to hear if anyone knows otherwise.

f0k:
> As far as I can tell, it's not possible to run single tests unless you remove the `setup.cfg` file, would like to hear if anyone knows otherwise.

That would be cool! I'm also tired of waiting for all the convolution tests to pass when I know I didn't touch any code related to this. [xdist](https://pypi.python.org/pypi/pytest-xdist)'s loop-on-fail mode helps a bit, but being able to specify a file on the command line would be nice. I'll remember the empty config file trick for the next time.

About the hook, I'm more comfortable running the tests on my own before committing, but I don't think it'd hurt to include this example. As it's just a few lines, I'd prefer to just include the lines directly in `development.rst` to avoid clobbering the project directory. (The current `ln` command seems to be broken anyway with the `../../`.)

MartinThoma:
The command `ln -s ../../pre-commit-hook.sh .git/hooks/pre-commit` is what I executed. It worked for me.

However, I just moved the file to `development.rst`.

benanne:
Looks good to me, I'll merge it tomorrow if there are no further comments.

f0k:
> The command `ln -s ../../pre-commit-hook.sh .git/hooks/pre-commit` is what I executed.

Ah, you're right. Relative links are pesky. A better understandable alternative would have been `ln -rs pre-commit-hook.sh .git/hooks/pre-commit`. But moving it into the documentation is nicer, thanks!

dnouri:
Regarding running a single test module only: that's possible using pytest's `-k` flag.  Say you want to run tests in `test_nonlinearities.py` only, then run `py.test -k test_nonlinearities`.  Or if you only want to run `test_leaky_rectify` then that's `py.test -k test_leaky_rectify`.

f0k:
@dnouri to the rescue! Thanks! :+1:

I always just saw the `usage: py.test [options] [file_or_dir] [file_or_dir] [...]` line and the `-k` filter didn't occur to me, but now that you mention it, it's obvious.

PR #214, opened 2015-04-19
add doc to DropoutLayer and GaussianNoiseLayer

skaae:
add docs to DropoutLayer and GaussianNoiseLayer

This is a cleaned up version of PR https://github.com/Lasagne/Lasagne/pull/212

### DIFF
--- a/lasagne/layers/noise.py
+++ b/lasagne/layers/noise.py
@@ -15,12 +15,58 @@ __all__ = [
 
 
 class DropoutLayer(Layer):
+    """Dropout layer [1]_,[2]_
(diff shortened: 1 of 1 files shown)


benanne:
Looks like all comments in the previous PR were addressed, so I'm fine with merging. Will wait till tomorrow in case there are further comments.

f0k:
I'm fine with it, too. Thanks, @skaee!

skaae:
np :) I'll do start on the init.py, but it will probably take a little longer because I have yet to read the references.

PR #217, opened 2015-04-21
Updates init

skaae:
Add's docstrings to initializers. 

TODO: Can someone check the Glorot and He descriptions?

### DIFF
--- a/lasagne/init.py
+++ b/lasagne/init.py
@@ -1,5 +1,11 @@
 """
-Functions to create initializers for parameter variables
+Functions to create initializers for parameter variables.
+
+Usage
+-------
+>>> from lasagne.layers import DenseLayer
+>>> from lasagne.init import Constant, Glorot
+>>> l1 = DenseLayer((100,20), num_units=50, W=GlorotUniform(), b=Constant(0.0))
 """
 
 import numpy as np
@@ -8,14 +14,48 @@ from .utils import floatX
 
 
 class Initializer(object):
+    """Initializer class
+
+    The :class:`Initializer` class represents a weight initializer used
+    to initialize weight parameters in a neural network layer. It should be
+    subclassed when implementing new types of weight initializers.
+
+    """
     def __call__(self, shape):
+        """
+        Makes :class:`Initializer` instances callable like a function, invoking
+        their :meth:`sample()` method.
+        """
         return self.sample(shape)
 
     def sample(self, shape):
+        """
+        Sample should return a theano.tensor of size shape and data type
+        theano.config.floatX.
+
+        Parameters
+        -----------
+        shape : tuple or int
+            Integer or tuple specifying the size of the returned
+            matrix.
+        returns : theano.tensor
+            Matrix of size shape and dtype theano.config.floatX.
+        """
         raise NotImplementedError()
 
 
 class Normal(Initializer):
+    """Sample initial weights from the Gaussian distribution
+
+    Initial weight parameters are sampled from N(mean, std).
+
+    Parameters
+    ----------
+    std : float
+        Std of initial parameters.
+    mean : float
+        Mean of initial parameters.
+    """
     def __init__(self, std=0.01, mean=0.0):
         self.std = std
         self.mean = mean
@@ -25,6 +65,22 @@ class Normal(Initializer):
 
 
 class Uniform(Initializer):
+    """Sample initial weights from the uniform distribution
+
+    Parameters are sampled from U(a, b).
+
+    Parameters
+    ----------
+    range : float or tuple
+        When std is None then range determines a, b. If range is a float the
+        weights are sampled from U(-range, range). If range is a tuple the
+        weights are sampled from U(range[0], range[1]).
+    std : float or None
+        If std is a float then the weights are sampled from
+        U(mean - np.sqrt(3) * std, mean + np.sqrt(3) * std).
+    mean : float
+        see std for description.
+    """
     def __init__(self, range=0.01, std=None, mean=0.0):
         import warnings
         warnings.warn("The uniform initializer no longer uses Glorot et al.'s "
@@ -50,7 +106,48 @@ class Uniform(Initializer):
 
 
 class Glorot(Initializer):
-    def __init__(self, initializer=Normal, gain=1.0, c01b=False):
+    """Glorot weight initialization [1]_
+
+    This is also known as Xavier initialization.
+
+    Parameters
+    ----------
+    initializer : lasagne.init.Initializer
+        Initializer used to sample the weights, must accept std in its
+        constructor to sample from a distribution with a given standard
+        deviation.
+    gain : float or 'relu'
+        When 'relu' the gain is set to sqrt(2), see notes.
+    c01b : bool
+        If lasagne.layers.cuda_convnet.Conv2DCCLayer is initialized with
+        dimshuffle=False, then c01b should be set to True.
+
+    References
+    ----------
+    [1] Glorot, Xavier, and Yoshua Bengio. "Understanding the difficulty of
+    training deep feedforward neural networks." International conference on
+    artificial intelligence and statistics. 2010.
+
+    Notes
+    ----------
+    For DenseLayers if gain='relu' and initializer is Uniform then the weights
+    are initialized as
+
+    :math:`a = \frac{sqrt{6}}{\sqrt{fan_{in}+n_{hid}}}`
+    :nath: `W \sim U[-a, a]`.
+
+    If gain=1 and initializer is Normal then the weights are initialized
+    as
+
+    :math: `std = \sqrt{\frac{2}{fan_{in}+n_{hid}}}`
+    :math: `W \sim N(0, std)`.
+
+    See Also
+    --------
+    GlorotNormal  : Shortcut with Gaussian initializer.
+    GlorotUniform : Shortcut with uniform initializer.
+    """
+    def __init__(self, initializer, gain=1.0, c01b=False):
         if gain == 'relu':
             gain = np.sqrt(2)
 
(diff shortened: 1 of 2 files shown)


lasagne/init.py
```diff
@@ -8,14 +14,48 @@
...
+
+    The :class:`Initializer` class represents a weight initializer used
+    to initialize weight parameters in a neural network layer. It should be
+    subclassed when implementing new types of weight initializers.
+
+    """
     def __call__(self, shape):
+        """
+           Subclasses of :class:`Layer` class use this method when they
+           initialize weight parameters.
```
f0k: Intended by three spaces too much, and `Layer` is wrong here :) I'd instead write:
```
Makes :class:`Initializer` instances callable like a function, invoking their :meth:`sample()` method.
```
(Is that the correct markup?)

lasagne/init.py
```diff
@@ -8,14 +14,48 @@
...
+        Sample should return a theano.tensor of size shape and data type
+        theano.config.floatX.
+
+        Parameters
+        -----------
+            shape : tuple or int
+                Integer or tuple specifying the size of the returned.
+                matrix.
+            returns : theano.tensor
+                 Matrix of size shape and datatype theano.config.floatX.
```
f0k: Indented by four spaces too much. `returned.` should be `returned` (without dot). `datatype` should probably be `dtype`.

lasagne/init.py
```diff
@@ -25,6 +65,22 @@ def sample(self, shape):
...
+    range : float or tuple
+        When std is None then range determines a, b. If range is a float the
+        weights are sampled from U(-range, range). If range is a tuple the
+        weights are sampled from U(range[0], range[1]).
+    std : float or None
+        If std is a float then the weights are sampled from
+        U(mean - np.sqrt(3) * std, mean + np.sqrt(3) * std).
+    mean : float
+        see std for description.
+    """
```
f0k: @benanne: Reading this, what about changing the functionality like this:
```
range : float or tuple
    If a float, weights are sampled from U(mean - range, mean + range).
    If a tuple, weights are sampled from U(range[0], range[1]).
std : float or None
    If a float, overrides `range` with `np.sqrt(3) * std`. This results in sampling
    from a uniform distribution with a standard deviation of `std`.
mean : float
    Mean of the sampled uniform distribution if `range` or `std` are given as a
    scalar, ignored otherwise.
```
The difference is that `mean` would be included if `range` is set to a scalar.

lasagne/init.py
```diff
@@ -50,6 +106,45 @@ def sample(self, shape):
...
 
 class Glorot(Initializer):
+    """Glorot weight initialization [1]_
+
+    This is also known as Xavier initialization.
+
+    Parameters
+    ----------
+    initializer : lasagne.init.Initializer
+        Initializer used to sample the weights. The default is Gaussian.
```
f0k: Suggestion:
Initializer used to sample the weights, must accept `std` in its constructor to sample from a distribution with a given standard deviation. Defaults to lasagne.init.Normal.

@benanne: Shouldn't the default be `Uniform` because that's what they do in the paper? Or shouldn't it have a default at all?

skaae: I agree on the Uniform default.

lasagne/init.py
```diff
@@ -75,16 +170,51 @@ def sample(self, shape):
...
+    """He weight initialization [1]_
+
+    Weights are initialized with std
+
+    :math:`\sigma = gain \sqrt{\frac{1}{fan_{in}}}`.
+
+    Parameters
+    ----------
+    initializer : Lasagne.init.Initializer
+        Initializer used to sample the weights. The default is Gaussian.
```
f0k: Same suggestion as above to make clear you can't use just *any* initializer here, but only particular ones (they could be given a common base class, but we don't want to over-engineer stuff).

lasagne/init.py
```diff
@@ -132,6 +279,16 @@ def sample(self, shape):
 
 
 class Sparse(Initializer):
+    """Initialize weights as sparse matrix.
+
+    Parameters
+    ----------
+    sparsity : float
+        Fraction of weights with non-zero value. Larger value gives less
+        sparsity.
```
f0k: The given fraction will be matched exactly for each column of the matrix.

skaae: Fraction of weights with non-zero value. Larger value gives less sparsity. The given fraction will be matched exactly for each column of the matrix.?

lasagne/init.py
benanne: Is this a common use case? I only added mean/std to support the modular Glorot / He initializers, So I'm not sure. People usually use distributions centered on 0 anyway. But if you think it's useful we can add it.

lasagne/init.py
benanne: Yeah, `Uniform` as default makes more sense. Although people should really be using `GlorotUniform` and not `Glorot` directly. We could also just get rid of the defaults (for both `Glorot` and `He`), I guess. I'm fine with whatever you guys decide :)

skaae: If no one complains i'll remove the defaults and make a comment in the doc that people should use the *Normal and *Uniform shortcuts.

benanne: Sounds good.

lasagne/init.py
skaae: I think sparsity is an ok name, I find nonzeros or fraction less descriptive.

lasagne/init.py
f0k: I think it's a good second use of the `mean` parameter now that we have it, it makes the documentation a bit easier, and it won't affect existing code. But I can add this after this PR.

lasagne/init.py
f0k: Yes, either this, or my second suggestion, whatever you find easier to understand. I just think it's important to note that it's not just the fraction of the full matrix, but even per column, and that it's not stochastic, but counted exactly.

f0k: > I think sparsity is an ok name, I find nonzeros or fraction less descriptive.

It's just that the meaning is inverted in that setting `sparsity` to a large value will give something not very sparse at all. That's a bit confusing.

lasagne/init.py
```diff
@@ -1,5 +1,11 @@
 """
-Functions to create initializers for parameter variables
+Functions to create initializers for parameter variables.
+
+Usage
+-------
+>>> from lasagne.layers import DenseLayer
+>>> from lasagne.init import Constant, Glorot
+>>> l1 = DenseLayer((100,20), num_units=50, W=Glorot(), b=Constant(0.0))
```
f0k: Now that you removed the default initializer, you cannot use `Glorot()` any more, but have to use one of the two shortcuts. Bonus points if you merge this with the previous commit :) (Just because it's a good opportunity to practice this. `git add`, `git commit --amend` and `git push --force` are your friends if you want to try it.)

skaae: I'll give it a try. I did manage to squash commits based on your previous suggestions.

benanne:
Looks like there are a few tests that try to instantiate `Glorot` and/or 'He' without arguments, so the fact that the default value has been removed is causing them to fail.

skaae:
ok. Should i modify the tests in this PR as well?

benanne:
Yeah,  you removed the default argument value in this PR, so you should update the tests to match :) Otherwise we can't merge this without breaking the build.

skaae:
* I fixed He and Glorot not having default initialiser
 * adds shape check to Glorot similarly to He initializer
 * add tests to get coverage to 100%

lasagne/tests/test_init.py
```diff
@@ -66,6 +85,10 @@ def test_glorot_normal_gain():
     assert -0.1 < sample.mean() < 0.1
     assert 0.9 < sample.std() < 1.1
 
+    sample = GlorotNormal(gain='relu').sample((100, 100))
+    assert -0.1 < sample.mean() < 0.1
+    assert 0.1 < sample.std() < 0.2
```
f0k: I think we should make that a lot stricter:
```python
assert -0.01 < sample.mean() < 0.01
assert 0.132 < sample.std() < 0.152
```

lasagne/tests/test_init.py
```diff
@@ -111,6 +134,10 @@ def test_glorot_uniform_gain():
     assert -1.0 <= sample.min() < -0.9
     assert 0.9 < sample.max() <= 1.0
 
+    sample = GlorotUniform(gain='relu').sample((100, 100))
+    assert -0.1 < sample.mean() < 0.1
+    assert 0.1 < sample.std() < 0.2
```
f0k: Same here (using the same target values).

lasagne/tests/test_init.py
```diff
@@ -156,6 +190,10 @@ def test_he_normal_gain():
     assert -0.1 < sample.mean() < 0.1
     assert 0.9 < sample.std() < 1.1
 
+    sample = HeNormal(gain='relu').sample((100, 100))
+    assert -0.1 < sample.mean() < 0.1
+    assert 0.1 < sample.std() < 0.2
```
f0k: Same here. But maybe we should use a shape that gives different results for Glorot and He, such as `(50,200)`. Adapt the tests if you're in the mood, otherwise I'll open a separate issue for that.

f0k:
> add tests to get coverage to 100%

Great, thanks a lot! :+1: Sorry for the extra requests regarding this, please feel free to leave that up for later or for somebody else.

skaae:
Its fine, I'll change the last test later today.

benanne:
Looks good to me! I'm okay with merging, will do so tomorrow if there are no further comments.

benanne:
Merging, thanks again!

Issue #203, opened 2015-04-12
Documentation: API reference to do list


benanne:
Here is a to do list for the API reference. Each of these could be a separate PR, or you can tackle multiple in a single PR. I guess it's a good idea to post here if you plan to work on one or more of these, so we don't end up doing any double work. Then I'll add your name on the list for an overview. Let's try to get this done within the next two weeks!

- [x] convert `lasagne.layers.helper` docs to numpydoc format (@MartinThoma)
- [x] convert `lasagne.layers.base` docs to numpydoc format (@MartinThoma)
- [x] convert `lasagne.layers.input` docs to numpydoc format (@MartinThoma)
- [x] complete `lasagne.layers.dense` docs and convert to numpydoc format (@skaae)
- [x] document `lasagne.layers.conv` (@benanne)
- [x] document `lasagne.layers.pool` (@ebenolson, @benanne)
- [x] document `lasagne.layers.noise` (@skaae)
- [x] complete `lasagne.layers.shape` docs and convert to numpydoc format (@skaae)
- [x] document `lasagne.layers.merge` (@skaae)
- [x] figure out how to make the docs for `corrmm`, `cuda_convnet` and `dnn` show up on readthedocs (these submodules cannot be imported without a gpu) (@benanne)
- [x] document `lasagne.layers.dnn` (@benanne)
- [x] document `lasagne.layers.corrmm` (@benanne)
- [x] document `lasagne.layers.cuda_convnet` (@benanne)
- [x] complete `lasagne.updates` docs and convert to numpydoc format (@skaae)
- [x] document `lasagne.init` (@skaae)
- [x] document `lasagne.nonlinearities` (@MartinThoma)
- [x] complete `lasagne.objectives` docs and convert to numpydoc format (@f0k)
- [x] complete `lasagne.regularization` docs and convert to numpydoc format (@skaae)
- [x] complete `lasagne.utils` docs and convert to numpydoc format (@JeffreyDF)
- [x] change the layout so it looks decent (@f0k)
- [x] get rid of string representations of objects in docstrings everywhere (e.g. `W=<lasagne.init.GlorotUniform object at 0x7f699026a690>`) (@benanne)
- [x] split `lasagne.layers` documentation into separate pages for each submodule (@f0k)

MartinThoma:
I can do the easy stuff (to numpydoc conversions of `lasagne.layers.helper`, `lasagne.layers.base`, `lasagne.layers.input`) today evening

[...]

PR #220, opened 2015-04-23
docs/user/about: Added comparision with Caffe

MartinThoma:
I've added a comparison about Caffe / Lasagne to the documentation as well as a little bit more about the "history" of Lasagne (who develops it and when did development start).
I think this information is interesting, because people can take a look at your websites and see what you're working on and get some confidence that Lasagne might be worth a try (I hope I didn't forget any core contributor).

The comparison of Caffe / Lasagne can certainly be better, but I think it's a good start. Other frameworks for which I we could eventually add comparisons are

* Torch: http://torch.ch/
* DL4J: http://deeplearning4j.org/

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -17,6 +17,7 @@ new Lasagne developers what to install and where to start.
 .. toctree::
   :maxdepth: 2
 
+  user/about
   user/installation
   user/tutorial
   user/development
--- a/docs/user/about.rst
+++ b/docs/user/about.rst
@@ -1,0 +1,36 @@
+About Lasagne
+=============
+
+Lasagne is a Python package for neural networks based on `Theano`_. It can
+execute the training on Nvidia GPUs without having to adjust the code. Its
+development was started by `Sander Dieleman`_ in September 2014. It is
+developed by many people via GitHub Pull requests, e.g. by `Daniel Nouri`_,
+`Colin Raffel`_, `Jan Schlüter`_ and some more.
+
+Comparison with Caffe
+---------------------
+
+`Caffe`_ is developed by the Berkeley Vision and Learning Center and released
+under the BSD 2-Clause license. Just like Lasagne, it can execute the neural
+network training on Nvidia GPUs. The development of Caffe started in September
+2013.
+
+Caffe is designed to be easy to train existing models and limited in terms of
+experimentation. Lasagne provides more flexibility at the cost of a being
+more difficult to use.
+
+Neural Networks are completely defined as Python code with Lasagne, whereas
+Caffe makes use of `JSON-like text files`_ to describe the topology.
+
+Lasagnes code base has about 3226 lines of Python code whereas Caffe has more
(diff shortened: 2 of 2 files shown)


f0k:
Sorry, except for the first sentence I'd have written everything differently and/or not included it in the user documentation at all. I'd list the contributors as part of the release notes, but everyone, not just the core team (Theano did that quite well for 0.6), and I wouldn't put a superficial comparison to Caffe as one of the first things a new user will see! In general, while I appreciate your input, I think we should focus on documenting the code before starting any meta-documentation -- please pick your favorite at #203 if you're in the mood for writing :) A nice "about" text or FAQ can wait until after the first release.

benanne:
I think there is room for something like this in the docs, but maybe not on the front page right now. It's a bit unfortunate that this would be the first thing the user sees. We don't really want to market ourselves as an alternative to Caffe, nor is Lasagne a 'reaction to it', so it seems out of place to put this much emphasis on such a comparison. 

But of course that's just down to adding more content to the front page first (a more neutral description of Lasagne and maybe a brief 'Getting started' section would be good for that). Eventually I think having a comparison to some other libraries in this space in our docs, highlighting which use cases each library is better suited for, would definitely be valuable.

f0k:
> a more neutral description of Lasagne and maybe a brief 'Getting started' section would be good for that

I like the structure of the [cudamat front page](https://github.com/cudamat/cudamat) (but I might be biased). I know we're talking about the front page for the documentation here, but that could be largely the same: A description of what it's about, a short code example illustrating the key aspect, and links to further information (installation, user guide, API reference).

> Eventually I think having a comparison to some other libraries in this space in our docs, highlighting which use cases each library is better suited for, would definitely be valuable.

I'd collect such meta-documentation in a FAQ at the *end* of the docs or in the wiki, along with an explanation of the name (that will be fun to write) and these sorts of things. "How is Lasagne different from pylearn2?", "How is Lasagne different from caffe?", ... But again, this can wait.

benanne:
> I like the structure of the cudamat front page (but I might be biased). 

It looks great, and gives you a rough idea of what the library is about at a glance. We should aim for the same effect.

> I'd collect such meta-documentation in a FAQ at the end of the docs or in the wiki, along with an explanation of the name (that will be fun to write) and these sorts of things. "How is Lasagne different from pylearn2?", "How is Lasagne different from caffe?", ... But again, this can wait.

Agreed, sounds good.

MartinThoma:
> I think we should focus on documenting the code before starting any meta-documentation -- please pick your favorite at #203 if you're in the mood for writing :)

I have pull request going for one of those, but there was an issue with possible rebasing due to #182. I'll rather wait until this is done before I start writing documentation. I am currently preparing a much simpler example using XOR. I could probably calculate the values for those by hand so that we can add more (meaningful) tests for very simple networks.

Ok. So I close this pull request and try the FAQ approach when I have time again / when the other stuff is done.

Issue #164, opened 2015-03-11
API: get_params()


f0k:
As observed in #141 (but also in #110 and https://github.com/benanne/Lasagne/issues/136#issuecomment-75169847), it's not entirely clear what `Layer.get_params()` means and what it should mean.

The documentation of `get_params()` says: "Returns a list of all the Theano variables that parameterize the layer.", and further below it says "it should be overridden in a subclass that has trainable parameters".
In addition, we have `get_bias_params()`, which is documented as: "Returns a list of all the Theano variables that are bias parameters for the layer.", and a note further below says "This is useful when specifying regularization (it is often undesirable to regularize bias parameters)."

Now when we include the use case of Batch Normalization, we actually have three overlapping classes of variables:

1. variables involved in the forward pass (currently `get_params()` according to its docstring)
2. variables to be updated wrt. the loss function (currently `get_params()` according to its usage in Lasagne)
3. variables to be updated wrt. the loss function that are not to be included in weight regularization (currently `get_bias_params()`)

Before the first release of Lasagne we should re-think the `get_params` and `get_bias_params` API. It obviously does not exactly match what comes up in the use cases. Should we possibly replace this with a single `get_params()` method with optional keyword arguments to filter out certain things? Most Layer classes would not need to bother with the keywords or just react on the bias-related keyword, so it wouldn't increase the complexity of adding a new Layer class. Shall we think that through?

benanne:
I agree, there is a distinction between the "parameters that should be learned" and the "parameters that should be serialized" that pops up sometimes. For most layers these are the same, but I guess this isn't always the case.

I like the idea of having a single method and controlling its behaviour with optional keyword arguments, which allows us to ignore this issue in 95% of use cases. It also allows easy extension later by defining new keyword arguments, in case we discover there are other kinds of parameter groupings as well.

The alternative seems to be to add an extra method to distinguish the two current `get_params()` use cases, but that makes the interface bigger so I don't like the idea much. It could also set a bad precedent.

So let's think this through indeed!

craffel:
[...]

PR #221, opened 2015-04-23
documentation for convolutional layers + user guide for layers

benanne:
I've only done `lasagne.layers.Conv1DLayer` so far (as well as the `conv_output_length` helper function), but feedback is welcome. I'll do `Conv2DLayer` next and then try to do the alternative convolution layers as well.


lasagne/layers/conv.py
```diff
@@ -9,6 +9,7 @@
 
 
 __all__ = [
+    "conv_output_length",
```
f0k: +1 for documenting it, but do we really want to export that as a first-class `lasagne.layers` citizen? I would have considered this to be an internal helper function to share code between convolution implementations.

lasagne/layers/conv.py
```diff
@@ -16,6 +17,54 @@
...
+
+        If 'valid', it is assumed that the convolution is only computed where
+        the input and the filter fully overlap.
+
+        If 'full', it is assumed that the convolution is computed wherever the
+        input and the filter overlap by at least one position.
+
+        If 'same', it is assumed that the convolution is computed wherever the
+        input and the filter overlap by half, which results in an output length
+        that is the same as the input length.
```
f0k: *overlap at least by half the filter size

lasagne/layers/conv.py
```diff
@@ -37,6 +86,97 @@ def conv_output_length(input_length, filter_size,
...
+
+        If 'valid', the convolution is only computed where the input and the
+        filter fully overlap.
+
+        If 'full', the convolution is computed wherever the input and the
+        filter overlap by at least one position.
+
+        If 'same', the convolution is computed wherever the input and the
+        filter overlap by half, which results in an output length that is the
+        same as the input length.
```
f0k: *overlap by at least half the filter size

lasagne/layers/conv.py
```diff
@@ -37,6 +86,97 @@ def conv_output_length(input_length, filter_size,
...
+        This should initialize the layer weights to a 3D array with shape
+        `(num_filters, num_input_channels, filter_length)`.
+
+    b : Theano shared variable, numpy array, callable or None
+        An initializer for the biases of the layer. If a Theano shared
+        variable is provided, it is used unchanged. If a numpy array is
+        provided, a shared variable is created and initialized with the
+        array. If a callable is provided, a shared variable is created and
+        the callable is called with the desired shape to generate suitable
+        initial values. The variable is then initialized with those values.
```
f0k: Very good, but I wonder whether we really want to copy that text for each and every parameter of each and every layer. `b` could point to `W`, or we could have that description in some central place (I don't know where, though).

lasagne/layers/conv.py
```diff
@@ -37,6 +86,97 @@ def conv_output_length(input_length, filter_size,
...
+
+    **kwargs
+        Any additional keyword arguments are passed to the `Layer` superclass.
+
+    Notes
+    -----
+    Theano's default convolution function (`theano.tensor.nnet.conv.conv2d`)
+    does not support the 'same' border mode by default. This layer emulates
+    it by performing a 'full' convolution and then slicing the result, which
+    may negatively affect performance.
```
f0k: To me, `cropping` would sound more to the point than `slicing` (I associate the latter with slicing something into multiple pieces).

lasagne/layers/conv.py
ebenolson: this seems problematic to me, what about even-sized filters?

how about something like this:

If 'same', it is assumed that the convolution is computed with the filter centered at each input position, producing an output of the same length. This is only possible for filters with an odd length.

f0k: The implementation does the correct thing even for even filter sizes. I'd call the mode you're describing `half`. But I was asking myself the same thing while reading the docstring -- does that actually work for even filter sizes as well? So maybe the documentation should be clear about this, possibly just by adding `... the same as the input length (both for odd and even filter sizes)`.

ebenolson: ah yes, this is coming back to me. I'd argue that it doesn't actually do the correct thing at the moment (it trims a `full` convolution, which gives a result that's not symmetric for even filters).

But either way the current wording implies that with a filter and an input of length 2, you'd get an output of length 3.

f0k: > which gives a result that's not symmetric for even filters

There's no alternative to this, is there?

> But either way the current wording implies that with a filter and an input of length 2, you'd get an output of length 3.

Ah yes. "At least half the filter size" is wrong for even filter sizes. So it would still need an addendum that says we're omitting the last position for even filter sizes, and similar descriptions for the actual Conv*DLayer docstrings.

ebenolson: > There's no alternative to this, is there?

I feel like the most correct thing would be to not support this mode for even filter_size, but the current behavior matches `numpy.convolve` so it's probably the right choice.

lasagne/layers/conv.py
benanne: maybe not. I added it there to make sure it shows up in the docs, but we can probably work around it if we don't want it in `lasagne.layers`. I'll remove it and see what happens.

lasagne/layers/conv.py
benanne: Maybe at the top level of the `layers` module we should explain some things about layers in general, including how learnable parameters are handled. Then we could explain all the different initialization mechanisms there and refer to it.

Some other things worth explaining are:
* the way keyword arguments are propagated by a `get_output()` call (e.g. `deterministic`)
* the way keyword arguments are passed to the base class by the constructors (e.g. to enable naming layers)
* what else?

f0k: Sounds good, as long as we link to it! (In a way that's possible from within a python session as well. When I need the docs for `Conv2DLayer`, I'd usually fire up an ipython session, import the `Conv2DLayer` and enter `Conv2DLayer?`. I can also do `lasagne.layers?` if I'm told to look there. I only use the web documentation when I'm looking for something I don't know the name of.)

benanne:
I've tried to address all comments so far. I've started writing an intro to the `layers` module as part of this PR, and it kind of got out of hand :) But I think it will be useful.

I have not been able to figure out how to refer to it from docstrings (to explain the different initialization options for example). Any ideas? Sphinx / restructuredtext / numpydoc is pretty opaque and unintuitive, it's really quite frustrating to get into it imo :(

MartinThoma:
> I have not been able to figure out how to refer to it from docstrings (to explain the different initialization options for example). Any ideas? Sphinx / restructuredtext / numpydoc is pretty opaque and unintuitive, it's really quite frustrating to get into it imo :(

What do you want to refer to from docstrings?

benanne:
In the section describing layer parameters (`W` and `b`), I would like to refer to the description of the various parameter initialization options at he top level of the `layers` module. I can't figure out how to make an 'anchor' at that position and then refer to it.

benanne:
Doctests are doing my head in! How can I split a line longer than 80 chars in a doctest properly, so the test passes and it doesn't look like crap? Tempted to just sprinkle `# doctest: +SKIP` everywhere...

MartinThoma:
> Doctests are doing my head in! How can I split a line longer than 80 chars in a doctest properly, so the test passes and it doesn't look like crap? Tempted to just sprinkle # doctest: +SKIP everywhere...

> EDIT: I guess I need ... instead of >>>.

Yes, you need `...` to continue lines (see https://docs.python.org/3/library/doctest.html#how-are-docstring-examples-recognized).

I'll try to find a solution for the anchor problem.

f0k:
> In the section describing layer parameters (`W` and `b`), I would like to refer to the description of the various parameter initialization options at he top level of the `layers` module. I can't figure out how to make an 'anchor' at that position and then refer to it.

How about describing the initialization options in `create_params` and point to that instead?
/edit: I see they are already described in `create_params` anyway, but your version may have been nicer. `W` and `b` could tell the user to check the documentation for `param` in `create_params`.

> I've started writing an intro to the `layers` module as part of this PR, and it kind of got out of hand :) But I think it will be useful.

Definitely! It will be useful, and it got out of hand :) I think it's a good text and should just be moved to the user guide instead. The module docstring probably shouldn't take up more than two or three terminal windows. Giving an example on instantiating a network in the docstring is a good idea, and maybe something pointing the user towards the three de-facto base classes `InputLayer`, `Layer` and `MergeLayer`, as well as a pointer towards `get_output()`.

benanne:
I guess it might be better to set up a separate PR for the user guide then? I think it makes more sense to refer to that than to the `create_param` docstring - you should only really have to worry about `create_param` if you want to implement your own custom layer.

f0k:
> I guess it might be better to set up a separate PR for the user guide then?

Probably yes, but I'd also be fine with including it here if that's easier.

> you should only really have to worry about `create_param` if you want to implement your own custom layer.

Hmm, that's a valid point. Still I think having the description of what can be passed for `W` and `b` readily available in the terminal would be important, duplicating it everywhere would be ugly, and having it in the module docstring would make it hard to find. As the three options for `W` and `b` (Theano variable, ndarray, callable) are already explained in `create_param` (because that's were they come from), the easiest would be to refer to that. Referring to the user guide makes sense for Sphinx, but is impossible for the terminal.

benanne:
> /edit: With #164, `Layer.create_param` will be replaced by `Layer.add_param`. A good option might be moving the "take a variable, ndarray or callable and return a variable" functionality out into an external helper function in `lasagne.utils` or `lasagne.init` and then refer to that. This can also happen before #164.

That sounds like a great idea. then `add_param` would just use that and in addition 'register' the parameter and handle the tagging.

f0k:
> `add_param` would just use that and in addition 'register' the parameter and handle the tagging.

Exactly, and name the parameter if needed (I'd do that in `add_param` after calling the helper function, à la: `if param.name is None: param.name = ...`).

benanne:
I moved everything I wrote about the layer classes to the user guide. It should be fleshed out a little (especially the part about creating your own layer classes) but I think it will do for now, there are more pressing matters :)

I think this is more or less ready, let me know what else needs to change (or if there are any proposed changes that I missed). I'll remove the WIP tag.

lasagne/layers/conv.py
```diff
@@ -37,6 +89,97 @@ def conv_output_length(input_length, filter_size,
...
+        filter fully overlap.
+
+        If 'full', the convolution is computed wherever the input and the
+        filter overlap by at least one position.
+
+        If 'same', the convolution is computed wherever the input and the
+        filter overlap by at least half the filter size, when the filter size
+        is odd. In practice, the input is zero-padded with half the filter size
+        at the beginning and half at the end (or one less than half in the case
+        of an even filter size). Thisresults in an output length that is the
```
f0k: *This results

docs/modules/layers.rst
```diff
@@ -49,6 +49,8 @@ Layer classes: convolutional layers
 .. autoclass:: Conv2DLayer
     :members:
 
+.. autofunction:: lasagne.layers.conv.conv_output_length
+
```
f0k: I guess I'd keep that one private, so users browsing the reference don't get confused. When implementing their own convolutional layer, they will probably use an existing implementation as the basis and see the helper function anyway. But I don't have a strong opinion on that, do as you please!

docs/user/layers.rst
```diff
@@ -0,0 +1,179 @@
...
+>>> l_in = lasagne.layers.InputLayer((100, 50))
+>>> l_hidden = lasagne.layers.DenseLayer(l_in, num_units=200)
+>>> l_out = lasagne.layers.DenseLayer(l_hidden, num_units=10,
+...                                   nonlinearity=T.nnet.softmax)
+
+The first layer of the network is an `InputLayer`, which represents the input.
+When creating an input layer, you should specify the shape of the input data.
+In this example, the input is a matrix with shape (100, 50), representing a
+batch of 100 data points, where each data point is a vector of length 50.
+the first dimension of a tensor is usually the batch dimension, following the
```
f0k: *The first dimension

lasagne/layers/conv.py
```diff
@@ -37,6 +89,97 @@ def conv_output_length(input_length, filter_size,
...
+        layer weights to a 3D array with shape
+        ``(num_filters, num_input_channels, filter_length)``.
+        See :meth:`Layer.create_param` for more information.
+
+    b : Theano shared variable, numpy array, callable or None
+        An initializer for the biases of the layer. If None is provided, the
+        layer will have no biases. This should initialize the layer biases to
+        a 1D array with shape ``(num_filters,)`` if `untied_biases` is set to
+        ``False``. If it is set to ``True``, its shape should be
+        ``(num_filers, input_length)`` instead.
```
f0k: *num_filters

lasagne/layers/conv.py
```diff
@@ -122,6 +272,94 @@ def get_output_for(self, input, input_shape=None, **kwargs):
...
+        layer weights to a 4D array with shape
+        ``(num_filters, num_input_channels, filter_height, filter_width)``.
+        See :meth:`Layer.create_param` for more information.
+
+    b : Theano shared variable, numpy array, callable or None
+        An initializer for the biases of the layer. If None is provided, the
+        layer will have no biases. This should initialize the layer biases to
+        a 1D array with shape ``(num_filters,)`` if `untied_biases` is set to
+        ``False``. If it is set to ``True``, its shape should be
+        ``(num_filers, input_height, input_width)`` instead.
```
f0k: *num_filters

f0k:
Just a few minor comments, please feel free to merge yourself after fixing them!

benanne:
Thanks! Merging :)

Issue #208, opened 2015-04-15
Missing stride argument in lasagne/layers/pool/FeaturePoolLayer


neilsummers:
Commit a0961b1cf451255ea3372787d6ee4c78cee700a8 modified lasagne/layers/pool.py
method ```pool_output_length``` to include a stride argument, but ```FeaturePoolLayer``` makes a call to this method without this argument, raising an exception.

More details:

```class FeaturePoolLayer(Layer)```
defines the method
```
def get_output_shape_for(self, input_shape):
        output_shape = list(input_shape)  # make a mutable copy
        output_shape[self.axis] = pool_output_length(input_shape[self.axis],
                                                     self.ds)
        return tuple(output_shape)
```

which makes a call to pool_output_length, but is missing the stride argument
```def pool_output_length(input_length, ds, st, ignore_border=True, pad=0):```

Should stride be set to ds if None like with ```MaxPool2DLayer```?

ebenolson:
Oops... I'm not sure there's a reason to support stride in `FeaturePoolLayer`, so just hardcoding it to use `st=ds` in that call is probably enough.

Feel free to make a PR, otherwise I'll do it later today. 

Even better would be a test case for `FeaturePoolLayer`

f0k:
> Should stride be set to ds if None like with `MaxPool2DLayer`?

Yes, it would be good to have `st=None` as a default in `pool_output_length()`, and there should be a test case for `FeaturePoolLayer.get_output_shape_for()`. A PR would be welcome!

neilsummers:
I can hopefully do a PR later today.

neilsummers:
From your comments above, I guess there are two different ways of doing this.
Would you rather that
a) ```pool_output_length()``` had the default argument of ```st=None```, and then set to ```st=ds``` in ```pool_output_length()```,
or
b) call ```pool_output_length()``` is called from ```FeaturePoolLayer.get_output_shape_for()``` with the argument ```st=None```?

f0k:
The second solution probably should have read "with the argument `st=ds`". In any case, I prefer the first solution: Have a default value `st=None`, and a `if st is None: st = ds` in `pool_output_length()`, but keep `FeaturePoolLayer.get_output_shape_for()` unchanged.

neilsummers:
Yes it should.
Will take me a bit longer to put a PR together with a test for ```FeaturePoolLayer```. If you can do it quickly then feel free.

ebenolson:
[...]

PR #222, opened 2015-04-23
Featurepool

ebenolson:
This is to close https://github.com/Lasagne/Lasagne/issues/208

I decided to simply remove the call to pool_output_length instead of changing its defaults because
- I'd prefer that function to require explicit parameters (and I removed the defaults that were there), since that avoids duplicating code for default handling and may help expose any bugs in layers calling it.

- Since `FeaturePoolLayer` doesn't use any of the more complicated parameters (and it doesn't seem like it will in the future) there's no real need for the helper function.

- The simple division is a bit easier to understand if you're just reading the code and aren't familiar with the helper function.

I also added a test for this layer, and rewrote some of the docstring and `get_output_for` to be a bit cleaner (in my opinion).

### DIFF
--- a/lasagne/layers/pool.py
+++ b/lasagne/layers/pool.py
@@ -256,15 +256,13 @@ class FeaturePoolLayer(Layer):
         return tuple(output_shape)
 
     def get_output_for(self, input, **kwargs):
-        num_feature_maps = input.shape[self.axis]
+        input_shape = tuple(input.shape)
+        num_feature_maps = input_shape[self.axis]
         num_feature_maps_out = num_feature_maps // self.pool_size
 
-        pool_shape = ()
-        for axis, length in enumerate(input.shape):
-            if axis == self.axis:
-                pool_shape += (num_feature_maps_out, self.pool_size)
-            else:
-                pool_shape += (length,)
+        pool_shape = (input_shape[:self.axis] +
+                      (num_feature_maps_out, self.pool_size) +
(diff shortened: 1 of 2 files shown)


lasagne/layers/pool.py
```diff
@@ -260,11 +260,11 @@ def get_output_for(self, input, **kwargs):
...
-        for k in range(self.axis):
-            pool_shape += (input.shape[k],)
-        pool_shape += (num_feature_maps_out, self.pool_size)
-        for k in range(self.axis + 1, input.ndim):
-            pool_shape += (input.shape[k],)
+        for axis, length in enumerate(input.shape):
+            if axis == self.axis:
+                pool_shape += (num_feature_maps_out, self.pool_size)
+            else:
+                pool_shape += (length,)
```
f0k: Looks cleaner. But why isn't this just `pool_shape = input.shape[:self.axis] + (num_feature_maps_out, self.pool_size) + input.shape[self.axis+1:]`?

f0k:
Very good, thanks! I agree with all your points.

lasagne/layers/pool.py
ebenolson: good point, I'll add that change.

benanne:
As do I. Nice work, once again :)

benanne:
Looks ready for merging, I will wait until tomorrow just in case :)

PR #223, opened 2015-04-26
Mock theano.config so GPU-dependent modules show up on readhtedocs.org

benanne:
I have no way to test this, I think... unless readthedocs.org can build docs off a branch as a test, I'll look into that (but I doubt it). Suggestions to maximize the chance of this working are very welcome :)

PR #224, opened 2015-04-29
Make lasagne.layers.corrmm / dnn / cuda_convnet docs show up on readthedocs.org, attempt 2

benanne:
This is a second attempt to make the docs for the alternative convolutional layer implementations show up on readthedocs.org.

~~I 'fixed' the issue by moving the imports of `GpuCorrMM` (corrmm) and `FilterActs` (cuda_convnet) into the methods were they are used. This allows readthedocs.org to import the modules and grab the docstrings.~~

~~The downside of this fix is that `lasagne.layers.cuda_convnet` will now no longer fail at import time if pylearn2 is not available. That's not good. I don't know how else to fix it though, and having the docs show up seems a bit more important. If someone can come up with a good work-around, let me know.~~

(See #223 for the previous attempt - it didn't work but I think it was a necessary step.)

EDIT: I changed the approach to avoid having to import these modules at runtime, based on @dnouri's suggestion in #223. This avoids the problem with the other approach, hopefully it works correctly on readthedocs.org as well.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -279,3 +279,10 @@ import theano.sandbox.cuda
 
 theano.config = Mock(device='gpu')
 theano.sandbox.cuda.dnn = Mock(dnn_available=lambda: True)
+
+import sys
+
+sys.modules['pylearn2.sandbox.cuda_convnet.filter_acts'] = \
+    Mock(FilterActs=None)
+
+sys.modules['theano.sandbox.cuda.blas'] = Mock(GpuCorrMM=None)


ebenolson:
Is it not possible to mock those imports on readthedocs?

benanne:
I guess it should be. I modified this PR to do that instead.

f0k:
Looks promising. Try it!

benanne:
Here goes nothing!

benanne:
... well, this is odd. It seems to have worked for `GpuCorrMM`, but not for the cuda-convnet based layers:

```
/var/build/user_builds/lasagne/checkouts/latest/docs/modules/layers.rst:118: WARNING: autodoc: failed to import module u'lasagne.layers.cuda_convnet'; the following exception was raised:
  File "/home/docs/checkouts/readthedocs.org/user_builds/lasagne/envs/latest/local/lib/python2.7/site-packages/sphinx/ext/autodoc.py", line 335, in import_object
    __import__(self.modname)
  File "/home/docs/checkouts/readthedocs.org/user_builds/lasagne/envs/latest/local/lib/python2.7/site-packages/Lasagne-0.1dev-py2.7.egg/lasagne/layers/cuda_convnet.py", line 14, in <module>
    from pylearn2.sandbox.cuda_convnet.filter_acts import FilterActs
ImportError: No module named pylearn2.sandbox.cuda_convnet.filter_acts
```

Still this error. Which doesn't really make sense because we just mocked exactly that module. Anybody have any idea what's going on here?

ebenolson:
I think that import will search for the intermediate modules `pylearn2`, `pylearn2.sandbox`, etc., so you need to mock those as well. Not sure why it would work for `GpuCorrMM` though.

dnouri:
Seems like Eben is right.  Note that pylearn2 isn't even installed, so it's not a matter of mocking a single module out of it, but mocking the whole package and all subpackages apparently.
```python
sys.modules['pylearn2'] = Mock()
sys.modules['pylearn2.sandbox'] = Mock()
sys.modules['pylearn2.sandbox.cuda_convnet'] = Mock()
sys.modules['pylearn2.sandbox.cuda_convnet.filter_acts'] = Mock()
from pylearn2.sandbox.cuda_convnet.filter_acts import FilterActs  # works
```
Possibly GPUCorrMM works because theano and all subpackages in that case are available.

benanne:
That sounds very plausible. I'll give that a try.

PR #225, opened 2015-04-29
added NonlinearityOnlyLayer for use with MultipleInputLayer

mheilman:
This adds a layer class that just performs a nonlinearity.
This is for use in conjunction with MultipleInputsLayer subclasses, which just add or concatenate outputs from layers.  For example, one might want to take inputs of length n and m, perform linear transformations on them using DenseLayer objects, then add the transformed outputs with an ElemwiseSumLayer, and then use this class to apply a nonlinearity (note that this would be different from applying the nonlinearity and then adding).  Alternatively, one could use a ConcatLayer followed by a DenseLayer, but concatenation is less efficient than adding.

### DIFF
--- a/lasagne/layers/dense.py
+++ b/lasagne/layers/dense.py
@@ -10,7 +10,7 @@ from .base import Layer
 __all__ = [
     "DenseLayer",
     "NINLayer",
-    "NonlinearityOnlyLayer"
+    "NonlinearityLayer"
 ]
 
 
@@ -19,7 +19,7 @@ class DenseLayer(Layer):
     A fully connected layer.
 
     :parameters:
-        - input_layer : `Layer` instance
+        - incoming : `Layer` instance
             The layer from which this layer will obtain its input
 
         - num_units : int
@@ -88,19 +88,11 @@ class DenseLayer(Layer):
         return self.nonlinearity(activation)
 
 
-class NonlinearityOnlyLayer(Layer):
+class NonlinearityLayer(Layer):
     """
-    A layer that just performs a nonlinearity.
-    This is for use in conjunction with MultipleInputsLayer subclasses,
-    which just add or concatenate outputs from layers.  For example, one
-    might want to take inputs of length n and m, perform linear transformations
-    on them using DenseLayer objects, then add the transformed outputs with an
-    ElemwiseSumLayer, and then use this class to apply a nonlinearity
-    (note that this would be different from applying the nonlinearity and
-    then adding).  Alternatively, one could use a ConcatLayer followed by
-    a DenseLayer, but concatenation is less efficient than adding.
+    A layer that just applies a nonlinearity.
 
-    - input_layer : `Layer` instance
+    - incoming : `Layer` instance
         The layer from which this layer will obtain its input
 
     - nonlinearity : callable or None
@@ -109,7 +101,7 @@ class NonlinearityOnlyLayer(Layer):
     """
     def __init__(self, incoming, nonlinearity=nonlinearities.rectify,
                  **kwargs):
-        super(NonlinearityOnlyLayer, self).__init__(incoming, **kwargs)
+        super(NonlinearityLayer, self).__init__(incoming, **kwargs)
         self.nonlinearity = (nonlinearities.identity if nonlinearity is None
                              else nonlinearity)
 
--- a/lasagne/layers/dense.py
+++ b/lasagne/layers/dense.py
@@ -19,8 +19,8 @@ class DenseLayer(Layer):
     A fully connected layer.
 
     :parameters:
-        - incoming : `Layer` instance
-            The layer from which this layer will obtain its input
+        - incoming : a :class:`Layer` instance or a tuple
+            the layer feeding into this layer, or the expected input shape
 
         - num_units : int
             The number of units of the layer
@@ -92,8 +92,8 @@ class NonlinearityLayer(Layer):
     """
     A layer that just applies a nonlinearity.
 
-    - incoming : `Layer` instance
-        The layer from which this layer will obtain its input
+    - incoming : a :class:`Layer` instance or a tuple
+        the layer feeding into this layer, or the expected input shape
 
     - nonlinearity : callable or None
         The nonlinearity that is applied to the layer activations. If None
--- a/lasagne/layers/dense.py
+++ b/lasagne/layers/dense.py
@@ -10,6 +10,7 @@ from .base import Layer
 __all__ = [
     "DenseLayer",
     "NINLayer",
+    "NonlinearityOnlyLayer"
 ]
 
 
@@ -55,10 +56,8 @@ class DenseLayer(Layer):
                  b=init.Constant(0.), nonlinearity=nonlinearities.rectify,
                  **kwargs):
         super(DenseLayer, self).__init__(incoming, **kwargs)
-        if nonlinearity is None:
-            self.nonlinearity = nonlinearities.identity
-        else:
-            self.nonlinearity = nonlinearity
+        self.nonlinearity = (nonlinearities.identity if nonlinearity is None
+                             else nonlinearity)
 
         self.num_units = num_units
 
@@ -89,6 +88,35 @@ class DenseLayer(Layer):
         return self.nonlinearity(activation)
 
 
+class NonlinearityOnlyLayer(Layer):
+    """
+    A layer that just performs a nonlinearity.
+    This is for use in conjunction with MultipleInputsLayer subclasses,
+    which just add or concatenate outputs from layers.  For example, one
+    might want to take inputs of length n and m, perform linear transformations
+    on them using DenseLayer objects, then add the transformed outputs with an
+    ElemwiseSumLayer, and then use this class to apply a nonlinearity
+    (note that this would be different from applying the nonlinearity and
+    then adding).  Alternatively, one could use a ConcatLayer followed by
+    a DenseLayer, but concatenation is less efficient than adding.
+
+    - input_layer : `Layer` instance
+        The layer from which this layer will obtain its input
+
+    - nonlinearity : callable or None
+        The nonlinearity that is applied to the layer activations. If None
+        is provided, the layer will be linear.
+    """
+    def __init__(self, incoming, nonlinearity=nonlinearities.rectify,
+                 **kwargs):
+        super(NonlinearityOnlyLayer, self).__init__(incoming, **kwargs)
+        self.nonlinearity = (nonlinearities.identity if nonlinearity is None
+                             else nonlinearity)
+
+    def get_output_for(self, input, **kwargs):
+        return self.nonlinearity(input)
+
+
 class NINLayer(Layer):
     """
     Network-in-network layer.
@@ -101,10 +129,8 @@ class NINLayer(Layer):
                  W=init.GlorotUniform(), b=init.Constant(0.),
                  nonlinearity=nonlinearities.rectify, **kwargs):
         super(NINLayer, self).__init__(incoming, **kwargs)
-        if nonlinearity is None:
-            self.nonlinearity = nonlinearities.identity
-        else:
-            self.nonlinearity = nonlinearity
+        self.nonlinearity = (nonlinearities.identity if nonlinearity is None
+                             else nonlinearity)
 
         self.num_units = num_units
         self.untie_biases = untie_biases


f0k:
I had one for exactly the same reason :) We always wanted to add that, the major point holding us back was that nobody came up with a good name. I think `NonlinearityOnlyLayer` isn't too bad. What do the others think? (Yay, a name discussion!)

lasagne/layers/dense.py
```diff
@@ -89,6 +88,35 @@ def get_output_for(self, input, **kwargs):
...
+    """
+    A layer that just performs a nonlinearity.
+    This is for use in conjunction with MultipleInputsLayer subclasses,
+    which just add or concatenate outputs from layers.  For example, one
+    might want to take inputs of length n and m, perform linear transformations
+    on them using DenseLayer objects, then add the transformed outputs with an
+    ElemwiseSumLayer, and then use this class to apply a nonlinearity
+    (note that this would be different from applying the nonlinearity and
+    then adding).  Alternatively, one could use a ConcatLayer followed by
+    a DenseLayer, but concatenation is less efficient than adding.
```
f0k: While the explanation is perfect to motivate this PR, I think it's a bit out of scope for the class docstring. I would just write `A layer that just applies a nonlinearity.`, and not give any use case, because there are plenty. The explanation of how to replace a ConcatLayer -> DenseLayer with DenseLayer -> ElementwiseSumLayer would be good for the (not yet existing) tutorial, though.
And last but not least, out of curiosity, have you actually profiled which one is faster? The first one involves a copy operation, but then allows a single dot product, which could actually be faster on GPU than adding up multiple dot products.

lasagne/layers/dense.py
```diff
@@ -89,6 +88,35 @@ def get_output_for(self, input, **kwargs):
...
+
+    - input_layer : `Layer` instance
+        The layer from which this layer will obtain its input
```
f0k: It should be `incoming`. You can just copy that from the base `Layer` docstring.

mheilman:
Ah, yeah, sorry.  I'll make those documentation fixes.

Regarding benchmarking, I threw together this little bit of code, and it looks like the summing is somewhat faster than concatenation, at least for a very simple case run on my laptop CPU.  I'll try to run it on a GPU hopefully tomorrow.

```
from lasagne.layers import DenseLayer,NonlinearityOnlyLayer,ConcatLayer,ElemwiseSumLayer,InputLayer 
import theano.tensor as T
from numpy.random import RandomState
from timeit import timeit

def setup():
    m1 = rng.rand(1000, 1000)
    m2 = rng.rand(1000, 1000)

    m1_sym = T.matrix()
    m2_sym = T.matrix()
    in_layer1 = InputLayer((1000, 1000))
    in_layer2 = InputLayer((1000, 1000))
    input_dict = {in_layer1: m1_sym, in_layer2: m2_sym}
    
    concat_layer = ConcatLayer([in_layer1, in_layer2])
    output_concat = DenseLayer(concat_layer, 100)
    foo_concat = theano.function([m1_sym, m2_sym],
                                 output_concat.get_output(input_dict),
                                 allow_input_downcast=True)

    dense1 = DenseLayer(in_layer1, 100, nonlinearity=None)
    dense2 = DenseLayer(in_layer2, 100, nonlinearity=None)
    output_sum = NonlinearityOnlyLayer(ElemwiseSumLayer([dense1, dense2]))
    foo_sum = theano.function([m1_sym, m2_sym],
                              output_sum.get_output(input_dict),
                              allow_input_downcast=True)

    return foo_sum, foo_concat, m1, m2

print("concat:", timeit('foo_concat(m1, m2)', number=100,
                        setup="from __main__ import setup; foo_sum, foo_concat, m1, m2 = setup()"))
print("sum:", timeit('foo_sum(m1, m2)', number=100,
                     setup="from __main__ import setup; foo_sum, foo_concat, m1, m2 = setup()"))
```

That printed out the following:
```
concat: 1.3420235139783472
sum: 0.6435205009765923
```

f0k:
Interesting, that's further apart than I had expected (although I'm not surprised that this approach is faster for the CPU). When you do the benchmark on the GPU, it's easier if you just run either of the functions in a simple `for _ in range(100):` loop and launch that script with `THEANO_FLAGS=profile=1,device=gpu,floatX=float32 CUDA_LAUNCH_BLOCKING=1 python thatscript.py` (once for the first, and once for the second function). This will give you a detailed listing of the runtime of involved Ops, of which you can ignore all `GpuToHost` and `GpuFromHost` ones. Otherwise you will need to take extra care to not include transfer costs in the `timeit` call.

benanne:
> I think `NonlinearityOnlyLayer` isn't too bad. What do the others think? (Yay, a name discussion!)

I'd go for `NonlinearityLayer` or even `NonlinLayer` instead. The "only" only makes sense when comparing it to `DenseLayer` or `Conv*DLayer`, and I don't think it should be seen as a "slimmed down" version of those layers. Also it saves four keystrokes :)

mheilman:
Thanks for the profiling tips.

I ran the following script with those flags on an EC2 GPU instance, once for summing, once for concatenation (the previous one I'd copied from an ipython nb and was missing some imports btw).

More detailed outputs are below, but with 1000 iterations, concatenation took about 7.5 seconds in function calls and summing took about 6.3 seconds.

```
from lasagne.layers import DenseLayer,NonlinearityOnlyLayer,ConcatLayer,ElemwiseSumLayer,InputLayer 
import theano.tensor as T
from numpy.random import RandomState
import theano
import sys

rng = RandomState(1234)
m1 = rng.rand(1000, 1000)
m2 = rng.rand(1000, 1000)

m1_sym = T.matrix()
m2_sym = T.matrix()
in_layer1 = InputLayer((1000, 1000))
in_layer2 = InputLayer((1000, 1000))
input_dict = {in_layer1: m1_sym, in_layer2: m2_sym}

if sys.argv[1] == "concat":
    concat_layer = ConcatLayer([in_layer1, in_layer2])
    output_concat = DenseLayer(concat_layer, 100)
    foo_concat = theano.function([m1_sym, m2_sym],
                                 output_concat.get_output(input_dict),
                                 allow_input_downcast=True)
    print("CONCAT")
    for _ in range(1000):
        foo_concat(m1, m2)
else:
    dense1 = DenseLayer(in_layer1, 100, nonlinearity=None)
    dense2 = DenseLayer(in_layer2, 100, nonlinearity=None)
    output_sum = NonlinearityOnlyLayer(ElemwiseSumLayer([dense1, dense2]))
    foo_sum = theano.function([m1_sym, m2_sym],
                              output_sum.get_output(input_dict),
                              allow_input_downcast=True)
    print("SUM"
    for _ in range(1000):
        foo_sum(m1, m2)
```

For summing, the  output was as follows:
```
...
SUM
Function profiling
==================
  Message: bench.py:32
  Time in 1000 calls to Function.__call__: 9.079201e+00s
  Time in Function.fn.__call__: 6.252115e+00s (68.862%)
  Time in thunks: 5.227695e+00s (57.579%)
  Total compile time: 1.142402e-01s
    Number of Apply nodes: 8
    Theano Optimizer time: 9.594989e-02s
       Theano validate time: 1.284599e-03s
    Theano Linker time (includes C, CUDA code generation/compiling): 1.142097e-02s
       Import time 2.271652e-03s

Time in all call to theano.grad() 0.000000e+00s
Class
---
<% time> <sum %> <apply time> <time per call> <type> <#call> <#apply> <Class name>
  63.3%    63.3%       3.311s       1.66e-03s     C     2000       2   theano.sandbox.cuda.basic_ops.GpuFromHost
  21.0%    84.4%       1.100s       1.10e-03s     C     1000       1   theano.sandbox.cuda.blas.GpuDot22
   9.3%    93.7%       0.487s       4.87e-04s     C     1000       1   theano.sandbox.cuda.blas.GpuGemm
   5.2%    98.9%       0.271s       2.71e-04s     C     1000       1   theano.sandbox.cuda.basic_ops.HostFromGpu
   1.0%    99.9%       0.052s       5.17e-05s     C     1000       1   theano.sandbox.cuda.basic_ops.GpuElemwise
   0.1%   100.0%       0.007s       3.47e-06s     C     2000       2   theano.sandbox.cuda.basic_ops.GpuDimShuffle
   ... (remaining 0 Classes account for   0.00%(0.00s) of the runtime)

Ops
---
<% time> <sum %> <apply time> <time per call> <type> <#call> <#apply> <Op name>
  63.3%    63.3%       3.311s       1.66e-03s     C     2000        2   GpuFromHost
  21.0%    84.4%       1.100s       1.10e-03s     C     1000        1   GpuDot22
   9.3%    93.7%       0.487s       4.87e-04s     C     1000        1   GpuGemm{inplace}
   5.2%    98.9%       0.271s       2.71e-04s     C     1000        1   HostFromGpu
   1.0%    99.9%       0.052s       5.17e-05s     C     1000        1   GpuElemwise{Composite{(i0 * ((i1 + i2 + i3) + Abs((i1 + i2 + i3))))}}[(0, 3)]
   0.1%   100.0%       0.007s       3.47e-06s     C     2000        2   GpuDimShuffle{x,0}
   ... (remaining 0 Ops account for   0.00%(0.00s) of the runtime)

Apply
------
<% time> <sum %> <apply time> <time per call> <#call> <id> <Apply name>
  33.6%    33.6%       1.754s       1.75e-03s   1000     1   GpuFromHost(<TensorType(float32, matrix)>)
  29.8%    63.3%       1.556s       1.56e-03s   1000     0   GpuFromHost(<TensorType(float32, matrix)>)
  21.0%    84.4%       1.100s       1.10e-03s   1000     4   GpuDot22(GpuFromHost.0, W)
   9.3%    93.7%       0.487s       4.87e-04s   1000     5   GpuGemm{inplace}(GpuDot22.0, TensorConstant{1.0}, GpuFromHost.0, W, TensorConstant{1.0})
   5.2%    98.9%       0.271s       2.71e-04s   1000     7   HostFromGpu(GpuElemwise{Composite{(i0 * ((i1 + i2 + i3) + Abs((i1 + i2 + i3))))}}[(0, 3)].0)
   1.0%    99.9%       0.052s       5.17e-05s   1000     6   GpuElemwise{Composite{(i0 * ((i1 + i2 + i3) + Abs((i1 + i2 + i3))))}}[(0, 3)](CudaNdarrayConstant{[[ 0.5]]}, GpuDimShuffle{x,0}.0, GpuDimShuffle{x,0}.0, GpuGemm{inplace}.0)
   0.1%   100.0%       0.005s       5.27e-06s   1000     3   GpuDimShuffle{x,0}(b)
   0.0%   100.0%       0.002s       1.68e-06s   1000     2   GpuDimShuffle{x,0}(b)
   ... (remaining 0 Apply instances account for 0.00%(0.00s) of the runtime)
```

And for concatenation:
```
...
CONCAT
Function profiling
==================
  Message: bench.py:22
  Time in 1000 calls to Function.__call__: 9.692008e+00s
  Time in Function.fn.__call__: 7.473125e+00s (77.106%)
  Time in thunks: 6.107281e+00s (63.014%)
  Total compile time: 9.842801e-02s
    Number of Apply nodes: 7
    Theano Optimizer time: 8.158374e-02s
       Theano validate time: 7.433891e-04s
    Theano Linker time (includes C, CUDA code generation/compiling): 1.050282e-02s
       Import time 2.248764e-03s

Time in all call to theano.grad() 0.000000e+00s
Class
---
<% time> <sum %> <apply time> <time per call> <type> <#call> <#apply> <Class name>
  54.2%    54.2%       3.313s       1.66e-03s     C     2000       2   theano.sandbox.cuda.basic_ops.GpuFromHost
  21.0%    75.3%       1.285s       1.29e-03s     C     1000       1   theano.sandbox.cuda.blas.GpuDot22
  20.4%    95.6%       1.243s       1.24e-03s     C     1000       1   theano.sandbox.cuda.basic_ops.GpuJoin
   3.5%    99.2%       0.214s       2.14e-04s     C     1000       1   theano.sandbox.cuda.basic_ops.HostFromGpu
   0.8%    99.9%       0.048s       4.79e-05s     C     1000       1   theano.sandbox.cuda.basic_ops.GpuElemwise
   0.1%   100.0%       0.003s       3.49e-06s     C     1000       1   theano.sandbox.cuda.basic_ops.GpuDimShuffle
   ... (remaining 0 Classes account for   0.00%(0.00s) of the runtime)

Ops
---
<% time> <sum %> <apply time> <time per call> <type> <#call> <#apply> <Op name>
  54.2%    54.2%       3.313s       1.66e-03s     C     2000        2   GpuFromHost
  21.0%    75.3%       1.285s       1.29e-03s     C     1000        1   GpuDot22
  20.4%    95.6%       1.243s       1.24e-03s     C     1000        1   GpuJoin
   3.5%    99.2%       0.214s       2.14e-04s     C     1000        1   HostFromGpu
   0.8%    99.9%       0.048s       4.79e-05s     C     1000        1   GpuElemwise{Composite{(i0 * ((i1 + i2) + Abs((i1 + i2))))}}[(0, 1)]
   0.1%   100.0%       0.003s       3.49e-06s     C     1000        1   GpuDimShuffle{x,0}
   ... (remaining 0 Ops account for   0.00%(0.00s) of the runtime)

Apply
------
<% time> <sum %> <apply time> <time per call> <#call> <id> <Apply name>
  28.4%    28.4%       1.734s       1.73e-03s   1000     2   GpuFromHost(<TensorType(float32, matrix)>)
  25.9%    54.2%       1.579s       1.58e-03s   1000     1   GpuFromHost(<TensorType(float32, matrix)>)
  21.0%    75.3%       1.285s       1.29e-03s   1000     4   GpuDot22(GpuJoin.0, W)
  20.4%    95.6%       1.243s       1.24e-03s   1000     3   GpuJoin(TensorConstant{1}, GpuFromHost.0, GpuFromHost.0)
   3.5%    99.2%       0.214s       2.14e-04s   1000     6   HostFromGpu(GpuElemwise{Composite{(i0 * ((i1 + i2) + Abs((i1 + i2))))}}[(0, 1)].0)
   0.8%    99.9%       0.048s       4.79e-05s   1000     5   GpuElemwise{Composite{(i0 * ((i1 + i2) + Abs((i1 + i2))))}}[(0, 1)](CudaNdarrayConstant{[[ 0.5]]}, GpuDot22.0, GpuDimShuffle{x,0}.0)
   0.1%   100.0%       0.003s       3.49e-06s   1000     0   GpuDimShuffle{x,0}(b)
   ... (remaining 0 Apply instances account for 0.00%(0.00s) of the runtime)
```

mheilman:
Better now?  Let me know if you have any other suggestions.

f0k:
> I'd go for `NonlinearityLayer` or even `NonlinLayer` instead. The "only" only makes sense when comparing it to `DenseLayer` or `Conv*DLayer`, and I don't think it should be seen as a "slimmed down" version of those layers.

On the other hand, `NonlinearityLayer` sounds as if it was a specialty of this layer to sport a nonlinearity. I think it's okay to regard `NonlinearityOnlyLayer` as a stripped-down version of usual layers that do something *besides* applying a nonlinearity.

By the way, if `nonlinearity` would default to `None` instead of the linear rectifier, we could also just add that keyword to every layer for convenience... :)

Anyway, I don't have a very strong opinion on the name, as much as I love name discussions.

---

> I ran the following script with those flags on an EC2 GPU instance, once for summing, once for concatenation

Cool. The interesting parts are for summing:
>```
21.0%    84.4%       1.100s       1.10e-03s   1000     4   GpuDot22(GpuFromHost.0, W)
 9.3%    93.7%       0.487s       4.87e-04s   1000     5   GpuGemm{inplace}(GpuDot22.0, TensorConstant{1.0}, GpuFromHost.0, W, TensorConstant{1.0})
```

And for concatenation:
>```
21.0%    75.3%       1.285s       1.29e-03s   1000     4   GpuDot22(GpuJoin.0, W)
20.4%    95.6%       1.243s       1.24e-03s   1000     3   GpuJoin(TensorConstant{1}, GpuFromHost.0, GpuFromHost.0)
```

You can nicely see that the sum gets compiled into a dot product and then a gemm call that adds its output to the output of the previous dot product, taking 1.6s in total, while the concatenation and single dot product takes 2.5s in total. Good to know!

mheilman:
I'm not sure I follow this comment.

> By the way, if nonlinearity would default to None instead of the linear rectifier, we could also just add that keyword to every layer for convenience... :)

Possibly related: I thought that maybe this nonlinearity functionality could just be integrated into `MultipleInputsLayer` (in `get_output`, I suppose), but having a separate class seemed conceptually a bit cleaner at the time.

Thanks for the more detailed analysis of the profiling output.  I'm still learning how to decipher that stuff.

f0k:
> Possibly related: I thought that maybe this nonlinearity functionality could just be integrated into  `MultipleInputsLayer` (in `get_output`, I suppose)

Yep, that's what I meant. `MultipleInputsLayer` (and *every* layer, for that matter) could get a `nonlinearity=` keyword in its constructor so you don't need an extra `NonlinearityLayer` just like you don't need it after a `DenseLayer` already. The problem is that `nonlinearity=` currently defaults to `lasagne.nonlinearities.rectify` wherever it is defined. That wouldn't be a sensible default for `ConcatLayer`, for example, and it would also be awkward to have it default to the rectifier for some layer classes and to the identity for others. If we generally default it to identity, we could add `nonlinearity=` to the base `Layer` class such that all classes would transparently support it.

mheilman:
Along the lines of the above, one could almost allow `DenseLayer` to take multiple inputs, and then one wouldn't need separate classes for merging.  It could just do what it does now if there is one input, but do what `ElemwiseSumLayer` does if there is more than one input (perhaps with a kwarg for doing what `ConcatLayer`does instead).

f0k:
> Just a thought.

Interesting one! You're probably right that it's better to just have all the building blocks. The use case of optimally merging multiple layers with a dot product wouldn't warrant a new class, but it would be an ideal candidate for a macro function!

benanne:
Looks good, merging! We might want to move it out of `dense.py` later since it doesn't really belong there (we can do that at any time since all layer classes are imported to `lasagne.layers` anyway).

PR #227, opened 2015-05-02
Pooling layer docs

benanne:
Cleaned up the docs for the pooling layers a bit further (@ebenolson already did a bunch of work on them before). This is as preparation for the corrmm/cuda_convnet/dnn docs, which I will probably do next.

### DIFF
--- a/lasagne/layers/pool.py
+++ b/lasagne/layers/pool.py
@@ -32,8 +32,8 @@ def pool_output_length(input_length, pool_size, stride,
     pad : integer
         The number of elements to be added to the input on each side.
     ignore_border: bool
-        If True, partial pooling regions will be ignored.
-        Must be True if pad != 0.
+        If ``True``, partial pooling regions will be ignored.
+        Must be ``True`` if ``pad != 0``.
 
     Returns
     -------
@@ -43,11 +43,11 @@ def pool_output_length(input_length, pool_size, stride,
 
     Notes
     -----
-    When `ignore_border == True`, this is given by the number of full
+    When ``ignore_border == True``, this is given by the number of full
     pooling regions that fit in the padded input length,
     divided by the stride (rounding down).
 
-    If `ignore_border == False`, a single partial pooling region is
+    If ``ignore_border == False``, a single partial pooling region is
     appended if at least one input element would be left uncovered otherwise.
     """
     if input_length is None or pool_size is None:
@@ -86,17 +86,17 @@ class MaxPool1DLayer(Layer):
     pool_size : integer or iterable
         The length of the pooling region
 
-    stride : integer, iterable or None
+    stride : integer, iterable or ``None``
         The stride between sucessive pooling regions.
-        If `None`, stride = pool_size.
+        If ``None`` then ``stride == pool_size``.
 
     pad : integer or iterable
         The number of elements to be added to the input on each side.
         Must be less than stride.
 
     ignore_border : bool
-        If True, partial pooling regions will be ignored.
-        Must be True if pad != 0.
+        If ``True``, partial pooling regions will be ignored.
+        Must be ``True`` if ``pad != 0``.
 
     **kwargs
         Any additional keyword arguments are passed to the :class:`Layer`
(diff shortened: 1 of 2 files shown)


benanne:
Anyone wanna have a quick look at this? Anyone who wants to, it's just useful to have a second set of eyes looking at it.  It would be appreciated! :)

lasagne/layers/pool.py
```diff
@@ -72,39 +72,45 @@ def pool_output_length(input_length, pool_size, stride,
...
+    incoming : a :class:`Layer` instance or tuple
+        The layer feeding into this layer, or the expected input shape. The
+        output of this layer should be a 3D tensor.
+
+    pool_size : integer or iterable
+        The length of the pooling region
+
+    stride : integer, iterable or None
+        The stride between sucessive pooling regions.
+        If `None`, stride = pool_size.
```
ebenolson: Is there a standard for when to use code quotes? Why here but not 2 lines above?

lasagne/layers/pool.py
```diff
@@ -326,9 +363,29 @@ def get_output_for(self, input, **kwargs):
...
+        (i.e. mean-pooling) and can be replaced by any other aggregation
+        function.
+
+    **kwargs
+        Any additional keyword arguments are passed to the :class:`Layer`
+        superclass.
+
+    Notes
+    -----
+    This layer requires that the size of the axis along which it groups units
```
ebenolson: Copypaste artifact

lasagne/layers/pool.py
```diff
@@ -217,28 +229,44 @@ def get_output_for(self, input, **kwargs):
...
+    :class:`Conv1DLayer` and :class:`Conv2DLayer`. The layer can be used to
+    implement maxout.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or tuple
+        The layer feeding into this layer, or the expected input shape.
+
+    pool_size : integer
+        the size of the pooling windows, i.e. the number of features / feature
```
ebenolson: we refer to them as pooling "regions", "windows", and "groups" in various places, probably should choose one.

ebenolson:
Made a couple minor comments, looks pretty good to me (I didn't try compiling it though, just looked through the diff).

lasagne/layers/pool.py
benanne: I'm not sure to be honest. I think you're even supposed to use double backticks (that's what you're talking about right?). But I think the rules for the actual type specification may be different. Maybe someone with more numpydoc experience can shed some light on this.

lasagne/layers/pool.py
benanne: Which line exactly? I think I'm reading over it.

lasagne/layers/pool.py
benanne: Good point. What name is most common in literature? I guess I'm fine standardizing on either "regions" or "windows" (I think "groups" is a bit of a vague term).

The term 'window' is also polyvalent to some extent, so I guess 'region' would be the least confusing. I'll go with that, unless anyone wants to have another name discussion of course ;)

benanne:
Thanks!

lasagne/layers/pool.py
ebenolson: The whole notes section, `GlobalPoolLayer` doesn't have a pool size

benanne: Whoops :) Thanks for catching this!

benanne:
I made the proposed changes. Since this is sort of small I'll merge it tomorrow provided that there are no further comments and Travis passes.

lasagne/layers/pool.py
f0k: Did I hear "name discussion"? :)

"groups" are used in cuda-convnet for grouped convolution, where you learn kernels for subsets of the input channels. In this case, you don't specify the size of the pooling region/window/thing, but the number of groups, and if you use the same number for each convolutional layer, you get a multi-column architecture.
So this notion of "groups" is closely related to what we're doing here if axis=1, but as it doesn't have to be axis=1 and if we want to have a common term, I think "region" is a good choice. "window" is something I would associate with the last axis/axes in bc01* notation.

PR #204, opened 2015-04-12
Documentation: Sphinx format to numpydoc format

MartinThoma:
See issue #203

PR #228, opened 2015-05-02
parameter registration / tagging system

benanne:
This is a PR to address #164. It should not be merged before #182 and #204 (and will require a rebase). I just wanted to get things going on this as well. Hopefully the rebase won't be too painful.

to do list:

- [x] extract `create_param()`, turn it into a helper function
- [x] add `Layer.add_param()`, `Layer.params`
- [x] replace `Layer.get_params()` (and remove `Layer.get_bias_params()`)
- [x] update all `Layer` subclasses to use `Layer.add_param()` instead of `Layer.create_param()` and supply the correct tags, also remove any `Layer.get_params()` and `Layer.get_bias_params()` implementations
- [x] update helper functions that get parameters to support tags
- [x] update all the tests: move `create_param` tests, add `Layer.add_param` tests, add `Layer.get_params` tests, add additional tests for using helper functions with tags
- [x] clean up (remove the commented code)
- [x] rebase after #182 is merged

after rebasing:

- [x] add deprecation placeholders
- [x] add (correctly formatted) docstrings for `Layer.add_param`, `Layer.get_params`

after rebasing again (to include all docs changes so far):

- [x] update docstrings for helper functions to mention the tags
- [x] update documentation and examples (if necessary)


benanne:
I think I'm more or less done with this, I just need to add / update a bunch of documentation. But since there is some overlap with #204 there, I will wait to spend time on that until #182 and #204 have been merged and I'm able to rebase this.

benanne:
Intersting Python 2/3 difference here: in Python 3, the `keys()` method of an `OrderedDict` returns a `KeysView`, not a list. This is now what `get_params()` returns, and this causes a bunch of problems.

The easiest fix is just to wrap it in `list(...)` to make sure `get_params()` returns a list in both Python 2 and Python 3. I'm okay with that solution, but before doing that I figured I'd ask if there are any caveats here, or if there is maybe a better solution. I have little to no experience with Python 3 so I wouldn't know.

dnouri:
I also don't know a better way to do this other than using `list(d.keys())` or `list(d)`.

benanne:
Adding that fix makes it pass Travis, which is nice! There is no point in reviewing this yet though (if anyone was planning on doing that), better to wait until after the imminent rebase.

benanne:
currently rebasing this... it's proving to be a great git exercise!

benanne:
Done! I hope I did everything right. Tests seem to pass now so that's good.

lasagne/layers/base.py
```diff
@@ -38,46 +36,33 @@ def __init__(self, incoming, name=None):
...
-
-        :note:
-            By default this returns an empty list, but it should be overridden
-            in a subclass that has trainable parameters.
+        TODO: docstring
         """
-        return []
+        result = list(self.params.keys())
 
-    def get_bias_params(self):
```
f0k: The base class should probably have a deprecated placeholder for now:
```python
def get_bias_params(self):
    warnings.warn(...)
    return self.get_params(regularizable=False)
```

lasagne/layers/helper.py
```diff
@@ -11,8 +12,6 @@
     "get_output",
     "get_output_shape",
     "get_all_params",
-    "get_all_bias_params",
-    "get_all_non_bias_params",
```
f0k: Again, we might want to have deprecated redirections for those.

lasagne/tests/layers/test_base.py
```diff
@@ -27,41 +32,69 @@ def test_layer_from_shape(self, layer_from_shape):
...
-        factory = Mock()
-        factory.return_value = array
-        result = layer.create_param(factory, (2, 3))
-        assert (result.get_value() == array).all()
-        factory.assert_called_with((2, 3))
-
     def test_named_layer(self):
         from lasagne.layers.base import Layer
         l = Layer(Mock(), name="foo")
         assert l.name == "foo"
```
f0k: Shouldn't this use the new `named_layer` fixture?

lasagne/tests/layers/test_helper.py
```diff
@@ -4,7 +4,7 @@
 import theano
 
 
-class TestGetAllLayers:
+class TestGetAllLayers(object):
```
f0k: Hmm, most of the tests have old-style classes... we should probably be consistent there. Is there any reason for using one or the other with pytest?

benanne: I do it by default nowadays. I guess old-style classes are frowned upon? I don't know :) Don't really care either way, as long as all classes in the library itself are new-style.

lasagne/utils.py
```diff
@@ -204,3 +204,76 @@ def compute_norms(array, norm_axes=None):
...
+
+    Parameters
+    ----------
+
+    spec : numpy array, Theano shared variable, or callable
+        One of three things:
+            * a numpy array with the initial parameter values
+            * a Theano shared variable representing the parameters
+            * a function or callable that takes the desired shape of
+              the parameter array as its single argument.
```
f0k: "and returns a numpy array"

lasagne/utils.py
```diff
@@ -204,3 +204,76 @@ def compute_norms(array, norm_axes=None):
...
+            raise RuntimeError("parameter array has shape %s, should be "
+                               "%s" % (spec.shape, shape))
+        return theano.shared(spec, name=name)
+
+    elif hasattr(spec, '__call__'):
+        arr = spec(shape)
+        if not isinstance(arr, np.ndarray):
+            raise RuntimeError("cannot initialize parameters: the "
+                               "provided callable did not return a numpy "
+                               "array")
```
f0k: This is independent of this PR, but as I saw it now, this check should probably go something like this instead:
```python
arr = spec(shape)
try:
    arr = floatX(arr)
except Exception:
    ... did not return an array-like ...
if arr.shape != shape:
    ... did not return correct shape ...
return theano.shared(arr, name=name)
```
This should also handle scalars correctly.

f0k:
Looks good so far! Cool that https://github.com/Lasagne/Lasagne/issues/164#issuecomment-84996815 seems to have been free of errors :)

lasagne/utils.py
benanne: If I include this change in this PR then I should rebase it to include the new floatX implementation, I guess.

benanne:
I've added deprecation placeholders and addressed your other comments, except the change to `create_param` to make it support scalar values. That depends on your recent change of the `floatX()` implementation, so I would have to rebase this PR first to add that in here. Maybe it's better kept as a separate thing anyway.

f0k:
Looks good! If you need help with the remaining TODOs, let me know.

benanne:
I should be able to finish it today. I'll go ahead and rebase it again as well so we can include the modifications to `create_param`.

benanne:
Looks like I need to wait for #204 to do the last two things, else I'll just be doing double work (or the rebase/merge will be painful).

benanne:
Rebased once more so I can update the docs later :)

lasagne/layers/base.py
```diff
@@ -39,49 +37,56 @@ def __init__(self, incoming, name=None):
...
 
-    def get_params(self):
+    def get_params(self, **tags):
         """
         Returns a list of all the Theano variables that parameterize the layer.
 
+        By default, all parameters that were registered with
+        ``Layer.add_param()`` will be returned, in the order they were
+        registered. The list can optionally be filtered by specifying tags
+        as keyword arguments.
```
f0k: I'm wondering if this should be targetted more towards users of layers, who don't know about `add_param`. An alternative might be:
```
By default, all parameters that participate in the forward pass will be returned (in the
order they were registered in the Layer's constructor via :meth:`add_param()`). The list
can optionally be filtered by specifying tags as keyword arguments. For example,
``trainable=True`` will only return trainable parameters, and ``regularizable=True``
will only return parameters that can be regularized (e.g, by L2 decay).
```
Feel free to mix and match if you agree.

benanne: Sure, looks good.

benanne:
I've updated the remaining docs, and in one go I fleshed out the custom layers documentation a bit as well, since I had to update it anyway.

docs/modules/layers.rst
```diff
@@ -7,6 +7,8 @@
 Helper functions
 ----------------
 
+.. autofunction:: get_output
+.. autofunction:: get_output_shape
```
f0k: Ah, good catch.

docs/user/custom_layers.rst
```diff
@@ -58,21 +60,100 @@ input as follows:
...
+network would). This matrix is a parameter of the layer. The shape of the
+matrix will be ``(num_inputs, num_units)``, where ``num_inputs`` is the
+number of input features and ``num_units`` has to be specified when the layer
+is created.
+
+.. code:: python
+
+    class DotLayer(lasagne.layers.Layer):
+        def __init__(self, num_units, W=lasagne.init.Normal(0.01), **kwargs):
+            super(DotLayer, self).__init__(**kwargs)
```
f0k: You're missing the `incoming` argument!

benanne: whoops :)

docs/user/custom_layers.rst
```diff
@@ -58,21 +60,100 @@ input as follows:
...
+parameter matrix. This property is available in all Lasagne layers, once the
+superclass constructor has been called.
+
+
+A layer with multiple behaviors
+-------------------------------
+
+Some layers can have multiple behaviors. For example, a layer implementing
+dropout should be able to be switched on or off. During training, we want it
+to apply dropout noise to its input, but during evaluation we don't want it to
```
f0k: *and upscale the surviving values

(The scaling should be mentioned somewhere to not confuse people familiar with the dropout paper.)

benanne: Ah yes, I guess so. It's in the code anyway so we might as well explain it :)

docs/user/layers.rst
```diff
@@ -143,10 +143,60 @@ Propagating data through layers
...
+>>> y1, y2 = lasagne.layers.get_output([l_out1, l_out2])
+
+You could also just call :func:`lasagne.layers.get_output()` twice:
+
+>>> y1 = lasagne.layers.get_output(l_out1)
+>>> y2 = lasagne.layers.get_output(l_out2)
+
+However, this is **not recommended**! Some network layers may have
+non-deterministic output, such as dropout layers. If you compute the network
+outputs with separate calls to :func:`lasagne.layers.get_output()`, they will
```
f0k: *If you create the output expressions

(I think it's important not to mix up the actual, nonsymbolic computation with the creation of the symbolic expressions.)

lasagne/layers/base.py
```diff
@@ -39,49 +37,58 @@ def __init__(self, incoming, name=None):
...
+    def get_params(self, **tags):
         """
         Returns a list of all the Theano variables that parameterize the layer.
 
+        By default, all parameters that participate in the forward pass will be
+        returned (in the order they were registered in the Layer's constructor
+        via :meth:`add_param()`). The list can optionally be filtered by
+        specifying tags as keyword arguments. For example, ``trainable=True``
+        will only return trainable parameters, and ``regularizable=True``
+        will only return parameters that can be regularized (e.g, by L2 decay).
```
f0k: *e.g.,

lasagne/layers/helper.py
```diff
@@ -312,9 +311,9 @@ def get_output_shape(layer_or_layers, input_shapes=None):
...
 
 
-def get_all_params(layer):
+def get_all_params(layer, **tags):
     """
-    This function gathers all learnable parameters of all layers below one or
+    This function gathers all parameters of all layers below one or
     more given :class:`Layer` instances, including the layer(s) itself. Its
     main use is to collect all parameters of a network just given the output
     layer(s).
```
f0k: I'd suggest to add the additional explanation on filtering from `Layer.get_params()` here as well, because `get_all_params()` is probably what people will learn about first.

lasagne/utils.py
```diff
@@ -205,3 +205,82 @@ def compute_norms(array, norm_axes=None):
...
+    Helper method to create Theano shared variables for layer parameters
+    and to initialize them.
+
+    Parameters
+    ----------
+
+    spec : numpy array, Theano shared variable, or callable
+        One of three things:
+            * a numpy array with the initial parameter values
+            * a Theano shared variable representing the parameters
```
f0k: I think for the markup to work there must be an new line before the list, and the list should not be intended (i.e., it should start on the same level as "One of three things").

f0k:
Almost there! As always, everything reads well.

benanne:
Thanks! It's surprising how easy it is to write docs when I'm doing it to avoid writing my dissertation instead ;)

f0k:
"lasagne/utils.py:220:1: W293 blank line contains whitespace" -- almost as bloody a beginner's mistake as mine earlier today ;)

benanne:
I've had that happen at least five times today. Sublime Text indents by default if you press enter. There must be a plugin that removes whitespace on blank lines on save or something :)

f0k:
It's the same with most editors, but the better ones have pep8 plugins that highlight all errors. What saved me with simpler editors was `git diff`, it highlights trailing white space in red (if you enabled colored diffs).

benanne:
I actually have a pep8/flake8 plugin for it but shortly after I installed it I turned off a few rules that were really doing my head in. 'Blank line contains whitespace' must have been one of them :p

benanne:
I'll leave it open till tomorrow afternoon in case there are any further comments.

skaae:
I think you forgot to add `get_all_bias_params` and `get_all_non_bias_params` to `__all__` in `helper.py`

benanne:
Looks like I did, good catch! I'll quickly fix it in master.

skaae:
btw I have question about add params.
In the old recurrent layer get params was

```
    def get_params(self):
        params = (helper.get_all_params(self.input_to_hidden) +
                  helper.get_all_params(self.hidden_to_hidden))
```

which i can nearly reproduce if i use
`get_all_bias_params` and `get_all_non_bias_params` and then add the parameters with the appropriate tags, but then i'll lose any extra tags that the user might have added. Can we add a helper function that returns the parameters and the tags for this situation?

benanne:
What about something like this?
```
    def get_params(self, **tags):
        params = (helper.get_all_params(self.input_to_hidden, **tags) +
                  helper.get_all_params(self.hidden_to_hidden, **tags))

        if self.learn_init:
            return params + self.get_init_params(**tags)
        else:
            return params
```

You should not use `get_all_bias_params` or `get_all_non_bias_params`, they are deprecation placeholders and will be removed.

skaae:
great, thanks

f0k:
I'd go a different route and modify `self.params` instead of overriding `get_params`:
```python
def __init__(...):
    ...
    self.params.update(self.input_to_hidden.params)
    self.params.update(self.hidden_to_hidden.params)
    initsomething = self.add_param(initsomething, initshape, trainable=learn_init)
```
That's way easier than checking the tags again in `self.get_init_params`, and my suggested implementation for the tagging system explicitly aimed at making `self.params` easily modifiable.

skaae:
I dont entirely get your second example. I have a related problem. 

In the LSTM layer i would like to optionally allow the user to provide an output network. The outptut network takes the current hidden state in the LSTM and calculates some output that is then used as input in the next timestep. This is typically used in language models.

The LSTM `get_params` needs to return all parameters that are registered for the LSTM and return the parameters of the `output_network` if it exists. 

I would like to be able to call `get_all_params` on self er something like that :)

I could use Sanders solution but then I would need to explicitly return all LSTM parameters.
At the moment I copied `get_params` from `base.py` and modified it to return output_network parameters, but that is pretty ugly.

f0k:
> I dont entirely get your second example.

Which part? It uses the base class implementation to return the `initsomething` parameter depending on the tags (I don't know what you called that parameter, that's why I referred to it as `initsomething`), and the `get_all_params` helper function to return the parameters of the embedded networks depending on the tags. Sander's suggestion was meant to do the same, but it failed to check the tags.

> I would like to be able to call  get_all_params  on self er something like that :)

That's what `super(name_of_the_current_class, self).get_params(**tags)` does.

PR #229, opened 2015-05-04
Utils docs

JeffreyDF:
Cfr. #203. 

I found it to be a bit tricky so it probably needs some work. I only changed the docs for compute_norms to the numpydoc format.

### DIFF
--- a/lasagne/utils.py
+++ b/lasagne/utils.py
@@ -5,17 +5,40 @@ import theano.tensor as T
 
 
 def floatX(arr):
-    """
-    Shortcut to turn a numpy array into an array with the
-    correct dtype for Theano.
+    """Converts numpy array to one with the correct Theano data-type.
+
+    Parameters
+    ----------
+    arr : numpy array
+        The array to be converted.
+
+    Returns
+    -------
+    numpy array
+        The input array in the ``floatX`` data-type configured for Theano.
     """
     return arr.astype(theano.config.floatX)
 
 
 def shared_empty(dim=2, dtype=None):
-    """
+    """Creates empty Theano shared variable.
+
     Shortcut to create an empty Theano shared variable with
     the specified number of dimensions.
+
+    Parameters
+    ----------
+    dim : int, optional
+        The number of dimensions for the empty variable, defaults to 2.
+    dtype : a numpy data-type, optional
+        The desired data-type for the variable, defaults to the Theano
+        ``floatX`` data-type.
+
+    Returns
+    -------
+    Theano shared variable
+        An empty Theano shared variable of data-type ``dtype`` with
+        `dim` dimensions.
     """
     if dtype is None:
         dtype = theano.config.floatX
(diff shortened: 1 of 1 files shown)


lasagne/utils.py
```diff
@@ -5,17 +5,40 @@
 
 
 def floatX(arr):
-    """
-    Shortcut to turn a numpy array into an array with the
-    correct dtype for Theano.
+    """Converts numpy array to one with the correct Theano data-type.
```
benanne: I think the phrase "Theano data-type" might confuse a few people, since we're talking about numpy dtypes here (not Theano tensors or shared variables). Refering to this as 'dtype' is probably clearer.

lasagne/utils.py
```diff
@@ -25,10 +48,21 @@ def shared_empty(dim=2, dtype=None):
 
 
 def as_theano_expression(input):
-    """
-    Wraps the given input as a Theano constant if it is not
+    """Wrap as Theano expression.
+
+    Wraps the given input as a Theano Constant if it is not
```
benanne: 'constant' doesn't need a capital here I think.

lasagne/utils.py
```diff
@@ -25,10 +48,21 @@ def shared_empty(dim=2, dtype=None):
...
     a valid Theano expression already. Useful to transparently
     handle numpy arrays and Python scalars, for example.
+
+    Parameters
+    ----------
+    input :
+        Expression to be converted to a Theano Constant.
```
benanne: Same here.

lasagne/utils.py
```diff
@@ -25,10 +48,21 @@ def shared_empty(dim=2, dtype=None):
...
+
+    Returns
+    -------
+    Theano symbolic constant
+        Theano Constant version of `input`.
```
benanne: and here :)

benanne:
Looks good, just two minor remarks!

lasagne/utils.py
JeffreyDF: Ah, yes, I didn't want to confuse it -- a TensorConstant -- with a "normal" constant. But it's probably better without any capitalisation. I'll remove it.

lasagne/utils.py
JeffreyDF: You're right. It's explained in the "returns" segment what exactly this dtype is.

lasagne/utils.py
```diff
@@ -25,10 +48,21 @@ def shared_empty(dim=2, dtype=None):
...
-    Wraps the given input as a Theano constant if it is not
+    """Wrap as Theano expression.
+
+    Wraps the given input as a Theano Constant if it is not
     a valid Theano expression already. Useful to transparently
     handle numpy arrays and Python scalars, for example.
+
+    Parameters
+    ----------
+    input :
```
JeffreyDF: I don't have dtype for this input yet. Anyone can help?

JeffreyDF:
Changed a few things based on the feedback from Sander.

lasagne/utils.py
benanne: Something like "number, numpy array or Theano expression" is fine I guess :)

benanne:
Looks good, will merge tomorrow if there are no further comments.

benanne:
For whatever reason GitHub doesn't show it, but Travis passed.

JeffreyDF:
>For whatever reason GitHub doesn't show it, but Travis passed.

Might be because of GitHub being down for a little while around that time but that's just a guess.

benanne:
Merging!

JeffreyDF:
Thanks!

PR #232, opened 2015-05-06
documentation for the cuDNN-based convolution/pooling layer implementations

benanne:
This will mostly be copy pasting I guess, but we should probably have dedicated docstrings for all of them anyway because they usually implement some extra features (and sometimes mis a few others).

### DIFF
--- a/lasagne/layers/dnn.py
+++ b/lasagne/layers/dnn.py
@@ -24,7 +24,49 @@ class DNNLayer(Layer):
 
 
 class Pool2DDNNLayer(DNNLayer):
+    """
+    2D pooling layer
 
+    Performs 2D mean- or max-pooling over the two trailing axes of a 4D input
+    tensor. This is an alternative implementation which uses
+    ``theano.sandbox.cuda.dnn.dnn_pool`` directly.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or tuple
+        The layer feeding into this layer, or the expected input shape. The
+        output of this layer should be a 4D tensor.
+
+    pool_size : integer or iterable
+        The length of the pooling region in each dimension
+
+    stride : integer, iterable or ``None``
+        The strides between sucessive pooling regions in each dimension.
+        If ``None`` then ``stride = pool_size``.
+
+    pad : integer or iterable
+        Number of elements to be added on each side of the input
+        in each dimension. Each value must be less than
+        the corresponding stride.
+
+    mode : string
+        Pooling mode, one of 'max', 'average_inc_pad' or 'average_exc_pad'.
+        Defaults to 'max'.
+
+    **kwargs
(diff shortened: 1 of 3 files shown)


benanne:
I added docstrings to the dnn-based pool/conv layers. I guess I'll do cuda-convnet and corrmm in separate PRs to keep them more manageable and easier to review. So this is ready for review :)

lasagne/layers/dnn.py
```diff
@@ -24,7 +24,49 @@ class DNNLayer(Layer):
...
+    ``theano.sandbox.cuda.dnn.dnn_pool`` directly.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or tuple
+        The layer feeding into this layer, or the expected input shape. The
+        output of this layer should be a 4D tensor.
+
+    pool_size : integer or iterable
+        The length of the pooling region in each dimension
```
f0k: I'd add: "If an integer, is promoted to a square pooling region. If an iterable, needs to have two elements."

lasagne/layers/dnn.py
```diff
@@ -24,7 +24,49 @@ class DNNLayer(Layer):
...
 
+    Performs 2D mean- or max-pooling over the two trailing axes of a 4D input
+    tensor. This is an alternative implementation which uses
+    ``theano.sandbox.cuda.dnn.dnn_pool`` directly.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or tuple
+        The layer feeding into this layer, or the expected input shape. The
+        output of this layer should be a 4D tensor.
```
f0k: Minor comment: I find "The output if this layer" a bit confusing. As the previous sentence allows both a layer or a shape tuple, does "this layer" refer to the `Pool2DDNNLayer` that is instantiated, maybe? I guess I'd just drop this, as the layer description above already mentions the 4D input tensor.

f0k:
Minor comments that apply to the convolution docstring in general, but nothing else. Feel free to address them or ignore them and merge as is!

lasagne/layers/dnn.py
dnouri: Why isn't it 'integer or 2-tuple' in the first place?

benanne: I guess it could also be a list or another iterable with 2 elements? :) I guess this also needs clarification in the original pooling layer docs then.

lasagne/layers/dnn.py
benanne: Fair enough, I'll get rid of it.

benanne:
tried to address the comments, since some of this applies to the regular pooling layer implementation as well, I modified it there as well. Will merge tomorrow if there are no further comments!

benanne:
Merging.

PR #235, opened 2015-05-10
lasagne.layers.corrmm.Conv2DMMLayer docs

benanne:
This one was easy, it turns out `Conv2DMMLayer` and `Conv2DDNNLayer` have the exact same interface so I could copy the docstring with almost no modifications.

I will merge this tomorrow unless there are any remarks.

### DIFF
--- a/lasagne/layers/corrmm.py
+++ b/lasagne/layers/corrmm.py
@@ -1,7 +1,3 @@
-"""
-GpuCorrMM-based convolutional layers
-"""
-
 import theano
 
 from .. import init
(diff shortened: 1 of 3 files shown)


lasagne/layers/corrmm.py
```diff
@@ -32,6 +28,105 @@ class MMLayer(Layer):
...
+
+        If 'same', the convolution is computed wherever the input and the
+        filter overlap by at least half the filter size, when the filter size
+        is odd. In practice, the input is zero-padded with half the filter size
+        at the beginning and half at the end (or one less than half in the case
+        of an even filter size). This results in an output length that is the
+        same as the input length (for both odd and even filter sizes).
+
+    untie_biases : bool, default False
+        If ``False``, the layer will have a bias parameter for channel, which
```
ebenolson: should be "per channel" or "for each channel"?

benanne: Thanks. This mistake is also in the DNN / regular convolution layer docs, so I've fixed it there as well.

benanne:
Merging.

PR #236, opened 2015-05-10
Dense

skaae:
I have a few question about the NINLayer. The original description was:

```    
    Like DenseLayer, but broadcasting across all trailing dimensions beyond the
    2nd.  This results in a convolution operation with filter size 1 on all
    trailing dimensions.  Any number of trailing dimensions is supported,
    so NINLayer can be used to implement 1D, 2D, 3D, ... convolutions.
```
As i understand it the NINLayer applies a dense layer to the 2. dimension. In case of a 2d input the NINlayer is identical to a denselayer. If the input has more than two dimensions the weight matrix is applied to all trailing dimensions? Can someone explain to me why that is similar to a convolution with filter size 1? 

Maybe the usage example should have examples on how to implement convolutions when they are mentioned in the description?

Is the `untie_biases`description accurate?

### DIFF
--- a/lasagne/layers/dense.py
+++ b/lasagne/layers/dense.py
@@ -18,36 +18,32 @@ class DenseLayer(Layer):
     """
     A fully connected layer.
 
-    :parameters:
-        - incoming : a :class:`Layer` instance or a tuple
-            the layer feeding into this layer, or the expected input shape
+    Parameters
+    -----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
 
-        - num_units : int
-            The number of units of the layer
+    num_units : int
+        The number of units of the layer
 
-        - W : Theano shared variable, numpy array or callable
-            An initializer for the weights of the layer. If a Theano shared
-            variable is provided, it is used unchanged. If a numpy array is
-            provided, a shared variable is created and initialized with the
-            array. If a callable is provided, a shared variable is created and
-            the callable is called with the desired shape to generate suitable
-            initial values. The variable is then initialized with those values.
+    W : Theano shared variable, numpy array or callable
+        An initializer for the weights of the layer. If a shared variable or a
+        numpy array is provided the shape should  be (num_inputs, num_units).
+        See :meth:`Layer.create_param` for more information.
 
-        - b : Theano shared variable, numpy array, callable or None
-            An initializer for the biases of the layer. If a Theano shared
-            variable is provided, it is used unchanged. If a numpy array is
-            provided, a shared variable is created and initialized with the
-            array. If a callable is provided, a shared variable is created and
-            the callable is called with the desired shape to generate suitable
-            initial values. The variable is then initialized with those values.
+    b : Theano shared variable, numpy array, callable or None
+        An initializer for the biases of the layer. If a shared variable or a
+        numpy array is provided the shape should be (num_units,).
+        If None is provided the layer will have no biases.
+        See :meth:`Layer.create_param` for more information.
 
-            If None is provided, the layer will have no biases.
 
-        - nonlinearity : callable or None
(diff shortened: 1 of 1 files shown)


benanne:
Your understanding is correct. The broadcasting across the trailing dimensions is equivalent to a convolution operation with 1x1x... filters. It basically means "apply this weight matrix at each spatial position", which is also what a convolution does.

lasagne/layers/dense.py
```diff
@@ -18,36 +18,38 @@ class DenseLayer(Layer):
...
+    Parameters
+    -----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
+
+    num_units : int
+        The number of units of the layer
+
+    W : Theano shared variable, numpy array or callable
+        An initializer for the weights of the layer. If a Theano shared
```
benanne: It's probably a good idea to reduce the copypasta here by referring to the Layer.create_param docs instead (as was done here: https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/conv.py#L141). This should probably also mention the expected shape of W, in case someone wants to provide an existing shared variable or a numpy array.

lasagne/layers/dense.py
```diff
@@ -18,36 +18,38 @@ class DenseLayer(Layer):
...
+        variable is provided, it is used unchanged. If a numpy array is
+        provided, a shared variable is created and initialized with the
+        array. If a callable is provided, a shared variable is created and
+        the callable is called with the desired shape to generate suitable
+        initial values. The variable is then initialized with those values.
+
+    b : Theano shared variable, numpy array, callable or None
+        An initializer for the biases of the layer. If a Theano shared
```
benanne: same here :)

lasagne/layers/dense.py
```diff
@@ -111,11 +115,56 @@ def get_output_for(self, input, **kwargs):
...
+    Parameters
+    -----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
+
+    num_units : int
+        The number of units of the layer
+
+    untie_biases : bool
+        If true a separate bias vector is used for dimension trailing the 2nd.
```
benanne: You might want to copy the `untie_biases` description from one of the convolution layers as well, see .e.g. https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/conv.py#L132

lasagne/layers/dense.py
```diff
@@ -111,11 +115,56 @@ def get_output_for(self, input, **kwargs):
...
+        dimension.
+
+    W : Theano shared variable, numpy array or callable
+        An initializer for the weights of the layer. If a Theano shared
```
benanne: Same here, better to refer to `Layer.create_param` docs instead and make sure to mention the expected shape

lasagne/layers/dense.py
```diff
@@ -111,11 +115,56 @@ def get_output_for(self, input, **kwargs):
...
+        variable is provided, it is used unchanged. If a numpy array is
+        provided, a shared variable is created and initialized with the
+        array. If a callable is provided, a shared variable is created and
+        the callable is called with the desired shape to generate suitable
+        initial values. The variable is then initialized with those values.
+
+    b : Theano shared variable, numpy array, callable or None
+        An initializer for the biases of the layer. If a Theano shared
```
benanne: and here as well

skaae:
I changed the description of untie_bises? is it better/correct? Also added expected shapes and references to create_param.

benanne:
Looks good to me!

benanne:
Merging!

skaae:
I forgot to mention in the Dense docs that dimensions trailing the third dimension are flattened.

PR #237, opened 2015-05-11
update docstrings for shape.py

skaae:


lasagne/layers/shape.py
```diff
@@ -32,35 +39,42 @@ class ReshapeLayer(Layer):
...
+        elements of the input tensor. Any remaining elements must be
+        positive integers directly giving the size of the corresponding
+        dimension.
+
+    Usage
+    ----------
+    >>> from lasagne.layers import InputLayer, ReshapeLayer
+    >>> l_in = InputLayer((32, 100, 20))
+    >>> l1 = ReshapeLayer(l_in, ((32, 50, 40)))
+    >>> l1.get_output_shape()
```
benanne: This usage example is being run as a doctest and it seems to be failing (according to Travis). I think you need to add the expected output for the `get_output_shape()` calls.

lasagne/layers/shape.py
```diff
@@ -233,6 +253,30 @@ def get_output_for(self, input, **kwargs):
...
+    Pad all dimensions except the first 'batch_ndim' with 'width'
+    zeros on both sides, or with another value specified in 'val'.
+
+    Parameters
+    -----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
+
+    width : int
+        Padding with
```
benanne: should be width, not with :)

benanne:
lgtm, apart from one typo and the failing doctest :)

lasagne/layers/shape.py
skaae: Ok, I removed thow because i thought it looked ugly :) I'll add them again.

skaae:
fixed the tests + the typo

benanne:
Merging!

Issue #133, opened 2015-02-09
Parametrized Nonlinearities


ebattenberg:
[MSR's new work](http://arxiv.org/abs/1502.01852) using parametric rectifiers has renewed my interest in this topic.  Previous work on piece-wise linear units seemed a bit overly complicated for the performance improvement it achieved, but this simpler 2-part linear unit seems like it would be fairly easy to implement, and it helped MSR set the new (new new new...) state-of-the-art on Imagenet.  

The only architectural consideration from a Lasagne standpoint would be the fact that the nonlinearity portion of a layer would have its own parameters.  I know that Maxout is a kind of parametrized nonlinearity where the parameters are the incoming weights (and it can be implemented by max pooling across channels after a linear layer).  And dropout has been implemented as a separate layer.  With these in mind, maybe it would make sense to just make a parametric ReLU just a separate layer rather than one of the simple parameter-free nonlinearity types that come along with existing layers.  

There is probably a lot more pressing working to be done as far as Lasagne is concerned, but I just wanted to make sure we were thinking about this possibility during the planned re-architecting of certain features.

benanne:
I've also been thinking about this, and I agree that this is probably better implemented as a `Layer` subclass, at least for now. If other interesting use cases for parameterised nonlinearities come up we can consider it, but I have a feeling they're going to add a lot of complexity (`get_all_params` would have to be modified to inspect layers' nonlinearities for example).

diogo149:
With the new Batch Normalization paper out, I've decided to try all of these alternative activations and do a comparison (and this could address https://github.com/benanne/Lasagne/issues/85 as well). (:

My current idea for an implementation would be to have all nonlinearities as layers for consistency (kind of like in caffe). It seems like it should be possible to have them both as a function and as a layer if backwards compatibility / syntax is an issue, though that violates Python's "There should be one-- and preferably only one --obvious way to do it."

I'd happily make the necessary changes (or learn why it's a bad idea).

benanne:
The current approach of having (simple) nonlinearities be part of a layer, with the rectifier as the default, is mainly because that's the most common use case. It's one of those issues where I think pragmatism trumps 'cleanliness' of the design. And it also makes it a bit easier to use custom nonlinearities. I like the 'sensible defaults' approach and I think rectification is the most sensible default thing to do (at least for now).

[...]

PR #243, opened 2015-05-14
Make lasagne.utils.floatX support scalars and avoid copies when possible

f0k:
This changes `lasagne.utils.floatX` to support scalars (see https://github.com/Lasagne/Lasagne/issues/133#issuecomment-102005873), and to not copy the input if it's already in the correct dtype.

### DIFF
--- a/lasagne/utils.py
+++ b/lasagne/utils.py
@@ -5,19 +5,20 @@ import theano.tensor as T
 
 
 def floatX(arr):
-    """Converts numpy array to one with the correct dtype.
+    """Converts data to a numpy array of dtype ``theano.config.floatX``.
 
     Parameters
     ----------
-    arr : numpy array
-        The array to be converted.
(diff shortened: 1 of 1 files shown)


benanne:
Maybe we should add a test or two to make sure we support scalar parameters in general? Seems like an easy thing to break inadvertently :) But we can do that in a separate PR maybe. I'm okay with merging this once Travis passes.

f0k:
> Maybe we should add a test or two to make sure we support scalar parameters in general? Seems like an easy thing to break inadvertently :)

The easiest (albeit a bit indirect) way would be to add a layer to Lasagne that uses scalar parameters, and have tests for that (as well as include that layer in tests of the various helper functions). We could also add direct tests for `lasagne.utils.floatX` et al., but I'm not sure how helpful that is.

In any case, I'll merge this one already.

PR #244, opened 2015-05-14
Handle scalars in count_params

instagibbs:
Simply returns 1 for '()' shapes.

### DIFF
--- a/lasagne/layers/helper.py
+++ b/lasagne/layers/helper.py
@@ -251,7 +251,7 @@ def count_params(layer):
     """
     params = get_all_params(layer)
     shapes = [p.get_value().shape for p in params]
-    counts = [np.prod(shape) for shape in shapes]
+    counts = [np.prod(shape) if shape != () else 1 for shape in shapes]
     return sum(counts)
 
 


lasagne/layers/helper.py
```diff
@@ -251,7 +251,7 @@ def count_params(layer):
     """
     params = get_all_params(layer)
     shapes = [p.get_value().shape for p in params]
-    counts = [np.prod(shape) for shape in shapes]
+    counts = [np.prod(shape) if shape != () else 1 for shape in shapes]
```
benanne: Are you sure this is necessary? I get:
```python
>>> import numpy as np
>>> np.prod(())
1.0
```

instagibbs: You're right. On that note it brings up another question. 

A blank numpy array is also computed as 1.

benanne: That's the behaviour I would expect though. What question does it bring up? :)

instagibbs: 0 parameters being counted as 1?

benanne: No, it's the shape that has 0 elements, not the parameter array itself.

instagibbs: You know what? I need more coffee. I was writing the wrong thing in my command line.

benanne: Haha, don't we all :)

PR #246, opened 2015-05-14
Clearer ReshapeLayer docstring (shape specification and usage example)

f0k:
As discussed in https://github.com/skaae/nntools/commit/0c8386d27cfc33f95ef22e39e11e6b5efff62b0d#commitcomment-11193981, this changes the `ReshapeLayer` docstring to be closer to my original version, but with reordered explanations and a bullet point list to make it easier to follow. Additionally, the second and third usage example are merged into a single one that shows all features without leading people into thinking it could be used for transposing data.

(What's the way to check if the bullet point markup is correct?)

### DIFF
--- a/lasagne/layers/shape.py
+++ b/lasagne/layers/shape.py
@@ -22,7 +22,7 @@ class FlattenLayer(Layer):
     Flatten all but the first dimension.
 
     See Also
-    ----------
+    --------
     flatten  : Shortcut
     """
     def get_output_shape_for(self, input_shape):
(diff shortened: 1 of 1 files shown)


benanne:
I usually just build the docs with `make html` and have a look :)

benanne:
lgtm by the way, feel free to merge unless you want to check the markup first, but I'm sure it'll be fine. Still plenty of documentation PRs to follow I'm guessing, we can always fix it later if necessary :)

f0k:
Fixed mine and @skaae's markup (note: to show up in Sphinx, section names must be "Examples" and "Notes", not "Usage" and "Note"). Merging now as the changes were outside of any code or doctests.

PR #247, opened 2015-05-14
[WIP] PReLULayer

instagibbs:
I have not written any tests yet, but would like to get this up to get some feedback.

I'm not sure if the dimensionality checking in get_output_for is necessary, but I'm using it as a proxy for "does this have channels or is it flattened input". 

Still not 100% this is right, as in my mnist testing the alpha value just keeps increasing over time.

Also, I'm not sure of etiquette when it comes to a large number of tiny commits, for example if they should be squashed.

### DIFF
--- a/lasagne/layers/__init__.py
+++ b/lasagne/layers/__init__.py
@@ -9,3 +9,4 @@ from .shape import *
 from .merge import *
 from .normalization import *
 from .prelu import *
+from .embedding import *
--- a/lasagne/layers/__init__.py
+++ b/lasagne/layers/__init__.py
@@ -8,3 +8,4 @@ from .pool import *
 from .shape import *
 from .merge import *
 from .normalization import *
+from .prelu import *
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -91,24 +91,25 @@ class Layer(object):
 
         return result
 
-    def get_output_shape(self):
+    def get_output_shape(self):  # pragma: no cover
         """
         Deprecated. Use `layer.output_shape`.
         """
         import warnings
         warnings.warn("layer.get_output_shape() is deprecated and will be "
                       "removed for the first release of Lasagne. Please use "
-                      "layer.output_shape instead.")
+                      "layer.output_shape instead.", stacklevel=2)
         return self.output_shape
 
-    def get_output(self, input=None, **kwargs):
+    def get_output(self, input=None, **kwargs):  # pragma: no cover
         """
         Deprecated. Use `lasagne.layers.get_output(layer, input, **kwargs)`.
         """
         import warnings
         warnings.warn("layer.get_output(...) is deprecated and will be "
                       "removed for the first release of Lasagne. Please use "
-                      "lasagne.layers.get_output(layer, ...) instead.")
+                      "lasagne.layers.get_output(layer, ...) instead.",
+                      stacklevel=2)
         from .helper import get_output
         return get_output(self, input, **kwargs)
 
@@ -233,11 +234,12 @@ class Layer(object):
 
         return param
 
-    def get_bias_params(self):
+    def get_bias_params(self):  # pragma: no cover
         import warnings
         warnings.warn("layer.get_bias_params() is deprecated and will be "
                       "removed for the first release of Lasagne. Please use "
-                      "layer.get_params(regularizable=False) instead.")
+                      "layer.get_params(regularizable=False) instead.",
+                      stacklevel=2)
         return self.get_params(regularizable=False)
 
 
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -239,7 +239,7 @@ class Layer(object):
 
             return theano.shared(utils.floatX(arr), name=name)
 
-        elif np.issubdtype(param, np.float) or np.issubdtype(param, np.int):
+        elif isinstance(param, (int, long, float, np.float16, np.float32, np.float64)):
             return theano.shared(utils.floatX(param), name=name)
 
         else:
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -200,7 +200,7 @@ class Layer(object):
             if param.shape != shape:
                 raise RuntimeError("parameter array has shape %s, should be "
                                    "%s" % (param.shape, shape))
-            return theano.shared(param, name=name)
+            return theano.shared(utils.floatX(param), name=name)
 
         elif hasattr(param, '__call__'):
             arr = param(shape)
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -212,7 +212,7 @@ class Layer(object):
             return theano.shared(utils.floatX(arr), name=name)
 
         elif isinstance(param, (int, long, float, np.float16, np.float32, np.float64)):
-            return theano.shared(utils.floatX(np.float32(param)), name=name)
+            return theano.shared(utils.floatX(param), name=name)
 
         else:
             raise RuntimeError("cannot initialize parameters: 'param' is not "
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -244,7 +244,7 @@ class Layer(object):
 
         else:
             raise RuntimeError("cannot initialize parameters: 'param' is not "
-                               "a numpy array, a Theano shared variable, a numpy scalar or a "
+                               "a numpy array, a Theano shared variable, a scalar, or a "
                                "callable")
 
 
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -240,7 +240,7 @@ class Layer(object):
             return theano.shared(utils.floatX(arr), name=name)
 
         elif isinstance(param, (int, long, float, np.float16, np.float32, np.float64)):
-            return theano.shared(utils.floatX(param), name=name)
+            return theano.shared(utils.floatX(np.float32(param)), name=name)
 
         else:
             raise RuntimeError("cannot initialize parameters: 'param' is not "
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -239,8 +239,8 @@ class Layer(object):
 
             return theano.shared(utils.floatX(arr), name=name)
 
-        elif np.issubdtype(param, np.float):
-            return theano.shared(param, name=name)
+        elif np.issubdtype(param, np.float) or np.issubdtype(param, np.int):
+            return theano.shared(utils.floatX(param), name=name)
 
         else:
             raise RuntimeError("cannot initialize parameters: 'param' is not "
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -1,6 +1,4 @@
-import numpy as np
-
-import theano
+from collections import OrderedDict
 
 from .. import utils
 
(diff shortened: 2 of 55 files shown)


benanne:
It would definitely be nice to squash this eventually :) I'll have a look at the code itself later today.

lasagne/layers/base.py
```diff
@@ -239,9 +239,12 @@ def create_param(self, param, shape, name=None):
 
             return theano.shared(utils.floatX(arr), name=name)
 
+        elif isinstance(param, (int, long, float, np.float16, np.float32, np.float64)):
```
benanne: This should be unnecessary, it will be taken care of in a separate PR without having to enumerate all possible scalar types :) See #243 and https://github.com/Lasagne/Lasagne/pull/228#discussion_r30338737

lasagne/layers/prelu.py
```diff
@@ -0,0 +1,60 @@
...
+    A layer used to implement Parametric Rectified Linear Units (PReLU).
+    This layer simply applies the parametric nonlinearity to the previous
+    layer's output. 
+
+    Parameters
+    -----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape.
+
+    alpha : double or None
```
benanne: I think `float` is more appropriate as a type name. But we haven't really standardized this across the docs yet. We probably should.

lasagne/layers/prelu.py
```diff
@@ -0,0 +1,60 @@
...
+        The initial "leakiness" of the rectifier which is then learned.
+        If argument not provided, initial leakiness will be set to 0.25.
+
+    unique_chan_param : boolean or None
```
benanne: Why should this be used only in the case of a convolutional layer? I see no objections to using this in combination with a dense layer (where each unit is just considered a separate channel). It will probably overfit completely but technically there's nothing wrong with it :) Or am I missing something? Also I'm not too sure about `unique_chan_param` as a name. I made a throwaway implementation of this a while back and I think I called it `tied`, which may be a little too concise :) ('tied' as in the parameters are tied across all channels, or not). Anybody have a better idea?

lasagne/layers/prelu.py
```diff
@@ -0,0 +1,60 @@
...
+        When true, this argument creates a vector of alphas rather than a
+        single theano scalar. Each scalar is applied to a separate channel
+        of the previous layer's output. Should only be used if the 
+        previous layer is a convolution. If argument not provided, it will
+        be set to False.
+
+    Usage
```
benanne: I believe this should be 'Examples' instead of 'Usage' for sphinx to pick it up.

benanne:
It might be a good idea to wait until #228 is merged and then rebase this, because some changes to the parameter creation API are coming soon.

lasagne/layers/prelu.py
JeffreyDF: In my own throwaway implementation I used "untie_channels", a bit like the "untie_biases" argument for the conv layers.

instagibbs: @benanne That would certainly simplify the code, and if that's fine... that's fine. I just generally write towards how *I'd* use it. Agreed on name. I'll probably go with @JeffreyDF 's idea.

lasagne/layers/base.py
instagibbs: For now I'll leave it in and remove it when that logic gets added.

lasagne/layers/prelu.py
f0k: > In my own throwaway implementation I used "untie_channels", a bit like the "untie_biases" argument for the conv layers.

Following that logic, it should be called `untie_alphas`, not `untie_channels`. Also see #249, maybe we should allow specifying exactly which axes to share alphas/biases over.

instagibbs: For the tuple case are we thinking of (False, True, True, False) if we want
to untie across dim 1 and 2?



Greg

f0k: I'm not sure. For `untie_biases`, it was always just about the spatial dimensions; there was no choice for the channels (or the batch dimension). Here it might be interesting to control this for every dimension (except for the batch dimension). Maybe we should allow giving a list of axes?

lasagne/layers/base.py
```diff
@@ -200,7 +200,7 @@ def create_param(self, param, shape, name=None):
             if param.shape != shape:
                 raise RuntimeError("parameter array has shape %s, should be "
                                    "%s" % (param.shape, shape))
-            return theano.shared(param, name=name)
+            return theano.shared(utils.floatX(param), name=name)
```
instagibbs: Without this it was creating float64s. It's less confusing if they're all casted using utils.floatX function. Obviously this section is changing anyways, but did this for now.

instagibbs:
Revamped the function after fighting with a bug. Layer now takes None, int, or list as argument for untied parameters. It removes duplicates from list, and fails if a non-valid(<0 or >= ndims) dimension is listed.

If there's a nicer short-hand for constructing the dimshuffle mask I'm all ears, but this one seems to work.

lasagne/layers/prelu.py
```diff
@@ -0,0 +1,72 @@
...
+    untie_alphas : int, array or None
+        When passed this argument creates a tensor of alphas rather than a
+        single theano scalar. The argument dictates which dimension(s)
+        the alphas are untied across. 
+
+    Examples
+    -------
+        >>> from lasagne.layers import InputLayer, DenseLayer, PReLULayer
+        >>> l_in = InputLayer((100, 20))
+        >>> l1 = DenseLayer(l_in, num_units=50, nonlinearity=identity)
```
f0k: This is probably going to fail without importing `identity`, but you can just supply `nonlinearity=None` instead.

lasagne/layers/prelu.py
```diff
@@ -0,0 +1,72 @@
...
+            self.alpha_shape = ()
+            self.alpha = self.add_param(utils.floatX(alpha), self.alpha_shape, name="alpha")
+        elif isinstance(untie_alphas, int) and untie_alphas >= 0 and untie_alphas < len(self.input_shape):
+            self.alpha_shape = tuple([self.input_shape[untie_alphas]])
+            self.alpha = self.add_param(utils.floatX(np.zeros(self.alpha_shape))+alpha, self.alpha_shape, name="alpha")
+        elif isinstance(untie_alphas, list) and np.min(untie_alphas) >= 0 and np.max(untie_alphas) < len(self.input_shape):
+            self.alpha_shape = tuple([self.input_shape[x] for x in set(untie_alphas)])
+            self.alpha = self.add_param(utils.floatX(np.zeros(self.alpha_shape))+alpha, self.alpha_shape, name="alpha")
+        else:
+            raise Exception('The untie_alphas parameter is badly formed.')
```
f0k: Looks good in principle, I'd just change a few things:
* test for `is None`, not `== None`
* modify `untie_alphas` and `alpha` first, then set `self.untie_alphas = untie_alphas` and `self.alpha = alpha` in the end
* make `alpha_shape` a local variable since you don't use it later

```python
if untie_alphas is None:
    alpha_shape = ()
else:
    untie_alphas = [untie_alphas] if isinstance(untie_alphas, int) else sorted(set(untie_alphas))
    alpha_shape = [self.input_shape[axis] for axis in untie_alphas]
alpha = np.zeros(self.alpha_shape, dtype=theano.config.floatX) + alpha
self.alpha = self.add_param(alpha, alpha_shape, name='alpha', regularizable=False)
```

lasagne/layers/prelu.py
```diff
@@ -0,0 +1,72 @@
...
+    Parameters
+    -----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape.
+
+    alpha : float or None
+        The initial "leakiness" of the rectifier which is then learned.
+        If argument not provided, initial leakiness will be set to 0.25.
+
+    untie_alphas : int, array or None
```
f0k: Should be `iterable of int` instead of `array`.

lasagne/layers/prelu.py
```diff
@@ -0,0 +1,72 @@
...
+            self.alpha_shape = tuple([self.input_shape[untie_alphas]])
+            self.alpha = self.add_param(utils.floatX(np.zeros(self.alpha_shape))+alpha, self.alpha_shape, name="alpha")
+        elif isinstance(untie_alphas, list) and np.min(untie_alphas) >= 0 and np.max(untie_alphas) < len(self.input_shape):
+            self.alpha_shape = tuple([self.input_shape[x] for x in set(untie_alphas)])
+            self.alpha = self.add_param(utils.floatX(np.zeros(self.alpha_shape))+alpha, self.alpha_shape, name="alpha")
+        else:
+            raise Exception('The untie_alphas parameter is badly formed.')
+
+    def get_params(self):
+        return [self.alpha]
```
f0k: Must be removed: https://groups.google.com/forum/#!topic/lasagne-users/nwfNFnC1ibs

lasagne/layers/prelu.py
```diff
@@ -0,0 +1,72 @@
...
+            untie_dim[self.untie_alphas] = 0
+            untie_dim = tuple(untie_dim)
+            return nonlinearities.LeakyRectify(self.alpha.dimshuffle(untie_dim))(input)
+        elif isinstance(self.untie_alphas, list):
+            for i, dim in enumerate(self.untie_alphas):
+                untie_dim[dim] = i
+            untie_dim = tuple(untie_dim)
+            return nonlinearities.LeakyRectify(self.alpha.dimshuffle(untie_dim))(input)
+        else:
+            return nonlinearities.LeakyRectify(self.alpha)(input)
```
f0k: Just a little shorter with my other change proposed above:
```python
if self.untie_alphas is None:
    alpha = self.alpha
else:
    pattern = ['x'] * input.ndim
    for idx, dim in enumerate(self.untie_alphas):
        pattern[dim] = idx
    alpha = self.alpha.dimshuffle(untie_dim)
return nonlinearities.LeakyRectify(self.alpha)(input)
```

lasagne/layers/__init__.py
```diff
@@ -8,4 +8,5 @@
 from .shape import *
 from .merge import *
 from .normalization import *
+from .prelu import *
```
f0k: I don't think this should become its own submodule of just a single layer. We could have a submodule for parameterized (or otherwise complex) nonlinearities, though. What would be a concise name?

ebenolson: It might make sense to group this together with `NonlinearityLayer`... not sure that really belongs in `dense.py` as it is currently.

benanne: I agree, that would make more sense. No idea what to call it though. `lasagne.layers.nonlinearities` would be easily confusable with `lasagne.nonlinearities`, so that's probably a bad idea.

ebenolson: Yeah, that would be confusing. What about `layers.elemwise`, `transfer` or `activation`?

benanne: `elemwise` works for me.

instagibbs: I like it, but perhaps elemwiselayer?

benanne: That doesn't make too much sense imo, the full 'path' would be `lasagne.layers.elemwise`, no need to tack another `layer` at the end :) Also it doesn't make sense to call the module `elemwiselayer` if we then proceed to put multiple `Layer` subclasses in it.

instagibbs: True if in the right folder it makes sense. Just wanted it to be very clearly different than nonlinearity.Then I'm agreed.

f0k: `elemwise` was the first that came to my mind as well, but that could be confusing with the `ElemwiseSumLayer` in `layers.merge`, and a `NonlinearityLayer` with `softmax` is not exactly performing an element-wise operation. Besides, we have other layers performing element-wise operations in `layers.noise` and `layers.normalization`.

`transfer` and `activation` would be better in this respect, the only trouble is that we never called it like that anywhere else in the code. (But that was exactly the point, otherwise we'd call it `layers.nonlinearities`).)

What other layers would be added to this module in future? We probably don't want to rename it later, nor introduce yet another submodule just because the name doesn't fit.

benanne: On the other hand, moving things around is not such a big deal because the user-facing API has all of them under `lasagne.layers` anyway (except for those in  `cuda_convnet`, `dnn`, `corrmm` because they have special dependencies).

f0k:
Alternative implementation merged in #446.

PR #248, opened 2015-05-14
Adam updates

skaae:


lasagne/updates.py
```diff
@@ -53,7 +54,7 @@
 
 import theano
 import theano.tensor as T
-
+import utils
```
benanne: Travis is failling only for Python 3, and it looks like it's because this should be a relative import. I think it should read `from . import utils` (not tested).

lasagne/updates.py
```diff
@@ -510,6 +512,52 @@ def adadelta(loss_or_grads, params, learning_rate=1.0, rho=0.95, epsilon=1e-6):
...
+    Parameters
+    ----------
+    learning_rate : float
+        Learning rate
+    beta_1 : float
+        Exponential decay rate for the first moment estimates.
+    beta_2 : float
+        Exponential decay rate for the second moment estimates.
+    epsilon : float
+        Constanst for numerical stability.
```
benanne: Constant*

lasagne/updates.py
```diff
@@ -510,6 +512,52 @@ def adadelta(loss_or_grads, params, learning_rate=1.0, rho=0.95, epsilon=1e-6):
...
+    Adam: A Method for Stochastic Optimization*,
+    http://arxiv.org/abs/1412.6980
+
+    Notes
+    ----------
+    The implementation does not include lambda as it is only used to
+    prove convergence of the algorithm.
+
+    Parameters
+    ----------
```
benanne: Could you also add `loss_or_grads` and `all_params` to the parameter list in the docs, for completeness' sake? I guess they can be copied from another function.

benanne:
Looks good! I went through the implementation with the paper, looks like everything matches so that's good (they changed the meaning of some parameters in the most recent version so I wanted to make sure).

It seems a bit odd to have the references and notes before the parameters in the docstring, maybe those should be swapped around (although sphinx may do this by itself anyway, I don't know).

lasagne/updates.py
skaae: Yes i forgot that, I'll add them and swap notes and parameters

lasagne/updates.py
```diff
@@ -510,6 +512,56 @@ def adadelta(loss_or_grads, params, learning_rate=1.0, rho=0.95, epsilon=1e-6):
...
+    References
+    ----------
+    [1] Diederik Kingma, Jimmy Ba,
+    Adam: A Method for Stochastic Optimization*,
+    http://arxiv.org/abs/1412.6980
+
+    Notes
+    ----------
+    The implementation does not include lambda as it is only used to
+    prove convergence of the algorithm.
```
f0k: As that's not explained in the paper, I'd add: "... of the algorithm (personal communication with the authors of [1]_)."

lasagne/updates.py
```diff
@@ -510,6 +512,56 @@ def adadelta(loss_or_grads, params, learning_rate=1.0, rho=0.95, epsilon=1e-6):
...
+        Exponential decay rate for the first moment estimates.
+    beta_2 : float
+        Exponential decay rate for the second moment estimates.
+    epsilon : float
+        Constant for numerical stability.
+
+    References
+    ----------
+    [1] Diederik Kingma, Jimmy Ba,
+    Adam: A Method for Stochastic Optimization*,
```
f0k: What's the asterisk?

skaae: dunno  :) Will remove and include your other suggestion

lasagne/updates.py
```diff
@@ -510,6 +512,56 @@ def adadelta(loss_or_grads, params, learning_rate=1.0, rho=0.95, epsilon=1e-6):
...
+    http://arxiv.org/abs/1412.6980
+
+    Notes
+    ----------
+    The implementation does not include lambda as it is only used to
+    prove convergence of the algorithm.
+    """
+    all_grads = get_or_compute_grads(loss_or_grads, params)
+    t_prev = theano.shared(utils.floatX(0.))
```
f0k: Cool, #243 put to use within less than 24 hours :)

skaae:
added @f0k suggestions.

f0k:
Thank you! Just one last thing: The docstring should have a "Returns" section after the "Parameters" section (copied from `sgd`, for example).

f0k:
Thanks, looks ready to merge now. The very first change will conflict with #251. I'm fine with merging yours first and rebasing my PR afterwards. @benanne, shall we merge?

benanne:
Fine with me, go for it!

PR #250, opened 2015-05-15
Documentation: Sphinx format to numpydoc format (rebased)

benanne:
This supersedes #204.


benanne:
I will merge this once Travis passes so we can get a move on :) I realise that it could be valuable to have the result of the rebase reviewed as well, but since it only concerns docstrings any mistakes can easily be fixed later and won't have a huge impact for now.

f0k:
Travis passes, merging.

PR #251, opened 2015-05-15
Docstring markup fixes: Section names, math and reference syntax

f0k:
Fixing several errors in the docstring markup. @skaae, please have a quick look at the diff to see what's important for future contributions!

Useful references:
* https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt
* https://github.com/numpy/numpy/blob/master/numpy/fft/info.py
* http://docs.scipy.org/doc/numpy/reference/routines.fft.html

Something's still wrong with how the references render in Sphinx, as well as the parameter descriptions (I think they are rendered as block quotes, with a grey bar on the left and a larger font size).

/edit: The most important gotchas: Section names need to follow the first document, reference definitions need to follow the first document, math needs to have backslashes escaped in docstrings.

### DIFF
--- a/docs/modules/init.rst
+++ b/docs/modules/init.rst
@@ -6,13 +6,31 @@
 .. autoclass:: Initializer
    :members:
 
-.. autoclass:: Normal
-   :members:
-
 .. autoclass:: Constant
    :members:
 
+.. autoclass:: Normal
+   :members:
+
 .. autoclass:: Uniform
+   :members:
+
+.. autoclass:: Glorot
+   :members:
+
+.. autoclass:: GlorotNormal
+   :members:
+
+.. autoclass:: GlorotUniform
+   :members:
+
+.. autoclass:: He
+   :members:
+
+.. autoclass:: HeNormal
+   :members:
+
+.. autoclass:: HeUniform
    :members:
 
 .. autoclass:: Orthogonal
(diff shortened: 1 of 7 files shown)


benanne:
The parameter descriptions are indeed rendered as block quotes. This seems to be intentional (look at the numpy documentation for example). In #203 I have a TODO item to create a custom stylesheet to make everything look a bit nicer. This seems to be how other libraries tackle this problem as well. I will probably base it on the custom CSS from scikit-bio, they made it look okay-ish, I think.

I have to say I've found it quite frustrating to work with Sphinx so far, it seems to be very difficult to make it do what you want it to do, the syntax is often cumbersome and making it look good is a challenge. I'm beginning to think we should have gone with another option :) But at least we're getting  close to where we want to be in terms of documentation now.

dnouri:
> I have to say I've found it quite frustrating to work with Sphinx so 
> far, it seems to be very difficult to make it do what you want it to 
> do, the syntax is often cumbersome and making it look good is a 
> challenge. I'm beginning to think we should have gone with another 
> option :) But at least we're getting close to where we want to be in 
> terms of documentation now.
> 
I'm curious: how many problem stem from using numpydoc versus vanilla 
Sphinx?

benanne:
Most of it has been bugging me since before we made the switch to numpydoc. If my memory doesn't betray me, anyway :)

benanne:
An 80-character line, @f0k! How dare you! Travis is pretty ruthless :)

f0k:
Woops! Used a different editor with the margin set to 80 and didn't run pytest myself... I've overwritten the offending commit.

PR #252, opened 2015-05-17
lasagne.layers.cuda_convnet documentation

benanne:
I've only done the convolution layer so far. It was mostly copy paste from the other implementations, but the `dimshuffle` and `partial_sum` arguments required some extra work. The notes section is also quite beefy because all the limitations of the cuda-convnet convolution implementation need to be explained.

I'll do the pooling layer and the bc01/c01b shuffle layers next.

### DIFF
--- a/lasagne/layers/cuda_convnet.py
+++ b/lasagne/layers/cuda_convnet.py
@@ -46,21 +46,21 @@ class Conv2DCCLayer(CCLayer):
     Parameters
     ----------
     incoming : a :class:`Layer` instance or a tuple
-        The layer feeding into this layer, or the expected input shape. The
-        output of this layer should be a 4D tensor, with shape
-        ``(batch_size, num_input_channels, input_height, input_width)``.
+        The layer feeding into this layer, or the expected input shape. This
+        layer expects a 4D tensor as its input, with shape
+        ``(batch_size, num_input_channels, input_rows, input_columns)``.
         If automatic dimshuffling is disabled (see notes), the shape should be
-        ``(num_input_channels, input_height, input_width, batch_size)``
+        ``(num_input_channels, input_rows, input_columns, batch_size)``
         instead (c01b axis order).
 
     num_filters : int
         The number of learnable convolutional filters this layer has.
 
-    filter_size : int or tuple of int
+    filter_size : int or iterable
         An integer or a 2-element tuple specifying the size of the filters.
         This layer does not support non-square filters.
 
-    stride : int or tuple of int
+    stride : int or iterable
         An integer or a 2-element tuple specifying the stride of the
         convolution operation. This layer does not support using different
         strides along both axes.
@@ -86,16 +86,16 @@ class Conv2DCCLayer(CCLayer):
         which is shared across all positions in this channel. As a result, the
         `b` attribute will be a vector (1D).
 
-        If True, the layer will have separate bias parameters for each
+        If ``True``, the layer will have separate bias parameters for each
         position in each channel. As a result, the `b` attribute will be a
         3D tensor.
 
     W : Theano shared variable, numpy array or callable
         An initializer for the weights of the layer. This should initialize the
         layer weights to a 4D array with shape
-        ``(num_filters, num_input_channels, filter_height, filter_width)``.
+        ``(num_filters, num_input_channels, filter_rows, filter_columns)``.
          If automatic dimshuffling is disabled (see notes), the shape should be
-        ``(num_input_channels, input_height, input_width, num_filters)``
+        ``(num_input_channels, input_rows, input_columns, num_filters)``
         instead (c01b axis order).
         See :func:`lasagne.utils.create_param` for more information.
 
@@ -104,14 +104,14 @@ class Conv2DCCLayer(CCLayer):
         layer will have no biases. This should initialize the layer biases to
         a 1D array with shape ``(num_filters,)`` if `untied_biases` is set to
         ``False``. If it is set to ``True``, its shape should be
-        ``(num_filters, input_height, input_width)`` instead.
+        ``(num_filters, input_rows, input_columns)`` instead.
         See :func:`lasagne.utils.create_param` for more information.
 
     nonlinearity : callable or None
         The nonlinearity that is applied to the layer activations. If None
         is provided, the layer will be linear.
 
-    pad : int, tuple of int or None
+    pad : int, iterable or None
         An integer or a 2-element tuple specifying the amount of zero-padding
         on each side. This may also be ``None``, in which case the correct
         amount of padding will be inferred from the specified ``border_mode``.
@@ -142,7 +142,7 @@ class Conv2DCCLayer(CCLayer):
     partial_sum : int or None (default: 1)
         This value tunes the trade-off between memory usage and performance.
         You can specify any positive integer that is a divisor of the output
-        feature map size (i.e. output height times output width). Higher
+        feature map size (i.e. output rows times output columns). Higher
         values decrease memory usage, but also performance. Specifying 0 or
         ``None`` means the highest possible value will be used. The Lasagne
         default of ``1`` gives the best performance, but also the highest
@@ -180,7 +180,7 @@ class Conv2DCCLayer(CCLayer):
     * the number of input channels must be even, or less than or equal to
       3.
     * if the gradient w.r.t. the input is to be computed, the number of
-      channels should be divisible by 4.
+      channels must be divisible by 4.
     * performance is optimal when the batch size is a multiple of 128 (but
       other batch sizes are supported).
     * this layer only works on the GPU.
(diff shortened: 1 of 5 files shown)


benanne:
I think this is ready for review. I also fixed up the `dense` docs a bit.

lasagne/layers/cuda_convnet.py
```diff
@@ -29,17 +29,177 @@
...
+    Performs a 2D convolution on its input and optionally adds a bias and
+    applies an elementwise nonlinearity.  This is an alternative implementation
+    which uses the cuda-convnet wrappers from pylearn2:
+    ``pylearn2.sandbox.cuda_convnet.filter_acts.FilterActs``.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape. The
+        output of this layer should be a 4D tensor, with shape
```
f0k: "The layer feeding into *this layer* [...] The output of *this layer* should be [...]"
This is a bit confusing, because "this layer" means different things in the two sentences. What about:
"This layer expects a 4D tensor, with shape ..."

lasagne/layers/cuda_convnet.py
```diff
@@ -29,17 +29,177 @@
...
+        ``(batch_size, num_input_channels, input_height, input_width)``.
+        If automatic dimshuffling is disabled (see notes), the shape should be
+        ``(num_input_channels, input_height, input_width, batch_size)``
```
f0k: Elsewhere we settled for `input_rows` and `input_columns` to not take a position on what's height and what's width.

lasagne/layers/cuda_convnet.py
```diff
@@ -29,17 +29,177 @@
...
+        same as the input length (for both odd and even filter sizes).
+
+    untie_biases : bool, default False
+        If ``False``, the layer will have a bias parameter for each channel,
+        which is shared across all positions in this channel. As a result, the
+        `b` attribute will be a vector (1D).
+
+        If True, the layer will have separate bias parameters for each
+        position in each channel. As a result, the `b` attribute will be a
+        3D tensor.
```
f0k: Minor thing: "True" should have double backticks, too.

lasagne/layers/cuda_convnet.py
```diff
@@ -29,17 +29,177 @@
...
+
+    W : Theano shared variable, numpy array or callable
+        An initializer for the weights of the layer. This should initialize the
+        layer weights to a 4D array with shape
+        ``(num_filters, num_input_channels, filter_height, filter_width)``.
+         If automatic dimshuffling is disabled (see notes), the shape should be
+        ``(num_input_channels, input_height, input_width, num_filters)``
```
f0k: Again, `height` and `width` should probably be `rows` and `columns`.

lasagne/layers/cuda_convnet.py
```diff
@@ -29,17 +29,177 @@
...
+    The cuda-convnet convolution implementation has several limitations:
+
+    * only square filters are supported.
+    * only identical strides in the horizontal and vertical direction are
+      supported.
+    * the number of filters must be a multiple of 16.
+    * the number of input channels must be even, or less than or equal to
+      3.
+    * if the gradient w.r.t. the input is to be computed, the number of
+      channels should be divisible by 4.
```
f0k: should be or must be?

lasagne/layers/cuda_convnet.py
```diff
@@ -29,17 +29,177 @@
...
+    with other layers that assume bc01 axis order.
+
+    However, these additional dimshuffle operations may sometimes negatively
+    affect performance. For this reason, it is possible to disable them by
+    setting ``dimshuffle=False``.
+
+    In this case, the user is expected to manually ensure that the input and
+    parameters have the correct axis order. :class:`ShuffleBC01ToC01BLayer` and
+    :class:`ShuffleC01BToBC01Layer` can be used to convert between bc01 and
+    c01b axis order.
```
f0k: It's a bit confusing that these are three separate paragraphs although they're all about the same issue. Does it look bad in Sphinx (worse than usual, I mean ;)) if you remove the empty lines in between?

lasagne/layers/cuda_convnet.py
```diff
@@ -196,7 +356,86 @@ def get_output_for(self, input, **kwargs):
...
+    -----
+    The cuda-convnet max-pooling implementation has several limitations:
+
+    * only square pooling regions are supported.
+    * only identical strides in the horizontal and vertical direction are
+      supported.
+    * only square inputs are supported. (This limitation does not exist for
+      the convolution implementation.)
+    * partial pooling regions are always ignored (``ignore_border`` is forced
+      to ``False``).
```
f0k: That's contradictory, isn't it? Do you mean "True" instead of "False", or "included" instead of "ignored"?

lasagne/layers/cuda_convnet.py
```diff
@@ -196,7 +356,86 @@ def get_output_for(self, input, **kwargs):
...
+    (or over axis 1 and 2 if ``dimshuffle=False``, see notes). This is an
+    alternative implementation which uses the cuda-convnet wrappers from
+    pylearn2: ``pylearn2.sandbox.cuda_convnet.pool.MaxPool``.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or tuple
+        The layer feeding into this layer, or the expected input shape.
+
+    pool_size : integer or iterable
```
f0k: Is it true that we allow any iterable here, but require a tuple for the convolutional layer?

lasagne/layers/cuda_convnet.py
```diff
@@ -196,7 +356,86 @@ def get_output_for(self, input, **kwargs):
...
+        using different strides along both axes.
+
+    pad : integer or iterable (default: 0)
+        This implementation does not support custom padding, so this argument
+        should always be set to ``0``. It exists only to make sure the
+        interface is compatible with :class:`lasagne.layers.MaxPool2DLayer`.
+
+    ignore_border : bool (default: False)
+        This implementation always ignores partial pooling regions, so this
+        argument should always be set to False. It exists only to make sure the
```
f0k: should or must?

lasagne/layers/cuda_convnet.py
```diff
@@ -196,7 +356,86 @@ def get_output_for(self, input, **kwargs):
...
+        two elements. This layer does not support non-square pooling regions.
+
+    stride : integer, iterable or ``None``
+        The strides between sucessive pooling regions in each dimension.
+        If ``None`` then ``stride = pool_size``. This layer does not support
+        using different strides along both axes.
+
+    pad : integer or iterable (default: 0)
+        This implementation does not support custom padding, so this argument
+        should always be set to ``0``. It exists only to make sure the
```
f0k: should or must?

lasagne/layers/cuda_convnet.py
benanne: You're right, we settled on rows / columns a while back but it looks like height / width snuck into a lot of docstrings through the library. I'll clean them all up while I'm at it (this was copy-pasted from `Conv2DLayer` anyway).

lasagne/layers/cuda_convnet.py
benanne: I think they mean the same thing in this context, but I'll change it to 'must' everywhere for consistency.

lasagne/layers/cuda_convnet.py
benanne: In my experience a lot of people find axis-juggling tricky to understand, so I think it's good to explain this in detail. The description of `dimshuffle` does not really explain why it exists in the first place, which is the main reason I added a note about it as well.

I merged all of it into one paragraph, seems to look fine.

lasagne/layers/cuda_convnet.py
benanne: No :) There is just very little consistency in how the various data types are named in docstrings. We should probably figure this out in a separate issue and decide on a 'standard'.

benanne:
I've tried to address all your comments. I'm not sure if I got all the "should or must" ones since it's hard to tell which line you are talking about in the GitHub interface.

f0k:
> I'm not sure if I got all the "should or must" ones since it's hard to tell which line you are talking about in the GitHub interface.

All right, I'll have a look! We should just generally try to avoid saying "should" if we actually mean "don't do this or we will inevitably raise an exception" :)

PS: Could you please remove the @ from the commit message? It nicely links to my profile, but it also gives me an email notification whenever anybody merges that commit into their fork... at least I *think* it does, I had some notifications like this coming up for other projects in the past. Thanks!

benanne:
Whoops, wasn't aware of that! I'll get rid of it.

lasagne/layers/cuda_convnet.py
f0k: > I merged all of it into one paragraph

No, you didn't :) Just for the convolution. Can you do the same here? Cheers!

f0k:
> Whoops, wasn't aware of that!

It's not really obvious why this should be a useful feature in commit messages anyway. And it's possible that additional notifications are only sent out if somebody does a non-fast-forward merge so the commit gets a different SHA. That can still get annoying.

As for the PR, except for one last comment, I'm fine with merging! Good job!

benanne:
Sweet :) Merging!

PR #253, opened 2015-05-18
step_scaling update rule + test

skaae:
Update rule that scales based on the combined norm.

Moved from https://github.com/craffel/nntools/pull/29

### DIFF
--- a/lasagne/updates.py
+++ b/lasagne/updates.py
@@ -20,13 +20,15 @@ Two functions can be used to further modify the updates to include momentum:
 * :func:`apply_momentum()`
 * :func:`apply_nesterov_momentum()`
 
-Finally, we provide a helper function to constrain the norm of a
-tensor variable:
+Finally, we provide two helper functions to constrain the norm of tensors:
 
 * :func:`norm_constraint()`
+* :func:`total_norm_constraint()`
 
-This can be used to constrain the norm of parameters (as an alternative
-to weight decay), or for a form of gradient clipping.
+:func:`norm_constraint()` can be used to constrain the norm of parameters
+(as an alternative to weight decay), or for a form of gradient clipping.
+:func:`total_norm_constraint() constrain the total norm of a list of tensors.
+This is often used when training recurrent neural networks.`
 
 Examples
 --------
@@ -67,6 +69,7 @@ __all__ = [
     "adadelta",
     "adam",
     "norm_constraint",
+    "total_norm_constraint"
 ]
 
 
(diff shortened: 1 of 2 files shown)


lasagne/updates.py
```diff
@@ -567,6 +575,48 @@ def adam(loss_or_grads, params, learning_rate=0.001, beta1=0.9,
...
+
+    """
+    if not isinstance(tensor_vars, (list, tuple)):
+        tensor_vars = [tensor_vars]
+
+    sq = 0
+    for tensor in tensor_vars:
+        sq += (tensor**2).sum()
+
+    norm = T.sqrt(sq)
```
f0k: better: `norm = T.sqrt(sum(T.sum(T.sqr(tensor)) for tensor in tensor_vars))`

lasagne/updates.py
```diff
@@ -567,6 +575,48 @@ def adam(loss_or_grads, params, learning_rate=0.001, beta1=0.9,
...
+    threshold = theano.shared(utils.floatX(threshold))
```
f0k: Creating a shared variable is a bit overkill. Just cast it to the correct dtype as in `norm_constraint`.

lasagne/updates.py
skaae: I think mine is more readable

lasagne/updates.py
skaae: Yes. I'll change that.

lasagne/updates.py
```diff
@@ -567,6 +575,48 @@ def adam(loss_or_grads, params, learning_rate=0.001, beta1=0.9,
...
+    scale = threshold / norm
+    multiplier = T.switch(norm < threshold, 1.0, scale)
+    tensor_vars_scaled = [step*multiplier for step in tensor_vars]
```
f0k: I'm wondering if this is better or worse than what we have in `norm_constraint`. It will fail for a norm of zero (that's bad), and it uses `T.switch` instead of `T.clip` for the case distinction (not sure which one is faster).

skaae: Would it help if added epsilon like in norm_constraint?

f0k:
Looks good in general, but it seems this does exactly the same as `norm_constraint`, but for a list of tensors instead of a single tensor. We shouldn't have two functions with different names and different interfaces for that, this will only confuse users. How do you feel about just extending `norm_constraint` to optionally accept a list of tensors instead of a single tensor? And have it optionally return the quantities you need for monitoring?

skaae:
Its fine by me. I'll modify `norm_constraint` instead. I would like to be able to print the combined norms during training.

benanne:
I'm just a little worried about making the code for `norm_constraint` too complicated by adding all these extra arguments to cater to different use cases. But I guess we can give it a try and see.

skaae:
What's the status on this?

benanne:
I don't know. I guess let's try to merge it with `norm_constraint`, there is a lot of overlap. Then we'll see if it's too complicated or not.

Maybe `norm_constraint` and `step_scaling` could be more specialized functions that both use the same underlying function? Something like that? I don't know :) `step_scaling` seems to me like it might not be the best name for this either, since it refers to a specific use case of this operation and not really the operation itself.

skaae:
Yes. Step scaling is maybe not the best name, but it might hint the user on the purpose. For me it is not clear that you could use norm_constraint on the gradients, where as step_scaling (or somthing similar) indicate that you can. 
I prefer to keep them seperate because they serve different purposes. They can ofcourse use the same underlying code.

f0k:
I'd prefer to merge them and document use cases in examples rather than giving two different names for the different use cases.

skaae:
I dont fell for strongly for any of them. So if you prefer to merge I can do that and demonstrate the use in the recurrent examples.

f0k:
:+1:

To summarize, what I'd do to `norm_constraint` would be:
* rename `tensor_var` to `var_or_vars` (to make clear it can be a single variable or a list). If it's a list, it should return a list, if it's a single variable, it should return a single variable.
* Change the default `norm_axes` value to `'auto'` (to make clear that this is a mode which automatically chooses the axes to normalize over, depending on the input shape(s))
* Introduce a new `norm_axes` value of `'all'` (which normalizes over all axes)
* Add a keyword argument `return_norms`, defaulting to `False`, which returns the norms if set to `True`. It should return a list of norms if `var_or_vars` is a list, and a single norm otherwise.

Hmm, wait, what is supposed to happen if `var_or_vars` is a list and `norm_axes` is not `'all'`? This wouldn't work, would it? `norm_axes='auto'` should default to "all axes" for this case then. Or maybe we should have a separate `total_norm_constraint` function for the case of constraining the total norm of multiple tensors after all...

skaae:
Using theano.clone seems to work. This also somehow adresses 

https://github.com/craffel/nntools/issues/20
https://github.com/craffel/nntools/pull/25

skaae:
Thanks for the corrections. I'll change to T.clip. I somehow forgot to remove the old code aswell.

f0k:
> I somehow forgot to remove the old code aswell. 

Ah yes. Didn't see that because I was inadvertently just commenting the new commit. I'd prefer moving this *after* `norm_constraint`, while you're at it. That seems like a more natural order.

lasagne/updates.py
```diff
@@ -567,6 +575,123 @@ def adam(loss_or_grads, params, learning_rate=0.001, beta1=0.9,
...
+    ----------
+    .. [1] Sutskever, I., Vinyals, O., & Le, Q. V. (2014): Sequence to sequence
+       learning with neural networks. In Advances in Neural Information
+       Processing Systems (pp. 3104-3112).
+    """
+    if not isinstance(tensor_or_tensor_vars, (list, tuple)):
+        tensor_or_tensor_vars = [tensor_or_tensor_vars]
+
+    norm = T.sqrt(T.sum([T.sum(tensor**2) for tensor in
+                         tensor_or_tensor_vars]))
```
f0k: Shouldn't the outer one just be `sum`, with the square brackets removed? I wonder if it's forming a tensor of the inner sums otherwise.

skaae: It also passes the tests. I cant find the correct name for theano scalar so its called that unless you have better name?

lasagne/updates.py
```diff
@@ -20,6 +20,13 @@
 * :func:`apply_momentum()`
 * :func:`apply_nesterov_momentum()`
 
+We additionally provide a function to scale the norm of a list of tensors:
+
+* :func:`step_scaling()`
+
+Step scaling can be used to scale the gradient steps when training recurrent
+neural networks.
+
```
f0k: You should have a look at the full diff yourself in between ;) I'd just add that to the next paragraph instead ("of a tensor variable or multiple tensor variables:")

lasagne/updates.py
```diff
@@ -66,6 +73,7 @@
     "rmsprop",
     "adadelta",
     "adam",
+    "step_scaling",
     "norm_constraint",
```
f0k: Needs to be updated as well.

f0k:
The code is perfect now, there are just two things left to update (see my comments) and I'd still prefer `total_norm_constraint` to come after `norm_constraint`.

benanne:
Very nice work :)

skaae:
I swapped the examples. Should squash all the commits?

lasagne/updates.py
```diff
@@ -20,13 +20,16 @@
 * :func:`apply_momentum()`
 * :func:`apply_nesterov_momentum()`
 
-Finally, we provide a helper function to constrain the norm of a
-tensor variable:
+Finally, we provide two helper functions to constrain the norm of a
+tensors:
```
f0k: "the norm of a tensors" -> "the norm of tensors"

f0k:
Sorry, there's one last mistake again :) I can also fix it while merging, if you're tired of it. Otherwise squashing the commits would be nice -- it's just a short added function without any other dependencies or refactorings, so we don't need a step-by-step commit history in the main branch. Thanks!

skaae:
Fixed the last typo and squashed commits.

f0k:
Cool! Thank you, that was a good collaboration :) Merging!

PR #254, opened 2015-05-18
embedding layer + test

skaae:
Moved from https://github.com/craffel/nntools/pull/28

### DIFF
--- a/lasagne/layers/embedding.py
+++ b/lasagne/layers/embedding.py
@@ -1,0 +1,61 @@
+import numpy as np
+import theano.tensor as T
+
+from .. import init
+from .base import Layer
+
+
+__all__ = [
+    "EmbeddingLayer"
+]
+
+
+class EmbeddingLayer(Layer):
+    """
+    A layer for word embeddings. The input should be a `theano.Tensor.ivector``
+    specifying which embeddings to use.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
+
+    input_size: int
+        The Number of different embeddings. The last embedding will have index
+        input_size - 1.
+
+    output_size : int
+        The size of each embedding.
+
+    W : Theano shared variable, numpy array or callable
+        The embedding matrix.
+
+    Examples
+    --------
+    >>> from lasagne.layers import EmbeddingLayer, InputLayer, get_output
+    >>> import theano
+    >>> x = T.ivector()
+    >>> l_in = InputLayer((3, ))
+    >>> W = np.arange(3*5).reshape((3, 5)).astype('float32')
+    >>> l1 = EmbeddingLayer(l_in, input_size=3, output_size=5, W=W)
+    >>> output = get_output(l1, x)
+    >>> f = theano.function([x], output)
+    >>> x_test = np.array([0, 2]).astype('int32')
+    >>> f(x_test)
+    array([[  0.,   1.,   2.,   3.,   4.],
+           [ 10.,  11.,  12.,  13.,  14.]], dtype=float32)
+    """
+    def __init__(self, incoming, input_size, output_size,
+                 W=init.Normal(), **kwargs):
+        super(EmbeddingLayer, self).__init__(incoming, **kwargs)
+
+        self.input_size = input_size
+        self.output_size = output_size
+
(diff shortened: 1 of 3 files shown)


lasagne/layers/embedding.py
```diff
@@ -0,0 +1,62 @@
...
+        The embedding matrix.
+
+    Examples
+    --------
+    >>> from lasagne.layers import EmbeddingLayer, InputLayer, helper
+    >>> import theano
+    >>> import theano.tensor as T
+    >>> x = T.ivector()
+    >>> l_in = InputLayer((3, ))
+    >>> W = W=np.arange(3*5).reshape((3, 5)).astype('float32')
```
f0k: Duplicate `W=`.

benanne:
Looks like you have a print statement in there that Python 3.4 doesn't like :)

lasagne/layers/embedding.py
```diff
@@ -0,0 +1,62 @@
...
+        The size of each embedding.
+
+    W : Theano shared variable, numpy array or callable
+        The embedding matrix.
+
+    Examples
+    --------
+    >>> from lasagne.layers import EmbeddingLayer, InputLayer, helper
+    >>> import theano
+    >>> import theano.tensor as T
```
f0k: I think you can omit this, as `T` is imported already. (It also works for `np`, doesn't it?)

lasagne/layers/embedding.py
```diff
@@ -0,0 +1,62 @@
...
+
+    output_size : int
+        The size of each embedding.
+
+    W : Theano shared variable, numpy array or callable
+        The embedding matrix.
+
+    Examples
+    --------
+    >>> from lasagne.layers import EmbeddingLayer, InputLayer, helper
```
f0k: I'd directly import `get_output` instead of `helper` (it's already available directly in `lasagne.layers`). Just for consistency across the library.

skaae:
updated with your sugesstions

lasagne/layers/embedding.py
```diff
@@ -0,0 +1,62 @@
...
+    >>> x = T.ivector()
+    >>> l_in = InputLayer((3, ))
+    >>> W = W=np.arange(3*5).reshape((3, 5)).astype('float32')
+    >>> l1 = EmbeddingLayer(l_in, input_size=3, output_size=5, W=W)
+    >>> output = helper.get_output(l1, x)
+    >>> f = theano.function([x], output)
+    >>> x_test = np.array([0, 2]).astype('int32')
+    >>> print f(x_test)
+    [[  0.   1.   2.   3.   4.]
+     [ 10.  11.  12.  13.  14.]]
```
f0k: The print statement will fail in Python 3. Can you just do `>>> f(x_test)` instead?

lasagne/layers/embedding.py
```diff
@@ -0,0 +1,62 @@
...
+
+    output_size : int
+        The size of each embedding.
+
+    W : Theano shared variable, numpy array or callable
+        The embedding matrix.
+
+    Examples
+    --------
+    >>> from __future__ import print_function
```
f0k: That's not actually needed here. `print(f(x_test))` is a valid Python 2 statement as well.

lasagne/layers/embedding.py
```diff
@@ -0,0 +1,62 @@
...
+    >>> x = T.ivector()
+    >>> l_in = InputLayer((3, ))
+    >>> W = np.arange(3*5).reshape((3, 5)).astype('float32')
+    >>> l1 = EmbeddingLayer(l_in, input_size=3, output_size=5, W=W)
+    >>> output = get_output(l1, x)
+    >>> f = theano.function([x], output)
+    >>> x_test = np.array([0, 2]).astype('int32')
+    >>> print(f(x_test))
+    [[  0.   1.   2.   3.   4.]
+     [ 10.  11.  12.  13.  14.]]
```
f0k: Can't you just do `>>> f(x_test)`? That would seem more natural to me.

f0k:
Thank you, just a final touch and I'd be fine with merging! :+1:

lasagne/layers/embedding.py
skaae: yes it passes the tests. Updating...

skaae:
final fix of example :)

benanne:
Alright then, merging.

PR #255, opened 2015-05-18
add SliceLayer + test

skaae:
Slicing layer. Thanks to @f0k for a lot of help!

The test might need a little refactoring but i'm not that familiar with pytest

### DIFF
--- a/lasagne/layers/shape.py
+++ b/lasagne/layers/shape.py
@@ -1,4 +1,5 @@
 import numpy as np
+import theano.tensor as T
 
 from ..theano_extensions import padding
 
@@ -14,6 +15,7 @@ __all__ = [
     "dimshuffle",
     "PadLayer",
     "pad",
+    "SliceLayer"
 ]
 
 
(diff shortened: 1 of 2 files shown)


lasagne/layers/shape.py
```diff
@@ -300,3 +302,64 @@ def get_output_for(self, input, **kwargs):
...
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
+
+    indices : int, list, tuple or slice instance
+        When indices is an int the corresponding index is selected from the
+        axis dimension and the axis dimension is dropped. When indices is list,
+        tuple or slice instance the axis dimension is kept even if it is
+        singleton dimension after slicing. Negative values behaves as in numpy
+        indexing.
```
f0k: I think this misses some bits and explains others in too much detail. What about:
```
If an ``int``, selects a single element from the given axis, dropping the axis. If a slice, selects all elements in the given range, keeping the axis. If a list or a tuple of ``start, stop, step``, will be converted to a slice.
```
Maybe we shouldn't allow tuples/lists because they could be expected to perform fancy indexing instead, especially along with the keyword name `indices`. And the explanation of what happens for "singleton dimensions" is probably best transported in the usage example! It's fully consistent with any indexing in Python, so I wouldn't put special emphasis on that here.

lasagne/layers/shape.py
```diff
@@ -300,3 +302,64 @@ def get_output_for(self, input, **kwargs):
...
+    >>> l_slice_ax0 = SliceLayer(l_inp, axis=0, indices=0)
+    >>> l_slice_ax1 = SliceLayer(l_inp, axis=1, indices=(1, 3))
+    >>> l_slice_ax2 = SliceLayer(l_inp, axis=-1, indices=-1)
+    >>> x = np.arange(np.prod(in_shp)).reshape(in_shp).astype('float32')
+    >>> get_output(l_slice_ax0, x).eval().shape
+    (3, 2)
+    >>> get_output(l_slice_ax1, x).eval().shape
+    (2, 2, 2)
+    >>> get_output(l_slice_ax2, x).eval().shape
+    (2, 3)
```
f0k: I'd simplify that a bit, and also reorder it so the output shape comes directly after the layer definition (similarly to how I did it for the ReshapeLayer). Also I'd use three different sizes for the three dimensions to make it clearer.
```
>>> l_in = InputLayer((2, 3, 4))
>>> SliceLayer(l_in, indices=0, axis=0).output_shape
... # equals input[0]
(3, 4)
>>> SliceLayer(l_in, indices=slice(1, 3), axis=1).output_shape
... # equals input[:, 1:3]
(2, 2, 4)
>>> SliceLayer(l_in, indices=slice(-2, -1), axis=-1)
... # equals input[..., -2:-1]
2, 3, 1
```
This also illustrates that a slice of only one element will still keep the axis. What do you think?

skaae: I think its good. I'll update. I Also added a check to only allow lists to be len 3.

f0k: Note that I edited my comment for the usage example, I think the second version is a bit clearer.

skaae:
I updated they examples with @f0k suggestions they are cleaner now, thanks. Also adds input checks.

lasagne/layers/shape.py
```diff
@@ -300,3 +302,69 @@ def get_output_for(self, input, **kwargs):
...
+
+    Examples
+    --------
+    >>> from lasagne.layers import SliceLayer, InputLayer, get_output
+    >>> in_shp = (2, 3, 2)
+    >>> l_inp = InputLayer(in_shp)
+    >>> l_slice_ax0 = SliceLayer(l_inp, axis=0, indices=0)
+    >>> l_slice_ax1 = SliceLayer(l_inp, axis=1, indices=(1, 3, 1))
+    >>> l_slice_ax2 = SliceLayer(l_inp, axis=-1, indices=-1)
+    >>> x = np.arange(np.prod(in_shp)).reshape(in_shp).astype('float32')
```
f0k: Lines 328--333 should be removed now.

lasagne/layers/shape.py
```diff
@@ -300,3 +302,69 @@ def get_output_for(self, input, **kwargs):
...
+    >>> SliceLayer(l_in, indices=slice(-2, None), axis=-1).output_shape
+    ... # equals input[..., -2:]
+    (2, 3, 2)
+    """
+    def __init__(self, incoming, indices, axis=-1, **kwargs):
+        super(SliceLayer, self).__init__(incoming, **kwargs)
+        if isinstance(indices, (int, slice)):
+            self.slice = indices
+        elif isinstance(indices, (list, tuple)) and len(indices) == 3:
+            self.slice = slice(*indices)
```
f0k: It should be `len(indices) in (1, 2, 3)` -- all of that is okay (sorry, wasn't clear about this in my docstring). But we should still think about removing support for lists and tuples entirely, to avoid the confusion with fancy indexing. @benanne, what do you think? Should we limit `indices` to `int` and `slice` or also allow `list` and `tuple` as a shortcut for `slice` creation? I'd vote for the former.

lasagne/layers/shape.py
```diff
@@ -300,3 +302,69 @@ def get_output_for(self, input, **kwargs):
...
+        else:
+            raise ValueError(
+                "Indices must be int, slice or a list/tuple of len 3")
```
f0k: "of len >=1 and <=3" (unless we drop support for this)

lasagne/layers/shape.py
skaae: I agree to drop support for lists, the most common use case is easily available with ints anyway.

lasagne/layers/shape.py
skaae: wops, sorry about that.

lasagne/layers/shape.py
f0k: Okay, cool! Feel free to go ahead and remove them then :) Don't forget to change the docstring (i.e., delete the last sentence and remove `list` and `tuple`) and the test (i.e., change the tuple to a slice). This will allow us to add lists for fancy indexing later if we wanted to, or for multi-axis slicing.

PS: Another fun way for the slice layer syntax would be:
```python
l_in = InputLayer((2,3,4))
l1 = SliceLayer(l_in)[:,-1]
```
That would probably raise some eyebrows, though.

skaae:
Removed support for lists. Docs and code reads more easily now.

f0k:
> Docs and code reads more easily now.

A lot! Merging, thanks again!

PR #256, opened 2015-05-19
generalize embedding

skaae:
The previous version of the embedding layer did only handle 1d output. Minor change that generalize it 
to any dimension input.


benanne:
Cool! It would be good to keep the tests from the 1D case as well though, instead of replacing them.

skaae:
ok, i'll clean up the tests tonight or tomorrow.

skaae:
adds 1d and 2d test case

f0k:
Looks good, Travis passes, merging.

PR #257, opened 2015-05-20
Change Sphinx theme to the one provided by readthedocs

f0k:
This changes the Sphinx theme to `sphinx_rtd_theme`. Content looks a lot nicer than for the bootstrap theme, it's also mobile-friendly (the sidebar collapses for narrow screens), and it's included with recent versions of sphinx (so you can also build it locally as before).
The next best bet would be the theme used by the Python documentation, with a sticky sidebar. I've added that one as well (called `"classic"` in Sphinx 1.3), but left it commented out. In my opinion, the readthedocs theme styles the "class lasagne.Xyz" lines much better, which is useful for pages that have a lot of classes (such as the `lasagne.layers` API reference).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,7 +14,6 @@
 
 # import sys
 # import os
-import sphinx_bootstrap_theme
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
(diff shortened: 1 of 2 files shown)


benanne:
Looks pretty good! Let's merge this when Travis passes.

f0k:
The Travis service hangs: https://www.traviscistatus.com/

benanne:
Bummer! Well, since it's only the docs config I guess it doesn't really matter? It's not like we test that anyway. Feel free to merge.

f0k:
Merging and hoping that it works like this on readthedocs as well!

benanne:
Always a gamble ;)

Issue #112, opened 2015-01-24
Improving test coverage


benanne:
As we discussed in #67, we want to encourage contributors to add tests to their pull requests as much as possible. But as long as the library has poor coverage, it's going to be hard to do this. So we should probably make a concentrated effort to get the coverage to an appropriate level.

@dnouri already did some great work setting things up and adding a ton of tests (see #40 and #47 among others).

Here's the current coverage report that I get when I run py.test:

```
Name                                 Stmts   Miss  Cover   Missing
------------------------------------------------------------------
lasagne/__init__                         7      0   100%   
lasagne/init                            63      3    95%   17, 44, 96
lasagne/layers/__init__                 12      0   100%   
lasagne/layers/base                     52      5    90%   163, 211, 216, 262, 289
lasagne/layers/conv                    116     94    19%   22-42, 45-46, 49, 52, 55-64, 69-92, 99-119, 122-123, 126, 129, 132-144, 149-173
lasagne/layers/corrmm                   69     51    26%   33-72, 75-76, 79, 82, 85-89, 92-107
lasagne/layers/cuda_convnet            175    140    20%   38-93, 96-101, 104, 107, 110-123, 126-151, 156-177, 180-195, 198-207, 222, 225, 236, 239, 254-273, 276, 279, 282, 285-297
lasagne/layers/dense                    66     29    56%   101-120, 123, 126, 129, 132-146
lasagne/layers/dnn                      96     70    27%   16, 32-35, 38-41, 44-46, 51, 57-103, 106-107, 110, 113, 116-120, 123-144
lasagne/layers/helper                   45      1    98%   58
lasagne/layers/input                    24      2    92%   42-43
lasagne/layers/merge                    39      9    77%   23-26, 63-66, 70-72
lasagne/layers/noise                    31      0   100%   
lasagne/layers/pool                     79     58    27%   20-22, 25-34, 37, 58-65, 69-71, 74-85, 102-108, 112-134, 143-144, 147, 150
lasagne/layers/shape                    28     13    54%   21, 24, 31-34, 37-44, 47
lasagne/nonlinearities                  10      0   100%   
lasagne/objectives                      15      0   100%   
lasagne/regularization                   8      4    50%   8-13
lasagne/tests/conftest                   6      0   100%   
lasagne/tests/layers/conftest            6      0   100%   
lasagne/tests/layers/test_base          90      0   100%   
lasagne/tests/layers/test_dense         55      0   100%   
[...]

PR #258, opened 2015-05-20
improve test coverage

benanne:
This is going to be a collection of tests to improve test coverage across the library. So far I've only done `base.py`, which should now be 100% covered when we get rid of the deprecated stuff. Other than that I'll be focusing mostly on the convolution implementations for now. If anyone wants to help we can try to divide the work a little. It would be nice to get the coverage up in the 90%+ range before we release the library :)

For clarity's sake, I'm of course aware that good coverage does not necessarily mean the tests are good - but it's an important first step!

### DIFF
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -91,7 +91,7 @@ class Layer(object):
 
         return result
 
-    def get_output_shape(self):
+    def get_output_shape(self):  # pragma: no cover
         """
         Deprecated. Use `layer.output_shape`.
         """
@@ -101,7 +101,7 @@ class Layer(object):
                       "layer.output_shape instead.")
         return self.output_shape
 
-    def get_output(self, input=None, **kwargs):
+    def get_output(self, input=None, **kwargs):  # pragma: no cover
         """
         Deprecated. Use `lasagne.layers.get_output(layer, input, **kwargs)`.
         """
@@ -233,7 +233,7 @@ class Layer(object):
 
         return param
 
-    def get_bias_params(self):
+    def get_bias_params(self):  # pragma: no cover
         import warnings
         warnings.warn("layer.get_bias_params() is deprecated and will be "
                       "removed for the first release of Lasagne. Please use "
--- a/lasagne/tests/layers/test_base.py
+++ b/lasagne/tests/layers/test_base.py
@@ -93,6 +93,10 @@ class TestLayer:
         A = named_layer.add_param(a, a_shape, name='A')
         assert A.name == 'layer_name.A'
 
+    def test_get_output_for_notimplemented(self, layer):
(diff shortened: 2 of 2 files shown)


dnouri:
This trick might be useful if you want to ignore coverage of deprecated functions or the like: http://nedbatchelder.com/code/coverage/excluding.html

benanne:
Good to know! Probably not worth it yet though, all of that code is going away in a matter of weeks anyway :) But I'll keep it in mind for later.

f0k:
> Probably not worth it yet though

I think it wouldn't hurt to add `#pragma: no cover` comments to the deprecated function definitions as part of the effort to increase coverage. The comments will go away along with the deprecated functions, and it will make it easier to hunt down the important uncovered cases. So don't hesitate to drop those comments where you see fit.

benanne:
True, I guess I can do it as I go through the coverage reports.

ebenolson:
Should probably ask, is the ultimate intention 100% coverage across the board? 
Testing for every little thing can get a bit silly, but it is a nice tangible goal.

The files currently needing coverage (most to least) are:
- [ ] lasagne/updates
- [ ] lasagne/layers/cuda_convnet
- [ ] lasagne/layers/pool
- [ ] lasagne/layers/helper
- [ ] lasagne/utils
- [ ] lasagne/layers/shape
- [ ] lasagne/layers/dnn
- [ ] lasagne/layers/conv
- [ ] lasagne/layers/corrmm
- [ ] ~~lasagne/objectives~~
- [ ] lasagne/layers/merge
- [ ] ~~lasagne/regularization~~
- [ ] lasagne/layers/normalization
- [ ] lasagne/layers/input

I'm planning to continue working my way up the list and grab some more low hanging fruit.

I'm ignoring objectives and regularization for now, since they will probably change more before first release.

benanne:
I don't think we need to worry about getting it to 100% right now, there are probably other matters that are more deserving of our attention. But it would be nice to have it somewhere in the 90%+ range before we release.

From that list I would focus on updates, helper and pool next if I were you :) I plan to sort out the convolution code myself since I know that part of the codebase very well. But if I'm not going fast enough to your liking feel free to take some of that off my hands as well ;)

ebenolson:
Sounds good. I think the conv stuff is going to be a bit of a pain, since most of what's uncovered are the various special restrictions that the generic tests avoid.

benanne:
Yep... probably :p I have a feeling there might be a few bugs lurking there as well though.

benanne:
Maybe the todo list should be moved to #112, so we don't lose track of it once this PR is merged.

ebenolson:
Makes sense, too bad it's not possible to edit other peoples comments.

benanne:
I'll merge this tomorrow and do the convolution tests in one or more separate PRs.

PR #259, opened 2015-05-20
fix signatures and a few other things in the docs

benanne:
This PR makes all function and class signatures look nice (part of #203). I also made `NonlinearityLayer` and `adam` show up in the docs.

### DIFF
--- a/docs/modules/layers.rst
+++ b/docs/modules/layers.rst
@@ -39,6 +39,9 @@ Layer classes: dense layers
 .. autoclass:: DenseLayer
    :members:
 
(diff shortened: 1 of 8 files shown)


f0k:
Looks better indeed! The only trouble is that we need to be extra careful to always update both signatures at once, and that it looks a tad silly in the terminal. I don't have a strong opinion on that, I'd be both fine with merging and with leaving it as it is. As you please! :)

benanne:
Yeah, we'll need to pay attention so they don't go out of sync. If a better solution to this problem comes along we should definitely consider it, but for now this will have to do.

PR #260, opened 2015-05-21
Extend and fix markup of LeakyRectify docstring

f0k:
Fixes the `Methods` section of the `LeakyRectify` docstring, documents it a bit more extensively, and properly documents the `leaky_rectify` shortcut (using the technique suggested in the numpydoc howto). While we're at it, also adds a `very_leaky_rectify` shortcut, as this seems to become more popular.

### DIFF
--- a/lasagne/nonlinearities.py
+++ b/lasagne/nonlinearities.py
@@ -83,29 +83,59 @@ def rectify(x):
 
 # leaky rectify
 class LeakyRectify(object):
-    """Implementation of a leaky rectifier
-    :math:`\\varphi(x) = \max(leakiness \cdot x, x)`.
+    """Leaky rectifier :math:`\\varphi(x) = \\max(\\alpha \\cdot x, x)`
+
+    The leaky rectifier was introduced in [1]_. Compared to the standard
+    rectifier :func:`rectify`, it has a nonzero gradient for negative input,
+    which often helps convergence.
 
     Parameters
     ----------
-    leakiness : float in [0, 1]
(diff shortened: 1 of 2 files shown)


lasagne/nonlinearities.py
```diff
@@ -83,29 +83,56 @@ def rectify(x):
...
-        a leakiness of 1 will lead to a linear activation function.
+        a leakiness of 1 will lead to a linear activation function,
+        and any value in between will give a leaky rectifier.
 
     Methods
     -------
     __call__(x)
-        Calculate the neuron's output by applying this leaky rectifier to the
-        activation `x`.
+        Calculate the neuron's output for activation `x`.
```
benanne: nitpicking here, but I guess this should be neurons' instead of neuron's, in the typical case :)

benanne:
Looks good! The 0.3 is somewhat arbitrary but I guess it's as good a value as any. Feel free to merge.

f0k:
> The 0.3 is somewhat arbitrary

0.3 is close to 1./3, which is backed by http://arxiv.org/abs/1409.6070 and http://benanne.github.io/2015/03/17/plankton.html. I just found 0.3 easier to write with our parameterization (as we have a multiplicative factor rather than a divisor). I'll change it to 1./3 and cite the paper.

lasagne/nonlinearities.py
f0k: It shouldn't mention neurons at all, I think. I'll change it to something more neutral.

PR #261, opened 2015-05-21
Trainable parameter

Erotemic:
I added the trainable parameter so Conv2DCCLayer. This lets the layer be excluded from training updates. I find this useful for when I initialize my first layer with weights from another network.

### DIFF
--- a/lasagne/layers/cuda_convnet.py
+++ b/lasagne/layers/cuda_convnet.py
@@ -152,6 +152,13 @@ class Conv2DCCLayer(CCLayer):
         `cuda-convnet documentation
         <https://code.google.com/p/cuda-convnet/wiki/LayerParams>`_.
 
+    trainable : bool (default True)
+        If ``True`` updates will be applied to the weights in this layer as
+        usual.
+
+        If ``False`` the weights in this layer will remain unchanged by
+        training.
+
     **kwargs
         Any additional keyword arguments are passed to the `Layer` superclass.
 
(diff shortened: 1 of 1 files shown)


benanne:
To be honest, I don't think this is the right way to approach this as it would require adding this parameter to all layers with parameters.

The easiest way to only train certain subsets of parameters is to manually construct the list of parameters instead of using `get_all_params()` on the output layer of the network.

Of course this gets a bit tedious. With the new parameter tagging system (see #228), what you could do instead is to mark the parameters you do not wish to train as non-trainable by removing the 'trainable' tag:
```python
layer.params[layer.W].remove('trainable')
```

Then just get the parameters with `get_all_params(..., trainable=True)` to only grab the trainable ones.

Maybe it would also be useful to extend `get_all_params` to allow for specifying layers whose parameters should not be included. But that makes the interface bigger and the code more complicated, and I'm not sure if this use case is common enough for that.

f0k:
Thank you -- but this is not the best solution to tackle this use case. For consistency, we would need to add this to every layer that has trainable parameters, making all the current and future layers more complex.
Two alternative suggestions:
* In your code, when collecting the parameters for training, use `get_all_layers` to collect the layers, exclude the ones you don't want to train on, and then collect the parameters for those layers manually instead of calling `get_all_params`.
* In your code, after creating the layer you don't want to train, tag its parameters as non-trainable.

You can even write a helper function for the second case:
```python
make_untrainable(layer):
    for param in layer.params:
        layer.params[param].discard('trainable')
    return layer
```
That would be used as:
```
layer = InputLayer((None, 10, 20, 30))
layer = Conv2DCCLayer(layer, ...)
make_untrainable(layer)
```

Erotemic:
Removing the tag seems reasonable. I'll do that instead.

Issue #74, opened 2014-12-18
First release


benanne:
People have already started using the library, so we should make an effort to put out a first release.I made a milestone to tag issues and pull requests that need to be sorted out before we can make a release (thanks to @craffel for the suggestion).

The most important things will be sorting out our test coverage, and writing some basic documentation. I've been adding some docstrings now and then, but progress is slow and we will probably need a concerted effort to get this done in a reasonable amount of time.

What else should we take care of for the first release? Are there any other issues that need to be tagged?

benanne:
We have less than a month left to get our first release ready at this point. Looking at the milestone, we're doing pretty well in the code department, with all necessary changes either having been carried out or at least planned. So the codebase should be ready soon. The main thing we still lack is complete API documentation, as well as something that can pass as a user guide for now :) At the very least we should have complete API docs and a rudimentary 'getting started' guide ready before we put this on PyPI.

What does everyone's schedule look like for the coming weeks? I realise the ISMIR deadline is around the same time, but I think we should all set aside some time to fill up holes in the docs.

f0k:
> What does everyone's schedule look like for the coming weeks? I realise the ISMIR deadline is around the same time, but I think we should all set aside some time to fill up holes in the docs.

Yes, the ISMIR deadline is slowly approaching. But I agree that we should start documenting things that we know won't change. And we should fix a documentation format and convert the documentation. What was the state on that? Did we agree to move to numpy's format?

benanne:
Yep, we agreed to switch to numpydoc in the dev meeting earlier this month. We didn't assign anyone to the task of converting the existing documentation though :p I guess it's pretty unrewarding work. I may end up doing it later this week as a form of procrastination, but if anyone else wants to do it that would be great (and I might get another few paragraphs of my dissertation written that way!)

ebenolson:
I'm happy to help out with this if you want to assign some things to me.

benanne:
That would be excellent! Once we've converted the existing docs to the numpydoc format, any function, class or method without a docstring will need one, so there should be plenty of opportunities to help out :)

f0k:
> Once we've converted the existing docs to the numpydoc format, any function, class or method without a docstring will need one

[...]

PR #262, opened 2015-05-21
Sphinx: Link each entity to its source code on github

f0k:
This links each entity to its source file on github. This should make it a lot easier for people to suggest edits to docstrings.
Numpy tweaked this further (https://github.com/numpy/numpy/blob/master/doc/source/conf.py#L286), in two respects:
* They link to the correct line in the file, while this PR just links to the file per se.
* They link to the correct version of the source code, while this PR just links to the latest master.

It should be possible to copy and adapt their code, but I don't have the time now -- I'd welcome any PRs to my PR!


benanne:
Looks good! It would be good to make it differentiate between master and release already, I have a feeling we'll forget about it if we wait until the release.

f0k:
> It would be good to make it differentiate between master and release already

Please have a look at the numpy code: They solve this via `numpy.__version__`. Maybe there's a different way when compiling online at readthedocs, but to produce the correct links locally, that's the way to go. I've added this to our release todo list: https://github.com/Lasagne/Lasagne/issues/74#issuecomment-104607646

f0k:
Rebased and amended. It now links to the correct line number in the correct file in the correct branch/tag. Ready for reviewing and merging.

docs/conf.py
```diff
@@ -61,10 +62,11 @@
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
+import lasagne
 # The short X.Y version.
-version = '0.1'
+version = '.'.join(lasagne.__version__.split('.', 3)[:2]).rstrip('dev')
 # The full version, including alpha/beta/rc tags.
-release = '0.1'
+release = lasagne.__version__
```
f0k: This saves another version string update that could easily be missed; numpy does it the same way. Let's see how readthedocs thinks about that. I've read that you may need to configure readthedocs to use `virtualenv`.

benanne: Why would it work with `virtualenv` but not otherwise? I'm not sure how it's currently set up - there seem to be multiple virtualenv-related settings, some of them are enabled and some aren't. It's a bit unclear what they all do. I remember having to tinker with them for a while before everything would work, though.

benanne:
Not much to say about this as I'm not very familiar with how Sphinx works. Everything looks fine to me.

docs/conf.py
f0k: This line requires "import lasagne" beforehand, maybe this is not universally fine. I also don't know exactly how Sphinx works. We can just merge and see what happens.

benanne: Fine by me :)

f0k:
Okay, merged and crossing fingers.

f0k:
Wait, it doesn't work for `lasagne.layers`, it thinks the definitions reside in `lasagne/layers.py` although they are defined in submodules imported in `lasagne/layers/__init__.py`. Sorry for not checking all the pages beforehand.

PR #264, opened 2015-05-23
Increase test coverage

ebenolson:
This brings the coverage of `dense.py` to 100%, adding tests for `NINLayer` and `NonlinearityLayer` and some small additions to `DenseLayer`. 

Also fixes a minor bug in `NINLayer`.

### DIFF
--- a/lasagne/layers/dense.py
+++ b/lasagne/layers/dense.py
@@ -176,11 +176,13 @@ class NINLayer(Layer):
         self.W = self.add_param(W, (num_input_channels, num_units), name="W")
         if b is None:
             self.b = None
-        elif self.untie_biases:
-            biases_shape = (num_units,) + self.output_shape[2:]
         else:
-            biases_shape = (num_units,)
-        self.b = self.add_param(b, biases_shape, name="b", regularizable=False)
+            if self.untie_biases:
+                biases_shape = (num_units,) + self.output_shape[2:]
+            else:
+                biases_shape = (num_units,)
+            self.b = self.add_param(b, biases_shape, name="b",
+                                    regularizable=False)
 
     def get_output_shape_for(self, input_shape):
         return (input_shape[0], self.num_units) + input_shape[2:]
--- a/lasagne/tests/layers/test_dense.py
+++ b/lasagne/tests/layers/test_dense.py
@@ -99,8 +99,8 @@ class TestDenseLayer:
         # Check that the input to the nonlinearity was what we expect
         # from dense layer, i.e. the dot product plus bias
         nonlinearity_arg = nonlinearity.call_args[0][0]
-        assert (nonlinearity_arg.eval() ==
-                np.dot(input.get_value().reshape(2, -1), W) + b).all()
+        assert np.allclose(nonlinearity_arg.eval(),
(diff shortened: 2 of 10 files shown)


ebenolson:
oops... I started from the latest master, need to redo this.

ebenolson:
Added tests for `theano_extensions`

lasagne/tests/layers/test_dense.py
```diff
@@ -1,5 +1,5 @@
 from mock import Mock
-import numpy
+import numpy as np
```
f0k: :+1: This has bugged me as well.

lasagne/tests/layers/test_dense.py
```diff
@@ -113,3 +115,173 @@ def test_named_layer_param_names(self, DenseLayer, dummy_input_layer):
...
+        X = input.get_value()
+        X = np.rollaxis(X, 1).T
+        X = np.dot(X, layer.W.get_value())
+        if layer.b is not None:
+            if layer.untie_biases:
+                X += layer.b.get_value()[:, np.newaxis].T
+            else:
+                X += layer.b.get_value()
+        X = np.rollaxis(X.T, 0, 2)
+        assert (nonlinearity_arg.eval() == X).all()
```
f0k: Travis failed here, it seems this is too strict. I'd suggest `np.allclose` or `np.testing.assert_array_almost_equal`. Not sure why it worked for the dense layer tests above.

ebenolson: Thanks, changed that.

benanne:
Sweet, nice work! Looks good, I'll have a more in-depth look tomorrow.

ebenolson:
Reworked `test_nonlinearities.py` to bring coverage to 100%

lasagne/layers/dense.py
```diff
@@ -176,11 +176,13 @@ def __init__(self, incoming, num_units, untie_biases=False,
...
-            biases_shape = (num_units,) + self.output_shape[2:]
         else:
-            biases_shape = (num_units,)
-        self.b = self.add_param(b, biases_shape, name="b", regularizable=False)
+            if self.untie_biases:
+                biases_shape = (num_units,) + self.output_shape[2:]
+            else:
+                biases_shape = (num_units,)
+            self.b = self.add_param(b, biases_shape, name="b",
+                                    regularizable=False)
```
benanne: Good catch :)

benanne:
Looks great! I didn't really see anything out of the ordinary in the new tests. One thing I find curious is that there are no changes to the 1D convolution implementations in `theano_extensions`, I was expecting there to be quite a few bugs in those (maybe only related to `border_mode != 'valid'` though).

I would be okay with merging this at this point, if you'd like to see your work so far merged sooner rather than later, and then you can continue your coverage quest in another PR. I think that would be useful, especially because there's a bugfix related to bias-less layers in here. Up to you though :)

ebenolson:
Merging now is fine, this got a little bigger than I expected. 

This PR was to your coverage branch, would you prefer I close it and open one to master?

benanne:
Oh, I hadn't even noticed! Yeah, maybe it's better to split this effort into a couple of PRs to master, otherwise it will be held back by my laziness ;)

PR #265, opened 2015-05-24
Extend test coverage

ebenolson:
This PR (previously https://github.com/Lasagne/Lasagne/pull/264) extends test coverage on dense.py, theano_extensions, nonlinearities.py, input.py, merge.py, normalization.py, shape.py and utils.py.

It also fixes a minor bug in NINLayer and remove an extra check from DimshuffleLayer

### DIFF
--- a/lasagne/tests/layers/test_dense.py
+++ b/lasagne/tests/layers/test_dense.py
@@ -99,8 +99,8 @@ class TestDenseLayer:
         # Check that the input to the nonlinearity was what we expect
         # from dense layer, i.e. the dot product plus bias
         nonlinearity_arg = nonlinearity.call_args[0][0]
-        assert (nonlinearity_arg.eval() ==
-                np.dot(input.get_value().reshape(2, -1), W) + b).all()
+        assert np.allclose(nonlinearity_arg.eval(),
+                           np.dot(input.get_value().reshape(2, -1), W) + b)
 
     def test_param_names(self, layer):
         assert layer.W.name == "W"
@@ -270,7 +270,7 @@ class TestNINLayer:
             else:
                 X += layer.b.get_value()
         X = np.rollaxis(X.T, 0, 2)
-        assert (nonlinearity_arg.eval() == X).all()
+        assert np.allclose(nonlinearity_arg.eval(), X)
 
     def test_param_names(self, layer):
         assert layer.W.name == "W"
--- a/lasagne/tests/layers/test_dense.py
+++ b/lasagne/tests/layers/test_dense.py
@@ -1,5 +1,5 @@
 from mock import Mock
-import numpy
+import numpy as np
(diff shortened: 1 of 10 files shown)


lasagne/tests/layers/test_dense.py
```diff
@@ -113,3 +115,173 @@ def test_named_layer_param_names(self, DenseLayer, dummy_input_layer):
...
+            )
+        assert layer.nonlinearity == lasagne.nonlinearities.identity
+
+    def test_get_output_for(self, layer_vars):
+        layer = layer_vars['layer']
+        nonlinearity = layer_vars['nonlinearity']
+
+        input = theano.shared(np.random.uniform(-1, 1, (2, 12)))
+        result = layer.get_output_for(input)
+        assert result is nonlinearity.return_value
```
f0k: `input = theano.tensor.matrix()` would be enough, and an `assert nonlinearity.called_with(input)` would be a nice addition!

lasagne/tests/layers/test_dense.py
```diff
@@ -113,3 +115,173 @@ def test_named_layer_param_names(self, DenseLayer, dummy_input_layer):
...
+
+class TestNonlinearityLayer:
+    @pytest.fixture
+    def NonlinearityLayer(self):
+        from lasagne.layers.dense import NonlinearityLayer
+        return NonlinearityLayer
+
+    @pytest.fixture
+    def layer_vars(self, dummy_input_layer):
+        from lasagne.layers.dense import NonlinearityLayer
```
f0k: Not critical at all, but you could use the fixture here instead of importing it again. (Same for `NINLayer` below.)

lasagne/tests/test_nonlinearities.py
```diff
@@ -1,17 +1,50 @@
...
+                                          nonlinearity)
+        np_nonlinearity = getattr(self, nonlinearity)
+
+        X = T.matrix()
+        X0 = lasagne.utils.floatX(np.random.uniform(-3, 3, (10, 10)))
+
+        theano_result = theano_nonlinearity(X).eval({X: X0})
+        np_result = np_nonlinearity(X0)
+
+        assert np.allclose(theano_result, np_result)
```
f0k: :+1: for generalizing the code a bit. Also the previous version relied on the functions working for nonsymbolic input, which is not actually what we needed to test.

f0k:
Went through everything, looks very good! I just added two very minor comments that you may address or ignore, as you please :) I'm fine with merging in any case.

benanne:
As am I. Will wait for those comments to be addressed to merge, but let me know in case you don't plan to do that :)

ebenolson:
Thanks, made those changes.

benanne:
Sweet, merging!

PR #266, opened 2015-05-24
Updates test coverage

ebenolson:
To add some coverage of the update rules, I wrote a test comparing their results to values calculated using the Torch optim package. 

In the process, I found a discrepancy with the ADAM updates, after looking at the paper I think the Torch version is correct but I'd appreciate a second opinion.

### DIFF
--- a/lasagne/tests/test_updates.py
+++ b/lasagne/tests/test_updates.py
@@ -1,6 +1,63 @@
 import pytest
+import numpy as np
+import theano
+import lasagne
 
 PCT_TOLERANCE = 1E-5
+
+
+class TestUpdateFunctions(object):
+    # These tests compare results on a toy problem
+    # to values calculated by the torch.optim package
+    torch_values = {'sgd': [0.81707280688755,
+                            0.6648326359915,
+                            0.5386151140949],
+                    'momentum': [0.6848486952183,
+                                 0.44803321781003,
+                                 0.27431190123502],
+                    'nesterov_momentum': [0.67466543592725,
+                                          0.44108468114241,
+                                          0.2769002108997],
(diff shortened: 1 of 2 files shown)


ebenolson:
:tired_face:  not sure how this happened, but I apparently deleted `total_norm_constraint()` by accident then reimplemented something similar...  will need to rework this tomorrow

benanne:
Good catch on the Adam implementation, I didn't spot that when it was merged. I guess those numerical tests are going to prove useful :)

benanne:
It would be useful to add a comment describing briefly how those values were obtained using Torch. That way if we decide to add more algorithms later we can update the tests more easily as well.

ebenolson:
Ok, I fixed the accidental deletion and added a link to a gist of the torch script I used.

skaae:
The update tests are nice. Good catch with the adam bug I don't know where i got the squared idea from :) I probably messed up in the dense subscript/superscript notation.

benanne:
Looks good! I'm okay with merging if there are no further comments.

lasagne/tests/test_updates.py
```diff
@@ -64,7 +139,7 @@ def test_norm_constraint_dim6_raises():
     assert "Unsupported tensor dimensionality" in str(excinfo.value)
 
 
-def total_norm_constraint():
+def test_total_norm_constraint():
```
f0k: Ooops, didn't see that when reviewing.

f0k:
Nice work! Also an interesting toy problem to demonstrate how advanced methods are independent to gradient magnitude.

benanne:
Alright then, merging!

PR #267, opened 2015-05-25
increase coverage for convolution layers

benanne:
This PR does a few things:
* fixes a bug in all convolution implementations which made bias-less layers unusable (the same one that @ebenolson fixed in `NINLayer`, see https://github.com/ebenolson/Lasagne/commit/ffce108e87d4f14a7e2c2499bb4ccb1c9c86f02e)
* moves the `border_mode` validity check into the constructor for all layers, so constructing a layer with an invalid border mode string fails as early as possible (previously this check was in `get_output_for()` and `get_output_shape_for()` for a few implementations)
* adds tests for bias-less layers (`b=None`), linear layers (`nonlinearity=None`) and invalid border mode checks

It brings coverage for `lasagne/layers/conv.py` up to almost 100%. It falls just two lines short of that: the `b=None` case in `get_output_for()` is not covered for `Conv1DLayer` and `Conv2DLayer`. I'm not sure how to go about covering these two lines effectively. Adding the `b=None` case to all test configurations in `conv2d_test_sets()` seems wasteful, it will increase the test duration by a lot and the benefit is negligible. It seems like a single "output test" should be sufficient. Any ideas?

I'll try to get the coverage for `corrmm`, `dnn` and `cuda-convnet` up to 100% as well in this PR (at least the convolution parts, might leave the pooling parts for later).

### DIFF
--- a/lasagne/layers/corrmm.py
+++ b/lasagne/layers/corrmm.py
@@ -18,7 +18,7 @@ __all__ = [
 ]
 
 
-if not theano.config.device.startswith("gpu"):
+if not theano.config.device.startswith("gpu"):  # pragma: no cover
     raise ImportError("requires a GPU to work")
 
 
--- a/lasagne/layers/corrmm.py
+++ b/lasagne/layers/corrmm.py
@@ -167,20 +167,21 @@ class Conv2DMMLayer(MMLayer):
                 self.pad = ((self.filter_size[0] - 1) // 2,
                             (self.filter_size[1] - 1) // 2)
             else:
-                raise RuntimeError("Unsupported border_mode for "
-                                   "Conv2DMMLayer: %s" % border_mode)
+                raise RuntimeError("Invalid border mode: '%s'" % border_mode)
         else:
             self.pad = as_tuple(pad, 2)
 
         self.W = self.add_param(W, self.get_W_shape(), name="W")
         if b is None:
             self.b = None
-        elif self.untie_biases:
-            biases_shape = (num_filters, self.output_shape[2],
-                            self.output_shape[3])
         else:
-            biases_shape = (num_filters,)
-        self.b = self.add_param(b, biases_shape, name="b", regularizable=False)
+            if self.untie_biases:
+                biases_shape = (num_filters, self.output_shape[2],
+                                self.output_shape[3])
+            else:
+                biases_shape = (num_filters,)
+            self.b = self.add_param(b, biases_shape, name="b",
+                                    regularizable=False)
 
         self.corr_mm_op = GpuCorrMM(subsample=self.stride, pad=self.pad)
 
--- a/lasagne/layers/corrmm.py
+++ b/lasagne/layers/corrmm.py
@@ -19,7 +19,7 @@ __all__ = [
 
 
 if not theano.config.device.startswith("gpu"):
-    raise ImportError("requires a GPU to work")
+    raise ImportError("requires a GPU to work")  # pragma: no cover
 
 
 # base class for all layers that rely on GpuCorrMM directly
--- a/lasagne/layers/corrmm.py
+++ b/lasagne/layers/corrmm.py
@@ -18,7 +18,7 @@ __all__ = [
 ]
 
 
-if not theano.config.device.startswith("gpu"):  # pragma: no cover
+if not theano.config.device.startswith("gpu"):
     raise ImportError("requires a GPU to work")
 
 
--- a/lasagne/layers/cuda_convnet.py
+++ b/lasagne/layers/cuda_convnet.py
@@ -466,7 +466,7 @@ class MaxPool2DCCLayer(CCLayer):
             stride = as_tuple(stride, 2)
             if stride[0] != stride[1]:
                 raise NotImplementedError("MaxPool2DCCLayer only supports "
-                                          "using the same stride in both, "
+                                          "using the same stride in both "
                                           "directions but stride=(%d, %d)"
                                           % stride)
             self.stride = stride[0]
--- a/lasagne/layers/cuda_convnet.py
+++ b/lasagne/layers/cuda_convnet.py
@@ -26,7 +26,7 @@ __all__ = [
 
 
 if not theano.config.device.startswith("gpu"):
-    raise ImportError("requires a GPU to work")
+    raise ImportError("requires a GPU to work")  # pragma: no cover
 
 
 # base class for all layers that use ops from pylearn2.sandbox.cuda_convnet
--- a/lasagne/layers/cuda_convnet.py
+++ b/lasagne/layers/cuda_convnet.py
@@ -256,7 +256,11 @@ class Conv2DCCLayer(CCLayer):
             else:
                 raise RuntimeError("Invalid border mode: '%s'" % border_mode)
         else:
-            self.pad = pad
+            pad = as_tuple(pad, 2)
+            if pad[0] != pad[1]:
+                raise RuntimeError("Conv2DCCLayer only supports square "
+                                   "padding, but pad=(%d, %d)" % pad)
+            self.pad = pad[0]
 
         if W is None:
             if dimshuffle:
--- a/lasagne/layers/cuda_convnet.py
+++ b/lasagne/layers/cuda_convnet.py
@@ -254,8 +254,7 @@ class Conv2DCCLayer(CCLayer):
                 # is probably not worth supporting.
                 self.pad = (self.filter_size - 1) // 2
             else:
-                raise RuntimeError("Unsupported border_mode for "
-                                   "Conv2DCCLayer: %s" % border_mode)
+                raise RuntimeError("Invalid border mode: '%s'" % border_mode)
         else:
             self.pad = pad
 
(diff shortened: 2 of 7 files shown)


lasagne/tests/layers/test_conv.py
```diff
@@ -108,6 +110,22 @@ def test_defaults(self, DummyInputLayer,
         except NotImplementedError:
             pass
 
+    def test_init_none_nonlinearity_bias(self, DummyInputLayer):
```
ebenolson: this test could perhaps be expanded to cover `get_output_for` as well.

ebenolson:
Nice, I should have thought to check for that bias bug in other layers as well.

Something I realized the other day: the current tests only pass because the test data are chosen to be positive, and biases are initialized to 0. Might want to change it so they test adding bias and doing the nonlinearity as well.

benanne:
Yeah, good call. It's a bit complicated though, I'm only just getting to grips with how the convolution tests actually work. I'll see what I can do in this regard but it may be more efficient for someone else to take a look at that in the meantime.

benanne:
Added a test configuration for the biasless case, so it is now tested for all layers. I also added some corrmm-specific tests for the 'pad' argument. `conv` and `corrmm` should now have 100% coverage. `cuda-convnet` is next up.

lasagne/layers/corrmm.py
```diff
@@ -18,7 +18,7 @@
 ]
 
 
-if not theano.config.device.startswith("gpu"):
+if not theano.config.device.startswith("gpu"):  # pragma: no cover
```
benanne: I'm not sure if this is the best we can do here - but it's hard to test this without spoofing. Might be worth it though, people could try to use these layers without a compatible GPU and we should make sure that they get an understandable error message in this case. Any ideas?

ebenolson: You could do something like this:

	def test_import_without_gpu_raises():
	    if theano.config.device.startswith("gpu"):
	        pytest.skip()
	    else:
	        with pytest.raises(ImportError):
	            import lasagne.layers.corrmm

benanne: Good idea.

benanne:
todo list for cuda_convnet-based code:

* [ ] test the `dimshuffle=False` case
* [ ]  test `MaxPool2DCCLayer` (can base this on the existing `MaxPool2DLayer` tests)
* [ ] test bc01 <-> c01b shuffle layers
* [ ] test `NINLayer_c01b` (can base this on the existing `NINLayer` tests)

benanne:
Added tests for the `dimshuffle=False` case, this should now be covered 100%. It probably still misses some things that should be tested, and to be honest `test_conv.py` is getting a bit messy, but I'll leave that for a later PR. I'll just focus on getting to 100% here, it seems to have helped us uncover more than a few bugs so far.

benanne:
I think I'm done with `cuda_convnet`, so just `dnn` left to do. I need to run the tests on another machine for that because the GPU in my laptop doesn't support cuDNN.

benanne:
We're now up to 100% coverage for everything except `regularization` and `objectives`, both of which will be reworked soon. I'm not sure if everything I added was done in the cleanest way possible, so comments on that are welcome.

lasagne/layers/cuda_convnet.py
```diff
@@ -617,7 +622,7 @@ class NINLayer_c01b(Layer):
         Any additional keyword arguments are passed to the `Layer` superclass.
     """
     def __init__(self, incoming, num_units, untie_biases=False,
-                 W=init.GlorotUniform(c01b=True), b=init.Constant(0.),
+                 W=init.GlorotUniform(), b=init.Constant(0.),
```
f0k: You removed that here, but not in the docstring.

lasagne/tests/layers/test_conv.py
```diff
@@ -190,3 +224,227 @@ def test_with_nones(self, Conv2DImpl, DummyInputLayer,
...
+        assert layer.get_params(_nonexistent_tag=True) == []
+        assert layer.get_params(_nonexistent_tag=False) == [layer.W, layer.b]
+
+
+class TestConvOutputLength:
+    def test_invalid_border_mode(self):
+        from lasagne.layers.conv import conv_output_length
+        with pytest.raises(RuntimeError) as exc:
+            conv_output_length(5, 3, 1, border_mode='_nonexistent_mode')
+        assert "Invalid border mode" in exc.value.args[0]
```
f0k: Everything else is not tested because it's tested implicitly via the conv layer tests?

lasagne/tests/layers/test_conv.py
```diff
@@ -190,3 +224,227 @@ def test_with_nones(self, Conv2DImpl, DummyInputLayer,
...
+            layer = Conv2DCCLayer(input_layer, num_filters=15,
+                                  filter_size=(3, 3))
+        assert ("Conv2DCCLayer requires num_filters to be a multiple of 16" in
+                exc.value.args[0])
+
+        with pytest.raises(RuntimeError) as exc:
+            layer = Conv2DCCLayer(input_layer, num_filters=16,
+                                  filter_size=(3, 3), pad=(1, 2))
+        assert ("Conv2DCCLayer only supports square padding" in
+                exc.value.args[0])
```
f0k: Wasn't there a limitation on the number of input channels as well? (1, 2, 3, or a multiple of 4.) I don't see that being tested here.

lasagne/tests/layers/test_conv.py
```diff
@@ -190,3 +224,227 @@ def test_with_nones(self, Conv2DImpl, DummyInputLayer,
...
+        assert layer.W.get_value().shape == (4, 3, 3, 16)
+        assert layer.b.get_value().shape == (16, 30, 30)
+
+    def test_dimshuffle_false_get_output_for(self, DummyInputLayer):
+        try:
+            from lasagne.layers.cuda_convnet import Conv2DCCLayer
+        except ImportError:
+            pytest.skip("cuda_convnet not available")
+
+        # this implementation needs to be
```
f0k: Come on! Don't keep us in suspense! :P

lasagne/tests/layers/test_pool.py
```diff
@@ -339,6 +339,56 @@ def test_get_output_shape_for(self, input_shape):
...
+        try:
+            from lasagne.layers.cuda_convnet import MaxPool2DCCLayer
+        except ImportError:
+            pytest.skip("cuda_convnet not available")
+
+        input_layer = self.input_layer((128, 4, 12, 12))
+
+        with pytest.raises(RuntimeError) as exc:
+            layer = MaxPool2DCCLayer(input_layer, pool_size=2, pad=2)
+        assert "MaxPool2DCCLayer does not support padding" in exc.value.args[0]
```
f0k: Hmm, funny. The original code has a `start` parameter that can be used for padding or cropping: https://code.google.com/p/cuda-convnet/wiki/LayerParams#Local_pooling_layer
But that's far outside this PR.

f0k:
Good job! We can make it to 100%!

lasagne/layers/cuda_convnet.py
benanne: whoops! I didn't expect this to bite us so soon, a better solution than #259 may be in order.

lasagne/tests/layers/test_conv.py
benanne: Yep. Which doesn't mean that it's not useful to add more tests for this function, I was just focusing on getting to 100% coverage.

lasagne/tests/layers/test_conv.py
benanne: There is indeed. `Conv2DCCLayer` does not check for this at the moment, and it should. One issue is that it should be a multiple of 4 if the gradient towards its input is to be computed (otherwise 1, 2, or 3 are also allowed). And we can't really check whether the user is going to do that. So we can only check for (1, 2, 3 or a multiple of 4) in the code, and the additional restriction when the gradient is required will just have to be added in the docstring.

lasagne/tests/layers/test_conv.py
benanne: This should have been something like "this implementation needs to be tested against `FilterActs` because testing it against `conv2d` leads to numerical precision errors". Later I realised that the tests in `TestConv2DLayerImplementations` (which already existed before) do actually test it against `conv2d` without problems. So I'm not sure where this issue actually came from in the end (but switching to `FilterActs` fixed it). This might be worth further investigation.

lasagne/tests/layers/test_conv.py
f0k: Fine by me to leave it at that for now.

lasagne/tests/layers/test_pool.py
benanne: Afaik this `start` parameter is not properly implemented in the `pylearn2` wrapper. iirc it exists, but it is ignored (would have to check again to be sure).

lasagne/tests/layers/test_conv.py
f0k: Sounds good!

benanne:
I think this is ready.

f0k:
Okay, all my comments were addressed and Travis passes, so I'm merging. Thanks!

PR #268, opened 2015-05-25
Coverage for pool.py, helper.py

ebenolson:
Adds a few tests to complete coverage for pool.py, and adds `pragma no cover` for the deprecated helper functions.

### DIFF
--- a/lasagne/layers/helper.py
+++ b/lasagne/layers/helper.py
@@ -116,7 +116,7 @@ def get_all_layers(layer, treat_as_input=None):
     return result
 
 
-def get_all_layers_old(layer):
+def get_all_layers_old(layer):  # pragma no cover
     """
     Earlier implementation of `get_all_layers()` that does a breadth-first
     search. Kept here to ease converting old models that rely on the order of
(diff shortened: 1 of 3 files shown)


ebenolson:
`helper.py` has full coverage except for https://github.com/ebenolson/Lasagne/blob/more_tests/lasagne/layers/helper.py#L302

I couldn't figure out how to trigger this, and it looks like it was copy-pasted from get_output. Is it a necessary check?

f0k:
> Is it a necessary check?

Good catch! The `try..except` clause around that is unneeded, it's impossible for a layer not to have a defined input shape. It seems getting the test coverage to 100% may uncover things that would go unnoticed at 90% :)

benanne:
We're already well into the 90% range thanks to @ebenolson's work, so we might as well go all the way!  Thanks a lot for this, I'll have a closer look tomorrow :)

benanne:
No further comments from me, I'm good with merging. Feel free to remove that `try..except` as well :)

ebenolson:
Great, made that change.

benanne:
Merging

PR #269, opened 2015-05-25
Change stacklevel for deprecation warnings to point at function call

f0k:
This sets `stacklevel=2` for the `get_output`, `get_output_shape` and `get_bias_params` deprecation warnings so they point at the offending deprecated function call (i.e., the point in code where you call a deprecated function) rather than the `warnings.warn` call (i.e., the point in Lasagne where we raise the warning). This makes it much easier to track down any remnants of the good old days, err, the old API in user code.


benanne:
Cool, merging!

PR #270, opened 2015-05-26
Make lasagne.utils.create_param() accept iterables as shapes

f0k:
This changes `lasagne.utils.create_param()` to accept arbitrary iterables as shapes, not only tuples. Before, when giving a `list` as a shape, the function would fail with an unhelpful "the provided callable did not return a value with the correct shape", because the list was later compared against a tuple. (I'd also be fine with a meaningful error message, but adding `shape = tuple(shape)` is easy enough and more convenient.)

Besides, I've changed the "Notes" section, which still assumed that `create_param()` was a `Layer` method.

### DIFF
--- a/lasagne/utils.py
+++ b/lasagne/utils.py
@@ -214,9 +214,8 @@ def create_param(spec, shape, name=None):
 
     Parameters
     ----------
-
     spec : numpy array, Theano shared variable, or callable
-        One of three things:
+        Either of the following:
 
         * a numpy array with the initial parameter values
         * a Theano shared variable representing the parameters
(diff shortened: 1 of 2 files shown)


lasagne/utils.py
```diff
@@ -244,11 +243,13 @@ def create_param(spec, shape, name=None):
 
     Notes
     -----
-    This method should be used in the constructor when creating a
-    :class:`Layer` subclass that has parameters. This enables the layer to
+    This method is called by :meth:`Layer.add_param()` in the constructor
```
benanne: Maybe we should change 'method' to 'function' as well, since technically a method is part of a class.

benanne:
Good catch, thanks for adding a test as well :) Feel free to merge.

lasagne/utils.py
f0k: Ah, of course. Will do, amend and merge (no need to wait for Travis again).

PR #272, opened 2015-05-26
Add lasagne.__version__ which mirrors the package's version.

dnouri:
As per the discussion in #263, this updates the version in setup.py to be pep440 compatible, and adds `lasagne.__version__`, which mirrors the package's version in setup.py.

I vaguely remember having some problems with `pkg_resources.get_version()` and readthedocs, but can't remember what it was.  I assume the Lasagne package is set up normally with readthedocs.  `get_version` would fail if it weren't.

### DIFF
--- a/lasagne/__init__.py
+++ b/lasagne/__init__.py
@@ -9,3 +9,8 @@ from . import objectives
 from . import regularization
 from . import updates
 from . import utils
+
+
+import pkg_resources
+__version__ = pkg_resources.get_distribution("Lasagne").version
+del pkg_resources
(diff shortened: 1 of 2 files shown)


lasagne/__init__.py
```diff
@@ -9,3 +11,6 @@
 from . import regularization
 from . import updates
 from . import utils
+
+
+__version__ = pkg_resources.get_distribution("Lasagne").version
```
f0k: Maybe add a `del pkg_resources` afterwards, so we don't export `lasagne.pkg_resources`? I like clean namespaces.

dnouri:
@f0k addressed your comment

f0k:
Great, thanks! I'd be fine with merging and hoping for the best on readthedocs.

benanne:
Alright, let's give it a go :)

Issue #6, opened 2014-09-11
Reproducibility, randomness


benanne:
For the sake of reproducibility It should be possible to run the same piece of code twice and get the exact same result.

In practice this means that any step involving randomness should be able to reuse the same seeds. The cleanest way to achieve this is to allow for custom RNG objects to be used for anything that involves randomness.

An interesting complication is that randomness occurs both on the numpy side of things (e.g. parameter initialization, the order in which training examples are presented), and on the Theano side (e.g. dropout and other forms of training noise).

So I'm wondering how we should implement this. We could make every class or method involving randomness take an optional `rng` argument, but that might get messy. Maybe it's better if it was possible to control this globally? I haven't really thought about this yet.

cancan101:
Right now for dropout, you are constructing a global `MRG_RandomStreams`:  https://github.com/benanne/Lasagne/blob/e090ad4d9cf650b9f452232e73c1e503b01d9460/lasagne/layers/noise.py#L9. 

This gets reproducibly seeded: https://github.com/Theano/Theano/blob/b320ab29a4dc38b4c0035e330eb7dbaa379249cc/theano/sandbox/rng_mrg.py#L1053.

The init code on the other hand uses numpy and does not appear to be seeded.

benanne:
Yeah, but that rng used for dropout should also be configurable imo. We need to work on that :) The init code also needs to use a shared, configurable rng (or it needs to be possible to pass in a custom rng, which is a little more flexible I suppose, or both).

f0k:
> The init code on the other hand uses numpy and does not appear to be seeded.

But it uses a global rng (`np.random`) just like the noise layers, so you can seed it to get reproducible results (`np.random.seed(...)`).

Still I think being able to pass an rng and srng wouldn't be too bad to have as an alternative option.

cancan101:
One other phenomenon I noticed is that at least using the cuDNN library, there seems to be some non determinism in the results from convolutions.

f0k:
> One other phenomenon I noticed is that at least using the cuDNN library, there seems to be some non determinism in the results from convolutions.

The reference says:
> By design, most of cuDNN API routines from a given version generate the same bit-
wise results at every run when executed on GPUs with the same architecture and
the same number of SMs. However, bit-wise reproducibility is not guaranteed across
versions, as the implementation of a given routine may change. With the current release,
the following routines do not guarantee reproducibility because they use atomic add
operations:
[...]

PR #273, opened 2015-05-28
move srng into noise layers, seed from numpy

ebenolson:
This partially addresses https://github.com/Lasagne/Lasagne/issues/6, which came up on the mailing list again a few days ago (https://groups.google.com/forum/#!topic/lasagne-users/85-6gxygtIo).

By making `_srng` a member of each layer which is initialized from numpy, the user only needs to call `np.random.seed` at the start of their code to get reproducible behavior.

### DIFF
--- a/lasagne/layers/noise.py
+++ b/lasagne/layers/noise.py
@@ -1,10 +1,9 @@
 import theano
+import numpy as np
 
 from .base import Layer
 
-# from theano.tensor.shared_randomstreams import RandomStreams
 from theano.sandbox.rng_mrg import MRG_RandomStreams as RandomStreams
-_srng = RandomStreams()
 
 
 __all__ = [
@@ -14,7 +13,15 @@ __all__ = [
 ]
 
 
-class DropoutLayer(Layer):
+class NoiseLayer(Layer):
+    """Base class for stochastic layers"""
+    def __init__(self, incoming, **kwargs):
+        super(NoiseLayer, self).__init__(incoming, **kwargs)
+
+        self._srng = RandomStreams(np.random.randint(1, 2147462579))
+
+
+class DropoutLayer(NoiseLayer):
     """Dropout layer [1]_,[2]_.
 
     Sets values to zero with probability p. See notes for disabling dropout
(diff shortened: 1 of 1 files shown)


f0k:
+1. This also circumvents the memory leak Colin found when instantiating a lot of dropout layers and discarding them (https://github.com/Theano/Theano/issues/2335#issuecomment-66855136).

f0k:
Thinking about it, do we really want the base class for just one line in the constructor? If we have a base class, we probably need to put it in `__all__` as well and give it a proper docstring that explains what it's for and that you shouldn't instantiate it. Right now it wouldn't even be included in Sphinx, but you would see that the two noise layers are derived from an undocumented `NoiseLayer` class. Maybe it's clearer for now to just instantiate `self._rng` in the two noise layer constructors directly.

ebenolson:
That's true, I was initially contemplating adding a method to set the seed, and it could simplify a potential future move away from the numpy internal rng. But with only two noise layers so far it's not really needed.

I also hoped it would help avoid situations where someone creates new stochastic layers that aren't seeded the same way.

f0k:
> That's true, I was initially contemplating adding a method to set the seed, and it could simplify a potential future move away from the numpy internal rng.

Okay, it would help stochastic layers outside of the Lasagne repository to seamlessly support different seeding methods if we ever extend the `NoiseLayer` base class in this respect. If we want that, we should document that the purpose of the `NoiseLayer` is more than saving a single line of code in the constructor. Right now I fear it mostly serves to obfuscate the code.

> I also hoped it would help avoid situations where someone creates new stochastic layers that aren't seeded the same way.

I think when somebody creates a new stochastic layer, that somebody will probably copy & paste the dropout layer or gaussian noise layer as a starting point because it explains the `deterministic` keyword argument. Without a `NoiseLayer` base class, this code is fully self-contained, and it's clear how to change the RNG if that's useful.

> But with only two noise layers so far it's not really needed.

Right. I think I'd prefer to drop it, because it makes the code slightly more difficult to understand. We can still add it back if we extend the API to allow users to specify a custom seeding RNG. But in general, I very much appreciate the foresight! A good `NoiseLayer` docstring would convince me as well, if you prefer to keep it!

ebenolson:
Let's go with simple for now.

 I think it would also be good to add a section to the docs somewhere on reproducibility, I guess `tutorial.rst` would be the place? Or could start a FAQ section.

f0k:
Looks good, thanks! I'll leave this up for Sander to merge in case he's got any comment.

> I think it would also be good to add a section to the docs somewhere on reproducibility, I guess tutorial.rst would be the place? Or could start a FAQ section.

Probably both places -- part of the Tutorial / the included MNIST example (that we don't have yet), and part of the FAQ (that we don't have yet). We'll keep that in mind.

benanne:
lgtm, merging

PR #274, opened 2015-05-29
Use OrderedDict for ADAM updates to be compatible with other updates.

JeffreyDF:
You couldn't apply momentum to ADAM updates because it still used a list for the updates. This converts the list to an OrderedDict.


f0k:
Oh, good catch. I guess we need more reviewers! Merging directly because it is a bugfix. We should extend the update tests with a test for the correct return type, though (#112).

JeffreyDF:
Thanks for the quick merge!

PR #275, opened 2015-05-29
remove travis_wait

ebenolson:
`travis_wait` isn't needed since none of the tests exceed 10 minute runtime, and its 20 minute timeout caused the last build to fail.

Off topic but does anyone know why 3.4 is so much slower than 2.7?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,5 +14,5 @@ install:
   - travis_retry conda install --yes python=$TRAVIS_PYTHON_VERSION pip numpy scipy
   - travis_retry pip install -r requirements.txt
   - travis_retry python setup.py dev
-script: travis_wait py.test --runslow
+script: py.test --runslow
 cache: apt


dnouri:
Can you explain how `travis_wait` caused the build to fail?

ebenolson:
`travis_wait` kills the `py.test --runslow` command if it exceeds 20 minutes runtime.

dnouri:
Oh, I thought that was 20 minutes runtime without output; must've misread.

f0k:
Is that killing behavior documented somewhere? From Travis' [documentation on timeouts](http://docs.travis-ci.com/user/build-timeouts/), it seems there is a timeout of 10 minutes without output, and `travis_wait` works around that by producing a single output line every minute for 20 minutes. It does not say that it kills the process afterwards. On the other hand, the global timeout for Travis jobs seems to be [50 minutes](http://docs.travis-ci.com/user/build-configuration/#Job-Timeouts) for public projects, so it probably shouldn't have killed the job that early.
In any case, I guess we can safely drop `travis_wait` because we shouldn't have any single test exceed ten minutes, even including Theano compile time.

ebenolson:
@f0k I didn't find much documentation on `travis_wait`, but the source is here: https://github.com/travis-ci/travis-build/blob/dd8234930168f302d625ece446c288548da66512/lib/travis/build/templates/header.sh#L108

f0k:
Ah, nice. So if we ever hit the "10-minutes-without-output" case, we can just reintroduce `travis_wait`, but give it a higher timeout than the default of 20 (like `travis_wait 50 py.test --runslow`).

PR #276, opened 2015-05-30
add total_norm_constraint to updates.rst

skaae:
add total_norm_constraint to updates.rst  and adds a few missing dots.

### DIFF
--- a/lasagne/updates.py
+++ b/lasagne/updates.py
@@ -7,28 +7,37 @@ rate for use with stochastic gradient descent.
 Update functions take a loss expression or a list of gradient expressions and
 a list of parameters as input and return an ordered dictionary of updates:
 
-* :func:`sgd()`
-* :func:`momentum()`
-* :func:`nesterov_momentum()`
-* :func:`adagrad()`
-* :func:`rmsprop()`
-* :func:`adadelta()`
-* :func:`adam()`
+.. autosummary::
+    :nosignatures:
+
+    sgd
+    momentum
+    nesterov_momentum
+    adagrad
+    rmsprop
+    adadelta
+    adam
 
 Two functions can be used to further modify the updates to include momentum:
 
-* :func:`apply_momentum()`
-* :func:`apply_nesterov_momentum()`
+.. autosummary::
+    :nosignatures:
+
+    apply_momentum
+    apply_nesterov_momentum
 
 Finally, we provide two helper functions to constrain the norm of tensors:
 
-* :func:`norm_constraint()`
-* :func:`total_norm_constraint()`
+.. autosummary::
+    :nosignatures:
+
+    norm_constraint
+    total_norm_constraint
 
 :func:`norm_constraint()` can be used to constrain the norm of parameters
 (as an alternative to weight decay), or for a form of gradient clipping.
-:func:`total_norm_constraint() constrain the total norm of a list of tensors.
-This is often used when training recurrent neural networks.`
+:func:`total_norm_constraint()` constrain the total norm of a list of tensors.
+This is often used when training recurrent neural networks.
 
 Examples
 --------
@@ -74,7 +83,7 @@ __all__ = [
 
 
 def get_or_compute_grads(loss_or_grads, params):
-    """Helper function returning a list of gradients.
+    """Helper function returning a list of gradients
 
     Parameters
     ----------
@@ -102,7 +111,7 @@ def get_or_compute_grads(loss_or_grads, params):
 
 
 def sgd(loss_or_grads, params, learning_rate):
-    """Stochastic Gradient Descent (SGD) updates.
+    """Stochastic Gradient Descent (SGD) updates
 
     Generates update expressions of the form:
 
@@ -132,7 +141,7 @@ def sgd(loss_or_grads, params, learning_rate):
 
 
 def apply_momentum(updates, params=None, momentum=0.9):
-    """Returns a modified update dictionary that includes momentum terms.
+    """Returns a modified update dictionary including momentum
 
     Generates update expressions of the form:
 
@@ -180,7 +189,7 @@ def apply_momentum(updates, params=None, momentum=0.9):
 
 
 def momentum(loss_or_grads, params, learning_rate, momentum=0.9):
-    """Stochastic Gradient Descent (SGD) updates with momentum.
+    """Stochastic Gradient Descent (SGD) updates with momentum
 
     Generates update expressions of the form:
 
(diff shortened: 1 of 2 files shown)


f0k:
Good catch!

If you feel like cleaning up [updates](http://lasagne.readthedocs.org/en/latest/modules/updates.html) some more, there's something wrong with the `total_norm_constraint()` listing in the beginning (a missing backtick?), and it would be cool to replace all the lists in the beginning with generated overview tables as in [nonlinearities](http://lasagne.readthedocs.org/en/latest/modules/nonlinearities.html). They should still be divided into three categories, so they can just become three tables, but we can omit the extra descriptions for `norm_constraint()` and `total_norm_constraint()` then (as they will be part of the last table).

Would you like to tackle that as well? Otherwise I can take care of it some time later.

skaae:
fixed the link. 

Would you like the table to be generated automatically? You can do that with something like:

```
#partial example...
.. autosummary::

   sgd
   momentum

here goes a description...

.. autosummary::

   nesterov_momentum
   adagrad

```

But I think that the updates currently looks much better than the auto generated tables.

f0k:
> Would you like the table to be generated automatically?

Yes, exactly as in [nonlinearities](http://lasagne.readthedocs.org/en/latest/modules/nonlinearities.html), which also uses autosummary: [nonlinearities.rst](https://raw.githubusercontent.com/Lasagne/Lasagne/master/docs/modules/nonlinearities.rst)

> But I think that the updates currently looks much better than the auto generated tables.

Hmm, I think the lists waste a lot of space that could be used for some additional information, and tables would be more in line with the nonlinearities module and the layers module (once we beautified that).

f0k:
Thinking again, is the `autosummary` directive even supported in module docstrings, or does it only work for `.rst` files? For `updates.py`, replacing the list with a `autosummary` wouldn't make the module docstring any less informative or harder to read in the terminal, but I don't know if Sphinx catches up the directive outside of `.rst` files.

skaae:
You can modify updates.rst to something like

https://gist.github.com/skaae/51390e25ff5f456f0a8f

Look at the raw file. 

Inserting the autosummary directly in the docstring of the py file is working. I prefer that option.

f0k:
> Inserting the autosummary directly in the docstring of the py file is working.

Cool!

> I prefer that option.

Yes, me too. It would make the HTML documentation nicer without impacting the readability/detail of the raw docstrings (as accessible from a terminal). Would be great if you could add that change to the PR here! (Otherwise I'll do it later, as you wish.)

skaae:
nice. I can add it tomorrow.

skaae:
Update PR to use autosummary. I also fixed a few errors in the total_norm_constraint doc

.idea/nntools.iml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="PYTHON_MODULE" version="4">
+  <component name="NewModuleRootManager">
+    <content url="file://$MODULE_DIR$" />
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+</module>
```
f0k: That shouldn't be part of Lasagne. Can you amend your last commit to not include this file?

lasagne/updates.py
```diff
@@ -132,8 +138,9 @@ def sgd(loss_or_grads, params, learning_rate):
 
 
 def apply_momentum(updates, params=None, momentum=0.9):
-    """Returns a modified update dictionary that includes momentum terms.
+    """Applies momentum
```
f0k: This is just copying the function name now -- was the previous version too long for the table? In that case, we should probably aim for something in between. I'll check how this PR renders in Sphinx and try to come up with something.

.idea/nntools.iml
skaae: whoops. I must have forgotten to uncheck it in the github app.

lasagne/updates.py
skaae: Yes. If the table gets too wide it'll get a scrollbar which look terrible. I changed the names slightly, maybe its better now.

skaae:
Removed the wrongly included file and changed the names of apply_momentum and apply_nesterow_momentum slightly.

skaae:
We can include the `:nosignatures:` command. It makes the tables more clean:

```
.. autosummary::
    :nosignatures:
```

skaae:
`EmbeddingLayer` is also missing from layers.rst. Should i add it here or do another PR?

benanne:
Maybe make a separate PR for that because we can merge that immediately.

skaae:
PR in https://github.com/Lasagne/Lasagne/pull/278

f0k:
> We can include the `:nosignatures:` command.

Good find. I've tested it myself and I'd prefer that here (but not in `nonlinearities`). This also creates more space for the short descriptions, maybe you can set `norm_constraint()` and `total_norm_constraint()` back to their original descriptions now?

Another thing I've noticed is that it does not create the correct reference links now. We either shouldn't have reference links in the one-line summary, or we should give them unique names instead of calling them `[1]` and `[2]` every time. The former is easier, and more in line with numpy's recommendations. So the references for adagrad etc. should be moved to the second sentence of the docstring.

skaae:
I get multiple defined referece warnings. Maybe that is the reason for the non working links?

Anyway I think we could just remove the references from the table. 
I'll use the shorter table and use the old descriptions.

f0k:
> I get multiple defined referece warnings. Maybe that is the reason for the non working links?

I get them, too, but usually across different files. I'm not sure how the links are supposed to be named -- numpy suggests `[1]`, `[2]`, but Sphinx seems to expect globally unique references such as `[HD]`. Maybe there's something wrong with our setup. The backlinks from the reference to the occurrence don't work correctly either, it always creates an extra backlink so we got the `(1, 2)` syntax even if the reference only occurred once.

> Anyway I think we could just remove the references from the table.

Yes, that will require moving them from the first class docstring sentence to the second.

skaae:
Updated with what we discussed above and squashed the commits.

benanne:
Nice! I have no additional comments here, so I'll let f0k merge this when he's happy ;)

lasagne/updates.py
```diff
@@ -574,6 +585,7 @@ def adam(loss_or_grads, params, learning_rate=0.001, beta1=0.9,
 def norm_constraint(tensor_var, max_norm, norm_axes=None, epsilon=1e-7):
     """Max weight norm constraints and gradient clipping
 
+    Max weight norm constraints and gradient clipping.
```
f0k: That's duplicated now. Always have a look at the full diff in the end :)

f0k:
Almost happy! Just one line left in the diff that shouldn't be there.

skaae:
Yes i missed that one. One second :)

f0k:
Thanks! Merging, no need to wait for Travis again.

PR #277, opened 2015-05-31
Temperature softmax

krallistic:
My current Implementation for custom temperature softmax. I copied the docstring/class style from the leaky rectify class. Not sure if its matches Lasagnes focus through.

Still have to find a reference for a low temperature softmax, i can remember seeing it somewhere

### DIFF
--- a/lasagne/nonlinearities.py
+++ b/lasagne/nonlinearities.py
@@ -3,6 +3,7 @@ Non-linear activation functions for artificial neurons.
 """
 
 import theano.tensor.nnet
+import theano.tensor as T
 
 
 # sigmoid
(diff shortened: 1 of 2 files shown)


benanne:
I'm not sure if I understand the use cases for this, could you elaborate?

As far as i can see there is no way to change the temperature after instantiation, except by directly changing the `temperature` attribute on the nonlinearity object. Is that the intended interface?

If the temperature is not meant to be changed after instantiation, it would be functionally equivalent to a regular softmax if used during training (e.g. if you set a temp of 0.1 the learned weights will just be 10 times smaller to compensate for it). As far as I know the point of changing temperatures in a softmax function is to make a model be more / less certain than it learnt to be.

f0k:
Thanks! The recently most popular reference would probably be Hinton's "Distilling Knowledge" paper: http://arxiv.org/abs/1503.02531
Ah wait, you have that one anyway.

I wonder if this should really be expressed as a nonlinearity, though. It would be nicer to have it in a form that allows controlling the temperature with a keyword argument on `get_output`, which means it should be a `Layer` class instead, something like:
```python
class TemperableSoftmaxLayer(Layer):
    def __init__(self, incoming, temperature=1, **kwargs):
        super(TemperableSoftmaxLayer, self).__init__(incoming, **kwargs)
        self.temperature = temperature
    def get_output_for(self, input, **kwargs):
        temperature = kwargs.get("temperature", self.temperature)
        if temperature != 1:
            input *= temperature
        return lasagne.nonlinearities.softmax(input)
```

benanne:
Ah yes, as a layer it makes more sense to me and we can indeed use a keyword argument. I think the convention is to divide by the temperature though, not multiply by it. That way high temperatures give rise to smoother distributions (more uncertainty) which matches the thermodynamic interpretation :)

f0k:
> I think the convention is to divide by the temperature though, not multiply by it.

Yes, I fixed that one minute after posting the comment :)

krallistic:
A Layer is much nicer since, like you mentioned, allows changing the temp, but i wasn't sure if it necessary for this really special case.

As far usage, besides the linked dark knowledge paper from Hinton i know for char based RNN, the temperature is used to create more wild/conservative outputs.

f0k:
> A Layer is much nicer since, like you mentioned, allows changing the temp, but i wasn't sure if it necessary for this really special case.

I think it poses some advantages. Do you want to take my code and turn it into a PR? You can overwrite this branch if you know how, or just ask me how to do it, or start a new PR. We haven't really decided on where to put these kinds of layers, though (i.e., layers that just compute a special kind of nonlinearity), see https://github.com/Lasagne/Lasagne/pull/247#discussion_r30728534 and following.

PR #278, opened 2015-05-31
add embedding layer to layers.rst + fix typo

skaae:


benanne:
Thanks!

lasagne/layers/embedding.py
```diff
@@ -18,7 +18,7 @@ class EmbeddingLayer(Layer):
     Parameters
     ----------
     incoming : a :class:`Layer` instance or a tuple
-        The layer feeding into this layer, or the expected input shape
+        The layer feeding into this layer, or the expected input shape.
```
f0k: We don't have that dot in any of our docstrings, though. I didn't place a dot because it's not a full sentence, but I capitalized the first letter because it didn't look well in Sphinx otherwise. We should probably agree on what we want, but not change it in a single place!

f0k:
Hey, you were too fast.

lasagne/layers/embedding.py
benanne: The current docs are rife with inconsistencies like this one. It bothers me personally because I'm a bit of a perfectionist in that sense, but objectively it's really not that important right now. We should focus on the big things and not spend all of our time 'cleaning things up' right now, if possible :)

skaae: ok. I added it because the other parameter decriptions in embedding layer ended with dot.

benanne:
well, I told him I'd merge it quickly :p that 'typo fix' probably shouldn't have snuck in there though, that was kind of the point of having a separate PR for it.

f0k:
Ah, doesn't matter, checking more docstrings, we have very different conventions. In `lasagne.layers.noise` we have a lower-case `the` and no dot, in `lasagne.layers.conv` we have a capital `T` and a dot, and `DenseLayer` has a capital `T` and no dot. Probably the first of those is the original string I wrote, and the others were copied and altered. Should be cleaned up at some point, but it's not crucial.

benanne:
Yeah, everyone just kind of does what they think is best because we have so many different styles already and no guidelines. We should sort that out eventually, but for now we should prioritize making the docs complete and usable, and all things considered a capital here or a superfluous period there is not going to make a difference. I have to keep telling myself that because as I said, it does bother me a bit :p

PR #280, opened 2015-05-31
use numpy for conv2d test sets

ebenolson:
This replaces the Theano-based conv2d test set with one using numpy.

The multiple calls to `conv2d(...).eval()` were a taking a lot of time, on my machine this change brings the test collection time from 33.59 to 0.51 seconds.

### DIFF
--- a/lasagne/tests/layers/test_conv.py
+++ b/lasagne/tests/layers/test_conv.py
@@ -1,13 +1,29 @@
-from mock import Mock
 import numpy as np
 import pytest
 import importlib
 import theano
-import theano.tensor as T
-from theano.tensor.nnet import conv2d
 
 import lasagne
 from lasagne.utils import floatX
+
+
+def conv2d(input, kernel, border_mode):
+    output = np.zeros((input.shape[0],
+                       kernel.shape[0],
+                       input.shape[2] + kernel.shape[2] - 1,
+                       input.shape[3] + kernel.shape[3] - 1,
+                       ))
+    for i in range(kernel.shape[2]):
+        for j in range(kernel.shape[3]):
+            k = kernel[:, :, i, j][:, :, np.newaxis, np.newaxis]
+            output[:, :, i:i + input.shape[2],
+                   j:j + input.shape[3]] += (input[:, np.newaxis] * k).sum(2)
+
+    if border_mode == 'valid':
+        trim = (kernel.shape[2] - 1, kernel.shape[3] - 1)
+        output = output[:, :, trim[0]:-trim[0], trim[1]:-trim[1]]
+
+    return output
 
 
 def conv2d_test_sets():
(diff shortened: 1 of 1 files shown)


benanne:
Could you maybe elaborate to show that this new set is equivalent to the old? Because I'm finding it hard to see that.

Also, I would prefer if this was merged after #267. It might need a rebase. And maybe we should also modify the additional cuda-convnet tests in this way, they now use `FilterActs` directly, which will probably be slow as well.

ebenolson:
Yeah, this only tests a square separable convolution, so it's not quite as strict as before. Although testing theano against theano isn't that great either. I'll see if I can do a 2d numpy convolution, but I'd rather not have to reimplement scipy.signal.convolve2d.

benanne:
I think I can live with the separable convolutions, but as a consequence this also restricts testing to square input, so we should also consider that. The previous set used non-square input (16, 23). Although this is probably fine too, all things considered.

ebenolson:
Ok, rewrote this to do the same tests using numpy.

lasagne/tests/layers/test_conv.py
```diff
@@ -20,9 +36,7 @@ def _convert(input, kernel, output, kwargs):
         for stride in [1, 2, 3]:
             input = np.random.random((3, 1, 16, 23))
             kernel = np.random.random((16, 1, 3, 3))
-            output = conv2d(input, kernel,
-                            border_mode=conv_mode,
-                            ).eval()
+            output = conv2d(input, kernel, border_mode=conv_mode)
             if border_mode == 'same':
```
benanne: Maybe this special case can now be moved into the `conv2d` function as well? I believe this was done because `T.nnet.conv.conv2d` doesn't support `border_mode='same'`, right?

benanne:
Excellent work, just one remark/question :)

ebenolson:
makes sense, done.

also updated `TestConv2DLayerImplementations` to use `get_output` helper.

benanne:
Thanks! I'm fine with merging, I'll wait a bit in case anyone else has any remarks.

lasagne/tests/layers/test_conv.py
```diff
@@ -191,7 +197,7 @@ def test_defaults(self, Conv2DImpl, DummyInputLayer,
                 W=kernel,
                 **kwargs
             )
-            actual = layer.get_output_for(input).eval()
+            actual = get_output(layer, input).eval()
```
f0k: Hmm, why this? The original version makes clear we're just propagating through this single layer to test its implementation of `get_output_for()`.

f0k:
Looks good! And it's really quite a bit faster at least when starting with an empty compilation cache. I've just got one minor question above.

lasagne/tests/layers/test_conv.py
benanne: That's a good point, I guess there is no need to test more than `get_output_for()` here.

ebenolson: oops, for some reason I was thinking it was using the deprecated method. 

I can change it back, but I should probably change `TestConv1DLayer` as well for consistency.

benanne: Yeah, better change it in both places then :)

ebenolson:
Ok I think that's set, thanks for reviewing.

f0k:
Thanks! :+1: Let's just wait for Travis, then merge.

benanne:
merging!

PR #281, opened 2015-05-31
Add coveralls.io

dnouri:
Apart from registering Lasagne on coveralls.io, this should be all that's needed for python-coveralls to do its work.

Related to #112: Improving test coverage.

### DIFF
--- a/.coveragerc-nogpu
+++ b/.coveragerc-nogpu
@@ -1,0 +1,6 @@
+[run]
+omit =
+    lasagne/tests/*
+    lasagne/layers/corrmm.py
+    lasagne/layers/cuda_convnet.py
+    lasagne/layers/dnn.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,9 @@ before_install:
 install:
   - travis_retry conda install --yes python=$TRAVIS_PYTHON_VERSION pip numpy scipy
   - travis_retry pip install -r requirements.txt
+  - travis_retry pip install python-coveralls
   - travis_retry python setup.py dev
 script: py.test --runslow
+after_success:
+  - coveralls
 cache: apt
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ install:
   - travis_retry pip install -r requirements.txt
   - travis_retry pip install python-coveralls
   - travis_retry python setup.py dev
-script: py.test --runslow
+script: py.test --runslow --cov-config=.coveragerc-nogpu
 after_success:
   - coveralls
 cache: apt
--- a/README.rst
+++ b/README.rst
@@ -4,6 +4,9 @@
 .. image:: https://travis-ci.org/Lasagne/Lasagne.svg?branch=master
     :target: https://travis-ci.org/Lasagne/Lasagne
 
+.. image:: https://img.shields.io/coveralls/Lasagne/Lasagne.svg
+    :target: https://coveralls.io/r/Lasagne/Lasagne
+
 Lasagne
 =======
 


benanne:
Sweet, thanks! Let's give that a try :)

benanne:
Unfortunately it looks like I'm running into this issue: https://github.com/lemurheavy/coveralls-public/issues/539

I'll check back tomorrow and try to add it again.

benanne:
Ok, I've managed to enable it now. I guess there is no real way to test this other than to merge it and see what happens?

f0k:
> I guess there is no real way to test this other than to merge it and see what happens?

I guess you could just [re-run Travis on this one](http://stackoverflow.com/a/17624403) so coveralls gets triggered, but I'd also be fine with just merging it.

dnouri:
Yes, I think if you rebuild in Travis, it should post to this pull request saying coverage was so-and-so.

benanne:
alright, gonna give that a go!

benanne:
One thing we should try to work around is that coveralls will probably report relatively poor coverage numbers because `cuda_convnet`, `corrmm` and `dnn` have no coverage (those tests do not run if no GPU is available). Practically this does not matter since we're aware of this shortcoming, but if people see those poor coverage numbers while browsing the repo it might create a bit of a negative perception.

ebenolson:
Looks like you could upload results from your machine manually: https://github.com/coagulant/coveralls-python/blob/master/coveralls/cli.py

benanne:
That's not really feasible though, unless we set up our own CI server. I don't have any Kepler / Maxwell GPUs to spare for that unfortunately :)

f0k:
Just to record it somewhere, the coverage build for this PR is this one: https://coveralls.io/builds/2703532

What we'd need to achieve would be to exclude `lasagne.layers.cuda_convnet`, `...dnn`, `...corrmm` when run on coveralls, but not when run locally (or exclude them when run without a GPU), and write an extra test for `conv_output_length()` covering this line: https://coveralls.io/builds/2703532/source?filename=lasagne%2Flayers%2Fconv.py#L80

We might be able to add a `.coveragerc` with an `[omit]` section and a nonstandard name so that it is not picked up by `pytest-cov`, then tell the `coveralls` command to use it via `--config_file`.
@dnouri, do you want to try that, or should I send a PR to this PR? We probably don't want to merge the PR as is and have a red badge in the README... Alternatively, we could merge it without the badge and then figure out ways to solve the problem of coveralls.io not having decent GPUs in all their VMs in a separate PR.

ebenolson:
It doesn't have to be CI though, you could just do it occasionally.

benanne:
Does coveralls actually do a separate run to determine coverage? I thought it integrates with Travis and gets the coverage info from there.

benanne:
@ebenolson but running it occasionally won't prevent the red badge from showing up, would it? For that, we would have to run it after every commit / merged PR. Or am I misunderstanding?

ebenolson:
I'm suggesting we remove the travis integration. Then update coverage.io manually using the CLI script.

f0k:
We should aim for CI. At the very least, that gives the benefit of coveralls urging people to add tests to their PRs instead of us.

> Does coveralls actually do a separate run to determine coverage?

No, it doesn't, but it passes the config file name to a `coverage` subclass called `coveralls`:
https://github.com/z4r/python-coveralls/blob/master/coveralls/__init__.py#L84
https://github.com/z4r/python-coveralls/blob/master/coveralls/control.py#L5
It then calls `load()` on that to load the coverage report from `pytest-cov`. The loading step may or may not use the configuration file:
https://bitbucket.org/ned/coveragepy/src/870358d48add0083efc76ba9b66da0d7b8277ea1/coverage/control.py?at=default#cl-604
The `coverage` API documentation doesn't give any details about `load()` either: http://nedbatchelder.com/code/coverage/api.html#api
So we can either continue digging through the code or try it.

f0k:
Got it figured out. @dnouri, will send you a PR in a minute!

f0k:
Looks good to me. Let's see what Travis and Coveralls say.

benanne:
Seems to work, coverage is now reported to be 98.94% which sounds about right :) I think this is ready to merge!

f0k:
Merging, thanks Daniel!

f0k:
@benanne: Is there any way for you to trigger a coverage report for the `master` branch when you log in? We now have a gray badge and clicking it doesn't seem to list any running builds. Maybe it will change once we merge something else into `master`, though.

ebenolson:
I think it will update when https://travis-ci.org/Lasagne/Lasagne/builds/65077542 finishes.

benanne:
Looks like it did :)

f0k:
Yay!

PR #282, opened 2015-06-02
Fix vertical alignment mismatch in sphinx documentation

f0k:
This provides a workaround for https://github.com/snide/sphinx_rtd_theme/issues/149, the first point on the list of documentation layout issues that bothered me: https://github.com/Lasagne/Lasagne/issues/203#issuecomment-103983492. See "Parameters: leakiness" on the following page if you're unsure what I'm talking about: http://lasagne.readthedocs.org/en/latest/modules/nonlinearities.html#lasagne.nonlinearities.LeakyRectify


benanne:
Good idea, looks terrible at the moment. Merging!

f0k:
Hooray, it works!

PR #284, opened 2015-06-02
Add license badge to README

f0k:
This adds a license badge to the README file, making it easier to see the license we're using.


benanne:
Cool :) Any others we should add? :p http://shields.io/

f0k:
No, I don't think so. We might just add one of the download counters once we have a release and enough downloads to boast about it :)

Issue #173, opened 2015-03-20
Objective class is confusing


ebenolson:
In particular the argument `input_layer`, which I think would make more sense if renamed to `output_layer`

Here is the current docstring:

>- input_layer : a `Layer` whose output is the networks prediction
        given its input

While this technically does indicate that `input_layer` should be the output layer of the network, I think something like this would be better:

- output_layer: The `Layer` whose output will be compared to the target.

Also, is there a reason not to specify `target` in the constructor?

benanne:
This is not quite a duplicate of #15, but there is quite some overlap (see my latest comment there).

The reason the argument is `input_layer` is because `Objective` used to be called `OutputLayer` in my galaxy challenge code, which Lasagne is largerly based on. Which made even less sense ;)

Regarding specifying `target` in the constructor, as I commented in the other issue we may just replace `Objective` classes with functions, so then this doesn't matter. But if we keep them as classes, I guess it would be useful to be able to set it in the constructor, for symmetry with `InputLayer`.

benanne:
Thinking about this again, I'm not sure if we should add the abilty to specify the `target` in the constructor. The use case for `Objective` is to avoid having to manually declare Theano variables. So adding the ability to specify a custom target variable kind of defeats the point.

If you wish to use a custom target variable, you can just write something like `loss = T.mean(T.nnet.binary_crossentropy(x, t))` directly. No need to use an `Objective` for that.

Are there any other uses cases for `Objective` instances?

(Renaming `input_layer` to `output_layer` is a good idea regardless which I will implement when I address the additional remarks from #15.)

ebenolson:
Sorry, I haven't tried to use `Objective` yet in my code so I'm a bit hazy on it. I think I suggested that change because looking at the current examples, it seemed like duplication could be reduced by putting those arguments in the constructor:

    loss_train = objective.get_loss(X_batch, target=y_batch)
    loss_eval = objective.get_loss(X_batch, target=y_batch,
                                   deterministic=True)

But it seems like the intended use is more like this?

    loss_train = objective.get_loss()
    loss_eval = objective.get_loss(deterministic=True)
    ...
    ...
    iter_train = theano.function(
        [batch_index], loss_train,
        updates=updates,
        givens={
            input_layer.input_var: dataset['X_train'][batch_slice],
            objective.target_var: dataset['y_train'][batch_slice],
        },
    )

f0k:
[...]

PR #286, opened 2015-06-03
Rework lasagne.objectives (remove Objective and MaskedObjective)

f0k:
This reworks the `lasagne.objectives` module as per our discussion in #173.
Todo list:
* [x] refactor, then deprecate Objective classes
* [x] add module docstring
* [x] add tests
* ~~update occurrences of Objective in the documentation~~ (wasn't mentioned anywhere else)
* [x] update MNIST example

### DIFF
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -123,16 +123,17 @@ def create_iter_functions(dataset, output_layer,
     batch_slice = slice(batch_index * batch_size,
                         (batch_index + 1) * batch_size)
 
-    objective = lasagne.objectives.Objective(output_layer,
-        loss_function=lasagne.objectives.categorical_crossentropy)
+    output = lasagne.layers.get_output(output_layer, X_batch)
+    loss_train = lasagne.objectives.categorical_crossentropy(output, y_batch)
+    loss_train = loss_train.mean()
 
-    loss_train = objective.get_loss(X_batch, target=y_batch)
-    loss_eval = objective.get_loss(X_batch, target=y_batch,
-                                   deterministic=True)
+    output_test = lasagne.layers.get_output(output_layer, X_batch,
+                                            deterministic=True)
+    loss_eval = lasagne.objectives.categorical_crossentropy(output_test,
+                                                            y_batch)
+    loss_eval = loss_eval.mean()
 
-    pred = T.argmax(
-        lasagne.layers.get_output(output_layer, X_batch, deterministic=True),
-        axis=1)
+    pred = T.argmax(output_test, axis=1)
     accuracy = T.mean(T.eq(pred, y_batch), dtype=theano.config.floatX)
 
     all_params = lasagne.layers.get_all_params(output_layer)
--- a/docs/modules/objectives.rst
+++ b/docs/modules/objectives.rst
@@ -3,6 +3,25 @@
 
 .. automodule:: lasagne.objectives
 
+
+Loss functions
+--------------
+
+.. autofunction:: binary_crossentropy
+.. autofunction:: categorical_crossentropy
+.. autofunction:: squared_error
+
+
+Aggregation functions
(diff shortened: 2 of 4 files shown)


f0k:
Ready for reviewing and merging! (Any code using `Objective` or `MaskedObjective` works as before, but remember to notify mailing list when merged.)

benanne:
Nice, I'll try to review it in detail this weekend.

Any reason you didn't include `make_var_like()` in one go? Seems like it would only add a few lines? But fair enough, we can separate it if you prefer it that way. Hopefully we won't forget about it though.

f0k:
> Any reason you didn't include `make_var_like()` in one go? Seems like it would only add a few lines?

The current example advocates just using `theano.tensor.matrix()` right away, and I don't necessarily think we should promote using `make_var_like()` -- it only works when the targets are the same shape as the outputs. So I'd just have added `make_var_like()` without changing the example, but then it's fully independent of what this PR does.

I'll be abroad from tomorrow until June 15, so please feel free to merge when you reviewed it! The only thing I'm unsure about is whether we want `weighted_aggregate` to be a separate function. We could just call that one `aggregate` and remove the simpler one. The `weights` parameter is optional anyway.

examples/mnist.py
```diff
@@ -123,16 +123,17 @@ def create_iter_functions(dataset, output_layer,
     batch_slice = slice(batch_index * batch_size,
                         (batch_index + 1) * batch_size)
 
-    objective = lasagne.objectives.Objective(output_layer,
-        loss_function=lasagne.objectives.categorical_crossentropy)
+    output = lasagne.layers.get_output(output_layer, X_batch)
+    loss_train = lasagne.objectives.categorical_crossentropy(output, y_batch)
+    loss_train = lasagne.objectives.aggregate(loss_train, 'mean')
```
benanne: I don't really see the value here of using `aggregate` instead of just writing `loss_train = T.mean(loss_train)`. If I understand correctly it does exactly the same thing anyway. Sure, `aggregate` is more flexible, but since we just want the mean here, why not write that directly so people realise everything is just passing around Theano expressions that they can do with as they please.

benanne:
I think getting rid of `aggregate` and renaming `weighted_aggregate` to `aggregate` is a good idea. As it stands, this function does nothing but wrap either `T.mean` or `T.sum`, and needlessly wrapping Theano is something we said we wouldn't do :)

One situation where it is useful is when you want to switch between different aggregation modes, then it's just a question of changing the `mode` argument, which is nice. But for that argument to make sense, `weighted_aggregate` should be the main function anyway.

That's really all I can think of, the docstrings and examples are great and the tests look good as well! Not sure what to do now since you'll be away for a week... maybe I'll amend it myself before merging. Probably not worth waiting till you're back.

f0k:
> Not sure what to do now since you'll be away for a week... maybe I'll amend it myself before merging. Probably not worth waiting till you're back.

What's the state on this? Shall I rebase, remove `aggregate` and rename `weighted_aggregate` to `aggregate`?

benanne:
Go for it, I haven't done anything.

f0k:
Rebased, changed MNIST example to just call `mean()` directly instead of using `aggregate()`, and merged `weighted_aggregate()` into `aggregate()`. Ready to merge when Travis passed. This should also raise coverage to 100%.

f0k:
Travis passed. @benanne, please merge when you're fine with it!

f0k:
Ah wait, missed an occurrence of `weighted_aggregate`...

f0k:
Ready to merge again, this time for real.

benanne:
Yay!

benanne:
Are you doing the [interface-change] post or should I do it? :)

f0k:
Will do it.

PR #289, opened 2015-06-07
Add requirements-dev.txt for testing requirements.

dnouri:
Necessary because different versions of pep8 will give different results.  See #271.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,9 @@
 -r requirements.txt
 mock
+numpydoc
 pep8==1.6.2
 pytest
 pytest-cov
 pytest-pep8
+Jinja2==2.7.3
+Sphinx==1.2.3
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,0 +1,6 @@
+-r requirements.txt
+mock
+pep8==1.6.2
+pytest
+pytest-cov
+pytest-pep8
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
   - conda update --yes conda
 install:
   - travis_retry conda install --yes python=$TRAVIS_PYTHON_VERSION pip numpy scipy
-  - travis_retry pip install -r requirements.txt
+  - travis_retry pip install -r requirements-dev.txt
   - travis_retry pip install python-coveralls
   - travis_retry python setup.py dev
 script: py.test --runslow --cov-config=.coveragerc-nogpu
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,0 @@
-Jinja2==2.7.3
-Sphinx==1.2.3
-#sphinx-bootstrap-theme
-numpydoc
-mock
--- a/docs/user/development.rst
+++ b/docs/user/development.rst
@@ -41,7 +41,8 @@ Tools
 
 * `pytest <http://pytest.org/>`_ for unit testing
 
-* Install testing dependencies with: ``pip install -r requirements-dev.txt``
+* Install dependencies for testing and documentation with: ``pip
+  install -r requirements-dev.txt``
 
 * GitHub for hosting the source code
 
@@ -80,12 +81,6 @@ Documentation
 -------------
 
 The documentation is generated with `Sphinx <http://sphinx-doc.org/latest/index.html>`_.
-On Debian derivates it can be installed with
-
-.. code:: bash
-
-    $ sudo apt-get install python-sphinx
-    $ sudo -H pip install numpydoc
 
 Sphinx makes use of `reStructured Text <http://openalea.gforge.inria.fr/doc/openalea/doc/_build/html/source/sphinx/rest_syntax.html>`_
 
--- a/docs/user/development.rst
+++ b/docs/user/development.rst
@@ -39,11 +39,12 @@ contribute.
 Tools
 -----
 
-* ``pytests`` for unit testing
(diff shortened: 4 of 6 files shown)


requirements-dev.txt
```diff
@@ -0,0 +1,6 @@
+-r requirements.txt
+mock
+pep8==1.6.2
```
f0k: How about a less strict `>=1.6.0`? The behaviour for E126 was changed in version 1.6.0 already: https://github.com/jcrocholl/pep8/blob/master/CHANGES.txt#L68
Of course, if we want to stay in perfect sync with Travis, we should probably require the latest released version and update the requirements file whenever a new pep8 version is released. Not sure what the best strategy is, you seem to be more versed in this.

requirements-dev.txt
```diff
@@ -0,0 +1,6 @@
...
+pytest
+pytest-cov
+pytest-pep8
```
f0k: We might want to list `sphinx` and `sphinx_rtd_theme` as well. Developers will usually need that to ensure they don't commit erroneous markup.

requirements-dev.txt
dnouri: Note that in `.travis.yml`, we're now using this `requirements-dev.txt` so local and Travis will have the same versions.

I can see your point about being less strict with the version requirement, but consider that 1.7.0 might introduce breaking changes again, and then we're left with the same problem again.

requirements-dev.txt
dnouri: There's actually a `docs/requirements.txt` as well.  I guess I'll try and consolidate the two then.

requirements-dev.txt
f0k: > Note that in `.travis.yml`, we're now using this `requirements-dev.txt` so local and Travis will have the same versions.

Ah, sorry, missed that we have full control over the environment for Travis.

> but consider that 1.7.0 might introduce breaking changes again

That's true. I guess requiring 1.6.2 is fine then.

dnouri:
Heads up: The last commit may require some fiddling with readthedocs.  It merges the requirements files for development and docs (that's `requirements-dev.txt` and `docs/requirements.txt`).

For now, I've moved the exact requirements over.  I'm actually not sure why sphinx_rtd_theme was commented out in `docs/requirements.txt`.  Anyone remember?

requirements-dev.txt
f0k: Ah, wasn't aware of that. Sorry, I'm not quite sure if they should be merged then -- maybe there are users who want to render the documentation, but who are not developers? But I'm fine with either.

f0k:
> I'm actually not sure why sphinx_rtd_theme was commented out in `docs/requirements.txt`. Anyone remember?

What's commented out is the `sphinx-bootstrap-theme`. I commented that out when switching to the default theme. When switching to the RTD theme, I saw that this is actually build into Sphinx: http://sphinx-doc.org/theming.html#builtin-themes

requirements-dev.txt
dnouri: I'm fine with either, too.  I guess the few extra requirements are probably okay for people who only care about docs but not pytest.

dnouri:
> What's commented out is the sphinx-bootstrap-theme. I commented that 
> out when switching to the default theme. When switching to the RTD 
> theme, I saw that this is actually build into Sphinx: 
> http://sphinx-doc.org/theming.html#builtin-themes
> 
Oh, woops.  OK, I guess we can remove the commented out 
sphinx-bootstrap-theme then.  :-)

f0k:
> OK, I guess we can remove the commented out sphinx-bootstrap-theme then. :-)

Yeah, probably. I've kept all options in `docs/conf.py` and so I kept that option in the requirements file as well, but without any comment, it doesn't really serve any purpose other than confusing us :)

> Heads up: The last commit may require some fiddling with readthedocs. It merges the requirements files for development and docs (that's `requirements-dev.txt` and `docs/requirements.txt`).

So maybe we should keep them separate, and just change the documentation for rendering the documentation to use `docs/requirements.txt`? And possibly have `requirements-test.txt` instead of `requirements-dev.txt`? I'm fine with anything, just throwing in some options.

dnouri:
> Yeah, probably. I've kept all options in docs/conf.py and so I kept 
> that option in the requirements file as well, but without any 
> comment, it doesn't really serve any purpose other than confusing us 
> :)
> 
If someone needs to remember those options, they should refer to the 
Git history.
>> Heads up: The last commit may require some fiddling with 
>> readthedocs. It merges the requirements files for development and 
>> docs (that's requirements-dev.txt and docs/requirements.txt).
>> 
> So maybe we should keep them separate, and just change the 
> documentation for rendering the documentation to use 
> docs/requirements.txt? And possibly have requirements-test.txt 
> instead of requirements-dev.txt? I'm fine with anything, just 
> throwing in some options.
> 
I guess two is less than three, and so I'm oh-so-slightly in favor of 
having two of these beasts instead of three.  I don't think many 
non-developers will want to render docs locally.

benanne:
Since @f0k said he's okay with anything I guess we're good to merge this? Or does this need any further work? Just checking to make sure.

dnouri:
Yes, it's good to merge.  readthedocs probably needs to be configured to use `requirements-dev.txt` now instead of `docs/requirements.txt`.  See https://github.com/Lasagne/Lasagne/pull/289#issuecomment-109799954

benanne:
Yep, just did that. So I'll merge this and we'll see what happens :)

PR #290, opened 2015-06-07
Fix github source code links generated by Sphinx

f0k:
Fixes the issue noted in https://github.com/Lasagne/Lasagne/pull/262#issuecomment-109764518: The source code file was not inferred correctly for `lasagne.layers.*` members. In addition, changes the github links to highlight the full range of the entity rather than just the first line. (This is something I should add to numpy as well... after my vacation.)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -108,26 +108,27 @@ pygments_style = 'sphinx'
 
 # Resolve function for the linkcode extension.
 def linkcode_resolve(domain, info):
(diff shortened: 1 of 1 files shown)


benanne:
Sweet! I'll merge this now since the links are currently broken on readthedocs.

benanne:
This is awesome, should be very useful for potential contributors as well! Nice work!

PR #293, opened 2015-06-09
Make regularization docs show up

benanne:
The PR with the reworked regularization code (#285) did not include the necessary files to make the module show up in the docs. This PR adds rudimentary versions of those. It's probably useful to flesh this out a bit, but let's do that at a later date.


benanne:
I will merge this so there is at least some documentation available, we can flesh this out later if need be.

PR #295, opened 2015-06-11
Check that parameter shapes match value shapes when loading weights

hjweide:
Recently I accidentally tried to load saved network parameters from a model with different numbers of units in the dense layers.  This passed by silently and I could not figure out why my network was not improving with more hidden units.  I think an error should be raised when loading saved weights changes the network structure.  This is closely related to https://github.com/Lasagne/Lasagne/pull/177.

### DIFF
--- a/lasagne/layers/helper.py
+++ b/lasagne/layers/helper.py
@@ -474,6 +474,7 @@ def set_all_param_values(layer, values, **tags):
     values : list of numpy.array
         A list of numpy arrays representing the parameter values, must match
         the number of parameters.
+        Every parameter's shape must match the shape of its new value.
 
     **tags (optional)
         tags can be specified to filter the list of parameters to be set.
@@ -486,7 +487,8 @@ def set_all_param_values(layer, values, **tags):
     Raises
     ------
     ValueError
-        If the number of values is not equal to the number of params.
+        If the number of values is not equal to the number of params, or
+        if a parameter's shape does not match the shape of its new value.
 
     Examples
     --------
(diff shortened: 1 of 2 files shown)


lasagne/layers/helper.py
```diff
@@ -503,5 +503,10 @@ def set_all_param_values(layer, values, **tags):
     if len(params) != len(values):
         raise ValueError("mismatch: got %d values to set %d parameters" %
                          (len(values), len(params)))
+
     for p, v in zip(params, values):
-        p.set_value(v)
+        if p.get_value().shape != v.shape:
+            raise ValueError("mismatch: parameter has shape %r but value to set \
+                              has shape %r" % (p.get_value().shape, v.shape))
```
benanne: It's probably better to do this instead:

```
raise ValueError("mismatch: parameter has shape %r but value to set "
                 "has shape %r" % (p.get_value().shape, v.shape))
```

Python will concatenate adjacent strings. Otherwise won't the string have a bunch of spaces in it because of the indentation? Either way, that's how we do it elsewhere in the code so if it doesn't make a difference, it would be nice to change it just for consistency's sake.

benanne:
Good call! Thanks for adding the tests in one go as well :)

I was trying to think of a case where `p.get_value().shape` would not correspond to the actual required parameter shape, but I can't think of any. That can only be the case if a user manually overrides the value of a parameter variable, in which case they are on their own anyway :)

One small remark, otherwise this looks good to merge to me. I'll leave it open for a bit so maybe someone else could have a look at it as well (although @f0k isn't around at the moment).

hjweide:
I agree that consistency is good and so I have made the requested changes.  I have also updated the function's documentation.

benanne:
If there are no further comments from anyone else I will merge this tomorrow.

benanne:
Merging!

f0k:
:+1:! Didn't know that Theano allows changing the shape of a shared variable in `set_value()`.

PR #301, opened 2015-06-12
add ElemwiseMergeLayer

skaae:
Add ElemwiseMergeLayer and ElemwiseSumLayer

Im not sure that i added ElemwiseSumLayer correctly to layers.rst

### DIFF
--- a/lasagne/layers/merge.py
+++ b/lasagne/layers/merge.py
@@ -6,6 +6,7 @@ from .base import MergeLayer
 __all__ = [
     "ConcatLayer",
     "concat",
+    "ElemwiseMergeLayer",
     "ElemwiseSumLayer",
 ]
 
@@ -40,13 +41,53 @@ class ConcatLayer(MergeLayer):
 concat = ConcatLayer  # shortcut
 
 
-class ElemwiseSumLayer(MergeLayer):
+class ElemwiseMergeLayer(MergeLayer):
+    """
+    This layer performs an elementwise merge of its input layers.
+    It requires all input layers to have the same output shape.
+
+    Parameters
+    ----------
+    incomings : a list of :class:`Layer` instances or tuples
+        the layers feeding into this layer, or expected input shapes,
+        with all incoming shapes being equal
+
+    merge_function : callable
+        the merge function to use. Should take two arguments and return the
+        updated value. Some possible merge functions are ``theano.tensor``:
+        ``mul``, ``add``, ``maximum`` and ``minimum``.
+
+    See Also
+    --------
+    ElemwiseSumLayer : Shortcut for sum layer.
+    """
+
+    def __init__(self, incomings, merge_function, **kwargs):
+        super(ElemwiseMergeLayer, self).__init__(incomings, **kwargs)
+        self.merge_function = merge_function
+
+    def get_output_shape_for(self, input_shapes):
+        if any(shape != input_shapes[0] for shape in input_shapes):
+            raise ValueError("Mismatch: not all input shapes are the same")
+        return input_shapes[0]
+
+    def get_output_for(self, inputs, **kwargs):
+        output = None
+        for input in inputs:
+            if output is not None:
+                output = self.merge_function(output, input)
+            else:
+                output = input
+        return output
+
+
+class ElemwiseSumLayer(ElemwiseMergeLayer):
     """
     This layer performs an elementwise sum of its input layers.
     It requires all input layers to have the same output shape.
 
     Parameters
-    -----------
+    ----------
     incomings : a list of :class:`Layer` instances or tuples
         the layers feeding into this layer, or expected input shapes,
         with all incoming shapes being equal
@@ -57,36 +98,28 @@ class ElemwiseSumLayer(MergeLayer):
         be included in the learnable parameters of this layer.
 
     Notes
-    -----------
+    -----
     Depending on your architecture, this can be used to avoid the more
     costly :class:`ConcatLayer`. For example, instead of concatenating layers
     before a :class:`DenseLayer`, insert separate :class:`DenseLayer` instances
     of the same number of output units and add them up afterwards. (This avoids
     the copy operations in concatenation, but splits up the dot product.)
     """
-
     def __init__(self, incomings, coeffs=1, **kwargs):
-        super(ElemwiseSumLayer, self).__init__(incomings, **kwargs)
+        super(ElemwiseSumLayer, self).__init__(incomings, T.add, **kwargs)
         if isinstance(coeffs, list):
             if len(coeffs) != len(incomings):
                 raise ValueError("Mismatch: got %d coeffs for %d incomings" %
                                  (len(coeffs), len(incomings)))
         else:
             coeffs = [coeffs] * len(incomings)
+
         self.coeffs = coeffs
 
-    def get_output_shape_for(self, input_shapes):
-        if any(shape != input_shapes[0] for shape in input_shapes):
-            raise ValueError("Mismatch: not all input shapes are the same")
-        return input_shapes[0]
-
     def get_output_for(self, inputs, **kwargs):
-        output = None
-        for coeff, input in zip(self.coeffs, inputs):
-            if coeff != 1:
-                input *= coeff
-            if output is not None:
-                output += input
-            else:
-                output = input
-        return output
+        # if needed multiply each input by its coefficient
+        inputs = [input * coeff if coeff != 1 else input
+                  for coeff, input in zip(self.coeffs, inputs)]
+
+        # pass scaled inputs to the super class for summing
+        return super(ElemwiseSumLayer, self).get_output_for(inputs, **kwargs)
--- a/docs/modules/layers.rst
+++ b/docs/modules/layers.rst
@@ -117,6 +117,9 @@ Layer classes: merge layers
 
 .. autoclass:: concat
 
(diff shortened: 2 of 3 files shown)


benanne:
Looks okay to me, will have a closer look tomorrow. I will leave this open until @f0k is back and let him merge it, just to make sure.

lasagne/layers/merge.py
```diff
@@ -6,7 +6,8 @@
 __all__ = [
     "ConcatLayer",
     "concat",
-    "ElemwiseSumLayer",
+    "ElemwiseMergeLayer",
+    "ElemwiseSumLayer"
```
f0k: We usually have a comma also after the last line so you can add things in any position with a single inserted line instead of having to modify two lines. (Just so you know for the future. Not crucial now.)

skaae: Ok. I have been wondering about the wiered commas :) thanks for the explanation.

f0k:
Looks good, there's only three points that warrant discussion:
1. `coeffs` seems a bit weird for anything but summing. I'm not sure if the implication of `ElemwiseMergeLayer([l1, l2, l3], coeffs=0.1, merge_function=T.mul)` is clear: It will compute `0.1*l1 * 0.1*l2 * 0.1*l3`. We could remove `coeffs` from `ElemwiseMergeLayer` and only add it to `ElemwiseSumLayer`, but then `ElemwiseSumLayer` needs a `get_output_for()` method like:
```python
    inputs = [coeff * input if coeff != 1 else input
              for coeff, input in zip(self.coeffs, inputs)]
    return super(ElemwiseSumLayer, self).get_output_for(inputs, **kwargs)
```
Is there any use case for having `coeffs` with a different merge function than summing? Is there any way to better document what's happening if you set it to a scalar and then use multiplication?
2. There should probably be a test for `ElemwiseMergeLayer`, right now it's only tested by means of the `ElemwiseSumLayer`. That gives 100% coverage alright, but it doesn't guarantee that `ElemwiseMergeLayer` cares about the `merge_function` argument at all. I think a quick test of `get_output_for` would be enough, there's no need to replicate all the other `ElemwiseSumLayer` tests for now.
3. `ElemwiseSumLayer` needs a docstring. I'd prefer a complete docstring instead of a mere "see ElemwiseMergeLayer", even if it's mostly a 1:1 copy. The `Notes` section should be kept in the `ElemwiseSumLayer` only since it doesn't really apply to the general `ElemwiseMergeLayer`.

skaae:
1.	I agree that `coeffs` is mostly used for summing. What do you think about moving coeffs to a separate layer? a simple layer that scales its input with a constant or a vector? It could be called `MultiplyConstantLayer` or `ScaleLayer`.  Similarly we could create `AddLayer`, `AbsLayer` and maybe `MaxLayers`? I don't know if there is any uses cases for the last ones though? 
If we separate out the scaling the name is also more precise. 
2.	Good point. I’ll write more tests when we have settled on a format.
3.	Agreed.

f0k:
> What do you think about moving coeffs to a separate layer? a simple layer that scales its input with a constant or a vector?

I think that would make two common use cases more complex:
* ``ElemwiseSumLayer([l1, l2, l3], coeffs=1./3)`` for averaging three layers would become ``ScaleLayer(ElemwiseSumLayer([l1, l2, l3]), 1./3)``.
* ``ElemwiseSumLayer([l1, l2, l3], coeffs=[0.2, 0.5, 0.3])`` for mixing three layers with specific weights would become ``ElemwiseSumLayer([ScaleLayer(l1, 0.2), ScaleLayer(l2, 0.5), ScaleLayer(l3, 0.3)])``.

So I'd still be in favor of moving `coeffs` to `ElemwiseSumLayer`. @benanne, what do you think?

On a different note, it would be good to mention `T.minimum` and `T.maximum` as further possible merge functions in the `ElemwiseMergeLayer` docstring -- people might try `T.min` or `T.max` otherwise.

benanne:
I'm with @f0k on all counts :)

skaae:
I prefer moving it coeffs to seprate layer. The most common usecase is probably `mean`. I dont think 
`ScaleLayer(ElemwiseSumLayer([l1, l2, l3]), 1./3).` is much worse can the current format? 

For your second example i actually prefer the verbosity of the scalinglayer because it is clear what is happening. 

From a user perspective I dont think its obvious that `ElemwiseSumLayer` would do weighted sum as well? Lastly separating the layers handles usescases we havent thougt about.

benanne:
I think if we're going to have things like `ScaleLayer` we're basically rebuilding Theano on top of Theano. It just feels needlessly verbose to me.

lasagne/layers/merge.py
```diff
@@ -68,22 +70,57 @@ class ElemwiseSumLayer(MergeLayer):
     before a :class:`DenseLayer`, insert separate :class:`DenseLayer` instances
     of the same number of output units and add them up afterwards. (This avoids
     the copy operations in concatenation, but splits up the dot product.)
```
f0k: As I said (https://github.com/Lasagne/Lasagne/pull/301#issuecomment-112519950), the `Notes` section should be moved to `ElemwiseSumLayer`, since it only applies to the case of summing. Also the underline is too long, please make it 5 characters (`-----`) while moving it, that would be nice!

lasagne/layers/merge.py
```diff
@@ -91,7 +128,7 @@ def get_output_for(self, inputs, **kwargs):
             if coeff != 1:
                 input *= coeff
             if output is not None:
-                output += input
+                output = self.merge_function(output, input)
             else:
                 output = input
         return output
```
f0k: I wouldn't duplicate the merge functionality here, but just use the function I suggested: https://github.com/Lasagne/Lasagne/pull/301#issuecomment-112519950

lasagne/layers/merge.py
skaae: I think it applies to all merge layers? I should probably change add to merge though.

lasagne/layers/merge.py
skaae: I think this version reads more easily. We could write out the list comprehension as a for-loop but then it would be nearly the same length.

lasagne/layers/merge.py
f0k: No, the `ElemwiseSumLayer` is the only way to avoid a `ConcatLayer` in front of a `DenseLayer`, because the matrix product in the `DenseLayer` can be expressed as a sum of some smaller matrix products.

skaae: True, it has to be rewritten for merge then. But you might as well use other merge functions to aviod concatenation.

lasagne/layers/merge.py
f0k: Okay, maybe I'm more used to reading list comprehensions, but the main point is not length, but keeping the implementations independent. Now you're coupling the subclass implementation to the super class implementation naming its attribute `merge_function`. It's not a big deal, but the software engineer in me cringes!
What about:
```python
    inputs = list(inputs)
    for idx in range(len(inputs)):
        if self.coeffs[idx] != 1:
            inputs[idx] = inputs[idx] * self.coeffs[idx]
    return super(ElemwiseSumLayer, inputs, **kwargs)
```
Although I'd really prefer:
```python
    # multiply each input by its coefficient if needed
    inputs = [input * coeff if coeff != 1 else input
              for coeff, input in zip(self.coeffs, inputs)]
    # pass inputs to the super class for summing
    return super(ElemwiseSumLayer, inputs, **kwargs)
```

lasagne/layers/merge.py
f0k: > But you might as well use other merge functions to aviod concatenation.

I can't think of any. And in any case, I think it's valuable to have the current note in the `ElemwiseSumLayer` documentation, because it explains an interesting and not-too-obvious use case for it. A more abstract / more general explanation for the `ElemwiseMergeLayer` would make it even less obvious.

lasagne/layers/merge.py
skaae: If you feel strongly about its fine :) There is no software engineer inside me.

I see your point with coupling.

lasagne/layers/merge.py
skaae: Sometimes you multiply inputs to get multiplicative interaction. Maybe its not that common but i have seen it used in some papers. Mostly related to RNNs though.

lasagne/layers/merge.py
f0k: > If you feel strongly about it then its fine :) There is no software engineer inside me.

I do :) I guess the list comprehension is easy to comprehend with the comment?

> I see your point with coupling.

The second problem with duplicating the code is that any changes in `ElemwiseMergeLayer` (whatever that could be) would need to be replicated in `ElemwiseSumLayer`, which may be forgotten. Again, that's the software engineer in me, `merge.py` is actually small and overseeable for practical matters. I'd still try to avoid code smell as much as possible.

skaae:
removing code smell :)

lasagne/layers/merge.py
f0k: > Sometimes you multiply inputs to get multiplicative interaction.

Sure, but then it cannot serve as a replacement for a `ConcatLayer`, can it? I'm maintaining my position that the `Notes` section only applies to the `ElemwiseSumLayer` :) Thanks for moving it!

f0k:
Thanks, looks very good now, merging!

PR #302, opened 2015-06-13
add dim arg to flatten

skaae:
adds a `dim` argument to flatten that lets the user decide from which dimension the input should be flattend, e.g. `dim=1`(default) has the same behavior as the current flatten layer.

This is usefull if you have recurrent layer that has convolutional output.

### DIFF
--- a/lasagne/layers/shape.py
+++ b/lasagne/layers/shape.py
@@ -21,17 +21,35 @@ __all__ = [
 
 class FlattenLayer(Layer):
     """
-    Flatten all but the first dimension.
+    A layer that flattens its input. The leading ``outdim-1`` dimensions of
+    the output will have the same shape as the input. The remaining dimensions
+    are collapsed into the last dimension.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape.
+    outdim : int
+        The number of dimensions in the output.
 
     See Also
     --------
     flatten  : Shortcut
     """
+    def __init__(self, incoming, outdim=2, **kwargs):
+        super(FlattenLayer, self).__init__(incoming, **kwargs)
+        self.outdim = outdim
+
+        if outdim < 1:
+            raise ValueError('Dim must be >0, was %i', outdim)
+
     def get_output_shape_for(self, input_shape):
-        return (input_shape[0], int(np.prod(input_shape[1:])))
+        shp = [input_shape[i] for i in range(self.outdim-1)]
+        shp += [int(np.prod(input_shape[(self.outdim-1):]))]
+        return tuple(shp)
 
     def get_output_for(self, input, **kwargs):
-        return input.flatten(2)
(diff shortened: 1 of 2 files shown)


lasagne/layers/shape.py
```diff
@@ -21,17 +21,33 @@
 
 class FlattenLayer(Layer):
     """
-    Flatten all but the first dimension.
+    Flatten all dimensions trailing ``dim``
```
benanne: I think this is ambiguous, to me it sounds like ``dim`` is not included (but it is, right?)

lasagne/layers/shape.py
```diff
@@ -21,17 +21,33 @@
...
     See Also
     --------
     flatten  : Shortcut
     """
+    def __init__(self, incoming, dim=1, **kwargs):
+        super(FlattenLayer, self).__init__(incoming, **kwargs)
+        self.dim = dim
+
+        if dim < 0:
+            raise ValueError('Dim must be >=0, was %s', str(dim))
```
benanne: Shouldn't this be:
```
raise ValueError('Dim must be >=0, was %d', dim)
```

Or does the `str` serve a purpose? I thought maybe it was to make it raise when someone passes something that isn't a number, but then `dim < 0` would also fail anyway.

benanne:
Two small remarks. It would be good to add a test or two for the dim argument as well.

lasagne/layers/shape.py
skaae: Yes, If dim is 0 all dimensions are flattened and 1 is the original behavior.

I'll try to come of with a more precise description.

lasagne/layers/shape.py
skaae: Yes it was meant to print what ever the user passed in, i'll fix it.

skaae:
added test. I changed the description but im not sure that its any better :)

benanne:
Maybe something like "flatten all dimensions following `dim` (including `dim` itself)" to avoid any ambiguity?

lasagne/layers/shape.py
```diff
@@ -21,17 +21,34 @@
...
     """
-    Flatten all but the first dimension.
+    Keep dimensions before ``dim`` and flatten other dimensions.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape.
+    dim : int
+        Dimensions before ``dim`` are kept. The reamining dimensions are
```
benanne: remaining*

benanne:
Actually, it might be a lot more intuitive to have `dim` be the number of dimensions that remains after the flattening, just like theano's `flatten` function does it. As a bonus it would be more consistent with Theano itself.

skaae:
Good call. I changed the behavior and argument names to be identical to T.flatten

benanne:
Looks great! Thanks for the extra tests as well :) I'll leave it open for a bit in case anyone else has comments.

benanne:
Merging.

PR #304, opened 2015-06-13
fix alias in docs

skaae:
- adds some missing layers
- removes see also when alias is printed in the docs
- fix bug in dropout ref

### DIFF
--- a/docs/modules/layers.rst
+++ b/docs/modules/layers.rst
@@ -78,7 +78,7 @@ Layer classes: noise layers
 .. autoclass:: DropoutLayer
     :members:
 
-.. autofunction:: dropout
+.. autoclass:: dropout
 
 .. autoclass:: GaussianNoiseLayer
     :members:
@@ -89,17 +89,25 @@ Layer classes: shape layers
 .. autoclass:: ReshapeLayer
     :members:
 
-.. autofunction:: reshape
+.. autoclass:: reshape
 
 .. autoclass:: FlattenLayer
     :members:
 
-.. autofunction:: flatten
+.. autoclass:: flatten
+
+.. autoclass:: DimshuffleLayer
+    :members:
+
+.. autoclass:: dimshuffle
 
 .. autoclass:: PadLayer
     :members:
 
-.. autofunction:: pad
+.. autoclass:: pad
+
+.. autoclass:: SliceLayer
+
 
 Layer classes: merge layers
 ---------------------------
@@ -107,13 +115,13 @@ Layer classes: merge layers
 .. autoclass:: ConcatLayer
     :members:
 
-.. autofunction:: concat
+.. autoclass:: concat
 
 .. autoclass:: ElemwiseSumLayer
     :members:
 
 Layer classes: embedding layers
----------------------------
+-------------------------------
 
 .. autoclass:: EmbeddingLayer
     :members:
@@ -137,4 +145,3 @@ Layer classes: embedding layers
 
 .. automodule:: lasagne.layers.dnn
     :members:
-
--- a/lasagne/layers/cuda_convnet.py
+++ b/lasagne/layers/cuda_convnet.py
@@ -99,10 +99,10 @@ class Conv2DCCLayer(CCLayer):
         An initializer for the weights of the layer. This should initialize the
         layer weights to a 4D array with shape
         ``(num_filters, num_input_channels, filter_rows, filter_columns)``.
-         If automatic dimshuffling is disabled (see notes), the shape should be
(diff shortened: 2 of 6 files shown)


skaae:
https://github.com/skaae/nntools/commit/ac8400fe4d5167a4b574a25b3ddc11ff6b54be09 fixes 2 warnings in the conv layer. I dont know if we want them?

.idea/.name
```diff
@@ -0,0 +1 @@
+nntools
```
benanne: Is this file supposed to be in here?

docs/modules/layers.rst
```diff
@@ -137,4 +145,5 @@ Layer classes: embedding layers
 
 .. automodule:: lasagne.layers.dnn
     :members:
+    :undoc-members: MaxPool2DDNNLayer
```
benanne: We should probably just add a docstring for this instead, I guess? Maybe not in this PR though :)

skaae:
I added the missing dot and removed the file that slipped in.

benanne:
needs a small rebase. I'd get rid of the undoc-members thing as well, I think this will only serve to confuse us later. We should just add some docs for it.

skaae:
Ok. I'll do i later today.

lasagne/layers/cuda_convnet.py
```diff
@@ -99,7 +99,9 @@ class Conv2DCCLayer(CCLayer):
         An initializer for the weights of the layer. This should initialize the
         layer weights to a 4D array with shape
         ``(num_filters, num_input_channels, filter_rows, filter_columns)``.
+
          If automatic dimshuffling is disabled (see notes), the shape should be
+
         ``(num_input_channels, input_rows, input_columns, num_filters)``
```
f0k: Why the extra blank lines (paragraphs)?

lasagne/layers/noise.py
```diff
@@ -14,7 +14,7 @@
 
 
 class DropoutLayer(Layer):
-    """Dropout layer [1]_,[2]_.
+    """Dropout layer [1]_, [2]_
```
f0k: Interesting -- was it the extra blank or the removed dot that fixed the reference?

lasagne/layers/cuda_convnet.py
skaae: Sphinx gives a really long warning if there is no line break. I guess it has to do with the long ``...``.
But we can remove them.

lasagne/layers/noise.py
skaae: The references need to separated by space.

lasagne/layers/cuda_convnet.py
f0k: +1 for removing Sphinx warnings! But what exactly does it say? Is this the only/correct way to fix it? What do you mean by "long `...`"?

skaae: I meant to write the " --- " string.
In any case I think it is the long 

```
``(num_filters, num_input_channels, filter_rows, filter_columns)``
```
that shinx is complaining about. At least i removs the error if you insert a line break.

f0k: >> But what exactly does it say?

> I think it is the long [...]

Went to check it myself. The exact error given by Sphinx is "68: ERROR: Unexpected indentation.", and it's because there is an extra space before "If automatic dimshuffling ...". Adding the blank lines worked because it makes the line stand as its own paragraph, so it doesn't have to match the indentation of the other paragraphs. The correct solution would be removing a space character before "If", not inserting blank lines. If you can fix that, I'm fine with merging (and we're down to zero complaints by Sphinx, yay)!

skaae: Nice. How did you spot that? Does the line number refer to a specific file or?

skaae:
I updated the PR with @f0k comment.

benanne:
Cool, lgtm! I'll leave it to @f0k to merge.

lasagne/layers/cuda_convnet.py
f0k: > Nice. How did you spot that?

Well, the error message clearly says "Unexpected indentation."

> Does the line number refer to a specific file or?

It refers to the line of the docstring where the error occurs. You need to add the line number of the first docstring line to convert it to file line numbers. It's a bit silly, the error message would be a lot more useful if it just printed the line of the error, or gave the file line number, instead of printing a list of all docstring lines.

f0k:
Great, thanks for all the fixes! Merging in 3, 2, 1...

PR #306, opened 2015-06-17
more tests for conv_output_length

ebenolson:
replaces the test of `conv_output_length` with a more complete version. Also changed the function to raise `ValueError` instead of `RuntimeError`.


f0k:
Thanks a lot! Looks good, Travis passes, Coverage increased to 100% as intended -- merging!

benanne:
Awesome!

PR #307, opened 2015-06-18
Update MaxPool2DCCLayer to use pool_output_length

hjweide:
MaxPoolConv2DCCLayer should allow inputs whose dimensions are not known at compile-time to be specified with None.  I updated MaxPoolConv2DCCLayer to use pool_output_length in a similar manner to how Conv2DCCLayer uses conv_output_length.

I set ignore_border=False and pad=0 because MaxPool2DCCLayer does not support ignore_border or padding.

This was inspired by the discussion on [lasagne-users](https://groups.google.com/forum/#!topic/lasagne-users/72e1-EEAZGM).

### DIFF
--- a/lasagne/layers/cuda_convnet.py
+++ b/lasagne/layers/cuda_convnet.py
@@ -8,6 +8,7 @@ from .. import nonlinearities
 from .base import Layer
 
 from .conv import conv_output_length
+from .pool import pool_output_length
 from ..utils import as_tuple
 
 from theano.sandbox.cuda.basic_ops import gpu_contiguous
@@ -506,10 +507,18 @@ class MaxPool2DCCLayer(CCLayer):
             num_input_channels = input_shape[0]
             input_rows, input_columns = input_shape[1:3]
 
-        output_rows = int(np.ceil(float(input_rows - self.pool_size +
-                                        self.stride) / self.stride))
-        output_columns = int(np.ceil(float(input_columns - self.pool_size +
-                                           self.stride) / self.stride))
+        output_rows = pool_output_length(input_rows,
+                                         pool_size=self.pool_size,
+                                         stride=self.stride,
+                                         ignore_border=False,
+                                         pad=0,
+                                         )
+        output_columns = pool_output_length(input_columns,
+                                            pool_size=self.pool_size,
+                                            stride=self.stride,
+                                            ignore_border=False,
+                                            pad=0,
+                                            )
 
         if self.dimshuffle:
             return (batch_size, num_input_channels, output_rows,
--- a/lasagne/tests/layers/test_pool.py
+++ b/lasagne/tests/layers/test_pool.py
@@ -180,7 +180,8 @@ class TestMaxPool1DLayer:
         assert np.allclose(numpy_result, layer_result)
 
     @pytest.mark.parametrize(
-        "input_shape", [(32, 64, 128), (None, 64, 128), (32, None, 128)])
+        "input_shape", [(32, 64, 128), (None, 64, 128), (32, None, 128),
+                        (32, 64, None)])
     def test_get_output_shape_for(self, input_shape):
         input_layer = self.input_layer(input_shape)
         layer = self.layer_ignoreborder(input_layer, pool_size=2)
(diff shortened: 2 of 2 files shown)


ebenolson:
Nice! If you have a chance, it would be good to add a test case covering this to `TestMaxPool2DCCLayer `and the others in `tests/layers/test_pool.py`.

benanne:
Thanks! As @ebenolson said, a test ensuring that `MaxPool2DCCLayer` supports `None` dimensions would be great so we don't end up breaking it again later :)

hjweide:
@ebenolson and @benanne, I just pushed some tests.  I'm not yet entirely familiar with how pytest works, so I would appreciate it if someone could just glance over my changes :)

ebenolson:
To test it, you'll need to add more asserts, like

    assert layer.get_output_shape_for((32, 64, 24, None)) == (32, 64, 12, None)

I'm actually not sure what the `pytest.mark.parametrize` is supposed to be accomplishing, it could probably be removed.

ebenolson:
Also, there seems to be a bug in `test_dimshuffle_false`, maybe you can fix that at the same time. Line 394

    actual = layer.get_output_for(input).eval()

Should be

    actual = layer.get_output_for(theano.shared(input)).eval()

benanne:
I'm not sure what you are talking about, the asserts are already there, aren't they? And the `parametrize` decorators definitely shouldn't be removed, they ensure that the tests are run with each of the specified input arguments (with different `input_shape`s in this case).

The bug in `test_dimshuffle_false` seems unrelated to this, so that should probably be fixed in a separate PR.

ebenolson:
But changing `input_shape` doesn't actually accomplish anything, since `get_output_shape_for` never refers to the layer's `input_shape`, only its argument. 

It might make more sense to rewrite the test to use the parametrized value though instead of having multiple asserts.

hjweide:
But doesn't `get_output_shape_for` reference `input_shape` through `layer`, which in turn references `input_shape` through `input_layer`?

ebenolson:
no, but the tests could be rewritten to compare `layer.output_shape`, which does function like that.

benanne:
> But changing `input_shape` doesn't actually accomplish anything, since `get_output_shape_for` never refers to the layer's `input_shape`, only its argument. 

That's definitely a bug, but the solution is to make it use `input_shape`, not throw away the `parametrize` decorator, I think. Clearly the intention was to test all these different shapes, even though that's not what's currently happening. This may be a little much to dump on @hjweide's shoulders though. But of course you're welcome to take a look at it if you want to. Otherwise someone else should probably fix this in a separate PR.

ebenolson:
> Clearly the intention was to test all these different shapes

As the one responsible for that addition, I'm not sure I agree about clarity :)
@hjweide if you want to fix this that would be great, otherwise I can take care of it.

hjweide:
If it's just restructuring the tests I don't mind including it in this PR, between what @benanne and @ebenolson have said I'm confused about what's supposed to happen.  Currently, I have the different input/output pairs in the `parametrize`, and I then have one assert that compares them:

```
@pytest.mark.parametrize(
    "input_shape,output_shape",
    [((32, 64, 24, 24), (32, 64, 12, 12)),
    ...
    ((32, 64, None, None), (32, 64, None, None))],
)
def test_get_output_shape_for(self, input_shape, output_shape):
    try:
          input_layer = self.input_layer(input_shape)
          layer = self.layer(input_layer,
                                     pool_size=(2, 2), stride=None)
          assert layer.get_output_shape_for(
                  input_shape) == output_shape
    except NotImplementedError:
          pytest.skip()
```

If this is correct, I'll go ahead and make these changes, else I'll leave it for someone with more experience using `pytest`.

hjweide:
Or should I remove the `parametrize` and use multiple asserts instead, each with its own input/output pair?

ebenolson:
Parametrize with pairs seems fine to me.

benanne:
Yeah this is... confusing. But it sounds like you guys are sorting it out :)

hjweide:
I've made the changes the way I understand the tests.  @ebenolson, please let me know if it's not correct :)  Once everyone is happy, I'll squash the commits.  (For the moment I have not fixed the bug that @ebenolson mentioned, I think we should do that in another PR.)

ebenolson:
Looks good to me! Your test reveals that we have the same issue in `MaxPool2DNNLayer`, but that can be fixed in another PR.

hjweide:
I've squashed the commits, so just waiting to see that Travis passes...

f0k:
The PR looks good to me!

> Should be
> ```actual = layer.get_output_for(theano.shared(input)).eval()```

If `input` is an ndarray, Theano will turn it into a constant expression, there's no need to wrap it as a shared variable. You can always do things like `T.matrix() * np.zeros(42)`. (If this is not what the bug is about: sorry, I haven't taken a look at that code, just guessed what you were after.)

ebenolson:
Never mind, I was running into a Theano bug that was fixed [here](https://github.com/Theano/Theano/commit/791b628b5e700c9f122602a8a92766f639dfbe8f).

benanne:
Looks good to me as well, so I'll merge it. Thanks again!

PR #308, opened 2015-06-20
Bugfix to discovering hidden factors demo

briancheung:
Thanks for making an awesome demo. I was surprised the difference in accuracy and blurriness of the reconstructions. I simply changed that value batchsize to a more standard 100 in this demo and that seems to help with the overfitting and reconstructions (validation accuracy around 96.5%). We tried 50 in the paper just to see if the XCov estimate was still reliable.

Since we worked from the MNIST yaml files already in pylearn2, you're right that there was some regularization features already enabled (max_weight_norm). We just kept it at the default value (I believe .9). Don't know how to do this in Lasagne and probably not necessary for a simple demo.


briancheung:
Also tried alpha=0, gamma=0, beta=10. which achieved about the same validation accuracy. alpha=0 and gamma=0 should turn it into a standard supervised MLP. So it's likely a standard overfitting issue with MLPs.

benanne:
Cool, thanks a lot for the feedback! We're going to be putting a bunch of examples and notebooks in our new 'Recipes' repository soon: https://github.com/Lasagne/Recipes

Once it's in there feel free to send a PR with any improvements you may have :) I'll try out your suggestions sometime as well!

briancheung:
Found a serious bug in this code which was causing the discrepancy in accuracy. The original code was training on 1/10th the training set! There's a mismatch between the batch size in mnist.py and the batch size in this demo. After the fix, the accuracy is around 98.4-98.5% which is slightly better than our paper :)

I found the bug because I was baffled by how this code was 8x faster than my own implementation. Things make more sense now. Would really appreciate this PR get merged in soon since the bug fix is fairly serious.

benanne:
Damn, good catch! Yeah, sorry for not merging this yet - the idea is to move this to the Recipes repository as soon as possible, but since people are probably already using this it's a good idea to merge it here. Thanks for digging into this!

benanne:
Okay, looks good. 200 epochs is probably a bit overkill now :) But I'll go ahead and merge it. Thanks again!

jsusskin:
Hey guys -- I implemented my own version of Brian's paper in Lasagne and then discovered this version the other day. Glad to see I was on the right track since my implementation is almost identical. One thing I've noticed is that training with the cross-covariance cost can easily cause the gradients to blow up (producing nan's) pretty easily. For example, if I add dropout layers, change from adadelta to nesterov momentum schemes, etc. I think this is because the linear units in the encoder layer are unbounded and can therefore result in large values in the cross covariance computation. I see that Brian used a max_weight_norm penalty, which may be helping to stabilize things. Just wanted to add that I found a similar strategy that helps stabilize things by simply clipping the relu's in the pre-encoder layers to [0, 3].

benanne:
Cool! I didn't notice any instability but I guess that's cause I stuck to adadelta (I wanted to reproduce the original as closely as possible). But good to know :) I'll be moving this over to Lasagne/Recipes soon. Just waiting for the changes to the examples to come through.

jsusskin:
BTW -- if you're interested, here are several threads I found that discuss nan blowup in the context of relu + categorical crossentropy loss:
https://groups.google.com/forum/#!topic/theano-dev/GaJwGw6emK0
https://groups.google.com/forum/#!topic/theano-users/pbbddYetkgM
https://groups.google.com/forum/#!topic/pylearn-dev/zEKrHiDJFz0
http://theano-users.narkive.com/oMCdyOzl/nan-gradients-using-binary-crossentropy-cost-function

briancheung:
I have noticed that softmax + cross entropy loss does result create instability in any MLP that uses rectified units. I suspect the softmax units are blowing up because the gradient mathematically simplifies to (sofmax_unit_i - target_i). See the last equation in http://www.willamette.edu/~gorr/classes/cs449/classify.html for reference.

I haven't checked if theano is actually doing the correct simplification for the gradient. But assuming it is, the softmax seems like a culprit for creating NaNs since its exponential w.r.t to the previous layer activations  (particularly bad with unbounded activation that @jsusskin mentioned). 

One easy fix that I tend to use now is either switching to HingeLoss or putting a batch normalization layer right before the softmax helps. Batch normalization actually fits quite well with the XCov loss too because it turns the XCov loss into something more akin to an XCorr loss.

sidps:
Have you experimented with Log-Softmax? I believe that's what is commonly used in Torch, and perhaps helps alleviate this problem (?) since it's in the log domain. It should also be faster to compute since log(e^x / sum e^x) = x - log(sum e^x), and the output is used directly (rather than log output) in calculating cross entropy. Sorry, just thinking aloud.

briancheung:
@sidps I'm not familiar with log-softmax, is that different from softmax w/ cross entropy loss? Or do you mean a layer that does log(softmax(x))? If it's the latter, what loss function do you pair that layer with?

jsusskin:
Hey @briancheung -- the log-softmax is just the log of the softmax, meaning log(exp(x)/sum(exp(x)) = x - log(sum(exp(x)). This is a nice formulation because it protects against taking the log of a particular x that may be saturated at 0. The log-softmax can be directly used in conjunction with crossentropy loss since crossentropy loss.
@sidps -- I'm not sure yet, but I think that theano may already recognize this pattern when optimizing the symbolic graph. Gotta look at that.

jsusskin:
On a related note, it seems that clipping the inputs into crossentropy solves this problem without adverse effects, at least for the experiments I've tried.

nouiz:
If Theano do not recognize the pattern, open an issue in Theano. We
recognize the softmax pattern and have many optimization for cross entropy,
but we could have missed some. No opt specific to log softmax except to use
it in cudnn v3 when it is released.

jsusskin:
@nouiz -- writing up an issue now. Thanks to @sidps for pointing out the idea to replace softmax with log-softmax. It turns out to produce proper stable gradients for very large inputs into the softmax. I'll copy the link to the issue here once it's written.

sidps:
@briancheung @jsusskin Sorry for not responding earlier. Glad it's working out, happy to help :).

briancheung:
Thanks for bringing this up @jsusskin and @sidps. So this protects against the result of softmax(x) being very close to 0, but won't it still be susceptible to very large x inputs?

jsusskin:
Large inputs will saturate to to 1.
the problem with unstable crossentropy is that the log(0) is undefined but
the log(1) is 0 and that's fine.

In general though, I think this problem does not happen in the case of code
derived from the lasagne mnist example. This is because the theano ops are
properly protecting against this problem when the target vectors are coded
as integer vectors representing the class labels. My understanding is that
it is only a problem when the targets are a matrix where rows are
observations and columns are class probabilities (e.g. one-hot coding).






I think the point is that softmax(x) can give numbers that are saturated at
0 or 1 if the net makes confident predictions, both of which tend to happen
when the inputs x have a large range (large difference btwn smallest and
largest value in the input vector).





The softmax(x) can saturate at 0 or 1
Very large inputs can lead to softmax outputs that saturate at 0 or 1
The linear activation into a softmax will result in outputs saturating at 0
or 1 if the inputs are large (depending on whether the weights are positive
or negative).

Issue #299, opened 2015-06-12
RFC: Unify conv/pool layers


craffel:
It has come up a couple of times as I have introduced people (cc @bmcfee, @moscow25, and issue #128) to Lasagne that someone is trying to use some code which uses, for example, `Conv2DDNNLayer` on a machine which has no GPU or does not have cudnn installed (this can happen when you're running someone else's code, or when you want to test out some code you wrote on your GPU box on your GPU-less laptop).  The simple fix is generally to just use `Conv2DLayer` instead, but it's not really a portable solution to tell people to modify code when it's being used on different systems.  A simple hack fix would be for all scripts which use Lasagne's layers which have optional dependencies to have something like this (as an example) at the top:
```Python
try:
    import lasagne.layers.dnn
    conv_layer = lasagne.layers.dnn.Conv2DNNLayer
except ImportError:
    conv_layer = lasagne.layers.Conv2DLayer
...
# All convolutional layers would then be instantiated like this
layer = conv_layer(incoming, ...)
```
I don't really think it's fair to encourage all of our users to do something this hacky, though, and I think we owe it to users to abstract this out.  To me, a good solution would be for there to be a single `Conv2DLayer` which had an additional kwarg such as `backend`, which could be, say, `'theano'`, `'cuda_convnet'`, `'dnn'`, etc., which chose which backend to use.  Given that the different backends have different requirements, we'd need to be diligent (as we already are) about throwing errors when a backend can't support the provided layer arguments.  The simplest way to implement this would be to basically just make `Conv2DLayer` a wrapper class around all of these layers, which threw warnings when it fell back on the theano backend because of something not being present.  I believe this also applies to pooling layers.  If and when theano does this for us, we can greatly simplify our code, but I think it's worth abstracting this out now instead of waiting.  I don't have enough in-depth knowledge of the different layers/backends to know if there is something making this not possible, so please let me know if there is.

ebenolson:
Another way to do this would be to add code to `cuda_convnet.py` and `dnn.py` which replaces those Layer classes with dummy wrappers around `Conv2DLayer` when the libraries are unavailable.

One complication is that `Conv2DLayer` doesn't support `pad` at the moment.

benanne:
[...]

PR #311, opened 2015-06-24
Merge border_mode with pad argument for convolutional layers

MartinThoma:
I also listed the arguments as suggested in https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

### DIFF
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -15,7 +15,7 @@ __all__ = [
 
 
 def conv_output_length(input_length, filter_size,
-                       stride, border_mode, pad=0):
+                       stride, pad=0):
     """Helper function to compute the output size of a convolution operation
 
     This function computes the length along a single axis, which corresponds
@@ -33,29 +33,23 @@ def conv_output_length(input_length, filter_size,
     stride : int
         The stride of the convolution operation.
 
-    border_mode : str, 'valid', 'full', 'same' or 'pad'
-        A string indicating the convolution border mode.
+    pad : int, 'full' or 'same' (default: 0)
+        By default, the convolution is only computed where the input and the
+        filter fully overlap (a valid convolution). When ``stride=1``, this
+        yields an output that is smaller than the input by ``filter_size - 1``.
+        The `pad` argument allows to implicitly pad the input with zeros,
+        extending the output size.
 
-        If 'valid', it is assumed that the convolution is only computed where
-        the input and the filter fully overlap.
+        A single integer results in symmetric zero-padding of the given size on
+        both borders.
 
-        If 'full', it is assumed that the convolution is computed wherever the
-        input and the filter overlap by at least one position.
+        ``'full'`` pads with one less than the filter size on both sides. This
+        is equivalent to computing the convolution wherever the input and the
+        filter overlap by at least one position.
 
-        If 'same', it is assumed that the convolution is computed wherever the
-        input and the filter overlap by at least half the filter size, when the
-        filter size is odd. In practice, the input is zero-padded with half the
-        filter size at the beginning and half at the end (or one less than half
-        in the case of an even filter size). This results in an output length
-        that is the same as the input length (for both odd and even filter
-        sizes).
-
-        If 'pad', zero padding of `pad` positions is assumed to be applied to
-        the input, and then a valid convolution is applied.
-
-    pad : int, optional (default 0)
-        If `border_mode` is set to 'pad', this is the size of the padding that
-        is applied on both sides of the input. Otherwise, this is ignored.
+        ``'same'`` pads with half the filter size on both sides (one less on
+        the second side for an even filter size). When ``stride=1``, this
+        results in an output size equal to the input size.
 
     Returns
     -------
@@ -65,21 +59,20 @@ def conv_output_length(input_length, filter_size,
     Raises
     ------
     RuntimeError
-        When an invalid border_mode string is specified, a `RuntimeError` is
-        raised.
+        When an invalid padding is specified, a `RuntimeError` is raised.
     """
     if input_length is None:
         return None
-    if border_mode == 'valid':
+    if pad == 'valid':
         output_length = input_length - filter_size + 1
-    elif border_mode == 'full':
+    elif pad == 'full':
         output_length = input_length + filter_size - 1
-    elif border_mode == 'same':
+    elif pad == 'same':
         output_length = input_length
-    elif border_mode == 'pad':
+    elif isinstance(pad, int):
         output_length = input_length + 2 * pad - filter_size + 1
     else:
-        raise ValueError('Invalid border mode: {0}'.format(border_mode))
+        raise ValueError('Invalid pad: {0}'.format(pad))
 
     # This is the integer arithmetic equivalent to
     # np.ceil(output_length / stride)
@@ -91,7 +84,7 @@ def conv_output_length(input_length, filter_size,
 class Conv1DLayer(Layer):
     """
     lasagne.layers.Conv1DLayer(incoming, num_filters, filter_size, stride=1,
-    border_mode="valid", untie_biases=False, W=lasagne.init.GlorotUniform(),
+    pad=0, untie_biases=False, W=lasagne.init.GlorotUniform(),
     b=lasagne.init.Constant(0.), nonlinearity=lasagne.nonlinearities.rectify,
     convolution=lasagne.theano_extensions.conv.conv1d_mc0, **kwargs)
 
@@ -117,23 +110,25 @@ class Conv1DLayer(Layer):
         An integer or a 1-element tuple specifying the stride of the
         convolution operation.
 
-    border_mode : str, one of 'valid', 'full', 'same'
-        A string indicating the convolution border mode.
+    pad : {0, 'full', 'same'} (default: 0)
+        By default, the convolution is only computed where the input and the
+        filter fully overlap (a valid convolution). When ``stride=1``, this
+        yields an output that is smaller than the input by ``filter_size - 1``.
+        The `pad` argument allows to implicitly pad the input with zeros,
+        extending the output size.
 
-        If 'valid', the convolution is only computed where the input and the
-        filter fully overlap.
-
-        If 'full', the convolution is computed wherever the input and the
+        ``'full'`` pads with one less than the filter size on both sides. This
+        is equivalent to computing the convolution wherever the input and the
         filter overlap by at least one position.
 
-        If 'same', the convolution is computed wherever the input and the
-        filter overlap by at least half the filter size, when the filter size
-        is odd. In practice, the input is zero-padded with half the filter size
-        at the beginning and half at the end (or one less than half in the case
-        of an even filter size). This results in an output length that is the
-        same as the input length (for both odd and even filter sizes).
+        ``'same'`` pads with half the filter size on both sides (one less on
+        the second side for an even filter size). When ``stride=1``, this
+        results in an output size equal to the input size.
 
-    untie_biases : bool, default False
+        Note that ``'full'`` and ``'same'`` can be faster than equivalent
+        integer values due to optimizations by Theano.
+
+    untie_biases : bool (default: False)
         If ``False``, the layer will have a bias parameter for each channel,
         which is shared across all positions in this channel. As a result, the
         `b` attribute will be a vector (1D).
@@ -181,12 +176,12 @@ class Conv1DLayer(Layer):
     Notes
     -----
     Theano's default convolution function (`theano.tensor.nnet.conv.conv2d`)
-    does not support the 'same' border mode by default. This layer emulates
+    does not support the 'same' pad by default. This layer emulates
     it by performing a 'full' convolution and then cropping the result, which
     may negatively affect performance.
     """
     def __init__(self, incoming, num_filters, filter_size, stride=1,
-                 border_mode="valid", untie_biases=False,
+                 pad=0, untie_biases=False,
                  W=init.GlorotUniform(), b=init.Constant(0.),
                  nonlinearity=nonlinearities.rectify,
                  convolution=conv.conv1d_mc0, **kwargs):
@@ -199,12 +194,12 @@ class Conv1DLayer(Layer):
         self.num_filters = num_filters
         self.filter_size = as_tuple(filter_size, 1)
         self.stride = as_tuple(stride, 1)
-        self.border_mode = border_mode
+        self.pad = pad
         self.untie_biases = untie_biases
         self.convolution = convolution
 
-        if self.border_mode not in ['valid', 'full', 'same']:
-            raise RuntimeError("Invalid border mode: '%s'" % self.border_mode)
+        if self.pad not in ['valid', 'full', 'same', 0]:
+            raise RuntimeError("Invalid pad: '%s'" % self.pad)
 
         self.W = self.add_param(W, self.get_W_shape(), name="W")
         if b is None:
@@ -232,7 +227,7 @@ class Conv1DLayer(Layer):
         output_length = conv_output_length(input_shape[2],
                                            self.filter_size[0],
                                            self.stride[0],
-                                           self.border_mode)
+                                           self.pad)
 
         return (input_shape[0], self.num_filters, output_length)
 
@@ -244,15 +239,15 @@ class Conv1DLayer(Layer):
 
         filter_shape = self.get_W_shape()
 
-        if self.border_mode in ['valid', 'full']:
+        if self.pad in [0, 'valid', 'full']:
             conved = self.convolution(input, self.W, subsample=self.stride,
                                       image_shape=input_shape,
                                       filter_shape=filter_shape,
-                                      border_mode=self.border_mode)
-        elif self.border_mode == 'same':
+                                      border_mode=self.pad or 'valid')
+        elif self.pad == 'same':
             if self.stride[0] != 1:
                 raise NotImplementedError("Strided convolution with "
-                                          "border_mode 'same' is not "
+                                          "pad 'same' is not "
                                           "supported by this layer yet.")
 
             conved = self.convolution(input, self.W, subsample=self.stride,
@@ -275,7 +270,7 @@ class Conv1DLayer(Layer):
 class Conv2DLayer(Layer):
     """
     lasagne.layers.Conv2DLayer(incoming, num_filters, filter_size,
-    stride=(1, 1), border_mode="valid", untie_biases=False,
+    stride=(1, 1), pad=0, untie_biases=False,
     W=lasagne.init.GlorotUniform(), b=lasagne.init.Constant(0.),
     nonlinearity=lasagne.nonlinearities.rectify,
     convolution=theano.tensor.nnet.conv2d, **kwargs)
@@ -302,23 +297,25 @@ class Conv2DLayer(Layer):
         An integer or a 2-element tuple specifying the stride of the
         convolution operation.
 
-    border_mode : str, one of 'valid', 'full', 'same'
-        A string indicating the convolution border mode.
+    pad : {0, 'full', 'same'} (default: 0)
+        By default, the convolution is only computed where the input and the
+        filter fully overlap (a valid convolution). When ``stride=1``, this
+        yields an output that is smaller than the input by ``filter_size - 1``.
+        The `pad` argument allows to implicitly pad the input with zeros,
+        extending the output size.
 
-        If 'valid', the convolution is only computed where the input and the
-        filter fully overlap.
-
-        If 'full', the convolution is computed wherever the input and the
+        ``'full'`` pads with one less than the filter size on both sides. This
+        is equivalent to computing the convolution wherever the input and the
         filter overlap by at least one position.
 
-        If 'same', the convolution is computed wherever the input and the
-        filter overlap by at least half the filter size, when the filter size
-        is odd. In practice, the input is zero-padded with half the filter size
-        at the beginning and half at the end (or one less than half in the case
-        of an even filter size). This results in an output length that is the
-        same as the input length (for both odd and even filter sizes).
+        ``'same'`` pads with half the filter size on both sides (one less on
+        the second side for an even filter size). When ``stride=1``, this
+        results in an output size equal to the input size.
 
-    untie_biases : bool, default False
+        Note that ``'full'`` and ``'same'`` can be faster than equivalent
+        integer values due to optimizations by Theano.
+
+    untie_biases : bool (default: False)
         If ``False``, the layer will have a bias parameter for each channel,
         which is shared across all positions in this channel. As a result, the
         `b` attribute will be a vector (1D).
@@ -363,12 +360,12 @@ class Conv2DLayer(Layer):
     Notes
     -----
     Theano's default convolution function (`theano.tensor.nnet.conv.conv2d`)
-    does not support the 'same' border mode by default. This layer emulates
+    does not support the 'same' pad by default. This layer emulates
     it by performing a 'full' convolution and then cropping the result, which
     may negatively affect performance.
     """
     def __init__(self, incoming, num_filters, filter_size, stride=(1, 1),
-                 border_mode="valid", untie_biases=False,
+                 pad=0, untie_biases=False,
                  W=init.GlorotUniform(), b=init.Constant(0.),
                  nonlinearity=nonlinearities.rectify,
                  convolution=T.nnet.conv2d, **kwargs):
@@ -381,12 +378,12 @@ class Conv2DLayer(Layer):
         self.num_filters = num_filters
         self.filter_size = as_tuple(filter_size, 2)
         self.stride = as_tuple(stride, 2)
-        self.border_mode = border_mode
+        self.pad = pad
         self.untie_biases = untie_biases
         self.convolution = convolution
 
-        if self.border_mode not in ['valid', 'full', 'same']:
-            raise RuntimeError("Invalid border mode: '%s'" % self.border_mode)
+        if self.pad not in ['valid', 'full', 'same', 0]:
+            raise RuntimeError("Invalid pad: '%s'" % self.pad)
 
         self.W = self.add_param(W, self.get_W_shape(), name="W")
         if b is None:
@@ -416,32 +413,32 @@ class Conv2DLayer(Layer):
         output_rows = conv_output_length(input_shape[2],
                                          self.filter_size[0],
                                          self.stride[0],
-                                         self.border_mode)
+                                         self.pad)
 
         output_columns = conv_output_length(input_shape[3],
                                             self.filter_size[1],
                                             self.stride[1],
-                                            self.border_mode)
+                                            self.pad)
 
         return (input_shape[0], self.num_filters, output_rows, output_columns)
 
     def get_output_for(self, input, input_shape=None, **kwargs):
-        # the optional input_shape argument is for when get_output_for is
+        # The optional input_shape argument is for when get_output_for is
         # called directly with a different shape than self.input_shape.
         if input_shape is None:
             input_shape = self.input_shape
 
         filter_shape = self.get_W_shape()
 
-        if self.border_mode in ['valid', 'full']:
+        if self.pad in [0, 'valid', 'full']:
             conved = self.convolution(input, self.W, subsample=self.stride,
                                       image_shape=input_shape,
                                       filter_shape=filter_shape,
-                                      border_mode=self.border_mode)
-        elif self.border_mode == 'same':
+                                      border_mode=self.pad or 'valid')
+        elif self.pad == 'same':
             if self.stride != (1, 1):
                 raise NotImplementedError("Strided convolution with "
-                                          "border_mode 'same' is not "
+                                          "pad 'same' is not "
                                           "supported by this layer yet.")
 
             conved = self.convolution(input, self.W, subsample=self.stride,
--- a/lasagne/layers/corrmm.py
+++ b/lasagne/layers/corrmm.py
@@ -30,7 +30,7 @@ class MMLayer(Layer):
 class Conv2DMMLayer(MMLayer):
     """
     lasagne.layers.Conv2DMMLayer(incoming, num_filters, filter_size,
-    stride=(1, 1), border_mode=None, untie_biases=False,
+    stride=(1, 1), pad=0, untie_biases=False,
     W=lasagne.init.GlorotUniform(), b=lasagne.init.Constant(0.),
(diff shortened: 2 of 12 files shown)


f0k:
Nice :+1:! But for consistency, could you please do the same to the remaining convolution layers (in `corrmm.py`, `cuda_convnet.py` and `dnn.py`), and to `conv_output_length()` (in `conv.py`)? Cheers!

MartinThoma:
Thank you :-)

Sure, I can do the same for the other layers.

Should I set the default of the other layers to `valid`? Currently, it is a bit inconsistent as `grep -nri "border_mode=" *` reveals. For `conv.py`, the default is `'valid'`, but for the others it is `None`.

## Conv2DLayer vs Conv2DDNNLayer

By the way, I would like some more information about the difference between `Conv2DDNNLayer` and `Conv2DLayer`. Currently, the only information within Lasagne I found is

> This is an alternative implementation which uses ``theano.sandbox.cuda.dnn.dnn_conv`` directly.

What does that imply? Is it faster? Why should / shouldn't I use `Conv2DDNNLayer` instead of `Conv2DLayer`?

(Should I probably open this discussion in the users group / open an issue for it? Seems a bit off-topic in this pull request.)

benanne:
There was a long discussion about this two weeks ago :) https://github.com/Lasagne/Lasagne/issues/299

In short: it's because the default Theano convolution interface has some limitations (like poor support for arbitrary padding / 'same' border_mode) that `dnn_conv` doesn't have. So if you have those features and you're using CuDNN, `Conv2DDNNLayer` allows you to access those features.

The Theano convolution interface is being revamped so we will soon be able to get rid of it.

f0k:
> For `conv.py`, the default is `'valid'`, but for the others it is `None`.

@benanne, is this because the other layers allow padding via the `pad` keyword, so `None` can either mean `'valid'` or `custom padding`? Shall we streamline this a bit, and merge `pad` with `border_mode` as in Theano (i.e., `border_mode` can be an `int` or `tuple` instead of a string for custom padding), or make custom padding require `border_mode='valid'` instead of `border_mode=None`? I find the current interface mildly confusing.

benanne:
Custom padding and `border_mode='valid'` are contradictory, so making that a requirement makes no sense to me. Otherwise I'm not married to any particular interface for these things. It just makes more sense to me to see `pad=3` rather than `border_mode=3` when I read it.

f0k:
> Custom padding and `border_mode='valid'` are contradictory

I read that as *"perform a valid convolution of the input padded with so and so many pixels"*. Having `border_mode` and `pad` seems like you could freely combine the two. Even if we don't allow so, having to set at least one of them to `None` seems weird to me (with the default being both of them set to `None`). I'd have expected the defaults `border_mode='valid', pad=0` and allow `border_mode='full', pad=0` to mean *"perform a full convolution of the input without any extra padding"*, `border_mode='valid', pad=(3,4)` to mean *"perform a valid convolution of the input padded with 3 and 4 rows and columns, respectively"* and disallow `border_mode='full', pad=3` because Theano doesn't support this and we can only emulate it if we know the filter size (oh, wait, we actually know the filter size in Lasagne, because the filter is part of the Layer...).

> It just makes more sense to me to see `pad=3` rather than `border_mode=3` when I read it.

Fully agreed. But then it would really be nice if `pad` was orthogonal to `border_mode`, rather than a different name for the same thing. What do you think about my way of reading it?

benanne:
> Fully agreed. But then it would really be nice if pad was orthogonal to border_mode, rather than a different name for the same thing. What do you think about my way of reading it?

True. But then it's probably easier to overload `border_mode` instead, even if it reads a bit less intuitively. Having `border_mode` and `pad` be completely orthogonal doesn't really add any useful functionality.

MartinThoma:
Another possibility could be renaming it, e.g `border`.

ebenolson:
I think it would be better to remove `border_mode`, and have `pad` accept `'same'` and `'full'` as shortcut strings.

benanne:
`border='same'`, `border=4` both make sense, I like that. Same for `pad='same'`, `pad=4` I guess :) Either of those would be fine.

f0k:
> But then it's probably easier to overload `border_mode` instead, even if it reads a bit less intuitively.

That was my reasoning for Theano, where we want them to be mutually exclusive (we don't necessarily know the filter shape at compile time, so combining custom padding with anything else than a valid convolution would complicate matters), and the easiest way to achieve that is by merging them into a single argument.

> Having `border_mode` and `pad` be completely orthogonal doesn't really add any useful functionality.

That's right, it just retains the readibility of the current version, and makes it easier to document/understand than the current version, but at the cost of complicating the code (`pad != 0` would need to check for `border_mode` even if it will hardly ever be combined with `border_mode != 'valid'` in practice).

> `border='same'`, `border=4` both make sense, I like that. Same for `pad='same'`, `pad=4` I guess :)

Same for me! +1 for merging them under a better name. In both cases, the default would be `0`, right? `pad='valid'` as a default wouldn't work that well, `border='valid'` would be an acceptable alternative to `border=0`.
In Theano, it will always be called `border_mode` for backwards compatibility, and `border` would be closer to that, but we don't have to follow Theano.
`pad` is a bit closer to what the border mode does conceptually (it pads the input before applying a valid convolution), but it's very nonstandard and the documentation might be confusing to people looking to do a full convolution ("`pad` controls the amount of zero-padding before applying a (valid) convolution. For a valid convolution, leave it at the default value of zero. For a full convolution, set it to `"full"`. [...]", compared to "`border` controls how the border is treated in applying the convolution. [...]").

> Either of those would be fine.

I'd still like to choose wisely :)

ebenolson:
> pad is a bit closer to what the border mode does conceptually (it pads the input before applying a valid convolution), but it's very nonstandard and the documentation might be confusing to people looking to do a full convolution

`pad` is what's used in Caffe, so I don't think it's that nonstandard. I'm guessing `border_mode` was adopted from scipy? 

Setting `border_mode=3` rubs me the wrong way, but I guess `border` and `pad` are pretty similar.

f0k:
> `pad` is what's used in Caffe, so I don't think it's that nonstandard.

Ah, didn't know that. But they don't mention the topic of valid vs. full convolutions, they just implicitly do valid convolutions and allow a custom integer for `pad`.

> I'm guessing `border_mode` was adopted from scipy?

Probably even numpy: http://docs.scipy.org/doc/numpy/reference/generated/numpy.convolve.html
Both in numpy/scipy, it's called `mode`.

> I guess `border` and `pad` are pretty similar.

Okay, probably it's just a matter of how we formulate the docstring. Shall we go for `pad` because it's shorter?
```
pad : int, tuple of int, "full" or "same"
    By default, the kernel is only applied at positions that do not extend past the input,
    yielding an output that is smaller than the input by ``filter_shape[0] - 1`` rows and
    ``filter_shape[1] - 1`` columns (a valid convolution), when ``stride=1``.
    To change that, the input can be implicitly padded with zeros to extend the output.
    A single integer results in symmetric zero-padding of the given size on all borders,
    a tuple of two integers allows different zero-padding of rows and columns.
    ``"full"`` pads with ``filter_shape[0] - 1`` rows and ``filter_shape[1] - 1`` columns
    on both sides (a full convolution), ``"same"`` pads such that the output size matches
    the input size when ``stride=1``.
    Note that ``"full"`` and ``"same"`` can be faster than equivalent integer values.
```
Not sure if that's easy enough to understand, but it's more exact than what we had before (which we can't reuse directly if we opt for `pad`).

MartinThoma:
I think `pad` is the better name than `border`:

* `pad` is used in Caffe ([example](https://gist.github.com/ksimonyan/fd8800eeb36e276cd6f9#file-vgg_cnn_s_deploy-prototxt-L79))
* `padW` and `padH` seem to be used by Torch7 ([source](https://github.com/torch/nn/blob/master/SpatialConvolutionMM.lua#L16))
* For me, `pad` is clearer than `border`.
* (`pad` is shorter than `border` - I don't think 3 characters matter much)

The only reason for border is that numpy uses it, right?

> `pad='valid'` as a default wouldn't work that well, `border='valid'` would be an acceptable alternative to `border=0`.

Wouldn't `pad=0` mean the same what currently is `border_mode='valid'`?

f0k:
> Wouldn't `pad=0` mean the same what currently is `border_mode='valid'`?

Yep, that's what I meant. I was just pondering whether `'valid'` would work as well, for those familiar with `border_mode` (from either numpy, scipy, or Theano). If we go for `pad`, the default should be `pad=0`.

MartinThoma:
@f0k Eventually we could give a warning when people use `pad='valid'`. Something like "`pad='valid'` is invalid. Use `pad=0` if you want to get a convolution only for those points where the filter and the input feature overlap."

But I don't think that is necessary. As Lasagne will fail loud and direct if we don't allow `pad='valid'`, people will take a look at the documentation. Lasagne is not 1.0 yet, so by now we don't have to be downwards compatible. I prefer the simpler option (allowing only `pad=0`). Especially, as it seems to be relatively obvious that `pad=0` means the same as `pad='valid'` would mean.
(And I guess people searching for "lasagne convolution border mode valid" will end up here.)

However, we should probably explicitly mention in the documentation of the `pad` parameter that `pad=0` will only apply the convolution where the input features and the filter overlap.

benanne:
Good arguments for `pad` and `pad=0` as default. It's got my vote :+1: 

I think we should probably still support `pad='valid'` as an alias, though. People might see `pad='same'` or `pad='full'` in the code and then try to change it to `'valid'`, because they know this option to work in Theano.

f0k:
> Good arguments for pad and pad=0 as default. It's got my vote :+1:

Let's go for this then, everybody involved in the discussion seems fine with it. Who will do it? I guess we should first converge on a docstring, it's a bit wasteful to do this via commits and discussions of the diff since the docstring is repeated all over the place. We need two versions: One that excludes custom padding (for the default `Conv2DLayer`), and one that includes it (for all others). I've posted an initial suggestion [further above](https://github.com/Lasagne/Lasagne/pull/311#issuecomment-115284099).

> People might see `pad='same'` or `pad='full'` in the code and then try to change it to `'valid'`, because they know this option to work in Theano.

If we officially support `valid`, we need to include it in the docstring, and `pad='valid'` looks a bit weird to me. What's *valid padding*, and why is valid padding the same as no padding? So what about not failing, but throwing a warning asking the user to better use `pad=0`?

benanne:
I'm okay with a warning. Saves us from having to document it but then at least it still works. Principle of least surprise etc. :)

MartinThoma:
>Who will do it?

I could do this in the current pull request.

> I guess we should first converge on a docstring, it's a bit wasteful to do this via commits and discussions of the diff since the docstring is repeated all over the place. We need two versions: One that excludes custom padding (for the default Conv2DLayer), and one that includes it (for all others). I've posted an initial suggestion further above.

I've added a few changes:

```
    pad : int, tuple of int, "full" or "same" (default: 0)
        By default, the kernel is only applied at positions that do not extend
        past the input, yielding an output that is smaller than the input by
        ``filter_shape[0] - 1`` rows and ``filter_shape[1] - 1`` columns when
        ``stride=1`` (a valid convolution). To change that, the input can be
        implicitly padded with zeros to extend the output.

        A single integer results in symmetric zero-padding of the given size on
        all borders, a tuple of two integers allows different zero-padding per
        dimension.

        ``"full"`` pads with ``filter_shape[0] - 1`` rows and ``filter_shape[1]
        - 1`` columns on both sides (a full convolution). This is equivalent to
          computing the convolution wherever the input and the filter overlap
          by at least one position.

         ``"same"`` pads such that the output size matches the input size when
        ``stride=1``.

        Note that ``"full"`` and ``"same"`` can be faster than equivalent
        integer values. A string indicating the convolution border mode.
```

f0k:
> I could do this in the current pull request, if that is ok.

Sure, wouldn't be a bad fit as we've got all the discussion in here! The title should probably become something like "Merge border_mode with pad argument for convolutional layers" then. Note that changing the API will also require changing the tests and possibly the examples (didn't check), but we're here to help if needed.

> I've added a few changes:

"To change that, the input can be implicitly padded with zeros to extend the output." I know I wrote this myself, but the last phrase should be "with zeros, extending the output." We can't have two "to ..." in there.
`filter_shape` should  be `filter_size` (didn't look up the argument name when writing my suggestion).
"computing the convolution" and the next two lines have extra indentation, Sphinx will bail out on this.
I'd move "(a full convolution)" to the second sentence (which is a good addition).
For the `same` convolution, maybe we should copy over the bit of the current docstring which details how we pad in case of an even filter size.
The very last bit "A string indicating the convolution border mode." is not a full sentence and I also don't think it adds anything useful, so I'd just omit it.

New suggestion (which is reusable for 1D, 2D and 3D convolution, in contrast to the former -- the only required change is the second paragraph):
```
    pad : int, tuple of int, 'full' or 'same' (default: 0)
        By default, the convolution is only computed where the input and the
        filter fully overlap (a valid convolution). When ``stride=1``, this
        yields an output that is smaller than the input by ``filter_size - 1``.
        The `pad` argument allows to implicitly pad the input with zeros,
        extending the output size.

        A single integer results in symmetric zero-padding of the given size on
        all borders, a tuple of two integers allows different symmetric padding
        per dimension.

        ``'full'`` pads with one less than the filter size on both sides. This
        is equivalent to computing the convolution wherever the input and the
        filter overlap by at least one position.

        ``'same'`` pads with half the filter size on both sides (one less on
        the second side for an even filter size). When ``stride=1``, this
        results in an output size equal to the input size.

        Note that ``'full'`` and ``'same'`` can be faster than equivalent
        integer values due to optimizations by Theano.
```

ebenolson:
> Note that ``'full'`` and ``'same'`` can be faster than equivalent
        integer values due to optimizations by Theano.

This seems unfortunate. Maybe we should recognize this and make the substitution.

f0k:
> This seems unfortunate. Maybe we should recognize this and make the substitution.

Good idea! I always forget that we know the filter size, so we can do some tricks that Theano cannot do. Where do we do the substitution, though? The easiest would be in the constructor, but people may be surprised if `layer.pad` is not what they gave as the `pad` argument. Although that also holds for upcasting to tuples... so let's just do it in the constructor (replace tuple by "full" or "same" if it happens to match)?

MartinThoma:
@f0k I like what you wrote, except for the beginning. I think the first sentence should describe what the parameter is. Something like

```text
    pad : int, tuple of int, 'full' or 'same' (default: 0)
        ``pad`` defines the padding which gets applied to the input feature
        vector before the convolution is computed.

        By default, the convolution is only computed where the input and the [...]
```

I'm not happy with my first sentence, but I hope you get my point.

(btw: This pull request exploded quite fast. I just wanted to change two lines in a fire-and-forget pull request, now it's a few dozen lines. Probably my first meaningful contribution to Lasagne :-) )

MartinThoma:
Do you have any preference regarding the argument positions? I would simply remove `border_mode`, resulting in corrmm.py

```python
    def __init__(self, incoming, num_filters, filter_size, stride=(1, 1),
                untie_biases=False, W=init.GlorotUniform(),
                 b=init.Constant(0.), nonlinearity=nonlinearities.rectify,
                 pad=None, flip_filters=False, **kwargs):
```

but I could also replace `border_mode` by `pad` and remove `pad` resulting in:

```python
    def __init__(self, incoming, num_filters, filter_size, stride=(1, 1),
                 pad=None, untie_biases=False, W=init.GlorotUniform(),
                 b=init.Constant(0.), nonlinearity=nonlinearities.rectify,
                 flip_filters=False, **kwargs):
```

benanne:
> (btw: This pull request exploded quite fast. I just wanted to change two lines in a fire-and-forget pull request, now it's a few dozen lines. Probably my first meaningful contribution to Lasagne :-) )

What? You wrote a ton of documentation before!

Our PRs and issues do have a tendency to turn into brainstorms... I guess that's not a bad thing as long as everyone who sends a PR has the patience to sit through them :p

benanne:
I didn't realise you posted two comments, sorry. Since these arguments are intended to be used as keyword arguments, I think it's okay to change the order. The 2nd proposal seems more logical to me (and we don't need to worry about backwards compatibility just yet).

f0k:
Yep, I'd go for the second option as well! (Sorry, have been abroad again.)

MartinThoma:
As @ebenolson suggested, the following sections should be removed and done automatically:

> Note that 'full' and 'same' can be faster than equivalent integer values due to optimizations by Theano.

However, I don't want to do it in this pull request. I just wanted to make sure that this change is done before #74 as it changes the API.

I just pushed the changes (it seems to work locally: 420 passed, 294 skipped).

lasagne/layers/conv.py
```diff
@@ -33,29 +33,27 @@ def conv_output_length(input_length, filter_size,
     stride : int
         The stride of the convolution operation.
 
-    border_mode : str, 'valid', 'full', 'same' or 'pad'
-        A string indicating the convolution border mode.
+    pad : int, tuple of int, 'full' or 'same' (default: 0)
```
f0k: Can't be a tuple of int in this function, only a single value. (Also needs slight adaptation of the text below.)

lasagne/layers/conv.py
```diff
@@ -33,29 +33,27 @@ def conv_output_length(input_length, filter_size,
...
-        the input, and then a valid convolution is applied.
+        ``'same'`` pads with half the filter size on both sides (one less on
+        the second side for an even filter size). When ``stride=1``, this
+        results in an output size equal to the input size.
 
-    pad : int, optional (default 0)
-        If `border_mode` is set to 'pad', this is the size of the padding that
-        is applied on both sides of the input. Otherwise, this is ignored.
+        Note that ``'full'`` and ``'same'`` can be faster than equivalent
+        integer values due to optimizations by Theano.
```
f0k: This doesn't affect the speed of `conv_output_length` at all, though, so this last paragraph should be removed.

lasagne/layers/conv.py
```diff
@@ -199,12 +198,14 @@ def __init__(self, incoming, num_filters, filter_size, stride=1,
...
+        self.pad = pad
         self.untie_biases = untie_biases
         self.convolution = convolution
 
-        if self.border_mode not in ['valid', 'full', 'same']:
-            raise RuntimeError("Invalid border mode: '%s'" % self.border_mode)
+        if self.pad not in ['valid', 'full', 'same', 0]:
+            raise RuntimeError("Invalid pad: '%s'" % self.pad)
+        if self.pad == 0:
+            self.pad = 'valid'
```
f0k: I'd probably do it the other way around, since officially (following the docstring), we only support `{0, 'full', 'same'}`. That will complicate the `self.convolution` call a little (`border_mode=pad or 'valid'`), but it might be surprising to have `Conv1DLayer.pad` be set to `valid` when you set it to `0`.

lasagne/layers/corrmm.py
```diff
@@ -149,25 +148,21 @@ def __init__(self, incoming, num_filters, filter_size, stride=(1, 1),
         self.untie_biases = untie_biases
         self.flip_filters = flip_filters
 
-        if border_mode is not None and pad is not None:
-            raise RuntimeError("You cannot specify both 'border_mode' and "
-                               "'pad'. To avoid ambiguity, please specify "
-                               "only one of them.")
-        elif border_mode is None and pad is None:
+        if pad is None:
```
f0k: This case shouldn't be supported at all, since the default is `0` anyway now.

lasagne/layers/corrmm.py
```diff
@@ -149,25 +148,21 @@ def __init__(self, incoming, num_filters, filter_size, stride=(1, 1),
...
             # no option specified, default to valid mode
             self.pad = (0, 0)
-        elif border_mode is not None:
-            if border_mode == 'valid':
+        elif isinstance(pad, basestring):
```
f0k: That's not really needed. Also you probably shouldn't replace `full` with a custom amount of padding, but leave it up to [GpuCorrMM](http://deeplearning.net/software/theano/library/sandbox/cuda/op.html#theano.sandbox.cuda.blas.GpuCorrMM) (but that was wrong before). I'd suggest:
```python
if pad == 'valid':
    self.pad = (0, 0)
elif pad == 'full':
    self.pad = 'full'
elif pad == 'same':
    self.pad = 'half'
else:
    self.pad = as_tuple(pad, 2)
```

lasagne/layers/cuda_convnet.py
```diff
@@ -71,23 +71,29 @@ class Conv2DCCLayer(CCLayer):
...
+        filter fully overlap (a valid convolution). When ``stride=1``, this
+        yields an output that is smaller than the input by ``filter_size - 1``.
+        The `pad` argument allows to implicitly pad the input with zeros,
+        extending the output size.
 
-        If 'valid', the convolution is only computed where the input and the
-        filter fully overlap.
+        A single integer results in symmetric zero-padding of the given size on
+        all borders, a tuple of two integers allows different symmetric padding
+        per dimension.
```
f0k: Different padding per dimension is not supported by this layer (this was included in the previous `pad` description), so the description is a bit misleading here. Maybe:
```
A single integer results in symmetric zero-padding of the given size on
all borders. This layer does not support using different amounts of
padding along both axes, but for compatibility to other layers you can
still specify the padding as a tuple of two same-valued integers.

File: lasagne/layers/cuda_convnet.py [hunk 7]
```diff
@@ -244,24 +241,20 @@ def __init__(self, incoming, num_filters, filter_size, stride=(1, 1),
...
+            elif pad == 'full':
                 self.pad = self.filter_size - 1
-            elif border_mode == 'same':
+            elif pad == 'same':
                 # only works for odd filter size, but the even filter size case
                 # is probably not worth supporting.
                 self.pad = (self.filter_size - 1) // 2
             else:
-                raise RuntimeError("Invalid border mode: '%s'" % border_mode)
+                raise RuntimeError("Invalid pad: '%s'" % pad)
```
f0k: I'd change the if/then/else construct a bit as suggested above, but maybe the `basestring` check is not that bad anyway. This time, `'full'` and `'same'` cannot be handled by the implementation and need to be converted to explicit padding as done here :+1:

lasagne/layers/dnn.py
```diff
@@ -220,34 +218,25 @@ def __init__(self, incoming, num_filters, filter_size, stride=(1, 1),
...
-        elif border_mode is not None:
-            if border_mode == 'valid':
+        elif isinstance(pad, basestring):
+            if pad == 'valid':
                 self.pad = (0, 0)
-                self.border_mode = 'valid'
-            elif border_mode == 'full':
+            elif pad == 'full':
                 self.pad = (self.filter_size[0] - 1, self.filter_size[1] - 1)
-                self.border_mode = 'full'
```
f0k: You should definitely set `self.pad = 'full'` in this case, to have `dnn_conv()` handle it.

MartinThoma:
Hmm ... I just realized that I had some errors in `get_output_shape_for` of `dnn.py`. I guess this was not covered by the tests as I tested on my small (non-GPU) machine. I'll test it on my GPU machine this evening and I'll add the changes suggested by @f0k then.

lasagne/layers/dnn.py
```diff
@@ -284,15 +273,11 @@ def get_output_shape_for(self, input_shape):
...
-        # if 'border_mode' is one of 'valid' or 'full' use that.
-        # else use pad directly.
-        border_mode = (self.border_mode if (self.border_mode is not None)
-                       else self.pad)
 
         conved = dnn.dnn_conv(img=input,
                               kerns=self.W,
                               subsample=self.stride,
-                              border_mode=border_mode,
+                              border_mode=pad,
```
f0k: This should fail the tests, `pad` is not a defined local variable. It's `self.pad`.

lasagne/tests/layers/test_conv.py
```diff
@@ -286,9 +298,9 @@ def test_pad(self, DummyInputLayer):
         input_layer = DummyInputLayer((1, 2, 3, 3))
         with pytest.raises(RuntimeError) as exc:
             layer = Conv2DDNNLayer(input_layer, num_filters=1,
-                                   filter_size=(3, 3), border_mode='valid',
+                                   filter_size=(3, 3),
                                    pad=(1, 1))
-        assert ("You cannot specify both 'border_mode' and 'pad'" in
+        assert ("You cannot specify both 'pad' and 'pad'" in
                 exc.value.args[0])
```
f0k: Uhm, are you sure you executed the tests locally? This is nonsense :) You should just remove the last 7 lines.

MartinThoma: This is exaclty what I meant when I wrote that these tests seem to be excluded as I don't have a GPU on my usual development machine :-) 

I already fixed that, but I want to test it on my GPU machine in a few hours before I commit / push it.

MartinThoma:
Ok, very strange. Even on my GPU computer with `py.test --runslow` it skipped quite a couple of tests. Is there an option to tell py.test to run all tests? Were they skipped because I run py.test before and nothing changed (for these tests)?

benanne:
Which tests were skipped specifically? Have you ensured CuDNN is set up correctly? Sounds like some of the tests might "give up" too easily :)

MartinThoma:
I'll test that tomorrow (my GPU machine is quite loud and I have about 30°C
in my room by now...)
I remember that some of the `theano_extensions` tests were skipped. But
I'll add a complete list tomorrow.

> Have you ensured CuDNN is set up correctly?

I don't know. How do I find out?

I do use Theano for neural network training on this machine and it works
fine (with GPU).


2015-07-01 22:49 GMT+02:00 Sander Dieleman <notifications@github.com>:

benanne:
If CuDNN is not set up correctly, Theano will silently fall back to `GpuCorrMM` if you're using `conv2d` afaik. I guess a good way to test it is to use `Conv2DDNNLayer` and see if the code runs (`dnn_conv` will fail if it is not set up correctly, there is no fallback). If those tests run it should be fine at any rate. And yeah, I appreciate that you may not want to stress your GPUs right now ;)

MartinThoma:
@benanne One reason seems to be that cuDNN is not installed. I've opened #316 because I think this should be mentioned there. 

https://gist.github.com/MartinThoma/68e5e78101a3fbaed334 is the complete list of my tests. I will try it again after I have cuDNN (which might take a bit... I don't know how fast they accept me as a "registered developer").

f0k:
Seems you addressed all my comments and I couldn't find anything new, thank you! I'd be fine with merging when you squash the commits. (Bonus points if you manage to keep the "sentences start with a capital letter ..." one separated, because it's independent of the API change. I don't know what would be a good way to achieve that, though, other than cherry-picking all commits in a particular order and then squashing the correct ones. Don't worry if that's too complicated.)

MartinThoma:
@f0k Did you test all of them? I still could not run all tests because of https://github.com/Lasagne/Lasagne/issues/316 (sorry, I didn't find the time by now to try what you suggested).

I just did 

```bash
$ git rebase HEAD~6
(squashing 5 of them, leaving the "sentences start..." commit)
$ git push origin master --force
```

but it seems not to be pushed. I can see 

```bash
$ git log
commit e75f5e4058abcf5d5b5bdbcc24b9d53a87695936
Author: Martin Thoma <info@martin-thoma.de>
Date:   Wed Jun 24 20:33:06 2015 +0200

    lasagne/layers: Merge border_mode with pad argument for convolutional layers; set pad=0 as default

commit 6541ab6c17997ecda4695ac889d0fa6d338cb4eb
Author: Martin Thoma <info@martin-thoma.de>
Date:   Thu Jun 25 09:11:39 2015 +0200

    layers: sentences start with a capital letter and and with a point; use (default: whatever) style consistently; use 'int' consistently (instead of 'integer'); True and False have capital letters in Python

commit 9f591a5f3a192028df9947ba1e4903b3b46e8fe0
Merge: d71b4ce 6ad0211
Author: Jan Schlüter <github@jan-schlueter.de>
Date:   Tue Jun 23 15:22:29 2015 +0300

    Merge pull request #301 from skaae/generalize_sumlayer
    
    add ElemwiseMergeLayer
    Closes #300.

commit 6ad0211c325aca1218a794b4513c036cb04b1325
Author: skaae <skaaesonderby@gmail.com>
Date:   Mon Jun 22 19:15:12 2015 +0200

    generalize summing layer

```

What did I do wrong? What should I do instead?

lasagne/layers/dnn.py
```diff
@@ -220,34 +218,22 @@ def __init__(self, incoming, num_filters, filter_size, stride=(1, 1),
...
+        if isinstance(pad, basestring):
+            if pad == 'valid':
                 self.pad = (0, 0)
-                self.border_mode = 'valid'
-            elif border_mode == 'full':
-                self.pad = (self.filter_size[0] - 1, self.filter_size[1] - 1)
-                self.border_mode = 'full'
-            elif border_mode == 'same':
+            elif pad == 'full':
+                self.pad = 'full'
```
ebenolson: This is causing an error in the calculation of `output_rows` and `output_columns` below. Also, `basestring` doesn't exist in python3, so that will need to be changed also.

f0k:
> ```
git push origin master --force
```
> but it seems not to be pushed.

Well surely it got pushed, but to `master`. You wanted to push it to `conv-doc` instead. Can you fix this? You seem to have done a `git pull` and `git push` later, so there's a merge commit in there which is worse than before... Also could you please remove the email addresses from [your comment](https://github.com/Lasagne/Lasagne/pull/311#issuecomment-121024171)? That's too easy for spammers to collect.

File: lasagne/layers/dnn.py (hunk 11 above)
Comment (f0k) [inline lasagne/layers/dnn.py]: So then maybe my other suggestion for the `if` clauses would be better to avoid the `basestring` check: https://github.com/Lasagne/Lasagne/pull/311#discussion_r33666980

**Comment by MartinThoma** [comment]
@f0k Wow ... I'm sorry, I keep making it worse when I try to fix the commits. I could create a new branch, manually add the changes and create a new pull request. Would that be ok?

**Comment by f0k** [comment]
Well, you had it in a good state once. You could try the following:
```
git checkout -b backup
git checkout conv-doc
git reset --hard 9f591a5
git cherry-pick e75f5e4
git cherry-pick 6541ab6
git push --force origin conv-doc
```
If those two commits (the ones you showed in your `git log` command above) really contain everything, this should do the trick.

MartinThoma:
@f0k I think it would be better if you could continue this. It seems as not all tests were working before my changes (all add a paste in a couple of minutes; I just let `py.test --runslow` run for the master branch).

f0k:
The errors are about cuDNN. You can still try what I wrote in the issue you opened: https://github.com/Lasagne/Lasagne/issues/316#issuecomment-118072158
Anyway, I'll continue this sometime soon if you don't have the time now.

PR #314, opened 2015-07-01
regularization doctoring

skaae:


skaae:
The funny naming was autocorrect :)

lasagne/regularization.py
```diff
@@ -1,3 +1,53 @@
+"""
+Functions to apply regularization to the weights in a network.
+
+We provide functions to calculate the L1 and L2 penalty. Penalty functions
+take a tensor as input and calculate the penalty contribution from that tensor:
+
+.. autosummary::
+    :nosignatures:
```
f0k: Maybe we could add the signatures here, because they're just a single `(x)`? For the other two tables, the signatures are probably too long.

f0k:
Looks good! And thanks, that was fast :)

lasagne/regularization.py
skaae: In my opinion it still look a little odd with signatures and `x` doesn't really convey any information. 
I can rename `x` to `tensor` and add the signature?

skaae: That would look like:
signatures

f0k: Hmm, not sure about changing the signature, we also have `x` in the `nonlinearities` module: http://lasagne.readthedocs.org/en/latest/modules/nonlinearities.html
And numpy also just uses `x`, e.g., in `np.square`: http://docs.scipy.org/doc/numpy/reference/generated/numpy.square.html
I think it's clear enough the way it is (but I still like `l1(tensor)`). @ebenolson, since you wrote this, do you have any opinion?

Otherwise let's just leave it as it is! (Without any signatures in the table, and with `x` instead of `tensor`.) The `apply_penaly` signature really seems too long for the summary table.

benanne:
lgtm as well

lasagne/regularization.py
ebenolson: I prefer short parameter names, and I doubt anyone is going to be confused by `x`. But I don't feel that strongly about it.

skaae: Ok, I guess that settles it on the current docstring?

f0k:
All right, merging! Thanks, @skaae!

PR #315, opened 2015-07-01
remove string representations of W normal init in embedding layer doc

skaae:


benanne:
Why should this be removed? Won't this mess up the layer's signature in the documentation?

skaae:
Maybe the description is bad. The PR changes the signature of the embedding layer from

```
lasagne.layers.EmbeddingLayer(incoming, input_size, output_size, W=<lasagne.init.Normal object at 0x7f8b6e104050>, **kwargs)
```

to 

```
lasagne.layers.EmbeddingLayer(incoming, input_size, output_size, W=lasagne.init.Normal(), **kwargs)
```

Its only a change in the docs and is similar to what is done for e.g. DenseLayer?

benanne:
right, it's adding the string, not removing it. The fact that it's highlighted in green in the diff and not in red should have tipped me off ;) Sorry for the confusion! Looks good to me, merging.

PR #318, opened 2015-07-03
fix Pool2DDNNLayer stride

ebenolson:
small fix to Pool2DDNNLayer

### DIFF
--- a/lasagne/layers/dnn.py
+++ b/lasagne/layers/dnn.py
@@ -72,7 +72,10 @@ class Pool2DDNNLayer(DNNLayer):
                  mode='max', **kwargs):
         super(Pool2DDNNLayer, self).__init__(incoming, **kwargs)
         self.pool_size = as_tuple(pool_size, 2)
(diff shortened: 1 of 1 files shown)


ebenolson:
also updated get_output_shape_for as in https://github.com/Lasagne/Lasagne/pull/307

benanne:
The formula used by `pool_output_length` is slightly different but should amount to the same thing for every possible input, as far as I can tell. But I'd be more comfortable merging this if someone else compared them as well, just to make sure :)

f0k:
The formulas are exactly the same. The former one here was:
```python
  (input_length + 2 * pad - pool_size) // stride + 1
```
And the one in `pool_output_length` is:
```python
  (input_length + 2 * pad - pool_size + 1 + stride - 1) // stride
= (input_length + 2 * pad - pool_size + stride) // stride
= (input_length + 2 * pad - pool_size) // stride + 1
```
Merging!

PR #319, opened 2015-07-03
one_hot docstring

mgrachten:
add note to one_hot function doc that refers potential users to the ability of categorical_crossentropy to handle integer vectors directly

### DIFF
--- a/lasagne/utils.py
+++ b/lasagne/utils.py
@@ -97,6 +97,16 @@ def one_hot(x, m=None):
     Theano tensor variable
         A Theano tensor variable of shape (``n``, `m`), where ``n`` is the
         length of `x`, with the one-hot representation of `x`.
+
+    Notes
+    -----
+
+    If your integer vector represents target class memberships, and you wish to
+    compute the cross-entropy between predictions and the target class
+    memberships, then there is no need to use this function, since the function
+    :func:`categorical_crossentropy()` can compute the cross-entropy from the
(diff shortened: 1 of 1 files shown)


benanne:
Thanks :)

lasagne/utils.py
```diff
@@ -97,6 +97,16 @@ def one_hot(x, m=None):
     Theano tensor variable
         A Theano tensor variable of shape (``n``, `m`), where ``n`` is the
         length of `x`, with the one-hot representation of `x`.
+
+    Notes
+    -----
+
```
f0k: @mgrachten: I think this empty line is what prevents the notes from showing up in the rendered documentation. You need to stick exactly to [the numpydoc howto](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) to make it work. Do you want to send a PR fixing this?

lasagne/utils.py
```diff
@@ -97,6 +97,16 @@ def one_hot(x, m=None):
...
+    If your integer vector represents target class memberships, and you wish to
+    compute the cross-entropy between predictions and the target class
+    memberships, then there is no need to use this function, since the function
+    :func:`categorical_crossentropy()` can compute the cross-entropy from the
+    integer vector directly.
+
```
f0k: We usually don't end with an empty line here (although this should not influence Sphinx at all).

PR #320, opened 2015-07-03
Let MaxPool2DLayer pass through mode argument

ebenolson:


benanne:
Hmm, it doesn't really to make sense to have a layer called `MaxPool2DLayer` that can also do average pooling. Maybe we should make a `Pool2DLayer` and have a `MaxPool2DLayer` that derives from it with the mode set to max-pooling, like we have for the CuDNN versions?

ebenolson:
Yeah, that makes more sense. Made the change.

lasagne/layers/pool.py
```diff
@@ -221,10 +227,22 @@ def get_output_for(self, input, **kwargs):
...
                                         ignore_border=self.ignore_border,
                                         padding=self.pad,
+                                        mode=self.mode,
                                         )
         return pooled
 
 
+class MaxPool2DLayer(Pool2DLayer):  # for consistency
+    def __init__(self, incoming, pool_size, stride=None,
+                 ignore_border=False, pad=(0, 0), mode='max', **kwargs):
```
benanne: this layer should not have the `mode` argument I think.

requirements.txt
```diff
@@ -1 +1 @@
-Theano==0.7
+git+https://github.com/Theano/Theano.git@4dacb9e7d04d842d8ffe2c948dbc90722ae990d4#egg=Theano
```
benanne: Is this necessary? I don't think we should require a git version of Theano for our imminent release. We should ensure everything works with 0.7.

ebenolson: 0.7 is months old at this point, and this feature doesn't exist in it. This is why I think releases are a bad idea :)

benanne: hmm, that complicates matters. Well at any rate I think it would be a bad idea for us to depend on a git version of Theano, for now. When people do `pip install Theano`, 0.7 is the version they will get.

lasagne/layers/pool.py
ebenolson: fixed

requirements.txt
ebenolson: I really don't think that makes sense. Theano went for over a year between releases, who knows when the next one will be. 

We should encourage people to use the git version, if we specify a commit in `requirements.txt`, it will be installed when people `pip install Lasagne`.

benanne: I think that's a discussion that deserves its own issue. Previously there was an implicit agreement that we would standardize on version 0.7 as soon as it was released. If we're going to change our stance on this we should discuss it thoroughly first.

ebenolson: Ok, didn't realize that. The reason I'm pushing for this because one of the examples I made for recipes is broken without this feature.

benanne: I see. Well, we should definitely consider it - I'm not strictly opposed to it. But I would like to hear as many of the core team members' input as possible (most notably @f0k and @dnouri who I think will have more insight than me into the implications of this change).

nouiz: We will try to make a release this summer. But it isn't sure
Le 4 juil. 2015 09:31, "Sander Dieleman" <notifications@github.com> a
écrit :

lasagne/layers/pool.py
```diff
@@ -161,14 +162,18 @@ class MaxPool2DLayer(Layer):
...
         Number of elements to be added on each side of the input
         in each dimension. Each value must be less than
         the corresponding stride.
 
-    ignore_border : bool
-        If ``True``, partial pooling regions will be ignored.
-        Must be ``True`` if ``pad != (0, 0)``.
+    mode : string
+        Pooling mode, one of 'max', 'average_inc_pad' or 'average_exc_pad'.
+        Defaults to 'max'.
```
f0k: From the [numpydoc HOWTO](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections), this should be:
```
    mode : {'max', 'average_inc_pad', 'average_exc_pad'}
        Pooling mode: max-pooling or mean-pooling including/excluding zeros in
        partial pooling regions. Defaults to 'max'.
```

lasagne/layers/pool.py
```diff
@@ -141,11 +141,12 @@ def get_output_for(self, input, **kwargs):
         return pooled[:, :, :, 0]
 
 
-class MaxPool2DLayer(Layer):
+class Pool2DLayer(Layer):
```
f0k: You will need to add this to `__all__` above, and to the documentation.

lasagne/layers/pool.py
```diff
@@ -221,10 +227,22 @@ def get_output_for(self, input, **kwargs):
                                         st=self.stride,
                                         ignore_border=self.ignore_border,
                                         padding=self.pad,
+                                        mode=self.mode,
                                         )
         return pooled
 
 
+class MaxPool2DLayer(Pool2DLayer):  # for consistency
```
f0k: This needs a docstring. Either copy what we had before, or tell that it's a shortcut for `Pool2DLayer` with `mode=max`. If possible, have a look at the documentation generated by Sphinx to check whether your solution is good.

requirements.txt
f0k: With #329, this isn't needed any more! I guess you will need to squash commits after removing this line, otherwise it will still not merge. Thanks!

ebenolson:
@f0k I think I addressed all your comments above. If it looks good I'll squash.

lasagne/layers/pool.py
```diff
@@ -221,10 +232,53 @@ def get_output_for(self, input, **kwargs):
...
+        Must be ``True`` if ``pad != (0, 0)``.
+    **kwargs
+        Any additional keyword arguments are passed to the :class:`Layer`
+        superclass.
+    Notes
+    -----
+    The value used to pad the input is chosen to be less than
+    the minimum of the input, so that the output of each pooling region
+    always corresponds to some element in the unpadded input region.
+    """
```
f0k: For some reason, all empty lines got missing here. Could you please fix this? I guess it won't work in Sphinx this way. (Just copy/paste it again from `Pool2DLayer`, change the first sentence and remove `mode` and `See also`.)

requirements.txt
```diff
@@ -1 +1 @@
-git+https://github.com/Theano/Theano.git@36d1eca8#egg=Theano==0.8.git
+git+https://github.com/Theano/Theano.git@36d1eca8#egg=Theano==0.8.git
```
f0k: Interesting. I guess this will vanish when you squash?

ebenolson: yeah I don't understand what happened there.

lasagne/layers/pool.py
ebenolson: very strange.. apparently copy/pasting from github ignores blank lines..

f0k:
Thanks, everything looks good for merging now! Can you please squash?

ebenolson:
squashed!

requirements.txt
f0k: Weird, it's still there! I'll have a look tomorrow.

ebenolson: It was missing a newline... fixed now.

f0k:
Great, thanks for fixing the last bit. Merging!

PR #321, opened 2015-07-03
modify theano_extensions and PadLayer to support arbitary padding

ebenolson:


benanne:
Looks okay to me. Not 100% sure about the wording in the docstring though. It says to pass either an integer, a tuple or a list of tuples, but  if I understand correctly any iterable / iterable of iterables should be accepted, right?

ebenolson:
A tuple of tuples would be fine, but a generic iterable won't work, because it indexes `width[k - batch_ndim]`. If you think it's important I can try to find a workaround.

benanne:
Nah, I guess that's fine then :)

f0k:
What about `2-tuple` instead of `tuple`? I find it a bit difficult to understand the docstring now because it seems it might accept an arbitrary-length list of arbitrary-length tuples. Also doesn't it accept a list of int as well (to specify a different padding per dimension, but use symmetric padding, i.e., the same padding width for the beginning and end)?

ebenolson:
No, it doesn't accept that form. Could copy the docstring from [np.pad](http://docs.scipy.org/doc/numpy/reference/generated/numpy.pad.html) since that's what I modeled it after, although I don't find that wording so clear either.

f0k:
> No, it doesn't accept that form.

Are you sure? [Line 327](https://github.com/Lasagne/Lasagne/pull/321/files#diff-3be3c9439d563fba8eae882eedc1c853R327) looks like it does:
```python
try:
    l[k], r[k] = width[k - batch_ndim]
except TypeError:
    try:
        l[k] = r[k] = width[k - batch_ndim]
    except TypeError:
        l[k] = r[k] = width
```
First it tries to interpret it as a list of 2-tuples, then as a list of int, then as an int.

lasagne/layers/shape.py
```diff
@@ -309,11 +314,20 @@ def __init__(self, incoming, width, val=0, batch_ndim=2, **kwargs):
...
             else:
-                output_shape += (s + 2 * self.width,)
+                try:
+                    l[k], r[k] = self.width[k - self.batch_ndim]
+                except TypeError:
+                    try:
+                        l[k] = r[k] = self.width[k - self.batch_ndim]
+                    except TypeError:
+                        l[k] = r[k] = self.width
+                output_shape += (input_shape[k] + l[k] + r[k],)
```
f0k: Using lists `l` and `k` when you're just interested in the current value seems a bit wasteful. What about:
```python
def get_output_shape_for(self, input_shape):
    if isinstance(self.width, int):
        widths = [self.width] * (len(input_shape) - self.batch_ndim)
    else:
        widths = self.width
    for k, w in enumerate(self.width):
        try:
            l, r = w
        except (TypeError, ValueError):
            l = r = w
        output_shape += l + r
```
This also handles arbitrary iterables for `self.width`. Note that your solution did not catch the `ValueError` for a wrong-length tuple.

ebenolson:
You're right, I was think of the case of passing a 2-tuple/list with different values for each edge but the same for each dimension.

lasagne/layers/shape.py
ebenolson: good point, I changed it to your method.

f0k:
> Could copy the docstring from np.pad since that's what I modeled it after, although I don't find that wording so clear either.

Hmm, yes, that one is not easy to read either. What about:
```
    width : int, tuple, or list of int/tuple
        Padding width. If an int, pads each axis symmetrically with the same
        amount in the beginning and end. If a tuple of two int, pads the
        beginning with the first and the end with the second amount. If a list,
        defines the padding width or padding widths separately for each axis.
```

ebenolson:
I like the wording, but it made me realize I've been coding something else :(
Ok, the possible cases are:

1. Symmetric padding, equal for each axis (width = int)
2. Symmetric padding, different for each axis (width = list/tuple of int)
3. Asymmetric padding, equal for each axis (width = list/tuple of int)
4. Asymmetric padding, different for each axis (width = list of list/tuple of int)

Right now, the code does 1, 2, and 4. I think 3 is a less useful case, but it's the one supported by np.pad so I should probably change it to that. Although maybe it would be less confusing to accept only 1 and 4.

f0k:
Ah, gotcha. I didn't realize that `(1, 2)` could be interpreted as both "pad with 1 in beginning and 2 in end, for each axis" or "symmetrically pad first axis with width 1, second axis with width 2".

> Although maybe it would be less confusing to accept only 1 and 4.

It would be less confusing in terms of the implementation, but it would leave people asking why there's nothing between "pad every edge of every axis the same" and "pad every edge of every axis differently".

What about supporting all four cases by sticking to the docstring (i.e., requiring a `tuple` for separate beginning/end padding, and requiring a not-`tuple` for separate padding per axis)?

benanne:
> What about supporting all four cases by sticking to the docstring (i.e., requiring a `tuple` for separate beginning/end padding, and requiring a not-`tuple` for separate padding per axis)?

I don't think it's a good idea to give different types of iterables (i.e. tuples and lists) different semantics.

skaae:
Have you considered writing the docstring as a bullet list describing each of the cases?

ebenolson:
What about separate kwargs to cover the different cases?
* `width = int` fully symmetric
* `width = (int, int, ...)` different for each axis
* `sides = (int, int)` asymmetric, equal for each axis
* `sides = ((int, int), ...)` fully asymmetric

f0k:
> Have you considered writing the docstring as a bullet list describing each of the cases? 

I know you like bullet lists, but I think the docstring is quite readable now (each sentence describes one case, without excessive vertical space). Anyway, what we're discussing is that the implementation doesn't match it (and cannot match it if we don't distinguish different kinds of iterables).

f0k:
> What about separate kwargs to cover the different cases?

I'd prefer to just drop case 3. We don't have to follow `np.pad`, I doubt everyone looking at `PadLayer` is familiar with the details of it, and case 4 allows to express case 3. I'd then change the docstring to:
```
    width : int, iterable of int, or iterable of tuple
        Padding width. If an int, pads each axis symmetrically with the same
        amount in the beginning and end. If an iterable of int, defines the
        symmetric padding width separately for each axis. If an iterable of
        tuples of two ints, defines a seperate padding width for each beginning
        and end of each axis.
```

ebenolson:
works for me, updated.

lasagne/theano_extensions/padding.py
```diff
@@ -7,29 +7,50 @@
...
+        output_shape[k + batch_ndim] += l + r
+        indices[k + batch_ndim] = slice(l, l + input_shape[k + batch_ndim])
 
+    indices = tuple(indices)
+    output_shape = tuple(output_shape)
+    if val:
+        out = T.ones(output_shape) * val
+    else:
+        out = T.zeros(output_shape)
+    print(output_shape)
```
f0k: `output_shape` and `print(output_shape)` should probably be removed before merging :)

ebenolson: oops :)
cleaned that up, I'll squash when there are no more comments.

ebenolson:
Hmm, for whatever reason indexing with a list works in Theano master but not 0.7. 
In the interests of pragmatism I put the conversion to tuple back in.

benanne:
Good call. Happy to see a consensus about the interface as well :) Great work!

lasagne/tests/test_theano_extensions.py
```diff
@@ -98,3 +98,33 @@ def test_pad(val, width=3, batch_ndim=2):
     X_pad_np = np.pad(X0, pads, mode='constant', constant_values=val)
 
     assert (X_pad_theano == X_pad_np).all()
+
+
+def test_pad_width_per_axis(val=0, batch_ndim=2):
```
f0k: The `batch_ndim` argument is a bit misleading, because setting it to a different value will only affect `X_pad_theano`, but not `X_pad_np` (i.e., it suggests the test was more flexible than it actually is). I'd just remove it. However, we don't have a test for `batch_ndim != 2` now, do we? Maybe just modify one of the tests for `batch_ndim == 1` as an easy way out?

ebenolson: good point, I parametrized batch_dim in the tests.

f0k:
> I parametrized batch_dim in the tests.

Nice! Didn't want to ask for it, but that's obviously the best solution. Good work, no more comments from me, I'm fine with squashing and merging!

ebenolson:
great, squashed.

f0k:
Travis passed, Sander seemed happy as well, merging! Thanks!

Issue #174, opened 2015-03-21
Draw network diagrams


ebenolson:
I wrote some functions to draw network diagrams with graphviz, like Caffe's `draw_net.py`. 

I don't think this belongs in the main library as pydot/graphviz adds a pretty big dependency, but it might make a good addition to the examples or some future collection of utilities. 

Code: https://github.com/ebenolson/Lasagne/blob/master/examples/draw_net.py

example output: https://github.com/ebenolson/Lasagne/blob/master/examples/mnist_conv.pdf

dimatura:
Nice!

benanne:
Very cool! I wonder if we can come up with some way to make all this code easily available to people without having all of it in the main library. I imagine people will start coming up with various utility functions and `Layer` subclasses that are too specialized to include in the main library, but other people might still find them useful. Maybe we need some kind of official list of 'extensions' somewhere? We could put it on the wiki or add it in the documentation.

f0k:
> I don't think this belongs in the main library as pydot/graphviz adds a pretty big dependency

Theano has `theano.printing.pydotprint`. I wouldn't mind including some printing functionality in Lasagne as well (a textual representation of a graph would be cool, too). As long as we don't auto-import it, I wouldn't worry about the extra dependency either.

jsusskin:
@ebenolson, nice work on the draw functions. In experimenting with it I found that for some reason if the I ran into an interesting annoying issue with the multi-line labels where if the shape='record' it simply ignores the \n character and renders the whole label on one line. This is on my macbook pro. Fortunately, I found that if I change the shape parameter to other node shapes like "box" or "ellipse" it renders multiline labels properly. You might want to do some testing to see if that's just something weird with my setup I guess.

PR #324, opened 2015-07-04
"Informal" string representation for Layer Class

sidps:
Based on [this mailing list post](https://groups.google.com/forum/#!searchin/lasagne-users/__repr__/lasagne-users/CtDfEC3wVnk/ahS2TefZt0cJ), I wrote a simple __str__ method for the base class which should return an easy to understand "informal" representation of the entire network contained under the specific Layer, similar to what Torch does.

For example, if a model such as the one in mnist.py example is built using

    >>> network = build_model(100, 10)

then one could

    >>> print(network)

to get

    input --> [0] --> [1] --> [2] --> [3] --> [4] --> [5] --> output 
    Details:
            [0]: InputLayer
            [1]: DenseLayer
            [2]: DropoutLayer
            [3]: DenseLayer
            [4]: DropoutLayer
            [5]: DenseLayer

To return more details for each layer, the corresponding layer class (that inherits from Layer) must override a method called __layer_str()

Note that this currently supports only completely sequential networks, but should be simple enough to extend.

### DIFF
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -53,8 +53,12 @@ class Layer(object):
             Please do NOT override this method, or __repr__!
             To create a representation of the layer, override __layer_str.
             """
+        # TODO: this supports sequential networks, add support for networks with merge layer
         network = get_all_layers(self)
-        representation = ''
+        representation = 'input --> '
+        representation += ''.join(['[' + str(i) + '] --> ' for i in range(len(network))])
+        representation += 'output \n' \
+                          'Details:'
         for i in range(len(network)):
             try:
                 layer_repr = network[i].__layer_str()
@@ -62,7 +66,7 @@ class Layer(object):
                 layer_repr = type(network[i]).__name__
             except:
                 raise
-            representation += '\n[{0}]: {1}'.format(i, layer_repr)
+            representation += '\n\t[{0}]: {1}'.format(i, layer_repr)
         return representation
 
     def __layer_str(self):
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -1,6 +1,7 @@
 from collections import OrderedDict
 
 from .. import utils
+from .helper import get_all_layers
 
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


ebenolson:
While a tidier `__repr__` for Layers would be nice, I think this would fit better as a utility function, or as a method of a hypothetical network class. 

I'm also a bit skeptical that this would be simple to extend to branched networks.

sidps:
That does seem like a better idea, I could certainly move this into a util function. However, any detailed representation of a layer would require implementing/overriding a specific method that the util function could call, perhaps the `__str__` itself?

I admit not having given much thought about branched networks. I'd initially conceived of looping through the layers list to detect a merge layer and taking it from there. Alternately, perhaps the `__str__` could be built recursively through the layers, where the merge layer handles the specifics of the merge representation? But the latter would be moving away from implementing this as a utility function. 

I'm not sure how this would work for networks that branch out and don't merge later. I'm perhaps not sure how such networks are handled in Lasagne either. I suppose you'd have to use a function handle for each branch?

ebenolson:
Have a look at the code of the [`get_all_layers`](https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/helper.py#L23) helper function for the current approach to branching layers. 

I was more wondering how you would lay out a complex network in text form. I did a bit of work on something similar a while back (https://github.com/Lasagne/Lasagne/issues/174), but I let a graph library do most of the job.

For simple linear networks, I don't really see the advantage over just printing a list. 
But I do think it would be nice to define `__str__` so instead of getting 

    <lasagne.layers.dnn.Conv2DDNNLayer at 0x7fa13df38e90>

we could have something like

    <Conv2DDNNLayer: filter_size=(3, 3), output_shape=(None, 11, 7, 7)>

sidps:
I second that, it would indeed be excellent if we could have informative `__str__`s. Perhaps it would be good to describe a standard as well, if this is going to be the approach in future.

For textual representation of the network, I initially intended mimicking torch loosely, but that doesn't have to be (I'm not sure they're immediately comprehensible either). An alternative would be to implement something like `git log --graph`, but handling cycles may not be straightforward.

sidps:
See [Recipes PR # 3](https://github.com/Lasagne/Recipes/pull/3).

Issue #322, opened 2015-07-04
Update mnist example to support step-by-step learning and remove inversion of control


buriy:
I'd like to improve mnist example to avoid unnecessary complication and allow easier reusability.

Currently we do:
```python
def train(iter_funcs, dataset, batch_size=BATCH_SIZE):
    for epoch in itertools.count(1):
        ..do training...
        yield result

for epoch in train(iter_funcs, dataset):
    print epoch
```
Instead I'd like to see:
```python
def train_epoch(iter_funcs, dataset, batch_size=BATCH_SIZE):
    ..do training...
    return result

for epoch in itertools.count(1):
    result = train_epoch(iter_funcs, dataset)
    print result
```
The current way of doing it looks like an unnecessary complication to me, especially for the people who want to feed network with a lot of data in macro batches that don't fit into memory.

I think the library-like continuous learning concept
```python
some_more_data, some_more_labels = grab_more_data_and_labels()
network.learn(some_more_data, some_more_labels)
```
is more general and scalable, compared to framework-like learning-once concept:
```
network.learn_until_done(all_data, all_labels)
```

ebenolson:
There's been discussion of this in https://github.com/Lasagne/Lasagne/issues/215, but I don't think anyone's working on it at the moment, so feel free to open a PR.

We're also starting to populate the [Recipes](https://github.com/Lasagne/Recipes) repo with other examples if you'd like to contribute something there.

benanne:
Totally agreed that we went a bit overboard with the current examples, they should be simplified. As @ebenolson said we're already discussing this but I don't think anyone is planning to work on it right now. So your help would be very welcome :)

f0k:
Well, I've been planning to simplify the example(s) last week, but had to postpone it. I'm again planning to do it this week, and fully agree with your second-to-last code block! The MNIST example will probably be converted to a flat file with as few indirections as possible anyway, focusing on an explanation of the library rather than giving a way to engineer your code for maximum reuse.

buriy:
f0k, Also I haven't found non-trivial for newscomers to understand how to reuse the pipeline for macro batches.
Recompiling the pipeline for each macrobatch isn't a good solution.
Would be cool if you showed the data reloading in the example.
I mean, input_layer.set_value(...) instead of rebuild_network for consequent batches.

f0k:
> Would be cool if you showed the data reloading in the example.

[...]

Issue #215, opened 2015-04-20
Modify examples to be simpler


diogo149:
I'm curious if other people would be interested in having examples in this fashion: https://github.com/enlitic/lasagne4newbs/blob/master/mnist_conv.py. I've found the existing examples to be a little confusing to people new to Lasagne, and made that version to help them out. I think the biggest downside would be teaching slightly less than optimal practices (specifically not transfering a large amount of input data to the GPU at once).

benanne:
I'm inclined to agree, we may have gone overboard cleaning them up a bit in the past (they used to be simpler before, e.g. https://github.com/Lasagne/Lasagne/commit/98f2ee75a1bf0b7887d10a1cd1075a3d7b3e3db7).

It might be better to dial that back a little and cut some corners in the interest of immediate clarity of the code, since the main purpose of the examples is to quickly show people how to get things done, after all. While all the functions are a good idea in practice, they do obscure the flow a bit.

One downside is that this will complicate testing of the examples. We currently have tests that run each example for 1 epoch to ensure everything still works. With scripts like this one, this is impossible (unless we make it accept command line arguments which in turn complicates the example again). I wonder if there's a way around this.

senbon:
Just thinking out loud here, but wouldn't it be an idea to move usage examples of Lasagne into another repository within the Lasagne organization?

This would further reduce the scope of this library while also simplify test coverage.

f0k:
> I wonder if there's a way around this.

We don't have to cut it back that far. Having a `main()` function is probably okay, and I think wrapping the model creation in a function is nice as well. What I find confusing are the indirections of `create_iter_functions()` and `train()` -- I need to flip between files and read a lot of code to understand what's going on, I need to know about generator functions, and I have to focus hard to keep in mind that despite the name, `create_iter_functions` doesn't have anything to do with Python's `iter`.

The only reason why we would want those indirections is to share code between the four examples (one for an MLP, three for CNNs), but the current examples are still a pretty complete failure at sharing code. What if we strip that back to a single `mnist.py` that has a parameter to choose which `build_model()` is executed: one for a simple MLP, one for a CNN? That example should have extensive comments similar to Diogo's one, to explain things in a tutorial style rather than just throwing the code at the unsuspecting :)

> I think the biggest downside would be teaching slightly less than optimal practices (specifically not transfering a large amount of input data to the GPU at once).

[...]

PR #328, opened 2015-07-08
Rewrite of MNIST example

f0k:
Includes both the dense and the convolutional architecture in a single example,
reduces indirections in the code and adds a lot of explanations (addressing #215, #322).

Also changes the code to not copy the dataset to GPU at once, but in batches -- this is easier to understand and modify for large datasets or on-the-fly augmentation. While at it, it also changes the code to shuffle the training data at each epoch. Finally, it changes the batch size from 600 to 500 so it evenly divides the train/val/test sizes (50,000, 10,000 and 10,000 examples each).

Although each of those changes should negatively affect performance, performance is almost the same as before (note: you need to run it as `mnist.py custom_mlp:2,512,0,0.5` to get the same architecture as in the original script).

The only question is whether we want the example file to be as extensively commented as I did now, or better reduce comments to a minimum and move the text to http://lasagne.readthedocs.org/en/latest/user/tutorial.html, interspersed with code snippets. I'm tending towards the latter, but I'm not sure if the example file still remains a good read then.


examples/mnist.py
```diff
@@ -1,240 +1,414 @@
...
+        print("       'custom_mlp:DEPTH,WIDTH,DROP_IN,DROP_HID' for an MLP")
+        print("       with DEPTH hidden layers of WIDTH units, DROP_IN")
+        print("       input dropout and DROP_HID hidden dropout,")
+        print("       'cnn' for a simple Convolutional Neural Network (CNN).")
+        print("EPOCHS: number of training epochs to perform (default: 500)")
+        return
+    if len(sys.argv) > 1:
+        model = sys.argv[1]
+    if len(sys.argv) > 2:
+        num_epochs = int(sys.argv[2])
```
dnouri: I guess let's move checking of `sys.argv` to the `if __name__ == '__main__'` block below.

f0k: Sounds good. Changed that (and included the change to `test_examples.py` I forgot to commit).

benanne:
Great work! I'll try to have an in-depth look at this ASAP.

f0k:
> I think I agree with your idea of moving most of the prose into a tutorial section. It also looks better if the example code can be 'glanced' :)

Did that now. It replicates a bit of what's in `docs/layers.rst`, but I think it's practical to explain right on spot how layers work. To distinguish the two, `docs/layers.rst` could be expanded a bit to include DAGs (i.e., branching by stacking multiple layers on top of the same source, and merging via one of the merge layers). That's not exactly critical for the first release, though.

Would be nice if somebody could at least superficially review my example and the tutorial that goes along with it.

moscow25:
As someone who recently (6 months back) got a running start from the MNIST
example, I both liked it as it were, and would have been even happier with
more inline commentary that I'm forced to see. It's not obvious at all, to
a beginner, what the different networks are than can be setup very quickly.
(Similarly, it wasn't clear to me that I don't need to load all my data
into theano.shared(), etc.)

Not sure it's possible to "glance" the code, in any case, and explaining
the options really does help provide a reference. The further that the
documentation goes from inline to out-of-line, the less likely it is to be
read, or maintained. I love the long explanations in "layers," "updates,"
etc.

Just my $0.02.

Great job, and thanks as always.

Best,
Nikolai

benanne:
I should have clarified that ASAP might not have been very soon at all... sorry! I was at ICML for the past week and I'm moving abroad this week so I can't help much for now. Anyone else up for it?

f0k:
> As someone who recently (6 months back) got a running start from the MNIST example, I both liked it as it were, and would have been even happier with more inline commentary that I'm forced to see.

Hmm, interesting! I thought it was really a bit hard to read. The way it is now the example code should be understandable to anybody familiar with neural networks and Theano, and the tutorial text gives more in-depth explanations and pointers where to learn about the preliminaries if needed.

> It's not obvious at all, to a beginner, what the different networks are than can be setup very quickly.

That's right, but teaching neural networks is clearly outside the scope of the example. We might add some neural network tutorials to [Lasagne/Recipes](https://github.com/Lasagne/Recipes), but this example targets people that know about neural networks and Theano and mostly want to learn about Lasagne.

> explaining the options really does help provide a reference

With moving most explanations from the code into the Sphinx docs, it's much easier to reference the Lasagne and Theano docs (e.g., the `InputLayer` or `DenseLayer` docstring, or what a Theano variable is).

> The further that the documentation goes from inline to out-of-line, the less likely it is to be read,

That's right, although the example provides a link to the tutorial, and the README file will probably also link to it. I could try to make the links more prominent.

> or maintained.

Yes, maintenance is clearly a pain in that form, because some of the code is now duplicated between the Python script and the ReST file. But the hope is that this will hardly need any changes, just like Theano's deep learning tutorials.

Thanks a lot for your input!

> Anyone else up for it?

Maybe @craffel, since he also wrote a tutorial at some point? (One that I really like, by the way, it would just need some updates to comply with API changes.) Do you want to have a quick look at the example and explanations? It doesn't need to be perfect for now, it just needs to be better than what we have.

docs/user/tutorial.rst
```diff
@@ -4,88 +4,611 @@
...
+``y_train.shape`` is simply ``(50000,)``, that is, it is a vector the same
+length of ``X_train`` giving an integer class label for each image -- namely,
+the digit between 0 and 9 depicted in the image (according to the human
+annotator who drew that digit).
+
+
+Building the model
+------------------
+
+This is were Lasagne steps in. It allows you to define an arbitrarily
```
ebenolson: s/were/where/

docs/user/tutorial.rst
```diff
@@ -4,88 +4,611 @@
...
+
+    def build_mlp(input_var=None):
+        l_in = lasagne.layers.InputLayer(shape=(None, 1, 28, 28),
+                                         input_var=input_var)
+
+The four numbers in the shape tuple represent, in order:
+``(batchsize, channels, rows, columns)``.
+Here we've set the batchsize to ``None``, which means the network will accept
+input data of arbitrary batchsize after compilation. If you know the batchsize
+beforehands and do not need this flexibility, you should give the batchsize
```
ebenolson: s/beforehands/beforehand/

docs/user/tutorial.rst
```diff
@@ -4,88 +4,611 @@
...
+The function begins like the others:
+
+.. code-block:: python
+
+    def build_cnn(input_var=None):
+        layer = lasagne.layers.InputLayer(shape=(None, 1, 28, 28),
+                                          input_var=input_var)
+
+We don't apply dropout to the inputs, as this tends to work less well for
+convolutional layers. Besides, instead of a :class:`DenseLayer
```
ebenolson: I'd remove 'Besides' here. Also, might be a little confusing using both 'kernels' and 'filters'.

docs/user/tutorial.rst
```diff
@@ -4,88 +4,611 @@
...
+Equipped with all the necessary Theano expressions, we're now ready to compile
+a function performing a training step:
+
+.. code-block:: python
+
+    train_fn = theano.function([input_var, target_var], loss, updates=updates)
+
+This tells Theano to generate and compile a function taking two inputs -- a
+mini-batch of images and a vector of corresponding targets -- and returning a
+single output: the training loss. Besides, each time it is invoked, it applies
```
ebenolson: I think think would read better with 'Additionally' instead of 'Besides'.

ebenolson:
:+1:  from me, for a basic example I think this is a big improvement. 

Aesthetically, my preference would be an IPython notebook, which could then be automatically converted to .py and [rst](http://sphinx-ipynb.readthedocs.org/en/latest/hello.html), but I think this arrangement is fine.

f0k:
Thanks for reviewing! I've addressed all your comments.

> Aesthetically, my preference would be an IPython notebook, which could then be automatically converted to .py and rst

Sounds good in theory, but that would give a little less flexibility -- now we can freely select and shorten the code excerpts to use in the tutorial, similar to Theano's deep learning tutorial. An IPython notebook might still be a good alternative to the Sphinx doc. It could do everything in a flat file, without any function definitions (except maybe for `iterate_minibatches()`), and just for the MLP. `mnist.py` would then be a somewhat more fleshed-out variant of what the notebook does, with three networks and command line support. I also like Colin's tutorial with its references to cooking. Countless possibilities!

I'll leave this open until tomorrow and then squash and merge it if there are no further comments.

craffel:
> Maybe @craffel, since he also wrote a tutorial at some point? (One that I really like, by the way, it would just need some updates to comply with API changes.)

It's up to date as of about two weeks ago, and I can update it again if needed.  I've given it as a tutorial 3 times and people seem to like it.  I have a version which is not online yet which includes recurrent layers too.

http://nbviewer.ipython.org/github/craffel/Lasagne-tutorial/blob/master/examples/tutorial.ipynb

f0k:
> It's up to date as of about two weeks ago, and I can update it again if needed.

But the version you linked to still uses `Layer.get_output()`, and the `Objective` class. You should try running it without `warnings.filterwarnings` once to see which interface changes need to be incorporated to be ready for the Lasagne release.

craffel:
> But the version you linked to still uses Layer.get_output(), and the Objective class. You should try running it without warnings.filterwarnings once to see which interface changes need to be incorporated to be ready for the Lasagne release.

Oh, sure, right.  I can/will change those.

Issue #323, opened 2015-07-04
What version of Theano should Lasagne require/support?


ebenolson:
I think Lasagne's `requirements.txt` should point to a git commit, updated with each PR to the latest version. This provides a fairly current Theano, while still making sure we don't pull in changes which might break Lasagne.

Arguments for this:
* 0.7 is several months old, with > 1000 commits since. 
* The last release gap was over 1 year, and I see no reason to expect that to change.
* While Theano's docs ostensibly refer to 0.7, they're actually compiled from master, and contain information on code that doesn't exist in 0.7. I think this is evidence that git is the definitive version.

benanne:
Note: this discussion came up in #320, but it's important enough to give it its own issue since previously there was an implicit agreement to standardize on version 0.7 once it was released.

I'll try to think of my own arguments and counter-arguments later :)

ebenolson:
Along with this, I think it should be a goal to keep Lasagne compatible with Theano master, and patch any problems that arise as soon as possible.

dnouri:
If we go with this, the docs will need fixing.  "pip install Lasagne" will have to be replaced with "pip install -r https://github.com/.../1.0/requirements.txt" and then "pip install Lasagne".  So I'd suggest making a separate pull request for this that also updates the docs.

There's a small trade off there: you can't tell users to "just use pip install" anymore.  But I think that's okay.

It may also trip up some people that use a Git checkout of Lasagne.  If the requirements.txt updates frequently as you do a git pull, I imagine it might happen that you forget to re-run the `pip install -r requirements.txt` to get the latest required Theano.  Maybe worthwhile mentioning that in the docs.

f0k:
Hmm... I'd generally encourage people to use the git version of Theano, but I don't necessarily think it's a good idea to make it a strict requirement. If it doesn't overly complicate matters, a good solution would be to fail gracefully for dependencies on features added since the latest release of Theano. This should make it easier to get people started, and allow people to upgrade Theano to the bleeding-edge version in case they need something that's not available in the latest release.

benanne:
I think I agree with @f0k here.

ebenolson:
> a good solution would be to fail gracefully for dependencies on features added since the latest release of Theano.

This would be messy. Code would be added to do version checks, then need to be removed once Theano releases. 

Also, for tiny feature additions like #320 it might not be terrible, but what about stuff like #310? You could no longer fix problems with Lasagne by getting changes introduced upstream.

[...]

PR #329, opened 2015-07-14
Require Git version of Theano.  For #320.  See #323.

dnouri:
For discussion, see #323.

`requirements.txt` has been changed to point to a recent Git version of Theano as per our discussion.  Install docs were updated, though necessary changes were minimal.  (Installation has become slightly more complicated though.)

I also did what @ebenolson suggested and removed `Theano` from dependencies in `setup.py`.  I've added a friendly error message for people who ignore the updated installation docs, run `pip install Lasagne` alone, and now get `ImportError theano`.  The point here is to avoid accidentally installing a too old version of Theano, which would then be hard to detect for us at runtime, and would lead to obscure errors as pointed out by @ebenolson.

### DIFF
--- a/docs/user/installation.rst
+++ b/docs/user/installation.rst
@@ -20,16 +20,13 @@ Install from PyPI
 .. note:: Lasagne hasn't been released as of now.  The only way to
           install it for now is from source.
 
-Simply run:
+To install release 0.1 of Lasagne from PyPI, run the following
+commands:
 
 .. code-block:: bash
 
-  pip install Lasagne
-
-It's also a good practice to install dependencies with exactly the
-same version numbers that the release was made with.  You can find the
-``requirements.txt`` that defines those version numbers in the top
-level directory of the Lasagne source tree.
+  pip install -r https://raw.githubusercontent.com/Lasagne/Lasagne/0.1/requirements.txt
+  pip install Lasagne==0.1
 
 Install from source
 ===================
@@ -42,10 +39,11 @@ copy is by cloning the git repository, for example:
   git clone https://github.com/Lasagne/Lasagne.git
 
 Once you have obtained a copy of the source, run the following command to
-install its dependencies:
+install Lasagne's dependencies:
 
 .. code-block:: bash
 
+  cd Lasagne
   pip install -r requirements.txt
 
 To install the Lasagne package itself, run:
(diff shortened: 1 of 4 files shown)


docs/user/installation.rst
```diff
@@ -20,16 +20,13 @@ Install from PyPI
...
 .. code-block:: bash
 
-  pip install Lasagne
-
-It's also a good practice to install dependencies with exactly the
-same version numbers that the release was made with.  You can find the
-``requirements.txt`` that defines those version numbers in the top
-level directory of the Lasagne source tree.
+  pip install -r https://raw.githubusercontent.com/Lasagne/Lasagne/1.0/requirements.txt
+  pip install Lasagne==1.0
```
f0k: Just to avoid possible confusion, I'd suggest to use `0.1` as the example and update it whenever we do a release. Otherwise I bet we'll soon have github issues / mails about `pip install Lasagne==1.0` not working :)

lasagne/__init__.py
```diff
@@ -2,6 +2,17 @@
...
 """
 
+try:
+    import theano
+except ImportError:  # pragma: no cover
+    raise ImportError("""Could not import Theano.
+
+Please make sure you install a recent enough version of Theano.  See
+section 'Install from PyPI' in the installation docs for more details.
+""")
```
f0k: Just for convenience, I'd replace the last `.` by a colon and add this URL: http://lasagne.readthedocs.org/en/latest/user/installation.html#install-from-pypi
This error message is targeted at users just doing `pip install lasagne`, and those may not know about the docs at all yet, so we can help them a little.

f0k:
Thanks, looks good to me! In the long term, we might want to expand the installation instructions to explain different options for Theano (`requirements.txt` or latest git version), and possibly also add some pointers about enabling GPU support. That's outside the scope of this PR, though.

dnouri:
Addressed your comments.  I forgot that first release is supposed to be 0.1, not 1.0.

f0k:
Great! Wanted to merge this, but just noticed that I got the reference wrong in https://github.com/Lasagne/Lasagne/issues/323#issuecomment-121201510 -- it's for #320, not #321. If you want to change the commit message and PR title, please go ahead, otherwise I'll just merge it as is.

dnouri:
Sorry, not sitting at my computer right now, so can't change the commit's title.  I'm fine if you want to merge as is.

PR #331, opened 2015-07-16
fix one_hot doc string note

mgrachten:
remove superfluous newline, fixed link to categorical_crossentropy function


f0k:
Thanks, LGTM, let's go for lunch!

PR #294, opened 2015-06-11
Add recurrent layers

craffel:
Addresses #17.  Includes layers for LSTM, GRU, "vanilla" (dense) RNN, and an RNN with arbitrary connection patterns.  Thanks to @skaae for the tests and the GRU layer (+more), and @JackKelly for some comment fixes and additional functionality.  Some misc comments that may warrant addressing -

* We added examples to the examples directory.  Do you want these?  The usage of the recurrent layers is different enough that an example is probably warranted, but there is also an example in the top docstring of the recurrent submodule, which covers the absolute basics in terms of how to reshape things to make it work.  Any examples should probably be supplemented with more fleshed-out stuff, like @skaae's penn treebank example and my speech rec example.
* We didn't resolve these two issues: https://github.com/craffel/nntools/issues/30 https://github.com/craffel/nntools/issues/15 which essentially are the same thing - i.e., the ability to use the RNNs for sequence prediction.  It's probably a feature we want to have, as it's a very hot topic at the moment.
* The LSTM layer (and the GRU layer, to some extent) have a LOT of `__init__` arguments because there are so many things to initialize.  Most of the time people initialize all weight matrices in the same way, although I'm sure that's not always true.  It's actually quite common to initialize the forget gate bias vector differently than the rest.
* The `peepholes` kwarg of the LSTM layer decides whether to include connections from the cell to the gates; sometimes people use these, sometimes they don't, so it's important that we include this functionality.  However, people also sometimes don't include other connections, and we don't have functionality for this.  What I'd really like is something where if one of the arguments is passed as `None`, then the connection is not used, but this would make the code pretty ugly - we'd essentially have to define a new dot method which does nothing when one of the arguments is `None` or something. So, in short, I'd rather not address this now.
* The `grad_clipping` argument in all layers clips the gradients, which is very common for RNNs.  I think, however, this is pretty common for other network architectures too, so what I'd really like to see is it be moved out to a global function in updates, or something.  But, this isn't something I'm familiar with enough to know how feasible that is, @skaae would be better to comment on that.  
* `hid_init`  (and `cell_init`) are allowed to be `TensorVariable`s, see our discussion here.  https://github.com/craffel/nntools/pull/27#issuecomment-107989355  To summarize that, I'd rather have them be able to be variable-size shared variables which are added with `add_param`, but we can't because of the way we allow for variable batch size.  Any ideas on that discussion are appreciated.

Ok, that's about it, let me know what you think!  Would be great to get this merged before the first release.

PR #335, opened 2015-07-20
Split layer docs

f0k:
This splits up the `lasagne.layers` Sphinx documentation across multiple pages; one per submodule (but taking care to present all classes as belonging to the `lasagne.layers` namespace if appropriate). It requires Sphinx 1.2.3 to build because of a regression in more recent versions (see https://github.com/Lasagne/Lasagne/issues/203#issuecomment-121310595), but we use that one on readthedocs anyway (just note that a more recent version will omit the autosummary tables, so make sure to use Sphinx version 1.2.3 or the latest git version if building locally).

I've tried a few different options for the TOC and layout, and this is the best I found. It has a bunch of autosummary tables for `lasagne.layers`, and the TOC is set up such that clicking `Convolutional layers` in the left-hand navbar takes you to the detail page for convolutional layers, not to the autosummary table within the `lasagne.layers` overview page (mainly because then we'd either have a second navbar entry for the detail page or have Sphinx whine about the detail page not being included in any TOC).

Finally, I did some minor docstring fixes that triggered Sphinx warnings or caught my eye in the overview page.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,7 +13,7 @@
 # serve to show the default.
 
 # import sys
-# import os
+import os
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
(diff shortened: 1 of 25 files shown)


benanne:
lgtm, I don't have time at the moment to review this in detail though. If you want to be on the safe side better have someone else check as well. Otherwise I'm okay to merge.

f0k:
All right, if anybody has the time to take a look, the following should do the trick:
```bash
git checkout -b f0k-split-layer-docs master
git pull https://github.com/f0k/Lasagne.git split-layer-docs
cd docs
make clean html
firefox _build/html/index.html &
```
Make sure to do `pip install Sphinx=1.2.3 sphinx_rtd_theme` if you have a different sphinx version.

f0k:
I've amended the first commit to include `sphinx_rtd_theme` in `requirements-dev.txt`. I hope readthedocs won't bark on this. I'll still wait with merging until #294 is merged, so the latter doesn't need to be rebased again. I'll rebase this one afterwards.

PR #336, opened 2015-07-21
Fix PEP-8 issues in mnist example

craffel:
The changes in https://github.com/Lasagne/Lasagne/pull/328 introduced some PEP-8 issues, this commit fixes them.

### DIFF
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -25,8 +25,7 @@ import theano.tensor as T
 import lasagne
 
 
-
-################### Download and prepare the MNIST dataset ###################
+# ################## Download and prepare the MNIST dataset ##################
 # This is just some way of getting the MNIST dataset from an online location
 # and loading it into numpy arrays. It doesn't involve Lasagne at all.
 
@@ -35,11 +34,13 @@ def load_dataset():
     if sys.version_info[0] == 2:
         from urllib import urlretrieve
         import cPickle as pickle
+
         def pickle_load(f, encoding):
             return pickle.load(f)
     else:
         from urllib.request import urlretrieve
         import pickle
+
         def pickle_load(f, encoding):
             return pickle.load(f, encoding=encoding)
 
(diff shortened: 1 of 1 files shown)


examples/mnist.py
```diff
@@ -100,25 +100,25 @@ def build_mlp(input_var=None):
     # Add a fully-connected layer of 800 units, using the linear rectifier, and
     # initializing weights with Glorot's scheme (which is the default anyway):
     l_hid1 = lasagne.layers.DenseLayer(
-            l_in_drop, num_units=800,
-            nonlinearity=lasagne.nonlinearities.rectify,
-            W=lasagne.init.GlorotUniform())
+        l_in_drop, num_units=800,
+        nonlinearity=lasagne.nonlinearities.rectify,
+        W=lasagne.init.GlorotUniform())
```
f0k: I'm fine with everything but these hanging indentation changes, they were more readable before. After a long discussion of the wording of hanging indentation rules in PEP8, the pep8 project decided to not report E126 by default from version 1.6.0 on. We then included the pep8 version to use in `requirements-dev.txt`: https://github.com/Lasagne/Lasagne/pull/289

So if you upgrade your pep8, you should be able to undo these changes! Please amend the commit instead of adding a new one, thanks!

f0k:
Sorry, didn't pep8 the file and it's not included in our continuous integration as well, thanks for checking! I'm fine with merging if you can address my comment above (i.e., only fix the missing empty lines and the missing blanks after the first `#` per comment line).

craffel:
Huh, interesting!  I disagree about double-indenting being clearer for hanging indent, but I guess it's too late to voice my opinion on that, and it no one will every agree completely on style conventions anyways.  I've reverted those changes.

f0k:
> I disagree about double-indenting being clearer for hanging indent, but I guess it's too late to voice my opinion on that

It's never too late, but I guess you're not enough people :) I think at least three of us were arguing that a single indent can easily be confused with starting the body of something. But yes, we'll never get everybody agree on everything. Thanks for reverting! I'll wait for Travis and merge.

PR #341, opened 2015-07-22
Fleshed out installation instructions

### LINKED ISSUE #316: cuDNN installation / configuration
I've recently seen that a lot of tests get skipped for me (see https://gist.github.com/MartinThoma/68e5e78101a3fbaed334). The / one reason is that I don't have CuDNN:

```bash
$ python mnist_conv_dnn.py
Using gpu device 0: GeForce GTX TITAN Black
  File "mnist_conv_dnn.py", line 4, in <module>
    from lasagne.layers import dnn  # fails early if not available
  File "/usr/local/lib/python2.7/dist-packages/Lasagne-0.1.dev-py2.7.egg/lasagne/layers/dnn.py", line 12, in <module>
    raise ImportError("dnn not available")  # pragma: no cover
ImportError: dnn not available
```

It should be made clear in the docs (e.g. on http://lasagne.readthedocs.org/en/latest/user/development.html) that you need to install it manually. Eventually we could also add some help how to install / configure it. Or simply link to http://deeplearning.net/software/theano/library/sandbox/cuda/dnn.html

### LINKED ISSUE #339: max_pool_2d() got an unexpected keyword argument 'mode'
getting this error while trying to run mnist.py cnn

f0k:
This adds a lot of information to the installation instructions about how to install requirements (and where to find more information), what the `requirements.txt` and Theano version is about, how to install Lasagne from PyPI, github, source or without pip, and how to get GPU support working (with pointers to where to find more information), including cuDNN. That's a lot, but I think it's good to have everything in a central place. It's probably tough to find all this when you're a Theano newbie, and the Theano installation instructions are not terribly clean / up-to-date.

Resolves #339, resolves #316. (This should make [github close these issues automatically](https://github.com/blog/1506-closing-issues-via-pull-requests) on merging.)

### DIFF
--- a/docs/user/installation.rst
+++ b/docs/user/installation.rst
@@ -4,50 +4,223 @@
 Installation
 ============
 
-Lasagne currently requires Python 2.7 or 3.4 to run. Some of Lasagne's
-dependencies such as ``numpy`` may require a C compiler to install.
+Lasagne has a couple of prerequisites that need to be installed first, but it
+is not very picky about versions. The single exception is Theano: Due to its
+tight coupling to Theano, you will have to install a recent version of Theano
+(usually more recent than the latest official release!) fitting the version of
+Lasagne you choose to install.
 
-For most installations, it is recommended to install Lasagne and its
-dependencies inside a virtualenv or a conda environment. The
-following commands assume that you have your environment active.
+Most of the instructions below assume you are running a Linux or Mac system;
+please do not hesitate to suggest instructions for Windows via the *Edit on
+GitHub* link on the top right!
 
-You can either install from PyPI or install from source, such as from
-a Git clone.
+If you run into any trouble, please check the `Theano installation instructions
+<http://deeplearning.net/software/theano/install.html>`_ which cover installing
+the prerequisites for a range of operating systems, or ask for help on `our
+mailing list <https://groups.google.com/d/forum/lasagne-users>`_.
 
-Install from PyPI
-=================
 
-.. note:: Lasagne hasn't been released as of now.  The only way to
-          install it for now is from source.
+Prerequisites
+=============
 
-To install release 0.1 of Lasagne from PyPI, run the following
-commands:
+Python + pip
+------------
+
+Lasagne currently requires Python 2.7 or 3.4 to run. Please install Python via
+the package manager of your operating system if it is not included already.
+
+Python includes ``pip`` for installing additional modules that are not shipped
+with your operating system, or shipped in an old version, and we will make use
+of it below. We recommend installing these modules into your home directory
+via ``--user``, or into a `virtual environment
+<http://www.dabapps.com/blog/introduction-to-pip-and-virtualenv-python/>`_
+via ``virtualenv``.
+
+C compiler
+----------
+
+Theano requires a working C compiler, and numpy/scipy require a compiler as
+well if you install them via ``pip``. On Linux, the default compiler is usually
+``gcc``, and on Mac OS, it's ``clang``. Again, please install them via the
+package manager of your operating system.
+
+numpy/scipy + BLAS
+------------------
+
+Lasagne requires numpy of version 1.6.2 or above, and Theano also requires
+scipy 0.11 or above. Numpy/scipy rely on a BLAS library to provide fast linear
+algebra routines. They will work fine without one, but a lot slower, so it is
+worth getting this right (but this is less important if you plan to use a GPU).
+
+If you install numpy and scipy via your operating system's package manager,
+they should link to the BLAS library installed in your system. If you install
+numpy and scipy via ``pip install numpy`` and ``pip install scipy``, make sure
+to have development headers for your BLAS library installed (e.g., the
+``libopenblas-dev`` package on Debian/Ubuntu) while running the installation
+command. Please refer to the `numpy/scipy build instructions
+<http://www.scipy.org/scipylib/building/index.html>`_ if in doubt.
+
+Theano
+------
+
+The version to install depends on the Lasagne version you choose, so this will
+be handled below.
+
+
+Stable Lasagne release
+======================
+
+.. note:: Lasagne hasn't been released yet. Please see the next section for
+    how to install the bleeding-edge version instead.
+
+Lasagne 0.1 requires a more recent version of Theano than the one available
+on PyPI. To install a version that is known to work, run the following command:
 
 .. code-block:: bash
 
   pip install -r https://raw.githubusercontent.com/Lasagne/Lasagne/0.1/requirements.txt
+
+.. warning::
+  An even more recent version of Theano will often work as well, but at the
+  time of writing, a simple ``pip install Theano`` will give you a version that
+  is too old.
+
+To install release 0.1 of Lasagne from PyPI, run the following command:
+
+.. code-block:: bash
+
   pip install Lasagne==0.1
 
-Install from source
-===================
+If you do not use ``virtualenv``, add ``--user`` to both commands to install
+into your home directory instead. To upgrade from an earlier installation, add
+``--upgrade``.
 
-Download and navigate to your copy of the Lasagne source. One way to obtain a
-copy is by cloning the git repository, for example:
+
+Bleeding-edge version
+=====================
+
+The latest development version of Lasagne usually works fine with the latest
+development version of Theano. To install both, run the following commands:
+
+.. code-block:: bash
+
+  pip install --upgrade https://github.com/Theano/Theano/archive/master.zip
+  pip install --upgrade https://github.com/Lasagne/Lasagne/archive/master.zip
+
+Again, add ``--user`` if you want to install to your home directory instead.
+
+
+Development installation
+========================
+
+Alternatively, you can install Lasagne (and optionally Theano) from source,
+in a way that any changes to your local copy of the source tree take effect
+without requiring a reinstall. This is often referred to as *editable* or
+*development* mode. Firstly, you will need to obtain a copy of the source tree:
 
 .. code-block:: bash
 
   git clone https://github.com/Lasagne/Lasagne.git
 
-Once you have obtained a copy of the source, run the following command to
-install Lasagne's dependencies:
+It will be cloned to a subdirectory called ``Lasagne``. Make sure to place it
+in some permanent location, as for an *editable* installation, Python will
+import the module directly from this directory and not copy over the files.
+Enter the directory and install the known good version of Theano:
 
 .. code-block:: bash
 
   cd Lasagne
   pip install -r requirements.txt
 
-To install the Lasagne package itself, run:
+Alternatively, install the bleeding-edge version of Theano as described in the
+previous section.
+
+To install the Lasagne package itself, in editable mode, run:
 
 .. code-block:: bash
 
-  python setup.py install  # or 'setup.py develop' if you're developing Lasagne itself
+  pip install --editable .
+
+As always, add ``--user`` to install it to your home directory instead.
+
+**Optional**: If you plan to contribute to Lasagne, you will need to fork the
+Lasagne repository on GitHub. This will create a repository under your user
+account. Update your local clone to refer to the official repository as
+``upstream``, and your personal fork as ``origin``:
+
+.. code-block:: bash
+
+  git remote rename origin upstream
+  git remote add origin https://github.com/<your-github-name>/Lasagne.git
+
+If you set up an `SSH key <https://help.github.com/categories/ssh/>`_, use the
+SSH clone URL instead: ``git@github.com:<your-github-name>/Lasagne.git``.
+
+You can now use this installation to develop features and send us pull requests
+on GitHub, see :doc:`development`!
+
+
+GPU support
+===========
+
+Thanks to Theano, Lasagne transparently supports training your networks on a
+GPU, which may be 10 to 50 times faster than training them on a CPU. Currently,
+this requires an NVIDIA GPU with CUDA support, and some additional software for
+Theano to use it.
+
+CUDA
+----
+
+Install the latest CUDA Toolkit and possibly the corresponding driver available
+from NVIDIA: https://developer.nvidia.com/cuda-downloads
+
+Closely follow the *Getting Started Guide* linked underneath the download table
+to be sure you don't mess up your system by installing conflicting drivers.
+
+After installation, make sure ``/usr/local/cuda/bin`` is in your ``PATH``, so
+``nvcc --version`` works. Also make sure ``/usr/local/cuda/lib64`` is in your
+``LD_LIBRARY_PATH``, so the toolkit libraries can be found.
+
+Theano
+------
+
+If CUDA is set up correctly, the following should print some information on
+your GPU (the first CUDA-capable GPU in your system if you have multiple ones):
+
+.. code-block:: bash
+
+  THEANO_FLAGS=device=gpu python -c "import theano; print theano.sandbox.cuda.device_properties(0)"
+
+To configure Theano to use the GPU by default, create a file ``.theanorc``
+directly in your home directory, with the following contents:
+
+.. code-block:: none
+
+  [global]
+  floatX = float32
+  device = gpu
+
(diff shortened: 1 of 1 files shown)


docs/user/installation.rst
```diff
@@ -4,50 +4,239 @@
...
+Prerequisites
+=============
 
-To install release 0.1 of Lasagne from PyPI, run the following
-commands:
+Python
+------
+
+Lasagne currently requires Python 2.7 or 3.4 to run. Please install Python via
+the package manager of your operating system if they are not included already.
```
benanne: not sure what 'they' refers to here, should this perhaps be 'it' (referring to Python) instead?

docs/user/installation.rst
```diff
@@ -4,50 +4,239 @@
...
+Use Python 2.7 if you're in doubt; it is still more widespread than Python 3.
```
benanne: I'm not sure if we should be opinionated about this... there is definitely a trend even in the scientific Python community to switch to Python 3 right now. Theano also has better support for it (or at least they are working on it) and I think most of the scientific Python ecosystem now runs flawlessly on Python 3. That said, I admit I haven't made the switch myself :)

docs/user/installation.rst
```diff
@@ -4,50 +4,239 @@
...
+
+C compiler
+----------
+
+Theano requires a working C compiler, and numpy/scipy require a compiler as
+well if you install them via ``pip``. On Linux, the default compiler is usually
+``gcc``, and on Mac OS, it's ``clang``. Again, please install them via the
+package manager of your operating system.
+
```
benanne: If we want to go for maximal user friendliness maybe we could have some Mac / Ubuntu instructions here as well (that combo should cover the largest part of our user base, probably). Just an idea, please don't feel obliged to add this in this PR. It's already quite comprehensive :)

docs/user/installation.rst
```diff
@@ -4,50 +4,239 @@
...
+numpy/scipy + BLAS
+------------------
+
+Lasagne requires numpy of version 1.6.2 or above, and Theano also requires
+scipy 0.11 or above. Numpy/scipy rely on a BLAS library to provide fast linear
+algebra routines. They will work fine without one, but a lot slower, so it is
+worth getting this right (but less importantly so if you plan to use a GPU).
```
benanne: 'but this is less important if you plan to use a GPU' reads a bit more smoothly imo

docs/user/installation.rst
```diff
@@ -4,50 +4,239 @@
...
+And then install via ``setup.py``:
 
 .. code-block:: bash
 
-  python setup.py install  # or 'setup.py develop' if you're developing Lasagne itself
+  cd Theano && python setup.py install && cd -
+  cd Lasagne && python setup.py install && cd -
+
+Append ``--prefix=~/.local`` after ``setup.py`` to install to your home
+directory, and replace ``install`` with ``develop`` to install in editable
```
benanne: I would reorder the parts of this sentence a bit, starting with "To install in editable mode, append...". Otherwise it's a bit unclear that these instructions are specific to that use case.

docs/user/installation.rst
```diff
@@ -4,50 +4,239 @@
...
+mode. Note that this will install the bleeding-edge versions of Theano and
+Lasagne.
+
+
+GPU support
+===========
+
+Thanks to Theano, Lasagne transparently supports training your networks on a
+GPU, which may be 10 to 50 times faster than training it on a CPU. Currently,
```
benanne: than training *them* on a CPU.

docs/user/installation.rst
```diff
@@ -4,50 +4,239 @@
...
+
+If CUDA is set up correctly, the following should print some information on
+your GPU (the first CUDA-capable GPU in your system if you have multiple ones):
+
+.. code-block:: bash
+
+  THEANO_FLAGS=device=gpu python -c "import theano; print theano.sandbox.cuda.device_properties(0)"
+
+To configure Theano to use the GPU by default, create a file ``.theanorc``
+directly in your home directory, with the following contents:
```
benanne: Windows users might be wondering what to do here. But there's a link to the Theano version of these instructions a few lines down, so I guess it's okay :)

docs/user/installation.rst
```diff
@@ -4,50 +4,239 @@
...
+
+.. code-block:: none
+
+  [global]
+  floatX = float32
+  device = gpu
+
+Optionally add ``allow_gc = False`` for some extra performance at the expense
+of some more GPU memory.
```
benanne: maybe "at the expensive of higher GPU memory usage". The way it's phrased right now makes it sound like there might be a modest increase, but it can actually be quite substantial :)

benanne:
This is excellent! Very nice work. Just a few remarks from me, mostly style and grammar.

docs/user/installation.rst
dnouri: I agree: Python 3 is largely ready.  It's unnecessary to recommend against it.  I use Python 3 for new projects and all the packages I care about have been ported already, if not years ago.

docs/user/installation.rst
```diff
@@ -4,50 +4,239 @@
...
+To install release 0.1 of Lasagne from PyPI, run the following command:
+
+.. code-block:: bash
+
   pip install Lasagne==0.1
 
-Install from source
-===================
+If you do not have administrative rights on your machine, add ``--user`` to
+both commands to install into your home directory instead. To upgrade from an
```
dnouri: I think we should recommend virtualenv instead of suggesting to install stuff site-wide or with ``--user``.  In my experience, installing stuff side-wide quickly becomes a mess with conflicts with your operating system's packages.  Another case is when you want to use two different versions of a package, say numpy, for two different projects that you develop.  It's when ``-user`` or ``sudo pip install`` quickly become a dead end.

docs/user/installation.rst
```diff
@@ -4,50 +4,239 @@
...
+  git remote add origin https://github.com/<your-github-name>/Lasagne.git
+
+If you set up an `SSH key <https://help.github.com/categories/ssh/>`_, use the
+SSH clone URL instead: ``git@github.com:<your-github-name>/Lasagne.git``.
+
+You can now use this installation to develop features and send us pull requests
+on GitHub, see :doc:`development`!
+
+  
+Installation without pip
```
dnouri: I think it's safe to drop the "Installation without pip" section altogether.  Or we call it "Lasange Windows XP edition".  I don't think there are such users.  :-)

(There may be Conda users.)

docs/user/installation.rst
f0k: Ah yes! I had something in plural first and overlooked this.

docs/user/installation.rst
f0k: All right, I'll remove it! I just remember someone somewhere was completely new to Python, just installed the one with the largest version number (which is the most reasonable thing to do usually) and was wondering why half of the things didn't work. But that was some time ago, so maybe that advice is unneeded now.

docs/user/installation.rst
f0k: I could add detailed instructions for Ubuntu, from zero to Python/gcc/pip/openBLAS/numpy/scipy/CUDA/cuDNN, but I think it would be a bit misplaced here (why detailed instructions for Ubuntu, but not for RedHat/CentOS/whatever, Max OS and Windows?). That might be something to put on an extra page, or the Wiki, so others can contribute similar guides for installing the prerequisites on other operating systems.

docs/user/installation.rst
f0k: Ah yes, "importantly" modifies "worth" now, doesn't it? I meant to modify "getting this right".

docs/user/installation.rst
f0k: Plus I warned them in the beginning that instructions are often specific to *nix systems :) I'd be fine with changing this, I just don't know where the file is to be placed on Windows.

docs/user/installation.rst
f0k: I mostly kept it because the previous "Install from source" section wasn't using `pip`, and I thought there was a reason for that. I don't know who wrote that, though.

docs/user/installation.rst
f0k: There was a very strong recommendation of using virtualenv or conda environments before. I removed it because I think it would make newbies uncomfortable to be told to better use something they've never heard of without explaining what it is, how to do it and why, and experienced users would think of using virtualenv on their own.
I'd welcome a more cautious hint in the beginning along the lines of "If you're developing different projects that may require different versions of numpy, scipy, or Theano, you might want to look into using virtualenv(link to a good guide on that) or conda environments(link to a good guide on that) instead of installing packages site-wide." I'm not using virtualenv myself, so I don't know what's a good formulation and what's a good introductory guide. Do you have any suggestions? Or do you want to send a PR after this one is merged?

docs/user/installation.rst
dnouri: I wrote the previous "Install from source".  Let's kick this section out then.

The remark about Conda users is only to say that there's other installation methods that are more common than the obscure "help I don't have pip".

docs/user/installation.rst
dnouri: I think a more correct cautious hint would be: "if you want to save yourself from dependency hell and fucking up your computer, do consider using virtualenv", with a link to some virtualenv resource.  I'm not familiar myself as to what the best virtualenv guide is but this one seems reasonable: http://www.dabapps.com/blog/introduction-to-pip-and-virtualenv-python/

I would personally drop mentioning ``--user`` or administrative rights.  Installing stuff site-global is harmful and shouldn't be recommended IMO.

ebenolson: I think teaching different package installation methods is a bit out of scope. Ditto for intricacies of installing BLAS/numpy/scipy. I would just leave that stuff out. cuDNN is more appropriate since it's pretty specific to Lasagne.

docs/user/installation.rst
```diff
@@ -4,50 +4,239 @@
...
+
+Lasagne 0.1 requires a more recent version of Theano than the one available
+on PyPI. To install a version that is known to work, run the following command:
 
 .. code-block:: bash
 
   pip install -r https://raw.githubusercontent.com/Lasagne/Lasagne/0.1/requirements.txt
+
+A more recent version will often work as well, but at the time of writing, a
+simple ``pip install Theano`` will give you a version that is too old.
```
ebenolson: I think this is the most crucial information in this section, it would be nice if it was highlighted a little more. I really like the [colored "! Warning" and "! Note" boxes here](http://blocks.readthedocs.org/en/latest/setup.html), does our theme have anything like that?

docs/user/installation.rst
f0k: > I think a more correct cautious hint would be

That's a bit strong again for my taste :)

For me, everything's fine without virtualenv, mostly because I've never had a package complain about a dependency being too *new*. So for me, it works well to install the pip version of a module if the one from the Ubuntu repository is too old, and then it automatically takes precedence over the repository version so I don't have to uninstall the latter (possibly leaving Ubuntu package dependencies unresolved).
For some modules like numpy and scipy, I do site-wise installs because on our GPU machines, all users (yes, all users, I've asked them) want the latest version, can't install it with `--user` (since home directories are shared between multiple machines that sometimes have different BLAS libraries to link against) and would be very annoyed if they had to learn virtualenv and create a different environment per server.
So for me, the only use case for virtualenv is for checking whether a particular piece of software I wrote can be deployed on machines running an older version of Python, numpy or scipy. But this is not one of my day-to-day activities.

Probably you had a different experience, but I'd still try not to scare people too much ("dependency hell" and "fucking up your computer" sounds scary, even if worded differently), just give them the option of using virtualenv and let them decide. It should be as easy as possible to get from zero to Lasagne.

> http://www.dabapps.com/blog/introduction-to-pip-and-virtualenv-python/

Thanks, looks very good at first glance! I think I'll put this in the beginning.

> prefix the `--user` trick with "should you for some reason not be using virtualenv" or some such.

Good idea! I guess I'd still include something about administrative rights, but along with a warning that site-wide installs are seldom a good idea (this is something I can agree on, it doesn't make life any harder for newbies to recommend against site-wide installs).

f0k: > I think teaching different package installation methods is a bit out of scope. Ditto for intricacies of installing BLAS/numpy/scipy. I would just leave that stuff out.

It's just that I assume there'll be several users who install Python, numpy, scipy and Theano just for the purpose of trying Lasagne, and there are several pitfalls that you'll never learn about if you just follow the basic installation instructions (install `requirements.txt`, then install Lasagne). Everything will seem to work fine without installing one of the more optimized BLAS libraries, it will just be slow.

docs/user/installation.rst
f0k: Yes, we're using exactly the same theme :)

ebenolson: Ah, I guess I should probably compile the docs before commenting :)

docs/user/installation.rst
```diff
@@ -4,50 +4,239 @@
...
+cuDNN
+-----
+
+NVIDIA provides a library for common neural network operations that especially
+speeds up Convolutional Neural Networks (CNNs). Again, it can be obtained from
+NVIDIA (after registering as a developer): https://developer.nvidia.com/cudnn
+
+To install it, copy the ``*.h`` files to ``/usr/local/cuda/include`` and the
+``*.so*`` files to ``/usr/local/cuda/lib64``.
+
```
ebenolson: I'm not sure if this matters, but I usually install cuDNN with `cp lib*` so that I get `libcudnn_static.a` as well as the `.so`

f0k: Doesn't matter for Theano, but it's a good idea anyway. I'll change that.

docs/user/installation.rst
dnouri: > For me, everything's fine without virtualenv, mostly because I've 
> never had a package complain about a dependency being too new. So for 
> me, it works well to install the pip version of a module if the one 
> from the Ubuntu repository is too old, and then it automatically 
> takes precedence over the repository version so I don't have to 
> uninstall the latter (possibly leaving Ubuntu package dependencies 
> unresolved).
> 
OK, fine.  If we can get rid of suggesting things should be installed 
with "sudo" (aka administrative rights), then we should be good.  I do 
sincerely think that it can fuck up your computer if you're sudo pip 
installing packages that your operating system depends on; Linux 
distros tend to use a lot of Python packages for core functionality 
these days.  (I realize that Theano and numpy probably don't qualify, 
but people will use this experience to install more common dependencies 
as well.)

As a side note: if you use virtualenv with the `--system-site-packages` 
flag, you can also go on using your Ubuntu's numpy.
> It should be as easy as possible to get from zero to Lasagne.
> 
... without screwing up your OS.  ;-)

f0k: Fully agreed. Hooray, consensus! :+1:

I'll update the text in the afternoon. It's a bit distressing that [coveralls.io has problems](http://status.coveralls.io/) and all pull requests are orange...

docs/user/installation.rst
```diff
@@ -4,50 +4,223 @@
...
+Lasagne currently requires Python 2.7 or 3.4 to run. Please install Python via
+the package manager of your operating system if it is not included already.
+
+Python includes ``pip`` for installing additional modules that are not shipped
+with your operating system, or shipped in an old version, and we will make use
+of it below. We recommend installing these modules into your home directory
+via ``--user``, or into a `virtual environment
+<http://www.dabapps.com/blog/introduction-to-pip-and-virtualenv-python/>`_
+via ``virtualenv``.
```
f0k: @dnouri, what do you think? I've also changed "If you don't have administrative rights" to "If you don't use `virtualenv`" further below.

dnouri: Looks good to me!

docs/user/installation.rst
benanne: Sure, that works :)

f0k:
Squashed commits. Will merge when Travis passes. (We don't actually need it, it's just touching the Sphinx docs, but a green checkmark looks better.)

benanne:
green check marks, merging :)

PR #342, opened 2015-07-22
Fleshed out README and design goals

### LINKED ISSUE #86: Example of Using L2 Regularizer
Currently none of the examples show using the L2 regularizer.

An example of using it in a model would be helpful.

f0k:
This fleshes out the README file to show an example upfront so you know what to expect, and to point to installation instructions, documentation and contribution guidelines (although the latter should probably be fleshed out in the documentation).

See https://github.com/f0k/Lasagne/tree/readme for how it would look.

Closes #86 in passing.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -14,23 +14,122 @@ Lasagne
 =======
 
 Lasagne is a lightweight library to build and train neural networks in Theano.
+Its main features are:
 
-For support, please refer to the lasagne-users mailing list: https://groups.google.com/forum/#!forum/lasagne-users
+* Supports feed-forward networks such as Convolutional Neural Networks (CNNs),
+  recurrent networks including Long Short-Term Memory (LSTM), and any
+  combination thereof
+* Allows architectures of multiple inputs and multiple outputs, including
+  auxiliary classifiers
+* Many optimization methods including Nesterov momentum, RMSprop and ADAM
+* Freely definable cost function and no need to derive gradients due to
+  Theano's symbolic differentiation
+* Transparent support of CPUs and GPUs due to Theano's expression compiler
 
-Documentation: http://lasagne.readthedocs.org/ (work in progress)
+Its design is governed by `six principles
+<http://lasagne.readthedocs.org/en/latest/user/development.html#philosophy>`_:
+
+* **Simplicity**: Be easy to use and easy to extend, with as few abstractions
+  as possible, to facilitate research
+* **Small interfaces**: Have as few classes and methods as possible
+* **Don't get in the way**: Do not obstruct users with features they decide not
+  to use, and allow all parts to be used independently
+* **Transparency**: Do not hide Theano behind abstractions, directly process and
+  return Theano expressions or Python / numpy data types
+* **Focus**: "Do one thing and do it well"
+* **Pragmatism**: Make common use cases easy, do not overrate uncommon cases
+
+
+Installation
+------------
+
+In short, you can install a known compatible version of Theano and the latest
+Lasagne development version via:
+
+.. code-block:: bash
+
+  pip install -r https://raw.githubusercontent.com/Lasagne/Lasagne/master/requirements.txt
+  pip install https://github.com/Lasagne/Lasagne/archive/master.zip
+
+For more details and alternatives, please see the `Installation instructions
+<http://lasagne.readthedocs.org/en/latest/user/installation.html>`_.
+
+
+Documentation
+-------------
+
+Documentation is available online: http://lasagne.readthedocs.org/
+
+For support, please refer to the `lasagne-users mailing list
+<https://groups.google.com/forum/#!forum/lasagne-users>`_.
+
+
+Example
+-------
+
+.. code-block:: python
+
+  import lasagne
+  import theano
+  import theano.tensor as T
+
+  # create Theano variables for input and target minibatch
+  input_var = T.tensor4('X')
+  target_var = T.ivector('y')
+
+  # create a small convolutional neural network
+  from lasagne.nonlinearities import leaky_rectify, softmax
+  network = lasagne.layers.InputLayer((None, 3, 32, 32), input_var)
+  network = lasagne.layers.Conv2DLayer(network, 64, (3, 3),
+                                       nonlinearity=leaky_rectify)
+  network = lasagne.layers.Conv2DLayer(network, 32, (3, 3),
+                                       nonlinearity=leaky_rectify)
+  network = lasagne.layers.Pool2DLayer(network, (3, 3), stride=2, mode='max')
+  network = lasagne.layers.DenseLayer(lasagne.layers.dropout(network, 0.5),
+                                      128, nonlinearity=leaky_rectify,
+                                      W=lasagne.init.Orthogonal())
+  network = lasagne.layers.DenseLayer(lasagne.layers.dropout(network, 0.5),
+                                      10, nonlinearity=softmax)
+
+  # create loss function
+  prediction = lasagne.layers.get_output(network)
+  loss = lasagne.objectives.categorical_crossentropy(prediction, target_var)
+  loss = loss.mean() + 1e-4 * lasagne.regularization.regularize_network_params(
+          network, lasagne.regularization.l2)
+
+  # create parameter update expressions
+  params = lasagne.layers.get_all_params(network, trainable=True)
+  updates = lasagne.updates.nesterov_momentum(loss, params, learning_rate=0.01,
+                                              momentum=0.9)
+
+  # compile training function that updates parameters and returns training loss
+  train_fn = theano.function([input_var, target_var], loss, updates=updates)
+
+  # train network (assuming you've got some training data in numpy arrays)
+  for epoch in range(100):
+      loss = 0
+      for input_batch, target_batch in training_data:
+          loss += train_fn(input_batch, target_batch)
+      print("Epoch %d: Loss %g" % (epoch + 1, loss / len(training_data)))
+
+  # use trained network for predictions
+  test_prediction = lasagne.layers.get_output(network, deterministic=True)
+  predict_fn = theano.function([input_var], T.argmax(test_prediction, axis=1))
+  print("Predicted class for first test input: %r" % predict_fn(test_data[0]))
+
+For a fully-functional example, see `examples/mnist.py <examples/mnist.py>`_,
+and check the `Tutorial
+<http://lasagne.readthedocs.org/en/latest/user/tutorial.html>`_ for in-depth
+explanations of the same. More examples, code snippets and reproductions of
+recent research papers are maintained in the separate `Lasagne Recipes
+<https://github.com/Lasagne/Recipes>`_ repository.
+
+
+Development
+-----------
 
 Lasagne is a work in progress, input is welcome.
 
-Design goals:
-
-* Simplicity: it should be easy to use and extend the library. Whenever a feature is added, the effect on both of these should be considered. Every added abstraction should be carefully scrutinized, to determine whether the added complexity is justified.
-
-* Small interfaces: as few classes and methods as possible. Try to rely on Theano's functionality and data types where possible, and follow Theano's conventions. Don't wrap things in classes if it is not strictly necessary. This should make it easier to both use the library and extend it (less cognitive overhead).
-
-* Don't get in the way: unused features should be invisible, the user should not have to take into account a feature that they do not use. It should be possible to use each component of the library in isolation from the others.
-
-* Transparency: don't try to hide Theano behind abstractions. Functions and methods should return Theano expressions and standard Python / numpy data types where possible.
-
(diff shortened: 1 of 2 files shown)


ebenolson:
Looks good, but I think the installation and documentation sections should be closer to the top. A blurb about [Recipes](https://github.com/Lasagne/Recipes) would be nice also.

f0k:
Agreed, the example scrolls other important aspects out of view. I can add a link to "Recipes" to the Example section. What'd be your suggested order of sections then?

ebenolson:
I'd go with something like
* Lasagne
* Installation
* Documentation
* Examples
* Design Goals
* Development

f0k:
Reordered. Maybe move "Design Goals" before "Installation"? It looks a bit lost where it is. (But it also is quite long and would push other sections outside the view.)

benanne:
Very nice work! Had a quick look at it but couldn't find anything to nitpick at :)

f0k:
I'm thinking about removing "Design Goals" (but linking to it, they're in the documentation anyway) and adding some bullet points about Lasagne's features instead. I'll try that later.

f0k:
I've removed the design goals and added a list of features instead. What do you think? Is there anything wrong/misleading/missing? (See https://github.com/f0k/Lasagne/blob/readme/README.rst for a nicely rendered version.)

README.rst
```diff
@@ -14,23 +14,103 @@ Lasagne
...
 
-For support, please refer to the lasagne-users mailing list: https://groups.google.com/forum/#!forum/lasagne-users
+* Support for feed-forward networks including Multi-Layer Perceptrons (MLPs)
+  and Convolutional Neural Networks (CNNs)
+* Support for recurrent networks including Long Short-Term Memory (LSTM) and
+  Gated Recurrent Units (GRUs) [coming soon!]
+* Support for architectures of multiple inputs and multiple outputs, including
+  auxiliary classifiers
+* Support for many optimization methods including Nesterov momentum, RMSprop
+  and ADAM
```
dnouri: I'd drop the 4x "Support for", it's repetitive.

README.rst
```diff
@@ -14,23 +14,103 @@ Lasagne
...
+
+  # create parameter update expressions
+  params = lasagne.layers.get_all_params(network, trainable=True)
+  updates = lasagne.updates.nesterov_momentum(loss, params, learning_rate=0.01,
+                                              momentum=0.9)
+
+  # compile training function that updates parameters and returns training loss
+  train_fn = theano.function([input_var, target_var], loss, updates=updates)
+
+For a fully-functional example, see `examples/mnist.py <examples/mnist.py>`_,
```
dnouri: I think this example ends somewhat abruptly.  I can imagine readers being left with a question of "now what?".  Why not add a few more lines that demonstrate how to actually use `train_fn` and then make predictions?

f0k: Yes, maybe... I mainly stopped because we're leaving the scope of Lasagne, and what's missing (reading data, training and testing) is included in the long example. The idea was to show how networks and update rules can be created, because that's all Lasagne can do for now, but of course that's only sufficient if you know Theano.

Are you thinking of something like:
```
# train network (assuming you've already got some training data as numpy arrays)
for epoch in range(100):
    loss = N = 0
    for input_batch, target_batch in training_data:
        loss += train_fn(input_batch, target_batch)
        N += 1
    print ("Epoch %d: Loss %g" % (epoch + 1, loss / N))

# use trained network for predictions
predict_fn = theano.function(
        [input_var],
        T.argmax(lasagne.layers.get_output(network, deterministic=True), axis=1))
print ("Predicted class for first ten test inputs: %r" % predict_fn(test_data[:10]))
```
Suggestions are highly welcome!

README.rst
f0k: That's right, but it felt weird to list "feed-forward networks" and "architectures" as features. We need a different introduction for the list then, one which fits "Feed-forward networks", "Optimization methods", but also "Designed to be easy to use and easy to extend". I'll figure something out.

README.rst
dnouri: > Are you thinking of something like:

Yes, I think these additional lines would be very instructive.

f0k:
Didn't find a good way to [drop the repeated "support for"](https://github.com/Lasagne/Lasagne/pull/342#discussion_r35404722), there's no unifying introductory sentence for all the bullet points then (especially the last one sticks out). Instead I merged the first two and reworded the other two. I dropped the explicit list of "Multi-Layer Perceptrons (MLPs), Convolutional Neural Networks (CNNs), Long Short-Term Memory (LSTM) and Gated Recurrent Units (GRUs)" because it was long and not exactly aesthetic, but I think these would be good keywords for google. Maybe we can later add a bullet point with "Includes examples for ..., ..., ..., ..." along with links once we have them.

I've also extended the example to include training and testing.

Any objections against squashing and merging?

benanne:
I kind of liked having the design goals up on the front page of the github repo, because I think they're really important and basically what set Lasagne apart from many other neural network libraries. This also ensures that everyone considering contributing has at least seen them. But I guess we should probably be shifting to catering towards users of the library, and not necessarily to contributors. So I'm okay with getting rid of them if most of you think that's better.

dnouri:
> I kind of liked having the design goals up on the front page of the github repo, because I think they're really important and basically what set Lasagne apart from many other neural network libraries.

I had the same thought.  +1 to putting design goals back.

ebenolson:
> Focus: follow the Unix philosophy of “do one thing and do it well”, with a strong focus on feed-forward neural networks.

I think this at least should be revised in light of #294.

benanne:
I guess so :) Although the design of the library is still pretty much focused on feedforward architectures, and RNNs were sort of built on top of that. I feel that some alternative libraries treat RNNs more like first-class citizens than we do, e.g. Blocks. But I'm okay with rephrasing it nevertheless.

f0k:
Ok, but where would you put the goals then? I think they're quite long in their current form, taking the focus from what I think are more important parts (the example and the installation instructions). I tried to include the most important design goals in the features list, namely the focus not only on easy usage (which doesn't set us apart), but also on extensibility (this is important when we're targeting researchers) and modularity (that's important as well for users coming from Theano and not wanting to completely switch over to a new library). Shall we continue that route, or include a somewhat shorter description list of design goals? Shall we put it after the list of features, or into the "development" section? Finally, does anybody of you want to help with this? (Copying and pasting the design goals from the Sphinx documentation is something I can easily do myself, but I don't think that's a good solution.)

craffel:
> I guess so :) Although the design of the library is still pretty much focused on feedforward architectures, and RNNs were sort of built on top of that. I feel that some alternative libraries treat RNNs more like first-class citizens than we do, e.g. Blocks. But I'm okay with rephrasing it nevertheless.

At this point I don't think there's any way that RNNs are not as much of first-class-citizens in `lasagne`, aside from the fact that you and @f0k don't use them ;) Given in particular that we have made a good handful of design decisions motivated by them (and will probably make a few more ).

benanne:
Depends how you look at it, really. I feel that the whole 'layer' paradigm is just inherently more suited to feedforward architectures than recurrent ones. You can nest recurrence within a layer, but that's not quite the same as treating it as a "first class citizen" imo.

Please don't get me wrong, this is not a value judgement of having recurrence in the library or anything, just an observation. I am in fact very happy that we have some high-quality code for supporting recurrent nets!

craffel:
> Depends how you look at it, really. I feel that the whole 'layer' paradigm is just inherently more suited to feedforward architectures than recurrent ones. You can nest recurrence within a layer, but that's not quite the same as treating it as a "first class citizen" imo.

Yes, but I think treating recurrence in another way is likely cumbersome; I see the layer architecture as just a straightforward way of building hierarchical networks in general - feedforward or recurrent.

> Please don't get me wrong, this is not a value judgement of having recurrence in the library or anything, just an observation. I am in fact very happy that we have some high-quality code for supporting recurrent nets!

No worries, I see your point.  I think the risk of stating that the main focus of the library is feedforward networks is that it will turn away people who are interested in recurrent networks needlessly, which is I think what @ebenolson is getting at.

f0k:
> I kind of liked having the design goals up on the front page of the github repo, because I think they're really important and basically what set Lasagne apart from many other neural network libraries.

Okay, I've put a shortened version of the design goals back in: https://github.com/f0k/Lasagne/tree/readme
I like the look, but I think they're not thought out well. I'll add some comments to the diff (although this PR is getting a bit out of hands).

README.rst
```diff
@@ -14,23 +14,122 @@ Lasagne
...
+* Transparent support of CPUs and GPUs due to Theano's expression compiler
 
-Documentation: http://lasagne.readthedocs.org/ (work in progress)
+Its design is governed by `six principles
+<http://lasagne.readthedocs.org/en/latest/user/development.html#philosophy>`_:
 
-Lasagne is a work in progress, input is welcome.
+* **Simplicity**: Be easy to use and easy to extend, with as few abstractions
+  as possible, to facilitate research
+* **Small interfaces**: Have as few classes and methods as possible
```
f0k: This is very closely related to "Simplicity". In the long version, it even says "This should make it easier to both use the library and extend it". I'd merge it with Simplicity or drop it.

README.rst
```diff
@@ -14,23 +14,122 @@ Lasagne
...
+* **Don't get in the way**: Do not obstruct users with features they decide not
+  to use, and allow all parts to be used independently
```
f0k: The first part is clear, but the second part is not really related to "Don't get in the way". I think it should be its own principle, something along modularity or reusability.

README.rst
```diff
@@ -14,23 +14,122 @@ Lasagne
...
+* **Transparency**: Do not hide Theano behind abstractions, directly process and
+  return Theano expressions or Python / numpy data types
```
f0k: The long version includes this in "Small interfaces" already: "Try to rely on Theano’s functionality and data types where possible, and follow Theano’s conventions". It's also included in "Simplicity": "Every added abstraction should be carefully scrutinized". And finally, it's also implied in allowing the parts to be used independently. Not sure what to make of that.

README.rst
```diff
@@ -14,23 +14,122 @@ Lasagne
...
+* **Focus**: "Do one thing and do it well"
```
f0k: I've removed talking about feedforward neural networks here in the shortened version. I'm thinking more along the lines of not aiming to provide a solution for everything involved in deep learning, but only facilitating some steps that would be more difficult in vanilla Theano (for now: creating a forward expression along with its learnable parameters, and creating update dictionaries).

README.rst
benanne: I'm okay with merging them :)

README.rst
benanne: Sure! The original list of 6 was the result of some stream-of-consciousness writing on my part, so the structure wasn't very well thought out.

f0k:
Thanks for the feedback, Sander! I've rewritten the design goals a bit (both in the README and in the documentation). I'm quite satisfied with the result.
The only thing I'm slightly unsure about is finding a noun capturing what was previously called **Don't get in the way**. For now I've chosen **Restraint**, in the sense that we try to constrain the effect of features on users, or in the sense that the library doesn't put every feature in every user's face. I was actually looking for something that describes how the interface is supposed to be clean and unobtrusive until you need something. Like being in a room with curtained walls, and you only pull aside the ones you want. I'm open to suggestions, but very keen on keeping it a noun, to go well with the others, or a single adjective.

In any case, I'd like to bring this PR to an end, squash and merge it, it wasn't intended to run forever! :)

benanne:
I thought "don't get in the way" was a little more immediately obvious. I'm not sure if there's a nice noun that captures the same essence (and that most people would understand). But the description is unambiguous so I think 'restraint' is fine for now :)

f0k:
> I thought "don't get in the way" was a little more immediately obvious.

It sure is, but stylistically it sticks out too much for my taste :)

Sooo... any objections against merging from anybody involved in the discussion? (I'll squash the first 6 commits into one before merging.)

ebenolson:
I think it would look better without the bold, it contrasts with the feature list above and draws focus.

I also think there's still a lot of overlap among the principles, but I don't have any very useful suggestions for that.

f0k:
> I think it would look better without the bold, it contrasts with the feature list above and draws focus.

Okay, I can take care of that when squashing. (I'll also remove the "[coming soon]".)

> I also think there's still a lot of overlap among the principles

Yes, but I think it's better than before, and some overlap may make things easier to understand. For example, what we have for "Transparency" is already implied by "Simplicity" and "Modularity", but I think it's still useful to write it out. And what we have for "Restraint" is an aspect of "Pragmatism", but for some design decisions it may be easier to point to the former than to the latter.

> but I don't have any very useful suggestions for that.

Good, let's leave it at that then! We can always change it later, PRs are for free on github ;)

morenoh149:
running the example I get `NameError: name 'training_data' is not defined` on line 40. Did I miss something?

PR #343, opened 2015-07-23
added support for setting a RandomState other than np.random

mheilman:
This is intended to address #6.

### DIFF
--- a/lasagne/__init__.py
+++ b/lasagne/__init__.py
@@ -18,6 +18,7 @@ from . import nonlinearities
 from . import init
 from . import layers
 from . import objectives
+from . import random
 from . import regularization
 from . import updates
 from . import utils
--- a/lasagne/init.py
+++ b/lasagne/init.py
@@ -4,13 +4,14 @@ Functions to create initializers for parameter variables.
 Examples
 --------
 >>> from lasagne.layers import DenseLayer
->>> from lasagne.init import Constant, Glorot
+>>> from lasagne.init import Constant, GlorotUniform
 >>> l1 = DenseLayer((100,20), num_units=50, W=GlorotUniform(), b=Constant(0.0))
 """
 
 import numpy as np
 
 from .utils import floatX
+from .random import get_rng
 
 
 class Initializer(object):
@@ -61,7 +62,7 @@ class Normal(Initializer):
         self.mean = mean
 
     def sample(self, shape):
-        return floatX(np.random.normal(self.mean, self.std, size=shape))
+        return floatX(get_rng().normal(self.mean, self.std, size=shape))
 
 
 class Uniform(Initializer):
@@ -101,7 +102,7 @@ class Uniform(Initializer):
         self.range = (a, b)
 
     def sample(self, shape):
-        return floatX(np.random.uniform(
+        return floatX(get_rng().uniform(
             low=self.range[0], high=self.range[1], size=shape))
 
 
@@ -317,9 +318,9 @@ class Sparse(Initializer):
 
         for k in range(n_outputs):
             indices = np.arange(n_inputs)
-            np.random.shuffle(indices)
+            get_rng().shuffle(indices)
             indices = indices[:size]
-            values = floatX(np.random.normal(0.0, self.std, size=size))
+            values = floatX(get_rng().normal(0.0, self.std, size=size))
             w[indices, k] = values
 
         return w
@@ -350,7 +351,7 @@ class Orthogonal(Initializer):
                                "supported.")
 
         flat_shape = (shape[0], np.prod(shape[1:]))
-        a = np.random.normal(0.0, 1.0, flat_shape)
+        a = get_rng().normal(0.0, 1.0, flat_shape)
         u, _, v = np.linalg.svd(a, full_matrices=False)
         # pick the one with the correct shape
         q = u if u.shape == flat_shape else v
--- a/lasagne/layers/noise.py
+++ b/lasagne/layers/noise.py
@@ -1,7 +1,7 @@
 import theano
-import numpy as np
 
 from .base import Layer
+from ..random import get_rng
 
 from theano.sandbox.rng_mrg import MRG_RandomStreams as RandomStreams
 
@@ -54,7 +54,7 @@ class DropoutLayer(Layer):
     """
     def __init__(self, incoming, p=0.5, rescale=True, **kwargs):
         super(DropoutLayer, self).__init__(incoming, **kwargs)
-        self._srng = RandomStreams(np.random.randint(1, 2147462579))
+        self._srng = RandomStreams(get_rng().randint(1, 2147462579))
         self.p = p
         self.rescale = rescale
 
@@ -112,7 +112,7 @@ class GaussianNoiseLayer(Layer):
     """
     def __init__(self, incoming, sigma=0.1, **kwargs):
         super(GaussianNoiseLayer, self).__init__(incoming, **kwargs)
-        self._srng = RandomStreams(np.random.randint(1, 2147462579))
+        self._srng = RandomStreams(get_rng().randint(1, 2147462579))
         self.sigma = sigma
 
     def get_output_for(self, input, deterministic=False, **kwargs):
--- a/lasagne/random.py
+++ b/lasagne/random.py
@@ -1,0 +1,36 @@
+"""
+A module with a package-wide random number generator,
+used for weight initialization and seeding noise layers.
+This can be replaced by a :class:`numpy.random.RandomState` instance with a
+particular seed to facilitate reproducibility.
+"""
+
+import numpy as np
+
+
+_rng = np.random
+
+
+def get_rng():
+    """Get the package-level random number generator.
+
+    Returns
+    -------
+    :class:`numpy.random.RandomState` instance
+        The :class:`numpy.random.RandomState` instance passed to the most
+        recent call of :func:`set_rng`, or ``numpy.random`` if :func:`set_rng`
+        has never been called.
+    """
+    return _rng
+
+
+def set_rng(new_rng):
+    """Set the package-level random number generator.
+
+    Parameters
+    ----------
+    new_rng : ``numpy.random`` or a :class:`numpy.random.RandomState` instance
+        The random number generator to use.
+    """
+    global _rng
+    _rng = new_rng
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -36,6 +36,7 @@ method, this part of the documentation is for you.
   modules/init
   modules/nonlinearities
   modules/objectives
+  modules/random
   modules/regularization
   modules/utils
 
--- a/docs/modules/random.rst
+++ b/docs/modules/random.rst
@@ -1,0 +1,7 @@
+:mod:`lasagne.random`
+====================
+
+.. automodule:: lasagne.random
+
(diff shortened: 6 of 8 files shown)


ebenolson:
I think it would be better to assign `rng = np.random` in `lasagne/__init__.py`, and omit the getter/setter.

dnouri:
> I think it would be better to assign rng = np.random in 
> lasagne/__init__.py, and omit the getter/setter.
> 

The problem with this is that code that does `from lasagne import rng` 
before other code that sets `lasagne.rng` to a custom value will still 
refer to the old `rng`.  It's error-prone.

lasagne/layers/noise.py
```diff
@@ -54,7 +54,7 @@ class DropoutLayer(Layer):
     """
     def __init__(self, incoming, p=0.5, rescale=True, **kwargs):
         super(DropoutLayer, self).__init__(incoming, **kwargs)
-        self._srng = RandomStreams(np.random.randint(1, 2147462579))
+        self._srng = RandomStreams(get_rng().randint(1, 2147462579))
```
f0k: Should we wrap this as `get_srng()`, or `create_srng()`?

f0k:
> code that does `from lasagne import rng`

Good catch! So we need to go with setters/getters.

In theory, we could even use different RNGs for weights and noise layers, via `get_rng('init')` and `get_rng('noise')`, and something like:
```python
rngs = {'': np.random}
def set_rng(rng, name=''):
    rngs[name] = rng
def get_rng(name=''):
    rngs.get(name, default=rngs[''])
```

Would that be helpful? It would be easier than explicitly passing RNGs for different purposes (e.g., as part of instantiating an initializer or layer), but almost as flexible.

dnouri:
> we could even use different RNGs for weights and noise layers

Your code looks good, though I can't think of a case where I'd want to do have different generators for init and noise.

ebenolson:
> The problem with this is that code that does `from lasagne import rng`
before other code that sets `lasagne.rng` to a custom value will still
refer to the old `rng`.  It's error-prone.

Good point, but I think it's more pythonic to just have an `rng.py` that consists of `from numpy import random as rng`, and refer to `rng.rng` where needed.

dnouri:
> Good point, but I think it's more pythonic to just have an rng.py that consists of from numpy import random as rng, and refer to rng.rng where needed.

I personally prefer from-imports that import module globals, so it would trip *me* up.  I don't think there's any official recommendation for one over the other.

lasagne/__init__.py
```diff
@@ -14,6 +14,7 @@
 else:
     del theano
 
+
```
f0k: why?

lasagne/init.py
```diff
@@ -4,13 +4,14 @@
 Examples
 --------
 >>> from lasagne.layers import DenseLayer
->>> from lasagne.init import Constant, Glorot
+>>> from lasagne.init import Constant
 >>> l1 = DenseLayer((100,20), num_units=50, W=GlorotUniform(), b=Constant(0.0))
```
f0k: It should be `Constant, GlorotUniform`. It still works without the import statement because it's the current module, but for people to be able to copy/paste it, we should include it.

f0k:
> Your code looks good, though I can't think of a case where I'd want to do have different generators for init and noise.

Me neither, at least none that is terribly realistic. It could be interesting when adding training loop support to Lasagne, though, and it was mostly meant as an additional argument for the getter/setter approach (because it allows us to extend that approach to support multiple rngs later, without breaking compatibility).

> I personally prefer from-imports that import module globals, so it would trip *me* up.

That's enough of a reason then!

I'm fine with merging, except that I'd like to hear other opinions on providing a `get_srng()` or `make_srng()` or `create_srng()` convenience function. Also I haven't checked test coverage yet, is it still 100%? coveralls is down for the moment.

ebenolson:
> I personally prefer from-imports that import module globals, so it would trip me up. I don't think there's any official recommendation for one over the other.

https://docs.python.org/2/howto/doanddont.html#from-module-import-name1-name2

dnouri:
> https://docs.python.org/2/howto/doanddont.html#from-module-import-name1-name2

Interesting.  Had no idea this is recommended against anywhere.  I guess I'll continue doing it though; saves me from typing so much.

mheilman:
@ebenolson, thanks for the suggestion on making it more pythonic.  I was having trouble working with the issue of importing module globals (e.g., how `from rng import rng` and then `rng = RandomState(42)` doesn't work), but I think it should work now.  I added a note in the docstring for `random.py` so that users hopefully don't make that mistake.

I also changed `rng.py` to `random.py` because I figured `lasagne.random.rng = rng1` looks better than `lasagne.rng.rng = rng1`.

dnouri:
Let's go back to using getter and setter here.  Because we might be able to decide what's Pythonic or not *for us*.  But how users will import and use that global is another story.  We should avoid this non-obvious pitfall; it's also a hard one to figure out.

+1 to renaming the module to `random`.

mheilman:
Sure, I can bring back the getter/setter.  I don't have a strong preference for either approach.

I'll wait a bit to see if anybody feels strongly either way.

dnouri:
> Incidentally, what I wanted to do at first was have a module-level property...

Interesting.  [It turns out](http://stackoverflow.com/questions/880530/can-python-modules-have-properties-the-same-way-that-objects-can/880751#880751) you can inherit from ModuleType and put an instance of your Module class (with properties!) into `sys.modules`.

But yeah, let's not go there.  :-)

ebenolson:
> Let's go back to using getter and setter here. Because we might be able to decide what's Pythonic or not for us. But how users will import and use that global is another story. We should avoid this non-obvious pitfall; it's also a hard one to figure out.

Ideally being pythonic is following an idiom that is widely known/accepted, not a personal judgment.
[I feel pretty certain what the right way to do this is](https://docs.python.org/2/faq/programming.html#how-do-i-share-global-variables-across-modules), but  it's really a very minor issue, so I'll leave it here.

dnouri:
@ebenolson This stuff is no way "widely known" and too easy to get wrong.  If another person can share their opinion, we can lay this issue to rest.

f0k:
I'm still on the getter/setter side a) to avoid the pitfall and b) to allow extending it to support multiple RNGs. I've also googled for module-level properties some days ago and found a similar stackoverflow thread as Daniel, but that's not a viable alternative :)
I'm fine with renaming the module to `lasagne.random` as well. And I'd still like to hear other's opinions on adding a `get_srng()`, `create_srng()` or `make_srng()` function performing what's currently done in the constructor of the two noise layers.

Hint: The easiest might be a) to `git reset --hard` to the "remove extra newline" commit, `git cherry-pick` the "added GlorotUniform ..." commit and then rename the module, or b) to `git revert` the "more pythonic" commit and then rename the module. Before merging, the commits should be squashed anyway.

benanne:
getter/setter seems like the safest bet, `lasagne.random` sounds good. But take my opinion with a grain of salt, I haven't had time to review the entire discussion.

mheilman:
I reverted the "more pythonic" commit, renamed `rng.py` to `random.py`, and rebased it into 1 commit.

I don't think methods like `get_srng` are necessary, but I don't mind adding them if people would prefer that.  Or, somebody else could add them pretty easily later in another PR if we change our minds.

lasagne/random.py
```diff
@@ -0,0 +1,26 @@
...
+import numpy as np
+
+
+_rng = np.random
+
+
+def get_rng():
+    """Get the package-level random number generator.
+    :return: a numpy.random or a numpy.random.RandomState object
+    """
```
f0k: We're using numpydoc throughout the package, see [the numpydoc howto](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) for a detailed treatment or any file in Lasagne for examples. Here, I'd suggest:
```
Get the package-level random number generator for Lasagne.

Returns
-------
:class:`numpy.random.RandomState` instance
    The :class:`numpy.random.RandomState` instance passed to the most
    recent call of :func:`set_rng`, or ``np.random`` if :func:`set_rng` has
    never been called.
```

f0k:
> I reverted the "more pythonic" commit, renamed `rng.py` to `random.py`, and rebased it into 1 commit.

Nice, thanks! I'm fine with postponing `get_srng()` until we have more use cases for it. The last request I'd like to make is changing the docstrings to numpydoc syntax, and including it in the documentation (this just needs a `docs/modules/random.rst` file that you can copy/paste and adapt from `utils.rst`, and a TOC entry in `docs/index.rst`, I guess I'd put it before `utils`). Oh yes, and including it in `__init__.py` (that one doesn't seem to have any particular order for some reason).

mheilman:
Ah, right, sorry about the improperly formatted docstrings, etc.  I'm not very familiar with numpydoc, so let me know if I need to correct anything.

lasagne/random.py
```diff
@@ -0,0 +1,36 @@
...
+
+def get_rng():
+    """Get the package-level random number generator.
+
+    Returns
+    -------
+    :class:`numpy.random.RandomState` instance
+        The :class:`numpy.random.RandomState` instance passed to the most
+        recent call of :func:`set_rng`, or `numpy.random` if :func:`set_rng` has
+        never been called.
```
f0k: `has` needs to be moved into the next line, otherwise it's too long (that's what Travis is complaining above).

While we're at it, `numpy.random` needs double apostrophes as in [my suggestion](https://github.com/Lasagne/Lasagne/pull/343#discussion_r35886511). Same for the module docstring above. The others are correct! (It takes a while to get used to Sphinx' ReST markup.)

mheilman:
Sorry again about the documentation issues.   I was confused by this sentence saying module names should be in single ticks: ```Use *italics*, **bold** and ``monospace`` if needed in any explanations (but not for variable names and doctest code or multi-line code). Variable, module, function, and class names should be written between single back-ticks (`numpy`)``` (https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt), but I guess that only works for classes and modules within the package (or with intersphinx?).

I built the docs (as I should have done before!), and I think it looks OK now.  Please let me know if you notice anything else.

By the way, when I ran sphinx, I needed to pip install `sphinx_rtd_theme`.  Would it be worth adding that to `requirements-dev.txt`?

f0k:
> I guess that only works for classes and modules within the package (or with intersphinx?).

Yes, I think that's only when using the `:mod:`, `:class:` or `:func:` prefixes. When not using a prefix, it will be treated as a variable name (so that's what should be used for referencing function arguments in a docstring). intersphinx would be useful to actually make `numpy.random` link to something, but we currently don't use it.

> By the way, when I ran sphinx, I needed to pip install `sphinx_rtd_theme`. Would it be worth adding that to `requirements-dev.txt`?

Yes... I removed it in #257 because it's included with Sphinx, but it's actually *not* included with the version of Sphinx we've got in `requirements-dev.txt` (my bad), so we should add it back. This should be a separate PR, though.

> I think it looks OK now.

Yes, thanks a lot! I'm fine with merging this :+1: Let's wait a bit if anybody else wants to comment.

f0k:
> This should be a separate PR, though.

Or maybe I'll just add that to #335 because it touches the configuration a bit.

f0k:
Merged, thanks again for your contribution!

PR #344, opened 2015-07-24
Merge border_mode with pad argument for convolutional layers

f0k:
Continuation of #311. Closes #311. (So github knows.)

I'm starting from the relevant changes done by Martin, dropping all unrelated docstring style changes (sorry, they don't really belong here and they caused a merge conflict).

I'll try fixing all Python3- and cuDNN-related bugs now.

### DIFF
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -14,8 +14,7 @@ __all__ = [
 ]
 
 
-def conv_output_length(input_length, filter_size,
-                       stride, pad=0):
+def conv_output_length(input_length, filter_size, stride, pad=0):
     """Helper function to compute the output size of a convolution operation
 
     This function computes the length along a single axis, which corresponds
@@ -37,7 +36,7 @@ def conv_output_length(input_length, filter_size,
         By default, the convolution is only computed where the input and the
         filter fully overlap (a valid convolution). When ``stride=1``, this
         yields an output that is smaller than the input by ``filter_size - 1``.
-        The `pad` argument allows to implicitly pad the input with zeros,
+        The `pad` argument allows you to implicitly pad the input with zeros,
         extending the output size.
 
         A single integer results in symmetric zero-padding of the given size on
@@ -110,11 +109,11 @@ class Conv1DLayer(Layer):
         An integer or a 1-element tuple specifying the stride of the
         convolution operation.
 
-    pad : {0, 'full', 'same'} (default: 0)
+    pad : {0, (0,), 'full', 'same', 'valid'} (default: 0)
         By default, the convolution is only computed where the input and the
         filter fully overlap (a valid convolution). When ``stride=1``, this
         yields an output that is smaller than the input by ``filter_size - 1``.
-        The `pad` argument allows to implicitly pad the input with zeros,
+        The `pad` argument allows you to implicitly pad the input with zeros,
         extending the output size.
 
         ``'full'`` pads with one less than the filter size on both sides. This
@@ -125,8 +124,7 @@ class Conv1DLayer(Layer):
         the second side for an even filter size). When ``stride=1``, this
         results in an output size equal to the input size.
 
-        Note that ``'full'`` and ``'same'`` can be faster than equivalent
-        integer values due to optimizations by Theano.
+        ``'valid'`` is an alias for ``0`` (no padding / a valid convolution).
 
     untie_biases : bool (default: False)
         If ``False``, the layer will have a bias parameter for each channel,
@@ -176,7 +174,7 @@ class Conv1DLayer(Layer):
     Notes
     -----
     Theano's default convolution function (`theano.tensor.nnet.conv.conv2d`)
-    does not support the 'same' pad by default. This layer emulates
+    does not support ``pad='same'`` by default. This layer emulates
     it by performing a 'full' convolution and then cropping the result, which
     may negatively affect performance.
     """
@@ -194,12 +192,17 @@ class Conv1DLayer(Layer):
         self.num_filters = num_filters
         self.filter_size = as_tuple(filter_size, 1)
         self.stride = as_tuple(stride, 1)
-        self.pad = pad
         self.untie_biases = untie_biases
         self.convolution = convolution
 
-        if self.pad not in ['valid', 'full', 'same', 0]:
-            raise RuntimeError("Invalid pad: '%s'" % self.pad)
+        if pad == 'valid':
+            self.pad = (0,)
+        elif pad in ('full', 'same'):
+            self.pad = pad
+        else:
+            self.pad = as_tuple(pad, 1, int)
+            if self.pad != (0,):
+                raise ValueError("Arbitrary padding is not supported.")
 
         self.W = self.add_param(W, self.get_W_shape(), name="W")
         if b is None:
@@ -224,10 +227,12 @@ class Conv1DLayer(Layer):
         return (self.num_filters, num_input_channels, self.filter_size[0])
 
     def get_output_shape_for(self, input_shape):
+        pad = self.pad if isinstance(self.pad, tuple) else (self.pad,)
+
         output_length = conv_output_length(input_shape[2],
                                            self.filter_size[0],
                                            self.stride[0],
-                                           self.pad)
+                                           pad[0])
 
         return (input_shape[0], self.num_filters, output_length)
 
@@ -239,12 +244,7 @@ class Conv1DLayer(Layer):
 
         filter_shape = self.get_W_shape()
 
-        if self.pad in [0, 'valid', 'full']:
-            conved = self.convolution(input, self.W, subsample=self.stride,
-                                      image_shape=input_shape,
-                                      filter_shape=filter_shape,
-                                      border_mode=self.pad or 'valid')
-        elif self.pad == 'same':
+        if self.pad == 'same':
             if self.stride[0] != 1:
                 raise NotImplementedError("Strided convolution with "
                                           "pad 'same' is not "
@@ -256,6 +256,12 @@ class Conv1DLayer(Layer):
                                       border_mode='full')
             shift = (self.filter_size[0] - 1) // 2
             conved = conved[:, :, shift:input.shape[2] + shift]
+        else:
+            border_mode = {(0,): 'valid', 'full': 'full'}[self.pad]
+            conved = self.convolution(input, self.W, subsample=self.stride,
+                                      image_shape=input_shape,
+                                      filter_shape=filter_shape,
+                                      border_mode=border_mode)
 
         if self.b is None:
             activation = conved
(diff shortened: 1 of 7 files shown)


f0k:
There was more left to do than anticipated.

* fixed all the bugs, docstrings, and tests
* made `self.pad` a tuple (`(0,)` or `(0, 0)` instead of `0`) for `Conv1DLayer` and `Conv2DLayer` to be consistent with other layer implementations and be ready for the future (`Conv2DLayer` will support custom padding when Theano adds it)
* restored Python3 compatibility by avoiding the `isinstance(pad, basestring)` check

For the last step, I first devised a nice solution with a `is_int_or_ints()` helper method that resulted in clauses like:
```python
if pad == 'valid':
    self.pad = (0, 0)
elif pad == 'full':
    ...
elif is_int_or_ints(pad):
    self.pad = as_tuple(pad, 2)
else:
    raise ValueError("Invalid pad: %r" % pad)
```
But then I realized that if we're really supporting any kind of iterable for `pad` (as for `filter_size` and `stride`), this won't work, because `is_int_or_ints()` would necessarily consume `pad` if it was a generator, leaving nothing for `as_tuple()`. So I moved the type check into `as_tuple()`, otherwise we'd have to move materialization of the iterable out of that call and replicate a lot of code. This gives a somewhat less informative error message if you give a funny string for `pad`, but I for one can live with that.

Ready to be reviewed!

lasagne/layers/conv.py
```diff
@@ -117,23 +109,22 @@ class Conv1DLayer(Layer):
         An integer or a 1-element tuple specifying the stride of the
         convolution operation.
 
-    border_mode : str, one of 'valid', 'full', 'same'
-        A string indicating the convolution border mode.
+    pad : {0, (0,), 'full', 'same'} (default: 0)
```
ebenolson: It seems a bit strange to me to have 0 here instead of something like 'int' or 'n'. Also, is it necessary to repeat the default from the signature? If so, it should probably be included for all arguments.

lasagne/layers/conv.py
```diff
@@ -33,29 +32,23 @@ def conv_output_length(input_length, filter_size,
     stride : int
         The stride of the convolution operation.
 
-    border_mode : str, 'valid', 'full', 'same' or 'pad'
-        A string indicating the convolution border mode.
+    pad : int, 'full' or 'same' (default: 0)
+        By default, the convolution is only computed where the input and the
+        filter fully overlap (a valid convolution). When ``stride=1``, this
+        yields an output that is smaller than the input by ``filter_size - 1``.
+        The `pad` argument allows to implicitly pad the input with zeros,
```
ebenolson: "allows you to" or maybe "allows implicitly padding"?

lasagne/layers/conv.py
ebenolson: Ok, I guess it's `0` because `nnet.conv2d` doesn't support nonzero padding yet? This seems pretty confusing to me.

I think it should either just take `['valid', 'full', 'same']`, or we should enable padding by wrapping the input in a `PadLayer`.

f0k: It can't take `valid` because that's not a valid keyword anymore. I could write `int, iterable of int` etc. and then note in the text that it must be zero, would that be better?
I'm not so sure about including the default everywhere either, Martin started that. I can remove it again.

lasagne/layers/conv.py
f0k: Mhm, isn't that all the same? (If it's wrong I'll gladly change it, but otherwise I'd better like to keep it because it's occurring in many places.)

lasagne/layers/conv.py
ebenolson: > It can't take valid because that's not a valid keyword anymore.

why not? it looks like it's still in the code below...

f0k: > wrapping the input in a `PadLayer`

Interesting idea. Might be a little unexpected to get two layers when you thought you were instantiating only one, though. We could instead just call `theano_extensions.padding.pad` as part of `get_output_for()`, that would fix the documentation problem and just be terribly slow until Theano catches up.

f0k: > why not? it looks like it's still in the code below...

Yes, we decided we'd keep it for compatibility reasons (people having the border mode as a positional argument), but have the official way be `pad=0`. Hmm, thinking about it, I guess we actually decided to give a warning in that case, but Martin didn't and I didn't remember to add it now.

ebenolson: Ah, ok. I think `pad=None` would be a little clearer, but `pad=0` works also. I do think it's confusing though to document multiple ways to to specify the same thing (`pad=0`, `pad=(0,)` and `pad=(0,0)` for the Conv2DLayer), even if they are all accepted.

lasagne/layers/conv.py
ebenolson: "Allows to" I don't think is grammatical.

lasagne/layers/conv.py
f0k: All right, what's the final suggestion then? Allow arbitrary padding via `theano_extensions`? Or remove `(0,)` or `(0, 0)` and only keep `0` in the docstring? Independently of that, I'll give a warning for `pad='valid'` in all layers.

lasagne/layers/conv.py
f0k: [You're right!](http://ell.stackexchange.com/questions/11193/allow-to-infinitive-substantive-verb-ing) Both your suggestions are grammatical, mine isn't. I thought "allows to do" and "allows doing" were interchangeable. Will fix the docstring and my [sprachgefühl](https://en.wiktionary.org/wiki/Sprachgef%C3%BChl).

lasagne/layers/conv.py
ebenolson: I think using `theano_extensions` would be the best option, so that it can have the same semantics as the other ConvLayers. Then when Theano catches up, it should be a simple speedup with no change visible to the user. It would be nice if you could reuse PadLayer, but I think you're right that returning two layers would be confusing.

f0k: > It would be nice if you could reuse PadLayer,

I'd still be practically reusing its `get_output_for()`, I just wouldn't use its `get_output_shape_for()`, but we've got `conv_output_length()` for that anyway. I think it's a lot cleaner to keep everything within the `Conv2DLayer` and `Conv1DLayer`, especially if we anticipate changing it later. Okay then, thanks for the feedback, will cope with this soon!

f0k:
Addressed what Eben and I discussed. Ready to be reviewed again! I can squash 71f... and 719... into ec99... before merging.

lasagne/layers/conv.py
```diff
@@ -242,25 +245,34 @@ def get_output_for(self, input, input_shape=None, **kwargs):
...
+            # no padding needed, or explicit padding of input needed
+            if self.pad == 'full':
+                border_mode = 'full'
+                pad = (0, 0)
+            elif self.pad == 'same':
+                border_mode = 'valid'
+                pad = self.filter_size[0] // 2, (self.filter_size[0] - 1) // 2
+            else:
+                border_mode = 'valid'
+                pad = (self.pad[0], self.pad[0])
```
ebenolson: is there a reason `pad = self.pad` wouldn't work here?

ebenolson: Never mind, I get it..

lasagne/layers/corrmm.py
```diff
@@ -51,30 +51,36 @@ class Conv2DMMLayer(MMLayer):
...
-        is odd. In practice, the input is zero-padded with half the filter size
-        at the beginning and half at the end (or one less than half in the case
-        of an even filter size). This results in an output length that is the
-        same as the input length (for both odd and even filter sizes).
+        ``'same'`` pads with half the filter size on both sides (one less on
+        the second side for an even filter size). When ``stride=1``, this
+        results in an output size equal to the input size.
 
-    untie_biases : bool, default False
+        Note that ``'full'`` and ``'same'`` can be faster than equivalent
```
ebenolson: is this still true?

lasagne/layers/conv.py
```diff
@@ -181,12 +175,12 @@ class Conv1DLayer(Layer):
     Notes
     -----
     Theano's default convolution function (`theano.tensor.nnet.conv.conv2d`)
```
ebenolson: This warning applies to everything but `pad=0` and `pad='full`, right?

ebenolson:
Made a few comments on the docs, but it looks good to me.

I'm not entirely happy about having an undocumented 'valid' mode though, I think it should either be supported or deprecated with a timeline for removal.

lasagne/layers/corrmm.py
f0k: Yes! (And I don't plan to change that as part of this PR, sorry... it's already taking too long.)

lasagne/layers/conv.py
f0k: Hmm, sort of. For `pad='same'` and `stride=(1,)`, it crops a full convolution. For `pad='same'` and `stride!=(1,)` or for custom padding, it explicitly pads the input. I can change the warning to:
```
Theano's underlying convolution (:func:`theano.tensor.nnet.conv.conv2d`)
only supports ``pad=0`` and ``pad='full'``. This layer emulates other modes
by cropping a full convolution or explicitly padding the input with zeroes.
```

f0k:
> I'm not entirely happy about having an undocumented 'valid' mode though, I think it should either be supported or deprecated

Hmm, then I'll support `'valid'` and add a sentence that it's an alias for `0`. I hope this makes clear that passing `pad='valid'` will not result in `layer.pad == 'valid'`, but in `layer.pad == (0, 0)`. I'd definitely want to avoid setting `self.pad = 'valid'` as this would complicate the code.

f0k:
Did [this](https://github.com/Lasagne/Lasagne/pull/344#issuecomment-127219076) and [that](https://github.com/Lasagne/Lasagne/pull/344#discussion_r36080240), [squashed something](https://github.com/Lasagne/Lasagne/pull/344#issuecomment-126388237), are you okay with merging now? Thanks for reviewing!

ebenolson:
looks good to me!

f0k:
Wait a bit... wanted to run the test suite again to be sure, but Theano suddenly fails to `import theano` with a weird compile error (`internal error: assertion failed: lower_expr: bad kind (/dvs/p4/build/sw/rel/gpu_drv/r340/r340_00/drivers/compiler_CUREL/edg/EDG_4.8/src/lower_il.c, line 15873)`).

This doesn't have anything to do with this PR, but I'd still like to run the test suite before merging. I'll try to get it to work again.

ebenolson:
strange... I ran the tests earlier today without trouble, but not since your latest commits.

f0k:
> I ran the tests earlier today without trouble

Including pylearn2?

One thing I've already noticed is that `GpuCorrMM` would like us to use `border_mode` instead of `pad`. I think Martin changed `border_mode=border_mode` to `pad=pad` instead of `border_mode=pad`. I'll fix that, but I'd like to make sure it's the only thing left to do.

ebenolson:
> Including pylearn2? 

I think so.. saw 100% coverage anyway.

> One thing I've already noticed is that GpuCorrMM would like us to use border_mode instead of pad.

Are you sure? It looks to me like `border_mode` is basically a placeholder : http://deeplearning.net/software/theano/library/sandbox/cuda/op.html#theano.sandbox.cuda.blas.GpuCorrMM

f0k:
> Are you sure?

Yes: https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/blas.py#L724-L729
It seems https://github.com/Theano/Theano/pull/2259 changed the underlying code of `GpuCorrMM` without updating my nicely written docstring, and I didn't intervene.

f0k:
Fixed that `border_mode` / `pad` confusion in `lasagne/layers/corrmm.py`, got Theano working again (`apt-get update && apt-get dist-upgrade && reboot` did the trick) and ran the full test suite without troubles and 100% coverage. Will merge when Travis passes.

>> Including pylearn2?

> I think so.. saw 100% coverage anyway.

If it's installed, it's tested. If it's not installed, it's not tested and does not count towards the coverage report -- at least that's how I think it's configured.

PR #348, opened 2015-07-29
optimise adam solver

kashif:
moved two calculations outside the for-zip loop


f0k:
Looks fine. Not sure if it's easier to read than before, but maybe it helps to make clear `t` and `a_t` are the same for every parameter.

Note that this doesn't actually optimize anything -- creating the expression graph may be a few microseconds faster, but runtime performance is exactly the same (the loop is only executed once to create the update dictionary which is then compiled into a function, it is not executed at each update step).

Any other opinions?

kashif:
true... just a bit clearer then

f0k:
Merging since there are no more comments, thank you!

kashif:
bitte!

PR #351, opened 2015-08-03
Correct typo in cuDNN instructions

### LINKED ISSUE #350: Typo in installation instructions
The instructions at docs/user/installation.rst contain the instruction:

    python -c "import theano; print theano.sandbox.cuda.dnn.dnn_available() or theano.sandbox.cuda.dnn.dnn_avalailable.msg"

Should the second "avalailable" be spelt "available"?
If I run it as written I get a Python error message, while if I run it with the normal spelling it tells me that the library is not supported by my card.

(It may be worth noting at this point in the instructions that compute capability 3.0 is required - and that you can determine which cards support this by looking at the list here: https://developer.nvidia.com/cuda-gpus )

peterderivaz:
Resolves #350.

### DIFF
--- a/docs/user/installation.rst
+++ b/docs/user/installation.rst
@@ -213,6 +213,9 @@ NVIDIA provides a library for common neural network operations that especially
 speeds up Convolutional Neural Networks (CNNs). Again, it can be obtained from
 NVIDIA (after registering as a developer): https://developer.nvidia.com/cudnn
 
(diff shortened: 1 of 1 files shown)


f0k:
Perfect, thank you!

f0k:
Oh, too quick, that's markdown code I gave you for the link, not ReST. Will fix it.

PR #360, opened 2015-08-06
Allow more than 1 feature dimension in CustomRecurrentLayer

### LINKED ISSUE #358: Allow (custom) recurrent layers to accept more than one feature dimension
As noted by @f0k here https://github.com/Lasagne/Lasagne/pull/294/files#r35102212, currently all the recurrent layers, including the `CustomRecurrentLayer`, squash the input past the third dimension.  The `CustomRecurrentLayer` doesn't need to do that; it should be able to accept N-D inputs and have, e.g., convolutional layers for its input-to-hidden and hidden-to-hidden connections.  I addressed that here: https://github.com/craffel/nntools/commit/1b55bbe9d910b7bc117040286ae0e6bc22053262 but it needs to be updated.  If and when there are more custom recurrent layer types #357, those should also be able to handle more-than-one-feature-dimension input.

craffel:
Resolves #358.  See the test for a usage example.

### DIFF
--- a/lasagne/layers/recurrent.py
+++ b/lasagne/layers/recurrent.py
@@ -2,8 +2,9 @@
 """
 Layers to construct recurrent networks. Recurrent layers can be used similarly
 to feed-forward layers except that the input shape is expected to be
-``(batch_size, sequence_length, num_inputs)``. The input is allowed to have
-more than three dimensions in which case dimensions trailing the third
+``(batch_size, sequence_length, num_inputs)``.   The CustomRecurrentLayer can
+also support more than one "feature" dimension (e.g. using convolutional
+connections), but for all other layers, dimensions trailing the third
 dimension are flattened.
 
 The following recurrent layers are implemented:
@@ -166,6 +167,22 @@ class CustomRecurrentLayer(MergeLayer):
     ...     lasagne.layers.InputLayer((None, n_hid)), n_hid)
     >>> l_rec = lasagne.layers.CustomRecurrentLayer(l_in, l_in_hid, l_hid_hid)
 
+    The CustomRecurrentLayer can also support "convolutional recurrence", as is
+    demonstrated below.
+
+    >>> n_batch, n_steps, n_channels, width, height = (2, 3, 4, 5, 6)
+    >>> n_out_filters = 7
+    >>> filter_shape = (3, 3)
+    >>> l_in = lasagne.layers.InputLayer(
+    ...     (n_batch, n_steps, n_channels, width, height))
+    >>> l_in_to_hid = lasagne.layers.Conv2DLayer(
+    ...     lasagne.layers.InputLayer((None, n_channels, width, height)),
+    ...     n_out_filters, filter_shape, pad='same')
+    >>> l_hid_to_hid = lasagne.layers.Conv2DLayer(
+    ...     lasagne.layers.InputLayer(l_in_to_hid.output_shape),
+    ...     n_out_filters, filter_shape, pad='same')
+    >>> l_rec = lasagne.layers.CustomRecurrentLayer(
+    ...     l_in, l_in_to_hid, l_hid_to_hid)
 
     References
     ----------
@@ -213,36 +230,84 @@ class CustomRecurrentLayer(MergeLayer):
             raise ValueError("Input sequence length cannot be specified as "
                              "None when unroll_scan is True")
 
-        # Check that output shapes match
-        if input_to_hidden.output_shape != hidden_to_hidden.output_shape:
+        # Check that the input_to_hidden connection can appropriately handle
+        # a first dimension of input_shape[0]*input_shape[1] when we will
+        # precompute the input dot product
+        if (self.precompute_input and
+                input_to_hidden.output_shape[0] is not None and
+                input_shape[0] is not None and
+                input_shape[1] is not None and
+                (input_to_hidden.output_shape[0] !=
+                 input_shape[0]*input_shape[1])):
+            raise ValueError(
+                'When precompute_input == True, '
+                'input_to_hidden.output_shape[0] must equal '
+                'incoming.output_shape[0]*incoming.output_shape[1] '
+                '(i.e. batch_size*sequence_length) or be None but '
+                'input_to_hidden.output_shape[0] = {} and '
+                'incoming.output_shape[0]*incoming.output_shape[1] = '
+                '{}'.format(input_to_hidden.output_shape[0],
+                            input_shape[0]*input_shape[1]))
+
+        # Check that the first dimension of input_to_hidden and
+        # hidden_to_hidden's outputs match when we won't precompute the input
+        # dot product
+        if (not self.precompute_input and
+                input_to_hidden.output_shape[0] is not None and
+                hidden_to_hidden.output_shape[0] is not None and
+                (input_to_hidden.output_shape[0] !=
+                 hidden_to_hidden.output_shape[0])):
+            raise ValueError(
+                'When precompute_input == False, '
+                'input_to_hidden.output_shape[0] must equal '
+                'hidden_to_hidden.output_shape[0] but '
+                'input_to_hidden.output_shape[0] = {} and '
+                'hidden_to_hidden.output_shape[0] = {}'.format(
+                    input_to_hidden.output_shape[0],
+                    hidden_to_hidden.output_shape[0]))
+
+        # Check that input_to_hidden and hidden_to_hidden output shapes match,
+        # but don't check a dimension if it's None for either shape
+        if not all(s1 is None or s2 is None or s1 == s2
+                   for s1, s2 in zip(input_to_hidden.output_shape[1:],
+                                     hidden_to_hidden.output_shape[1:])):
             raise ValueError("The output shape for input_to_hidden and "
-                             "input_to_hidden must be equal, but "
-                             "input_to_hidden.output_shape={} and "
-                             "hidden_to_hidden.output_shape={}".format(
+                             "hidden_to_hidden must be equal after the first "
+                             "dimension, but input_to_hidden.output_shape={} "
+                             "and hidden_to_hidden.output_shape={}".format(
                                  input_to_hidden.output_shape,
                                  hidden_to_hidden.output_shape))
+
+        # Check that input_to_hidden's output shape is the same as
+        # hidden_to_hidden's input shape but don't check a dimension if it's
+        # None for either shape
+        if not all(s1 is None or s2 is None or s1 == s2
+                   for s1, s2 in zip(input_to_hidden.output_shape[1:],
+                                     hidden_to_hidden.input_shape[1:])):
+            raise ValueError("The output shape for input_to_hidden "
+                             "must be equal to the input shape of "
+                             "hidden_to_hidden after the first dimension, but "
+                             "input_to_hidden.output_shape={} and "
+                             "hidden_to_hidden.input_shape={}".format(
+                                 input_to_hidden.output_shape,
+                                 hidden_to_hidden.input_shape))
 
         if nonlinearity is None:
             self.nonlinearity = nonlinearities.identity
         else:
             self.nonlinearity = nonlinearity
 
-        # Get the input dimensionality and number of units based on the
-        # expected output of the input-to-hidden layer
-        self.num_inputs = np.prod(input_shape[2:])
-        self.num_units = input_to_hidden.output_shape[-1]
-
         # Initialize hidden state
         if isinstance(hid_init, T.TensorVariable):
-            if hid_init.ndim != 2:
+            if hid_init.ndim != len(hidden_to_hidden.output_shape):
                 raise ValueError(
                     "When hid_init is provided as a TensorVariable, it should "
-                    "have 2 dimensions and have shape (num_batch, num_units)")
+                    "have the same shape as hidden_to_hidden.output_shape")
             self.hid_init = hid_init
         else:
             self.hid_init = self.add_param(
-                hid_init, (1, self.num_units), name="hid_init",
-                trainable=learn_init, regularizable=False)
+                hid_init, (1,) + hidden_to_hidden.output_shape[1:],
+                name="hid_init", trainable=learn_init, regularizable=False)
 
     def get_params(self, **tags):
         # Get all parameters from this layer, the master layer
@@ -256,7 +321,8 @@ class CustomRecurrentLayer(MergeLayer):
         # The shape of the input to this layer will be the first element
         # of input_shapes, whether or not a mask input is being used.
         input_shape = input_shapes[0]
-        return input_shape[0], input_shape[1], self.num_units
+        return ((input_shape[0], input_shape[1]) +
+                self.hidden_to_hidden.output_shape[1:])
 
     def get_output_for(self, inputs, **kwargs):
         """
@@ -286,28 +352,27 @@ class CustomRecurrentLayer(MergeLayer):
         # Retrieve the mask when it is supplied
         mask = inputs[1] if len(inputs) > 1 else None
 
-        # Treat all dimensions after the second as flattened feature dimensions
-        if input.ndim > 3:
-            input = T.flatten(input, 3)
-
         # Input should be provided as (n_batch, n_time_steps, n_features)
         # but scan requires the iterable dimension to be first
         # So, we need to dimshuffle to (n_time_steps, n_batch, n_features)
-        input = input.dimshuffle(1, 0, 2)
-        seq_len, num_batch, _ = input.shape
+        input = input.dimshuffle(1, 0, *range(2, input.ndim))
+        seq_len, num_batch = input.shape[0], input.shape[1]
 
         if self.precompute_input:
             # Because the input is given for all time steps, we can precompute
             # the inputs to hidden before scanning. First we need to reshape
-            # from (seq_len, batch_size, num_inputs) to
-            # (seq_len*batch_size, num_inputs)
-            input = T.reshape(input,
-                              (seq_len*num_batch, -1))
+            # from (seq_len, batch_size, trailing dimensions...) to
+            # (seq_len*batch_size, trailing dimensions...)
+            # This strange use of a generator in a tuple was because
+            # input.shape[2:] was raising a Theano error
+            trailing_dims = tuple(input.shape[n] for n in range(2, input.ndim))
+            input = T.reshape(input, (seq_len*num_batch,) + trailing_dims)
             input = helper.get_output(
                 self.input_to_hidden, input, **kwargs)
 
-            # Reshape back to (seq_len, batch_size, num_units)
-            input = T.reshape(input, (seq_len, num_batch, -1))
+            # Reshape back to (seq_len, batch_size, trailing dimensions...)
+            trailing_dims = tuple(input.shape[n] for n in range(1, input.ndim))
+            input = T.reshape(input, (seq_len, num_batch) + trailing_dims)
 
         # We will always pass the hidden-to-hidden layer params to step
         non_seqs = helper.get_all_params(self.hidden_to_hidden)
@@ -354,8 +419,13 @@ class CustomRecurrentLayer(MergeLayer):
         if isinstance(self.hid_init, T.TensorVariable):
             hid_init = self.hid_init
         else:
-            # Dot against a 1s vector to repeat to shape (num_batch, num_units)
-            hid_init = T.dot(T.ones((num_batch, 1)), self.hid_init)
+            # The code below simply repeats self.hid_init num_batch times in
+            # its first dimension.  Turns out using a dot product and a
+            # dimshuffle is faster than T.repeat.
+            dot_dims = (list(range(1, self.hid_init.ndim - 1)) +
+                        [0, self.hid_init.ndim - 1])
+            hid_init = T.dot(T.ones((num_batch, 1)),
+                             self.hid_init.dimshuffle(dot_dims))
 
         if self.unroll_scan:
             # Retrieve the dimensionality of the incoming layer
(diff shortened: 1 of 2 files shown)


f0k:
Excellent! Will have a closer look tomorrow!

craffel:
Hm, interesting, Travis doesn't like it; it looks like the conv backend being used on Travis is different from mine and is unhappy:
```
E           ValueError: The hardcoded shape for the batch size (2) isn't the run time shape (6).
E           Apply node that caused the error: ConvOp{('imshp', (4, 5, 6)),('kshp', (3, 3)),('nkern', 7),('bsize', 2),('dx', 1),('dy', 1),('out_mode', 'full'),('unroll_batch', 2),('unroll_kern', 1),('unroll_patch', False),('imshp_logical', (4, 5, 6)),('kshp_logical', (3, 3)),('kshp_logical_top_aligned', True)}(Reshape{4}.0, W)
E           Toposort index: 25
E           Inputs types: [TensorType(float64, 4D), TensorType(float64, 4D)]
E           Inputs shapes: [(6, 4, 5, 6), (7, 4, 3, 3)]
E           Inputs strides: [(960, 240, 48, 8), (288, 72, 24, 8)]
E           Inputs values: ['not shown', 'not shown']
E           Outputs clients: [[Subtensor{::, ::, int64:int64:, int64:int64:}(ConvOp{('imshp', (4, 5, 6)),('kshp', (3, 3)),('nkern', 7),('bsize', 2),('dx', 1),('dy', 1),('out_mode', 'full'),('unroll_batch', 2),('unroll_kern', 1),('unroll_patch', False),('imshp_logical', (4, 5, 6)),('kshp_logical', (3, 3)),('kshp_logical_top_aligned', True)}.0, Constant{1}, ScalarFromTensor.0, Constant{1}, ScalarFromTensor.0)]]
E           
E           Backtrace when the node is created:
E             File "/home/travis/build/Lasagne/Lasagne/lasagne/layers/conv.py", line 460, in get_output_for
E               border_mode='full')
E           
E           HINT: Use the Theano flag 'exception_verbosity=high' for a debugprint and storage map footprint of this apply node.
```

craffel:
Ah, I see what's happening.  Before pre-computing the input to the RNN layer, there's a reshape which squashes the first two input dimensions:
https://github.com/craffel/nntools/blob/recurrent_any_hid_shape/lasagne/layers/recurrent.py#L304
This makes the input to the convolution be `n_batch*n_seq`, rather than `batch_size` as I defined it.  You might think it could be fixed by changing the input shape to the `l_in_to_hid` convolutional layer so it expects `n_batch*n_seq` but then the recurrent layer complains (as it probably should) that the output shape of `l_in_to_hid` and `l_hid_to_hid` is different.  Not sure what the cleanest way to solve this is, because it really comes down to my invalid assumption that all of the ops we're using (like dot or conv) should allow the first dimension to vary.

craffel:
Hah, wait a second, the first dimension should just be set to `None` for the recurrent connection layers, and then it works fine.  I changed the test to do that and it should pass on Travis.  This is an interesting failure mode though.

f0k:
> You might think it could be fixed by changing the input shape to the `l_in_to_hid` convolutional layer so it expects `n_batch*n_seq` but then the recurrent layer complains (as it probably should) that the output shape of `l_in_to_hid` and `l_hid_to_hid` is different.

Well, `l_in_to_hid` would need a `ReshapeLayer` on top that gets back to `(n_batch, n_seq, ...)`. That one could also be added automatically if `precompute_input == True`.

> This is an interesting failure mode though.

It should definitely be documented that `precompute_input=True` requires a different input shape for `l_in_to_hid`! Everything else should be unaffected (i.e., the output shape of `l_in_to_hid` and the input and output shapes of `l_hid_to_hid` should all be equal, and all be the same no matter if `precompute_input` is set or not).

craffel:
> Well, l_in_to_hid would need a ReshapeLayer on top that gets back to (n_batch, n_seq, ...). That one could also be added automatically if precompute_input == True.

It already reshapes back to `(n_batch, n_seq, ...)`: https://github.com/craffel/nntools/blob/recurrent_any_hid_shape/lasagne/layers/recurrent.py#L310

> It should definitely be documented that precompute_input=True requires a different input shape for l_in_to_hid!

Ok, I will document it.

> Everything else should be unaffected (i.e., the output shape of l_in_to_hid and the input and output shapes of l_hid_to_hid should all be equal, and all be the same no matter if precompute_input is set or not). It's not actually required to set the batch size to None, that's just the easy way out.

No, the output shape should actually be different when `precompute_input=True` (assuming the batch dimension is not `None`): `l_in_to_hid` should produce accept inputs and produce outputs with `shape[0] = n_batch*n_seq` when `precompute_input = True` and `shape[0] = n_batch` when `precompute_input = False`, while `l_hid_to_hid` should always accept inputs and produce outputs with `shape[0] = n_batch`.  I will add logic to check for this.

f0k:
> It already reshapes back

Yep, I was just thinking of automatically adding this to the network in the constructor already so it seamlessly passes your `l_in_to_hid.output_shape` checks afterwards.

>  I will add logic to check for this.

That's probably a better solution, yes!

craffel:
> Yep, I was just thinking of automatically adding this to the network in the constructor already so it seamlessly passes your l_in_to_hid.output_shape checks afterwards.

It still wouldn't pass the check - a reshape layer would need to reshape it to `(n_seq, n_batch,  ...)` when `precompute_input = True` while `l_hid_to_hid` should never have a sequence dimension.

I will push the checks soon.

craffel:
Hm, an interesting thing; I implemented the test which requires that `input_to_hidden.output_shape[0] is None` or `input_to_hidden.output_shape[0] == n_batch*n_seq ` when `precompute_input = True`, and it causes a bunch of the tests to fail.  It's arguably overly strict; many Lasagne layers happily ignore when you tell them you're going to give them one batch size and you give them another (as they should):

```
In [1]: import lasagne
In [2]: l_in = lasagne.layers.InputLayer((10, 20))
In [3]: l_dense = lasagne.layers.DenseLayer(l_in, 30)
In [4]: out = lasagne.layers.get_output(l_dense)
In [5]: import theano
In [6]: import numpy as np
In [7]: out.eval({l_in.input_var: np.zeros((10, 20), dtype=theano.config.floatX)}).shape
Out[7]: (10, 30)
In [8]: out.eval({l_in.input_var: np.zeros((100, 20), dtype=theano.config.floatX)}).shape
Out[8]: (100, 30)
```

So, I guess the question is wether to be overly strict and throw errors when we don't necessarily always need to, or just to be careful about documentation so that when a layer is used (e.g. Conv2DLayer with a certain backend) which doesn't allow the batch size to change, the user can understand why the error that comes up is coming up.  Any thoughts?

ebenolson:
I feel like this was discussed in the context of Conv layers a while ago, but I don't recall what was decided. Anyway, my opinion is that strict with good documentation/error messages is better (explicit is better than implicit). A fair amount of work has been done to make `None`function for varying input shapes wherever possible, people should be encouraged to use that properly.

benanne:
Agreed with Eben, it's better to complain early than to crash somewhere further down the line with an obscure unrelated error. Now that I've been getting acquainted with Torch I'm only becoming more convinced that this is a good idea :p

f0k:
+1 for encouraging people to specify shapes correctly including `None`s. I'd still have dense layers work happily with "wrong" batch sizes, because it would mean extra effort for us to check for that. Maybe let's just go the easiest route: if a strict check makes your implementation easier, go for it, otherwise don't?

f0k:
> It's arguably overly strict

Reading that bit again, I think that check is perfectly reasonable. We shouldn't acept a layer of a different shape just because it *might* work anyway!

craffel:
Ok, sounds like the consensus is for the overly strict check.  I will update the PR soon.  As a side note, I'd be curious to know which layers would fail with an unexpected batch size.

f0k:
> As a side note, I'd be curious to know which layers would fail with an unexpected batch size.

I think it's only the legacy convolution, but it's hard to introduce this into a graph these days when you're running on GPU (you'd have to give `THEANO_FLAGS=optimizer_excluding=conv_dnn,conv_gemm`). For CPU, the legacy convolution is the only implementation there is for now, so there you'd always run into trouble.

Ah wait, there's also the noise layers. If the input shape is fully specified in the constructor, it will be passed on to the symbolic random generator, which allows it to pick the correct number of parallel random streams to run. If the run-time shape doesn't match the compile-time shape, it will fail.

lasagne/layers/recurrent.py
```diff
@@ -166,6 +167,22 @@ class CustomRecurrentLayer(MergeLayer):
...
+    >>> n_batch, n_steps, n_channels, width, height = (2, 3, 4, 5, 6)
+    >>> n_out_filters = 7
+    >>> filter_shape = (3, 3)
+    >>> l_in = lasagne.layers.InputLayer(
+    ...     (n_batch, n_steps, n_channels, width, height))
+    >>> l_in_to_hid = lasagne.layers.Conv2DLayer(
+    ...     lasagne.layers.InputLayer((None, n_channels, width, height)),
+    ...     n_out_filters, filter_shape, pad='same')
+    >>> l_hid_to_hid = lasagne.layers.Conv2DLayer(
+    ...     lasagne.layers.InputLayer((None, n_out_filters, width, height)),
```
f0k: What about using `l_in_to_hid.output_shape` instead of `(None, n_out_filters, width, height)` here? That's a bit more flexible (i.e., it will still work with different padding for `l_in_to_hid` or different convolution layer implementations (some don't do `pad='same'` correctly for even filter sizes, they actually do `pad='half'` instead)).

lasagne/layers/recurrent.py
```diff
@@ -213,12 +230,47 @@ def __init__(self, incoming, input_to_hidden, hidden_to_hidden,
...
 
-        # Check that output shapes match
-        if input_to_hidden.output_shape != hidden_to_hidden.output_shape:
+        # Check that the input_to_hidden connection can appropriately handle
+        # a first dimension of input_shape[0]*input_shape[1] when we will
+        # precompute the input dot product
+        if (self.precompute_input and
+                input_to_hidden.output_shape[0] is not None and
+                (input_to_hidden.output_shape[0] !=
+                 input_shape[0]*input_shape[1])):
```
f0k: You also need to add `and input_shape[0] is not None and input_shape[1] is not None`, otherwise you can't do the multiplication and have to skip the early check.

lasagne/layers/recurrent.py
```diff
@@ -213,12 +230,47 @@ def __init__(self, incoming, input_to_hidden, hidden_to_hidden,
...
+                'input_to_hidden.output_shape[0] must equal '
+                'hidden_to_hidden.output_shape[0] but '
+                'input_to_hidden.output_shape[0] = {} and '
+                'hidden_to_hidden.output_shape[0] = {}'.format(
+                    input_to_hidden.output_shape[0],
+                    hidden_to_hidden.output_shape[0]))
+
+        # Check that input_to_hidden and hidden_to_hidden output shapes match
+        if (input_to_hidden.output_shape[1:] !=
+                hidden_to_hidden.output_shape[1:]):
```
f0k: You should also allow `None`s here, i.e., do the following instead:
```python
        if not all(s1 is None or s2 is None or s1 == s2
                   for s1, s2 in zip(input_to_hidden.output_shape[1:],
                                     hidden_to_hidden.output_shape[1:]):
```

lasagne/layers/recurrent.py
```diff
@@ -230,18 +282,19 @@ def __init__(self, incoming, input_to_hidden, hidden_to_hidden,
         # Get the input dimensionality and number of units based on the
         # expected output of the input-to-hidden layer
         self.num_inputs = np.prod(input_shape[2:])
```
f0k: That will break if `input_shape[2:]` contains `None`s. Is `self.num_inputs` needed anywhere?

lasagne/layers/recurrent.py
```diff
@@ -230,18 +282,19 @@ def __init__(self, incoming, input_to_hidden, hidden_to_hidden,
...
-        self.num_units = input_to_hidden.output_shape[-1]
+        self.hid_shape = input_to_hidden.output_shape
```
f0k: Wouldn't `hidden_to_hidden.output_shape` be better? The former depends on `self.precompute_input`, the latter is always the same.

f0k:
Looks good except that `None`s could be in all places, not only the first dimension, and should be handled correctly (see comments). And the convolution example could be made a little simpler by using `l_in_to_hid.output_shape` for `l_hid_to_hid`'s input layer. Almost good for merging now!

lasagne/layers/recurrent.py
craffel: Sure, changed this.

lasagne/layers/recurrent.py
craffel: Good catch!

lasagne/layers/recurrent.py
craffel: So many `None`s!  Fixed.

lasagne/layers/recurrent.py
craffel: I guess not, I think that's cruft from an old way that `self.hid_init` was initialized.  Usually pyflakes warns me about this stuff, but I guess it doesn't mind unused `self` variables (which makes sense).

lasagne/layers/recurrent.py
craffel: Good call.  

> Or maybe set it to input_to_hidden.output_shape[1:]

I like the syntax of `hidden_to_hidden.output_shape` more because it's hard to know what to name "the hidden shape after the first dimension".

craffel:
> Looks good except that Nones could be in all places, not only the first dimension, and should be handled correctly (see comments).

Thanks, fixed and pushed these changes.

>  Almost good for merging now!

Sorry, not quite yet - I realized yesterday that I also should be checking that `input_to_hidden.output_shape[1:] == hidden_to_hidden.input_shape[1:]` (ignoring `None`s) and `hidden_to_hidden.output_shape[1:] == hidden_to_hidden.input_shape[1:]` (which is implicitly tested after we test for `input_to_hidden.output_shape[1:] == hidden_to_hidden.output_shape[1:]` `input_to_hidden.output_shape[1:] == hidden_to_hidden.input_shape[1:]`.)  I will push this test and a test for the test and then I think it may be ready!

craffel:
Ok, pushed one last check!  Hope this covers our bases!

lasagne/layers/recurrent.py
```diff
@@ -213,35 +230,88 @@ def __init__(self, incoming, input_to_hidden, hidden_to_hidden,
...
+                input_shape[0] is not None and
+                input_shape[1] is not None and
+                (input_to_hidden.output_shape[0] !=
+                 input_shape[0]*input_shape[1])):
+            raise ValueError(
+                'When precompute_input == True, '
+                'input_to_hidden.output_shape[0] must equal '
+                'incoming.output_shape[0]*incoming.output_shape[1] '
+                '(i.e. batch_size*sequence_length) or be None but '
+                'input_to_hidden.output_shape = {} and '
```
f0k: You mean `input_to_hidden.output_shape[0] = {}` here!

lasagne/layers/recurrent.py
```diff
@@ -213,35 +230,88 @@ def __init__(self, incoming, input_to_hidden, hidden_to_hidden,
...
-                             "input_to_hidden.output_shape={} and "
-                             "hidden_to_hidden.output_shape={}".format(
+                             "hidden_to_hidden must be equal after the first "
+                             "dimension, but input_to_hidden.output_shape={} "
+                             "and hidden_to_hidden.output_shape={}".format(
                                  input_to_hidden.output_shape,
                                  hidden_to_hidden.output_shape))
 
+        # Check that input_to_hidden's output shape is the same as
+        # hidden_to_hidden's input shape but don't check a dimension of it's
```
f0k: "of it's" -> "if it's"

lasagne/layers/recurrent.py
```diff
@@ -213,35 +230,88 @@ def __init__(self, incoming, input_to_hidden, hidden_to_hidden,
...
 
         # Initialize hidden state
         if isinstance(hid_init, T.TensorVariable):
-            if hid_init.ndim != 2:
+            if hid_init.ndim != len(self.hid_shape):
                 raise ValueError(
                     "When hid_init is provided as a TensorVariable, it should "
-                    "have 2 dimensions and have shape (num_batch, num_units)")
+                    "have the same shape as input_to_hidden.output_shape and "
+                    "hidden_to_hidden.output_shape")
```
f0k: But `input_to_hidden.output_shape` and `hidden_to_hidden.output_shape` are not necessarily equal to begin with, depending on `precompute_input`. I'd just mention `hidden_to_hidden.output_shape` here and also set `self.hid_shape = hidden_to_hidden.output_shape` instead of `self.hid_shape = input_to_hidden.output_shape` (see https://github.com/Lasagne/Lasagne/pull/360#discussion_r36614207, that one seems to have disappeared... actually it seems you already updated the comment, but not the assignment).

lasagne/tests/layers/test_recurrent.py
```diff
@@ -94,15 +94,78 @@ def test_recurrent_init_val_error():
...
+        n_out_filters, filter_shape, pad='same')
+    l_hid_to_hid = lasagne.layers.Conv2DLayer(
+        lasagne.layers.InputLayer((None, n_out_filters, width, height)),
+        n_out_filters, filter_shape, pad='same')
+    l_rec = lasagne.layers.CustomRecurrentLayer(
+        l_in, l_in_to_hid, l_hid_to_hid)
+    out = theano.function([l_in.input_var], lasagne.layers.get_output(l_rec))
+    out_shape = out(np.zeros((n_batch, n_steps, n_channels, width, height),
+                             dtype=theano.config.floatX)).shape
+    assert out_shape == (n_batch, n_steps, n_out_filters, width, height)
```
f0k: In addition, you should check that `l_rec.output_shape` matches the expected shape. And you will see that it doesn't :) (I had overlooked this until now! `get_output_shape_for()` still accesses `self.num_units` instead of `self.hid_shape`. And maybe `self.hid_shape` isn't that useful anyway, maybe `get_output_shape_for()` should just use `self.l_hid_to_hid.output_shape[1:]` directly, and the constructor should use `self.output_shape` or `self.l_hid_to_hid.output_shape`, depending on what's needed.)

lasagne/tests/layers/test_recurrent.py
```diff
@@ -94,15 +94,78 @@ def test_recurrent_init_val_error():
...
+        lasagne.layers.InputLayer((None, n_channels, width + 1, height)),
+        n_out_filters, filter_shape, pad='same')
+    with pytest.raises(ValueError):
+        l_rec = lasagne.layers.CustomRecurrentLayer(
+            l_in, l_in_to_hid, l_hid_to_hid)
+
+    # And, the output shape of input_to_hidden must match the input shape
+    # of hidden_to_hidden past the first dimension.  By not using padding,
+    # the output of l_in_to_hid will be cropped, which will make the
+    # shape inappropriate.
```
f0k: "the output of l_in_to_hid" -> "the output of l_hid_to_hid"

The former wouldn't be a problem at all, the latter is.

f0k:
Thanks again, we're almost there! Only minor things left, except for one major issue: `CustomRecurrentLayer.get_output_shape_for()` has been forgotten :) Doesn't look too difficult, though.

lasagne/layers/recurrent.py
craffel: Good point, thanks, fixed.

lasagne/layers/recurrent.py
craffel: Oops, thanks, fixed.

lasagne/layers/recurrent.py
craffel: >  I'd just mention hidden_to_hidden.output_shape here

Ok, changed that.

> actually it seems you already updated the comment, but not the assignment).

That's bizarre, I remember changing that.  I guess I will remove `hid_shape` anyways based on your comment below.

lasagne/tests/layers/test_recurrent.py
craffel: It's a somewhat unclear comment; the exception being tested for here is testing that the output shape of `l_in_to_hid` is the same as the input shape of `l_hid_to_hid`, it's not testing anything about the output shape of `l_hid_to_hid`.  When both `l_in_to_hid` and `l_hid_to_hid` are using padding, their output shape will be the same (so the test above will pass), but the input shape of `l_hid_to_hid` is different from the output shape of `l_in_to_hid`, which will raise the exception we're testing for.

f0k: Ah, I see! Had mistaken this for testing something else, didn't concentrate enough at 36°C.

lasagne/tests/layers/test_recurrent.py
craffel: Ok, removed `self.hid_shape` and `self.num_units` and added a test.

craffel:
Ok, pushed these changes.  Let me know what you think.

lasagne/layers/recurrent.py
f0k: Hmm, no, it's not!

craffel: Haha, there were two places with this issue!  Hold on, let me fix this one too and rebase.

f0k:
Great, looks good except that this one seems to have gone lost: https://github.com/Lasagne/Lasagne/pull/360#discussion_r36754629
When you fixed this, can you please squash your commits into one? Then Travis can already run over the final commit we'd merge!

craffel:
Squashed.

f0k:
Great work! :+1:
If there are no further comments, I'll merge this tomorrow. Everything looks good to me now.

PR #361, opened 2015-08-06
Transformer

skaae:
Spatial transformer Layer

TODO/questions

* Tests. Added simple tests with https://github.com/skaae/nntools/commit/27faf05049c1889676ac2394d00b1b1677a72724
* I get an error when generating the docs. It seems to be related to objectives.rst??
* I put the TransformerLayer in a new file: transform.py. Do someone have a better suggestion?

### DIFF
--- a/docs/modules/layers.rst
+++ b/docs/modules/layers.rst
@@ -149,6 +149,7 @@
     :nosignatures:
 
     InverseLayer
+    TransformerLayer
 
 
 .. rubric:: :doc:`layers/corrmm`
--- a/docs/modules/layers/special.rst
+++ b/docs/modules/layers/special.rst
@@ -7,3 +7,7 @@ Special-purpose layers
 
 .. autoclass:: InverseLayer
     :members:
+
+.. autoclass:: TransformerLayer
+    :members:
+
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -1,10 +1,12 @@
 import theano
+import theano.tensor as T
 
 from .base import MergeLayer
 
 
 __all__ = [
     "InverseLayer",
+    "TransformerLayer",
 ]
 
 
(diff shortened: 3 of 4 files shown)


skaae:
I cleaned up the docs and added som very simple tests.

f0k:
> I get an error when generating the docs. It seems to be related to objectives.rst??

Oh, my bad. Didn't see that when doing #359, I only ran the tests :-/ Fixed in #364!

skaae:
Thanks. I'll wait for https://github.com/Lasagne/Lasagne/pull/364 to be merged

f0k:
> Thanks. I'll wait for #364 to be merged

Merged! If you want, you can rebase your branch now:
```
git pull upstream master --rebase
git push --force origin
```

skaae:
And no doc errors, except for my own...

skaae:
I think this is ready for review

lasagne/layers/transform.py
```diff
@@ -0,0 +1,200 @@
...
+    Parameters
+    ----------
+    input : a :class:`Layer` instance
+        The input where the affine transformation is applied. This should
+        have convolution format, i.e (num_batch, channels, height, width).
+    localization_network: : a :class:`Layer` instance
+        The network that calculates the parameters of the affine
+        transformation. See the example for how to initialize to the identity
+        transform. Note that all parameters are clipped to lie in the range
+        ``[-1, 1]``, similar to [1]_.
```
f0k: Does that mean you're clipping the values in the code? Wouldn't it be more efficient to just require the `localization_network` to output values in [-1, 1] in the first place?

skaae: Yes you could do that. But if you happen to output values larger than -1,+1 i think you would get indexing errors which are pretty hard to debug.

lasagne/layers/transform.py
```diff
@@ -0,0 +1,200 @@
...
+    def __init__(
+            self, input, localization_network, downsample_factor=1, **kwargs):
+        super(TransformerLayer, self).__init__(
+            [input, localization_network], **kwargs)
+        self.downsample_factor = downsample_factor
+
+        input_shp, loc_shp = self.input_shapes
+
+        if loc_shp[-1] != 6 or len(loc_shp) != 2:
+            raise ValueError("The localization network shoud have "
```
f0k: *must have

lasagne/layers/transform.py
```diff
@@ -0,0 +1,200 @@
...
+                             "outputformat [num_batch, 6]")
```
f0k: *output_shape (batchsize, 6)

(Although I'm not sure if there's any consensus of how we call the first dimension, we could check the convolutional layers.)

lasagne/layers/transform.py
skaae: But i'm very open to suggestions. Not sure if i have done it in the best way.

lasagne/layers/transform.py
```diff
@@ -0,0 +1,200 @@
...
+        if len(input_shp) != 4:
+            raise ValueError('The input network must have 4 dimensions: '
+                             '[num_batch, channels, height width]')
```
f0k: Just checked the convolutional layer. I'd change it to:
```
The input network must have a 4-dimensional output shape:
(batch_size, num_input_channels, input_rows, input_columns)
```

lasagne/layers/transform.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def get_output_shape_for(self, input_shapes):
+        shp = input_shapes[0]
+        dsf = self.downsample_factor
+        return tuple(list(shp[:2]) + [int(s//dsf) for s in shp[2:]])
```
f0k: Easier:
```python
return shp[2:] + tuple(int(s//df) for s in shp[2:])
```

lasagne/layers/transform.py
skaae: Changed

lasagne/layers/transform.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def get_output_for(self, inputs, deterministic=False, **kwargs):
+        # theta should be shape (num_batch, 2, 3)
```
f0k: That's not in line with what you have above. It seems theta should have `(batch_size, 6)`.

skaae: Yeah. Its in old comment i forgot to remove.

lasagne/layers/transform.py
```diff
@@ -0,0 +1,200 @@
...
+
+def _transform(theta, input, downsample_factor):
+    num_batch, num_channels, height, width = input.shape
+    theta = T.reshape(theta, (-1, 2, 3))
+
+    height_f = T.cast(height, 'float32')
+    width_f = T.cast(width, 'float32')
+
+    out_height = T.cast(height_f // downsample_factor, 'int64')
+    out_width = T.cast(width_f // downsample_factor, 'int64')
```
f0k: Why do you cast it to float, then perform an integer division (with `//`) and cast it back? Can't you just do the integer division directly on the integer shapes instead?

skaae: All the casting is to avoid float64.

Oddly

```
    out_height = height // downsample_factor
    out_width = width // downsample_factor
```

results in float64.

lasagne/layers/transform.py
```diff
@@ -0,0 +1,200 @@
...
+    x0 = T.cast(T.floor(x), 'int64')
+    x1 = x0 + 1
+    y0 = T.cast(T.floor(y), 'int64')
+    y1 = y0 + 1
+
+    # note sure if this is nessesary...
+    x0 = T.clip(x0, zero, max_x)
+    x1 = T.clip(x1, zero, max_x)
+    y0 = T.clip(y0, zero, max_y)
+    y1 = T.clip(y1, zero, max_y)
```
f0k: Well, it's only necessary if you don't trust the user to provide a network with a correct range of output values ([see above](https://github.com/Lasagne/Lasagne/pull/361/files#r36521015)). I'm not sure how much the impact on performance is.

lasagne/layers/transform.py
```diff
@@ -0,0 +1,200 @@
...
+
+    # The input is [num_batch, height, width, channels]. We do the lookup in
+    # the flattened input, i.e [num_batch*height*width, channels]. We need
+    # to offset all indices to match the flat version
+    dim2 = width
+    dim1 = width*height
+    base = _repeat(
+        T.arange(num_batch, dtype='int32')*dim1, out_height*out_width)
+    base_y0 = base + y0*dim2
+    base_y1 = base + y1*dim2
```
f0k: Again, this is something that might profit from precomputing in numpy if width and height are fixed at compile time.

skaae: I think without precomputing the network supports changing batch sizes.

lasagne/layers/transform.py
skaae: Yes. One problem with not having the clip is that you can't use a licalization network with linear output as in the paper.

lasagne/layers/transform.py
f0k: Maybe an option `clip_transform=True` that clips transformation parameters when activated (which should be the default), but that can be disabled for improved performance if your network has tanh output, for example.

skaae: I'm not sure, but i think that performance drop from using clipping is negligible?

lasagne/layers/transform.py
f0k:
```
In [1]: import theano
Using gpu device 0: GeForce GT 640 (CNMeM is disabled)

In [2]: x = theano.tensor.matrix()

In [3]: x.shape[0].dtype
Out[3]: 'int64'

In [4]: (x.shape[0] // 2).dtype
Out[4]: 'int64'

In [5]: (x.shape[0] // 1.5).dtype
Out[5]: 'float64'
```

Okay, if `downsample_factor` is not an int, you'd get `float64`. Still it doesn't help to cast to `float32` before, it's enough just to cast to `int64` afterwards!

lasagne/layers/transform.py
f0k: Ah, there's also the batch size involved here... I bet that repetition could be avoided. Hmm. But we can leave performance optimizations for a later PR. For now I'm fine if we can just get rid of some casts!

lasagne/layers/transform.py
skaae: Did a little bit of experiments. If I do

```
    # int dimensions
    out_height = T.cast(height // downsample_factor, 'int64')
    out_width = T.cast(width // downsample_factor, 'int64')
```

Then the division before cast will be float64 which does not happen if you use width_f and hight_f

f0k: > Then the division before cast will be float64 which does not happen if you use width_f and hight_f

That's right, but why is that important? If you cast to `int64` it will end up on the CPU anyway, does it hurt to have `float64` in between then?

skaae: hmm ok. I just ran it with the 'warn_float64' flag and got rid of all float64.

Will it still be copied to cpu if i use int32?

f0k: > Will it still be copied to cpu if i use int32?

Not sure. But I'm also not so sure if it is helpful to have that shape computation on GPU.

If you don't need integers in the end, you could also just cast the symbolic shapes and downsample factor to `floatX`:
```python
height = T.cast(height, theano.config.floatX)
width = T.cast(width, theano.config.floatX)
downsample_factor = np.asarray(downsample_factor, dtype=theano.config.floatX)
```

skaae: I need the iteger height and width when i offset the indices.

f0k: But that's in `_interpolate`, where you compute everything again (it might be nice to avoid that).

> Will it still be copied to cpu if i use int32?

I've just checked, it seems float32 divided by int32 also gives float64. int16 would work. But I think we should probably just do that on CPU and have it copy back the end result?

skaae: I updated interpolate to take out_height and out_width as inputs. Do you suggest that i also reuse width_f and height_f?

lasagne/layers/transform.py
f0k: Could be. Let's just leave that for later!

lasagne/layers/transform.py
f0k: > I updated interpolate to take out_height and out_width as inputs.

Looks nicer!

> Do you suggest that i also reuse width_f and height_f?

Mhm, no, that would probably be overkill.

lasagne/layers/transform.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def _interpolate(im, x, y, out_height, out_width):
+    # *_f are floats
+    num_batch, height, width, channels = im.shape
+    height_f = T.cast(height, 'float32')
+    width_f = T.cast(width, 'float32')
+    zero = T.zeros([], dtype='int64')
+    max_y = T.cast(im.shape[1] - 1, 'int64')
+    max_x = T.cast(im.shape[2] - 1, 'int64')
```
f0k: You don't need to cast `max_y` and `max_x`:
```
In [22]: (x.shape[0] - 1).dtype
Out[22]: 'int64'
```

lasagne/layers/transform.py
```diff
@@ -0,0 +1,198 @@
...
+from .base import MergeLayer
+
+__all__ = [
+    "TransformerLayer",
+]
+
+
+class TransformerLayer(MergeLayer):
+    """
+    Spatial transformer layer [1]_.
```
f0k: The reference shouldn't be part of the first sentence, otherwise you get "reference not found" errors in Sphinx when it's compiling the overview tables for http://lasagne.readthedocs.org/en/latest/modules/layers.html.

It'd actually be good to have a second paragraph of two or three sentences describing what kind of spatial transform this actually performs (i.e., what do the six values mean). That's where you could put the reference.

lasagne/layers/transform.py
f0k: Good point. Please just change the comment to something else then :) Thanks!

skaae:
Just realized that what if have said about the range of the affine parameters was not correct. The parameters are free to be any value. The clips are to ensure that we do not look up illegal indices. 

I removed the references to [-1, +1] from the documetation. ]

lasagne/layers/transform.py
```diff
@@ -0,0 +1,207 @@
...
+                             "output shape [num_batch, 6]")
+        if len(input_shp) != 4:
+            raise ValueError('The input network must have a 4-dimensional '
+                             'output shape:(batch_size, num_input_channels, '
+                             'input_rows, input_columns)')
+
+    def get_output_shape_for(self, input_shapes):
+        shp = input_shapes[0]
+        dsf = self.downsample_factor
+        return shp[:2] + tuple(int(s//dsf) for s in shp[2:])
```
f0k: Just realized that this won't work with Nones. You'd need:
```python
return (shp[:2] + tuple(None if s is None
                        else int(s//dsf) for s in shp[2:]))
```

lasagne/layers/transform.py
```diff
@@ -0,0 +1,207 @@
...
+
+    The layer applies an affine transformation on the input. The affine
+    transformation is parameterized with six learned parameters [1]_.
+    The output is interpolated with a bilinear transformation.
+
+    Parameters
+    ----------
+    input : a :class:`Layer` instance
+        The input where the affine transformation is applied. This should
+        have convolution format, i.e (num_batch, channels, height, width).
```
f0k: A bit picky, but it should be `i.e.,` :)

lasagne/layers/transform.py
f0k: And maybe we should't use integer division if we allow floats for downsample factors?

f0k:
Great, I like your documentation extensions! Just two minor things, and then I'd be all for squashing and merging!

lasagne/layers/transform.py
skaae: For non integer downsample factor i just floor the output shapes. Is that the question?

lasagne/layers/transform.py
skaae: Thanks :)

skaae:
I think this is done if any one else whant to have a look?

I would like to test the code on MNIST before I merge it. I have used the code at it worked fine, but just in case I have messed something up :)

lasagne/layers/transform.py
```diff
@@ -0,0 +1,208 @@
...
+    The layer applies an affine transformation on the input. The affine
+    transformation is parameterized with six learned parameters [1]_.
+    The output is interpolated with a bilinear transformation.
+
+    Parameters
+    ----------
+    input : a :class:`Layer` instance
+        The input where the affine transformation is applied. This should
+        have convolution format, i.e. (num_batch, channels, height, width).
+    localization_network: : a :class:`Layer` instance
```
f0k: There's a colon too much (the one in `localization_network:`). Didn't see that before, sorry. Guess it might confuse Sphinx. Feel free to amend your commit if you have the time, otherwise we can leave that for some general docstring proofreading sometime.

f0k:
Looks good except for one minor thing I've overlooked before, sorry!

And when you're testing again on MNIST, could you try replacing the integer `//` divisions by normal  `/` divisions? It should be the same. The integer divisions suggest that `downsample_factor` would be an integer, which would render the type casting unneeded. So I'd prefer using `/` for the division if possible!

skaae:
I fixed the last docstring error. Using `/` does work, which makes sense. Thanks for the detailed review :) 

I'll get back when i have run the MNIST test. I'll probably add a recipe to reproduce the results in the paper.

f0k:
> I'll get back when i have run the MNIST test.

Cool! Feel free to commit / squash / push so Travis can run already. (Just to be sure, note that there are two occurrences of `//` in the code.)

f0k:
Any news? Can we merge this today, or do you still want to test something?

skaae:
No. Im still in Montreal but i’m going home today. I’ll be able to test it in the weekend.

skaae:
I ran a test on cluttered MNIST and the layer seems to be working fine. I'm fine with merging. 

The experiment is here: 

https://github.com/skaae/Recipes/blob/spatial_transform/examples/spatial_transformer_network.ipynb

The last plot shows that its does zoom as expected.

f0k:
Cool, thanks a lot! Will merge this after #382. Your recipe looks very nice as well.

f0k:
Thinking again, didn't we want to move that to `special.py` (https://github.com/Lasagne/Lasagne/pull/373#issuecomment-130781053)? If so, you'd need to rename `transform.py`, `test_transform.py` and `docs/layers/transform.rst` accordingly, and update the references in `layers/__init__.py` and `docs/layers.rst`. In #373, we've chosen "Special-purpose layers" for the header in `docs/layers/special.rst`: https://github.com/Lasagne/Lasagne/pull/373/files#diff-5e96c229ca5e2f81af27a8a50443dec8R1

Do you want to rename it here (and squash or amend), or shall we merge #373, rebase yours and move it? I guess the former would be less work for you, then I can care about rebasing #373.

skaae:
Ah yes forgot that. Go ahead and merge #373. I can rebase this when you have merged #373.

f0k:
All right, merged it. Rebasing this one should just trigger trivial conflicts in docs/modules/layers.rst and lasagne/layers/__init__.py. I guess it will be easiest to not add `transform` to those two files at all when you solve the conflict, then move over your code from `*transform.*` to `*special.*` and amend your commit.

skaae:
I moved it to special.py . Hopefully Travis passes :)

f0k:
Cool, that was fast! Looks good to me, let's see what Travis and Coveralls think!

f0k:
Merging, thanks a lot!

We can now figure out if there are ways to make it faster if the image shape is known at compile time (which is probably a common use case).

PR #366, opened 2015-08-07
Attribute code to "Lasagne contributors"

f0k:
Changes attributions in code, documentation and license to include all Lasagne contributors, not only (or not primarily) Sander. He's still explicitly named as the one who started the project, though!

Copied some sections about copyright and contribution agreement from Caffe's license file. Not sure if we need to include Caffe's license file now because Caffe's license text is covered by Caffe's license text? %-)

I've chosen our google group as the email address to contact the authors under. Seemed like the most reasonable option. (Theano has its *theano-dev* group there, but we don't have nor need one. Blocks and Fuel don't give an email address at all, which would also be an option.)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,8 @@ tests_require = [
 setup(
     name="Lasagne",
     version=version,
-    description="neural network tools for Theano",
+    description="A lightweight library to build and train neural networks "
+                "in Theano",
     long_description="\n\n".join([README, CHANGES]),
     classifiers=[
         "Development Status :: 3 - Alpha",
@@ -40,8 +41,8 @@ setup(
         "Topic :: Scientific/Engineering :: Artificial Intelligence",
         ],
     keywords="",
-    author="Sander Dieleman",
-    author_email="sanderdieleman@gmail.com",
+    author="Lasagne contributors",
+    author_email="lasagne-users@googlegroups.com",
     url="https://github.com/Lasagne/Lasagne",
     license="MIT",
     packages=find_packages(),
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,13 @@
 The MIT License (MIT)
 
-Copyright (c) 2014 Sander Dieleman
+Copyright (c) 2014-2015 Lasagne contributors
+
+Lasagne uses a shared copyright model: each contributor holds copyright over
+their contributions to Lasagne. The project versioning records all such
+contribution and copyright details.
+By contributing to the Lasagne repository through pull-request, comment,
+or otherwise, the contributor releases their content to the license and
+copyright terms herein.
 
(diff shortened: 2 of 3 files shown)


moscow25:
Great. So there will be a standard way to cite in publication?

benanne:
Looks good!

setup.py
```diff
@@ -27,7 +27,8 @@
 setup(
     name="Lasagne",
     version=version,
-    description="neural network tools for Theano",
+    description="A lightweight library to build and train neural networks "
+                "in Theano",
     long_description="\n\n".join([README, CHANGES]),
     classifiers=[
         "Development Status :: 3 - Alpha",
```
benanne: Maybe this needs an update as well? :p

benanne:
> Great. So there will be a standard way to cite in publication?

Yep, as soon as the library is released we'll obtain a DOI, which makes the repository citable.

skaae:
Im not sure this is related but the Core-team link in https://github.com/orgs/Lasagne/teams/core-team does not work?

f0k:
> Im not sure this is related but the Core-team link [...] does not work?

It works for me... what does it look like for you? Is there an error message?

skaae:
I get a 404 for this link: https://github.com/orgs/Lasagne/teams/core-team

f0k:
> I get a 404

Bummer, teams are only visible to organization owners and team members ("secret" teams) or to all organization members ("closed" teams), but not to everybody: https://developer.github.com/changes/2015-06-24-api-enhancements-for-working-with-organization-permissions/#team-privacy

That's not a useful link to give to users that may not even have a github account. We can only explicitly list the names instead then, I guess. Not sure if that's a good spot for that, though.

ebenolson:
What about linking to this page instead: https://github.com/Lasagne/Lasagne/graphs/contributors

I can still see it when signed out, plus it lists everyone who's ever had a commit merged.

f0k:
> What about linking to this page instead

That one is linked as well, see http://lasagne.readthedocs.org/en/latest/user/development.html
Shall we just not list the core team (including Soren) at all or list them by name somewhere else (either in that file or in CHANGES.txt or CONRIBUTORS.txt or something)?

benanne:
I think there should be a public list somewhere at least.

f0k:
> I think there should be a public list somewhere at least.

Okay, can we leave that for another PR then? Most of the ["Development" page](http://lasagne.readthedocs.org/en/latest/user/development.html) is badly-written anyway, so it could be addressed as part of rewriting this.

benanne:
Sure.

setup.py
f0k: Don't know. If at all, I'd change that along with the version. We'd need to check if there are any definitions/recommendations on when to use which status.

PR #369, opened 2015-08-12
Implemented 2D upscaling layer

Britefury:
Implemented Upscale2DLayer, a layer that performs in inverse operation of pooling layers.
Useful in segmentation approaches such as [1], [2] and [3].

The layer was added to pool.py. Tests are included.

[1] U-Net: Convolutional Networks for Biomedical Image Segmentation, http://arxiv.org/abs/1505.04597
[2] Hypercolumns for Object Segmentation and Fine-grained Localization, http://arxiv.org/abs/1411.5752
[3] Image Segmentation with Cascaded Hierarchical Models and Logistic
Disjunctive Normal Networks, http://www.cv-foundation.org/openaccess/content_iccv_2013/papers/Seyedhosseini_Image_Segmentation_with_2013_ICCV_paper.pdf

### DIFF
--- a/lasagne/layers/pool.py
+++ b/lasagne/layers/pool.py
@@ -10,6 +10,7 @@ __all__ = [
     "MaxPool1DLayer",
     "MaxPool2DLayer",
     "Pool2DLayer",
+    "Upscale2DLayer",
     "FeaturePoolLayer",
     "FeatureWTALayer",
     "GlobalPoolLayer",
@@ -289,6 +290,54 @@ class MaxPool2DLayer(Pool2DLayer):
 
 # TODO: add reshape-based implementation to MaxPool*DLayer
 # TODO: add MaxPool3DLayer
+
+
+class Upscale2DLayer(Layer):
+    """
+    2D upscaling layer layer
+
+    Performs 2D upscaling over the two trailing axes of a 4D input tensor.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or tuple
+        The layer feeding into this layer, or the expected input shape.
+
+    scale_factor : integer or iterable
+        The scale factor in each dimension. If an integer, it is promoted to
+        a square scale factor region. If an iterable, it should have two
+        elements.
+
+    **kwargs
+        Any additional keyword arguments are passed to the :class:`Layer`
+        superclass.
+    """
+
+    def __init__(self, incoming, scale_factor, **kwargs):
+        super(Upscale2DLayer, self).__init__(incoming, **kwargs)
+
+        self.scale_factor = as_tuple(scale_factor, 2)
+
+        if self.scale_factor[0] < 1 or self.scale_factor[1] < 1:
+            raise ValueError('Scale factor must be >= 1, not {0}'.format(
+                self.scale_factor))
+
+    def get_output_shape_for(self, input_shape):
+        output_shape = list(input_shape)  # copy / convert to mutable list
+        if output_shape[2] is not None:
+            output_shape[2] *= self.scale_factor[0]
+        if output_shape[3] is not None:
+            output_shape[3] *= self.scale_factor[1]
+        return tuple(output_shape)
+
+    def get_output_for(self, input, **kwargs):
+        a, b = self.scale_factor
+        upscaled = input
+        if b > 1:
+            upscaled = T.extra_ops.repeat(upscaled, b, 3)
+        if a > 1:
+            upscaled = T.extra_ops.repeat(upscaled, a, 2)
+        return upscaled
 
 
 class FeaturePoolLayer(Layer):
--- a/lasagne/tests/layers/test_pool.py
+++ b/lasagne/tests/layers/test_pool.py
@@ -70,6 +70,19 @@ def max_pool_2d_ignoreborder(data, pool_size, stride, pad):
     return data_pooled
 
 
+def upscale_2d_shape(shape, scale_factor):
+    return (shape[0], shape[1],
+            shape[2] * scale_factor[0], shape[3] * scale_factor[1])
+
+
+def upscale_2d(data, scale_factor):
+    upscaled = np.zeros(upscale_2d_shape(data.shape, scale_factor))
+    for j in range(scale_factor[0]):
+        for i in range(scale_factor[1]):
+            upscaled[:, :, j::scale_factor[0], i::scale_factor[1]] = data
+    return upscaled
+
+
 class TestFeaturePoolLayer:
     def pool_test_sets():
         for pool_size in [2, 3]:
@@ -467,6 +480,67 @@ class TestMaxPool2DNNLayer:
         except NotImplementedError:
             raise
         #    pytest.skip()
+
+
+class TestUpscale2DLayer:
+    def scale_factor_test_sets():
+        for scale_factor in [2, 3]:
+                yield scale_factor
+
+    def input_layer(self, output_shape):
+        return Mock(output_shape=output_shape)
+
+    def layer(self, input_layer, scale_factor):
+        from lasagne.layers.pool import Upscale2DLayer
+        return Upscale2DLayer(
+            input_layer,
+            scale_factor=scale_factor,
+        )
+
+    def test_invalid_scale_factor(self):
+        from lasagne.layers.pool import Upscale2DLayer
+        inlayer = self.input_layer((128, 3, 32, 32))
+        with pytest.raises(ValueError):
+            Upscale2DLayer(inlayer, scale_factor=0)
+        with pytest.raises(ValueError):
+            Upscale2DLayer(inlayer, scale_factor=-1)
+        with pytest.raises(ValueError):
+            Upscale2DLayer(inlayer, scale_factor=(0, 2))
+        with pytest.raises(ValueError):
+            Upscale2DLayer(inlayer, scale_factor=(2, 0))
+
+    @pytest.mark.parametrize(
+        "scale_factor", list(scale_factor_test_sets()))
+    def test_get_output_for(self, scale_factor):
+        input = floatX(np.random.randn(8, 16, 17, 13))
+        input_layer = self.input_layer(input.shape)
+        input_theano = theano.shared(input)
+        result = self.layer(
+            input_layer,
+            (scale_factor, scale_factor),
+        ).get_output_for(input_theano)
+
+        result_eval = result.eval()
+        numpy_result = upscale_2d(input, (scale_factor, scale_factor))
+
+        assert np.all(numpy_result.shape == result_eval.shape)
+        assert np.allclose(result_eval, numpy_result)
+
+    @pytest.mark.parametrize(
+        "input_shape,output_shape",
+        [((32, 64, 24, 24), (32, 64, 48, 48)),
+         ((None, 64, 24, 24), (None, 64, 48, 48)),
+         ((32, None, 24, 24), (32, None, 48, 48)),
+         ((32, 64, None, 24), (32, 64, None, 48)),
+         ((32, 64, 24, None), (32, 64, 48, None)),
+         ((32, 64, None, None), (32, 64, None, None))],
+    )
+    def test_get_output_shape_for(self, input_shape, output_shape):
+        input_layer = self.input_layer(input_shape)
+        layer = self.layer(input_layer,
+                           scale_factor=(2, 2))
+        assert layer.get_output_shape_for(
+            input_shape) == output_shape
 
 
 class TestFeatureWTALayer(object):
--- a/docs/modules/layers.rst
+++ b/docs/modules/layers.rst
@@ -80,6 +80,7 @@
     MaxPool1DLayer
     MaxPool2DLayer
     Pool2DLayer
+    Upscale2DLayer
     GlobalPoolLayer
     FeaturePoolLayer
     FeatureWTALayer
--- a/docs/modules/layers/pool.rst
+++ b/docs/modules/layers/pool.rst
@@ -14,6 +14,9 @@ Pooling layers
 .. autoclass:: Pool2DLayer
     :members:
 
+.. autoclass:: Upscale2DLayer
+    :members:
+
 .. autoclass:: GlobalPoolLayer
     :members:
 


benanne:
Thanks! What type of upscaling does this do? There are many different approaches being used in literature (nearest neighbour, bed of nails, ...) so the naming should probably be a bit more specific.

It looks like this is just doing nearest neighbour, right? In that case it might be more efficient to implement it using `T.tile` and `T.reshape`, to avoid the repeated `set_subtensor` calls.

f0k:
> There are many different approaches being used in literature (nearest neighbour, bed of nails, ...) so the naming should probably be a bit more specific.

Maybe it should just get a `mode` kwarg if we ever extend it to other variants. After all, everything will be the same except for `get_output_for()`, so we don't really need separate classes.

> In that case it might be more efficient to implement it using T.tile and T.reshape, to avoid the repeated set_subtensor calls.

That should be profiled before merging, otherwise we might forget. I'd prefer to provide more or less optimal implementations in the library -- one of the points of a library is to centralize the effort of finding good implementations :) Geoffrey, can you do this?

----

On a related note, a colleague and me have implemented a general `Unlayer` that inverts the operation of a given Layer using its gradient. We could add that as well. For pooling, it would use the max-pooling switches from the original pooling layer for the reconstruction (i.e., it's only useable in an auto-encoder setting, where the Layer to invert comes before the Unlayer inverting it).

benanne:
Cool, `Unlayer` sounds like it would be a very useful addition as well :)

Britefury:
@f0k: I have to admit I am not certain as to how to profile Theano code. I have access to machines with GPUs, its just a matter of the best approach. I could attempt it if you advise me.

JeffreyDF:
You can do profiling by using 

import os
theano.config.profile = True
os.environ['CUDA_LAUNCH_BLOCKING'] = '1'

in your main file.

Then, when the script is finished, or you interrupt it, it should print a profile output at the end. See also the [profiling page in the Theano tutorial](http://deeplearning.net/software/theano/tutorial/profiling.html). (Jan knows a lot more about this so hopefully he'll correct me if I'm missing something.)

f0k:
What Jeffrey suggested works fine, I just tend to do that on the command line instead (`THEANO_FLAGS=profile=1 CUDA_LAUNCH_BLOCKING=1 python yourscript.py`).

Ideally, you'd prepare a script that just applies that layer to some data for 10 or 100 times, then you can play around with the implementation. Timing the backward pass is a little tricky, but doable as well. Something like this should work (based on [a mailing list post of mine](https://groups.google.com/d/msg/lasagne-users/t_rMTLAtpZo/P_rGw85QEgAJ)):

```python
import sys
import numpy as np
import theano
import theano.tensor as T
from theano.sandbox.cuda import as_cuda_ndarray_variable
import lasagne

scale_factor = map(int, sys.argv[1].split('x'))
input_shape = (64, 3, 32, 32)
layer = lasagne.layers.Upscale2DLayer(input_shape, scale_factor)
output_shape = layer.output_shape

x = T.tensor4()
y = layer.get_output_for(x)
gy = T.tensor4()
gx = theano.grad(None, wrt=x, known_grads={y: gy})

X = np.random.randn(*input_shape).astype(theano.config.floatX)
Y = np.random.randn(*output_shape).astype(theano.config.floatX)
fn_fwd = theano.function([], as_cuda_ndarray_variable(y), givens={x: X})
fn_bwd = theano.function([], as_cuda_ndarray_variable(gx), givens={x: X, gy: Y},
                         on_unused_input='ignore')

fn = fn_fwd if sys.argv[2] == 'fwd' else fn_bwd
for _ in range(10):
    fn()
```

Now you can launch this with a scale factor (e.g., 2x2, or 3x3, or 1x5) as the first argument and "fwd" or "bwd" as the second argument, along with the environment variables I've mentioned (or by further instructing the code as Jeffrey suggested). (It's untested, though, let me know if you can't get it to run.)

Britefury:
@benanne Yes, its only doing nearest neighbour. I couldn't see how to achieve the desired result with `T.tile` but `T.extra_ops.repeat` does the trick.

@f0k I have implemented both approaches and benchmarked them with the code you provided. It seems that `repeat` is much faster on the forward pass, but `set_subtensor` may be faster on the backward pass. To be honest I am not entirely confident in my results, so I think I need to re-do some of my experiments. As a consequence, the latest push contains both approaches, with a class attribute selecting which one. The `repeat` based approach is the default.
I have added `Upscale2DLayer` to the docs.

benanne:
I'm surprised to hear that the backward pass of `set_subtensor` is actually faster. I think we need to do some careful timing of both approaches for different input dimensions and settle on one of them. Supporting both needlessly complicates the code.

Britefury:
I've been looking further into profiling.
My initial results didn't have enough runs to smooth out inconsistencies. I was also reading off the 'Time in thunks' value. Is this what we are interested in? Or would I want to ignore certain ops such as GpuAlloc?

Britefury:
@benanne @f0k Would you mind very much if I just removed the `set_subtensor` based upscaling approach and settled on using `T.repeat`; it seems more likely that this could be an optimised path going into the future. I have done this in the latest commit.

f0k:
> I was also reading off the 'Time in thunks' value. Is this what we are interested in? Or would I want to ignore certain ops such as GpuAlloc?

The script I provided is set up such that it shouldn't include any host/device transfers (since inputs are on the GPU and outputs stay on the GPU), so yes, it should be fine to compare the total time (including GpuAllocs and such).

> Would you mind very much if I just removed the `set_subtensor` based upscaling approach and settled on using `T.repeat`

If that seems to be the faster one, I'm fine! :) At least the code is very readable.

lasagne/layers/pool.py
```diff
@@ -291,6 +292,51 @@ def __init__(self, incoming, pool_size, stride=None,
 # TODO: add MaxPool3DLayer
 
 
+class Upscale2DLayer (Layer):
```
f0k: For consistency, there shouldn't be a space before `(Layer)`.

lasagne/layers/pool.py
```diff
@@ -291,6 +292,51 @@ def __init__(self, incoming, pool_size, stride=None,
...
+
+    Performs 2D upscaling over the two trailing axes of a 4D input tensor.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or tuple
+        The layer feeding into this layer, or the expected input shape.
+
+    scale_factor : integer or iterable
+        The scale factor in each dimensions. If an integer, it is promoted to
```
f0k: *in each dimension (not "dimensions")

lasagne/layers/pool.py
```diff
@@ -291,6 +292,51 @@ def __init__(self, incoming, pool_size, stride=None,
...
+        super(Upscale2DLayer, self).__init__(incoming, **kwargs)
+
+        self.scale_factor = as_tuple(scale_factor, 2)
+
+    def get_output_shape_for(self, input_shape):
+        output_shape = list(input_shape)  # copy / convert to mutable list
+        output_shape[2] = input_shape[2] * self.scale_factor[0] \
+            if input_shape[2] is not None else None
+        output_shape[3] = input_shape[3] * self.scale_factor[1] \
+            if input_shape[3] is not None else None
```
f0k: If `input_shape[2] is None`, then `output_shape[2] is None` already, no need to set it again. So you can simplify it to:
```
if output_shape[2] is not None:
    output_shape[2] *= self.scale_factor[0]
if output_shape[3] is not None:
    output_shape[3] *= self.scale_factor[1]
return tuple(output_shape)
```

lasagne/layers/pool.py
```diff
@@ -291,6 +292,51 @@ def __init__(self, incoming, pool_size, stride=None,
...
+
+    **kwargs
+        Any additional keyword arguments are passed to the :class:`Layer`
+        superclass.
+    """
+
+    def __init__(self, incoming, scale_factor, **kwargs):
+        super(Upscale2DLayer, self).__init__(incoming, **kwargs)
+
+        self.scale_factor = as_tuple(scale_factor, 2)
```
f0k: Possibly add a check that both factors are >= 1 (things will get funny if they aren't).

Britefury:
@f0k I've fixed the issues you mentioned.

I haven't done any more benchmarking since my last inconclusive benchmarking session, the results of which can be found in this [Google Spreadsheet](https://docs.google.com/spreadsheets/d/1SnIfxyNv1oeqX1WmhY7SYL8PQL6twA-RPv0Sxue1TmI/edit?usp=sharing). The benchmarking was performed on an nVidia 980. I hope the results are clear. A brief explanation just in case:
- each coloured block represents results, one cell for each scale factor, row by column
- each cell contains thunk time in seconds
- a group of three coloured blocks is top block: forward pass, middle block: backward pass, bottom block: forward + backward pass added together
- the top 3 blocks use `set_subtensor`, the bottom 3 `repeat`.

I am not confident in the quality of the results since they seem susceptible to noise. This is demonstrated by the differences in values for scale factor of e.g. 2x1 and 1x2; they should be very close but often are not. Either the difference in memory striding is making a huge difference or noise is playing a very large factor. Given that results can change if the exact same experiment is run a few times, I think noise is a problem here. I think that in order to get more accurate results, I need to increase the number of repetitions from 1,000 to >10,000. Early tests seem to indicate that this can improve matters.

f0k:
Interesting. It really seems the backward pass is the culprit. And it seems `set_subtensor` would win for small scale factors (up to 2x2, 2x3, 3x2). But that's something that should be solved by Theano, and I agree that `T.repeat` is something that could be optimized upstream more readily than a `set_subtensor`-based approach. Thank you for the detailed report!

Finally, can you please squash this one as well?
```bash
git reset --soft 706636e
git commit --amend --no-edit
git push --force origin upscale-layer
```

Thanks for the great work!

Britefury:
Okay, done. :)
Pleasure to contribute :)

lasagne/tests/layers/test_pool.py
```diff
@@ -469,6 +482,75 @@ def test_get_output_shape_for(self, input_shape, output_shape):
...
 
 
+class TestUpscale2DLayer:
+    def scale_factor_test_sets():
+        for scale_factor in [2, 3]:
+                yield scale_factor
+
+    def input_layer(self, output_shape):
+        return Mock(get_output_shape=lambda: output_shape,
+                    output_shape=output_shape)
```
f0k: Please remove the `get_output_shape` argument and only keep the `output_shape` one, I've gone rid of the last remnants in 25fd8c3! (No blame on you, you started your branch a day before that.)

lasagne/tests/layers/test_pool.py
```diff
@@ -469,6 +482,75 @@ def test_get_output_shape_for(self, input_shape, output_shape):
...
+                (scale_factor, scale_factor),
+            ).get_output_for(input_theano)
+
+            result_eval = result.eval()
+            numpy_result = upscale_2d(input, (scale_factor, scale_factor))
+
+            assert np.all(numpy_result.shape == result_eval.shape)
+            assert np.allclose(result_eval, numpy_result)
+        except NotImplementedError:
+            pytest.skip()
```
f0k: No need for that `try`/`except` here, you don't raise `NotImplementedError` anywhere. This was just needed to cater for different capabilities of the pooling layer implementations.

lasagne/tests/layers/test_pool.py
```diff
@@ -469,6 +482,75 @@ def test_get_output_shape_for(self, input_shape, output_shape):
...
+    )
+    def test_get_output_shape_for(self, input_shape, output_shape):
+        try:
+            input_layer = self.input_layer(input_shape)
+            layer = self.layer(input_layer,
+                               scale_factor=(2, 2))
+            assert layer.get_output_shape_for(
+                input_shape) == output_shape
+        except NotImplementedError:
+            pytest.skip()
```
f0k: Again, no need for `try`/`except` here.

lasagne/tests/layers/test_pool.py
```diff
@@ -469,6 +482,75 @@ def test_get_output_shape_for(self, input_shape, output_shape):
...
+    def layer(self, input_layer, scale_factor):
+        from lasagne.layers.pool import Upscale2DLayer
+        return Upscale2DLayer(
+            input_layer,
+            scale_factor=scale_factor,
+        )
+
+    def test_invalid_scale_factor(self):
+        from lasagne.layers.pool import Upscale2DLayer
+        from mock import Mock
```
f0k: This import is unused, please remove! (Also it's redundant, `Mock` is imported at the top of the file.)

lasagne/layers/pool.py
```diff
@@ -291,6 +292,55 @@ def __init__(self, incoming, pool_size, stride=None,
...
+        return tuple(output_shape)
+
+    def get_output_for(self, input, **kwargs):
+        y = self.scale_factor[0]
+        x = self.scale_factor[1]
+        upscaled = input
+        if x > 1:
+            upscaled = T.extra_ops.repeat(upscaled, x, 3)
+        if y > 1:
+            upscaled = T.extra_ops.repeat(upscaled, y, 2)
```
f0k: That's a bit confusing. At first glance, it looks like the two scale factors are applied to the two dimensions in the wrong order. Why not:
```python
    x, y = self.scale_factor
    upscaled = input
    if x > 1:
        upscaled = T.extra_ops.repeat(upscaled, x, 2)
    if y > 1:
        upscaled = T.extra_ops.repeat(upscaled, y, 3)
```
If you're worried about a discrepancy with image height and width, call the factors `f1` and `f2` or `a` and `b`.

f0k:
Sorry, found some more things on re-reading the code! After fixing and before committing, you can just do the following to amend the single commit you have:
```bash
git commit -a --amend --no-edit
git push --force origin upscale-layer
```
Cheers!

Britefury:
Done.

Britefury:
@f0k the build failed due to http://deeplearning.net not being available for the MNIST example during the Python 2.7 test.

f0k:
Seems to be up again. I'll close and reopen the PR to rerun Travis.

f0k:
Merged, thanks again!

PR #373, opened 2015-08-13
added UnLayer to base.py and tests

neosatrapahereje:


docs/modules/layers.rst
```diff
@@ -43,7 +43,7 @@
 
     Layer
     MergeLayer
-
+    UnLayer
```
f0k: You removed an empty line here, please add it back for consistency!

lasagne/layers/base.py
```diff
@@ -1,17 +1,21 @@
...
     "Layer",
     "MergeLayer",
+    "UnLayer",
 ]
 
 
 # Layer base class
 
 class Layer(object):
+
```
f0k: Please don't insert empty lines where they haven't been any. Either configure your editor to not perform style changes, or remove them before committing.

lasagne/layers/base.py
```diff
@@ -300,3 +307,50 @@ def get_output_for(self, inputs, **kwargs):
         `NotImplementedError`.
         """
         raise NotImplementedError
+
+
+class UnLayer(MergeLayer):
+
+    """
+    lasagne.layers.UnLayer(incoming, layer, **kwargs)
```
f0k: You can omit the signature here. It's only needed where the automatically generated signature doesn't look good because it contains objects (such as weight initializers). Also please remove the empty line before the docstring, just to be consistent with what we have elsewhere.

lasagne/layers/base.py
```diff
@@ -300,3 +307,50 @@ def get_output_for(self, inputs, **kwargs):
...
+
+    The :class:`UnLayer` class performing inverse operations
+    for a single layer of a neural network: transposed layer
+    for a :class:`DenseLayer`, deconvolutional layer for
+    :class:`Conv2DLayer`, :class:`Conv1DLayer`; or
+    an unpooling layer for :class:`MaxPool2DLayer`.
```
f0k: "performing" -> "performs"

Also please add a sentence on *how* it does this. Something like: "It does so by applying the partial derivative of the layer to be inverted with respect to its input."

lasagne/layers/base.py
```diff
@@ -300,3 +307,50 @@ def get_output_for(self, inputs, **kwargs):
...
+
+    It is specially useful for building (convolutional)
+    autoencoders with tied parameters.
```
f0k: I'd add something like: "Note that if the layer to be inverted contains a nonlinearity and/or a bias, the :class:`UnLayer` will include the derivative of that in its computation. If you want to avoid that, each encoder layer should have its nonlinearities and biases moved into a :class:`NonlinearityLayer`." (That will require adding a bias to `NonlinearityLayer`.)

lasagne/layers/base.py
```diff
@@ -300,3 +307,50 @@ def get_output_for(self, inputs, **kwargs):
...
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance
+        The layer feeding into this layer.
```
f0k: That's incomplete. Please copy from any of the other layers (e.g., from `Layer` above).

lasagne/layers/base.py
```diff
@@ -300,3 +307,50 @@ def get_output_for(self, inputs, **kwargs):
...
+    layer : a :class:`Layer` instance
+        The layer with respect to which the instance of the
+        UnLayer is inverse to.
+
+
```
f0k: There's one newline too much here.

lasagne/layers/base.py
```diff
@@ -300,3 +307,50 @@ def get_output_for(self, inputs, **kwargs):
...
+    Examples
+    --------
+    >>> import lasagne
+    >>> from lasagne.layers import InputLayer, Conv2DLayer, DenseLayer, UnLayer
+    >>> l_in = InputLayer((100, 3, 28, 28))
+    >>> l1 = Conv2DLayer(l_in, num_filters=16, filter_size=5)
+    >>> l2 = DenseLayer(l1, num_units=20)
+    >>> l_u = UnLayer(l2, l1)  # As Deconv2DLayer
+
```
f0k: A newline too much.

lasagne/layers/base.py
```diff
@@ -300,3 +307,50 @@ def get_output_for(self, inputs, **kwargs):
...
+    """
+
+    def __init__(self, incoming, layer, **kwargs):
+
```
f0k: A newline too much.

lasagne/tests/layers/test_base.py
```diff
@@ -5,6 +5,7 @@
 
 
 class TestLayer:
+
```
f0k: Again, please remove all newlines automatically added by your editor.

lasagne/tests/layers/test_base.py
```diff
@@ -145,3 +147,67 @@ def test_get_output_shape_for_notimplemented(self, layer):
...
+        nonlinearity = identity
+
+        shape = (10, 12)
+        l_in = InputLayer(shape=shape)
+
+        layer = DenseLayer(
+            l_in,
+            num_units=3,
+            b=b,
+            nonlinearity=nonlinearity,
```
f0k: I'd directly write `b=None` and `nonlinearity=identity` here, as that's very essential for your `test_get_output_for()`.

lasagne/tests/layers/test_base.py
```diff
@@ -145,3 +147,67 @@ def test_get_output_shape_for_notimplemented(self, layer):
...
+        assert unlayer.get_output_shape_for(
+            [(34, 55, 89, 144), (5, 8, 13, 21), (1, 1, 2, 3)]) == (1, 1, 2, 3)
+
+    def test_get_output_for(self, unlayer_vars):
+        from lasagne.layers.helper import get_output
+        unlayer = unlayer_vars['unlayer']
+        layer = unlayer_vars['layer']
+        W = layer.W.get_value()
+
+        input = theano.shared(numpy.ones((10, 12)))
```
f0k: `numpy.random.randn(*layer.input_shape)` would be nicer!

f0k:
Thanks for contributing our UnLayer! :) Got a couple of comments, but let's go for lunch first.

For others: I wasn't sure where to suggest to put this... it doesn't fit any of the existing categories. It's also not really a base layer because nobody is supposed to subclass it (or at least not more than a DenseLayer or something). We could open up a `lasagne/layers/special.py` for things that don't fit anywhere else and won't attract a bunch of layers to accompany them. Or we could put that one into `lasagne/layers/autoenc.py` in case we'd also add a simple `TransposedDenseLayer` and some helper functions for turning a network into an autoencoder.

benanne:
Very cool, thanks for contributing! I know name discussions are getting old, but maybe we should think about a more descriptive name for this as well? `UnLayer` sorta makes sense if you already know what it's supposed to be doing, but by itself it's a little vague.

f0k:
Would `UndoLayer` be better? A more accurate name would be `GradientLayer` or something, but then it's not so clear what it's useful for. Other possibilities are `ReverseLayer` or `BackwardLayer`, but they could be confused with doing something to time sequences.

What about the location? Shall we start a `lasagne/layers/autoenc.py` for that?

ebenolson:
`InverseLayer`?

benanne:
I'm okay with putting this in `autoenc.py`, although I suppose it has other use cases besides autoencoders, so maybe this name is even a bit restrictive :) You could do next-frame prediction in video with an architecture that uses deconv / depool layers for example.

I like `InverseLayer`.

f0k:
> although I suppose it has other use cases besides autoencoders

Hmm, but note that this layer can only be used to unpool / deconvolve if you've also got a layer that pools / convolves something of the shape you want to output. So it's a bit trickier to use outside of a encoder/decoder architecture, although it should be possible in general. (It will get horribly inefficient if the layers you want to invert are neither part of the forward pass nor their underlying Theano ops all support shape inference, but it would still work.)

I mainly like `autoenc.py` because, off the top of my head, that's the only category that we could put it into that might attract additional layers and macro functions. We could also put it in `special.py` and move it somewhere else as soon as it makes sense. `special.py` could be a container for all layers that would stay alone in any more specific category for the foreseeable future.

benanne:
Right, I neglected that this does weight sharing by definition. All in all I'm fine with `autoenc.py`. It's something we can change more easily since people are supposed to use the `lasagne.layers` namespace anyway.

ebenolson:
> We could also put it in special.py and move it somewhere else as soon as it makes sense. special.py could be a container for all layers that would stay alone in any more specific category for the foreseeable future.

I think `autoenc.py` is fine, but I like this idea - PReLU and I think some other things also need a home at some point.

skaae:
I put the TransformerLayer, https://github.com/Lasagne/Lasagne/pull/361,  in transform.py, a bit overkill as i don't know any other layers that would fit in that category?

Maybe `special.py` could host both layers?

f0k:
Okay, we've moved it to a new `special.py` now (Carlos sits two rooms from mine, if anybody is confused). We can still move it to `autoenc.py` once there's a use for it. And he renamed it to `InverseLayer`, and we squashed the commits.

I'm fine with merging. (The only question left is whether the shared variable in the last test should better be np.rand().astype(theano.config.floatX) instead. It will now always run on the CPU.)

f0k:
Merging since there were no more comments and other PRs want to add something to `special.py` as well.

benanne:
Sweet! I think it would be useful to add a runnable example for this layer at some point.

f0k:
My colleague has some helper methods for turning a network into an auto-encoder. After #378, these could be added as well, and then we could have a Recipe for a convolutional AE.

benanne:
Sweet :)

PR #375, opened 2015-08-13
Prepare for first release

f0k:
This adds a changelog as suggested in https://github.com/Lasagne/Lasagne/issues/74#issuecomment-130331059, updates `requirements.txt` to a Theano version supporting cuDNN v3 as suggested in https://github.com/Lasagne/Lasagne/issues/74#issuecomment-130390810, and changes the version number from `0.1.dev` to `0.1`.

It assumes we're going to tag the release as `v0.1` (that's how numpy and blocks do it, for example), not as `0.1` (that's how, hmm... some projects do it like this). I'm fine with either way. @dnouri might have a preference?

It would be nice to have a DOI badge in the README of the first release, but I'm not sure if it's possible to add the DOI badge before we tag the release, which we need to do in order to get a DOI. I'll investigate.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,0 +1,35 @@
+Changelog
+---------
+
+0.1 (2015-08-13)
+~~~~~~~~~~~~~~~~
+
+First release.
+
+* core contributors, in alphabetical order:
+
+  * Eric Battenberg (@ebattenberg)
+  * Sander Dieleman (@benanne)
+  * Daniel Nouri (@dnouri)
+  * Eben Olson (@ebenolson)
+  * Aäron van den Oord (@avdnoord)
+  * Colin Raffel (@craffel)
+  * Jan Schlüter (@f0k)
+  * Søren Kaae Sønderby (@skaae)
+
+* extra contributors, in chronological order:
+
+  * Daniel Maturana (@dimatura): documentation, cuDNN layers, LRN
+  * Jonas Degrave (@317070): get_all_param_values() fix
+  * Jack Kelly (@JackKelly): help with recurrent layers
+  * Gábor Takács (@takacsg84): support broadcastable parameters in lasagne.updates
+  * Diogo Moitinho de Almeida (@diogo149): MNIST example fixes
+  * Brian McFee (@bmcfee): MaxPool2DLayer fix
+  * Martin Thoma (@MartinThoma): documentation
+  * Jeffrey De Fauw (@JeffreyDF): documentation, ADAM fix
+  * Michael Heilman (@mheilman): NonlinearityLayer, lasagne.random
+  * Gregory Sanders (@instagibbs): documentation fix
+  * Jon Crall (@erotemic): check for non-positive input shapes
+  * Hendrik Weideman (@hjweide): set_all_param_values() test, MaxPool2DCCLayer fix
+  * Kashif Rasul (@kashif): ADAM simplification
+  * Peter de Rivaz (@peterderivaz): documentation fix
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,11 @@
+include *.rst
+include *.txt
+include LICENSE
+
+recursive-include lasagne/tests *.py
+include .coveragerc
+recursive-include examples *.py
+recursive-include docs *.rst conf.py *.css Makefile
+
+recursive-exclude * __pycache__
+recursive-exclude * *.py[co]
--- a/docs/user/installation.rst
+++ b/docs/user/installation.rst
@@ -70,15 +70,12 @@ be handled below.
 Stable Lasagne release
 ======================
 
-.. note:: Lasagne hasn't been released yet. Please see the next section for
-    how to install the bleeding-edge version instead.
-
 Lasagne 0.1 requires a more recent version of Theano than the one available
 on PyPI. To install a version that is known to work, run the following command:
 
 .. code-block:: bash
 
-  pip install -r https://raw.githubusercontent.com/Lasagne/Lasagne/0.1/requirements.txt
+  pip install -r https://raw.githubusercontent.com/Lasagne/Lasagne/v0.1/requirements.txt
 
 .. warning::
   An even more recent version of Theano will often work as well, but at the
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-git+https://github.com/Theano/Theano.git@36d1eca8#egg=Theano==0.8.git
+git+https://github.com/Theano/Theano.git@15c90dd3#egg=Theano==0.8.git
--- a/setup.py
+++ b/setup.py
@@ -7,8 +7,7 @@ version = '0.1.dev'
 here = os.path.abspath(os.path.dirname(__file__))
 try:
     README = open(os.path.join(here, 'README.rst')).read()
-    CHANGES = ''
-    # CHANGES = open(os.path.join(here, 'CHANGES.txt')).read()
+    CHANGES = open(os.path.join(here, 'CHANGES.rst')).read()
 except IOError:
     README = CHANGES = ''
 
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ import os
 from setuptools import find_packages
 from setuptools import setup
 
-version = '0.1.dev'
+version = '0.1'
 
 here = os.path.abspath(os.path.dirname(__file__))
 try:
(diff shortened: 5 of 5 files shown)


craffel:
> It would be nice to have a DOI badge in the README of the first release, but I'm not sure if it's possible to add the DOI badge before we tag the release, which we need to do in order to get a DOI. I'll investigate.

So, you can hook the repository into zenodo before a release, but you may need to make a tag before you can have zenodo generate a DOI.  Should be pretty simple to find out https://guides.github.com/activities/citable-code/

CHANGES.rst
```diff
@@ -0,0 +1,35 @@
...
+  * Eric Battenberg (@ebattenberg)
+  * Sander Dieleman (@benanne)
+  * Daniel Nouri (@dnouri)
+  * Eben Olson (@ebenolson)
+  * Aäron van den Oord (@avdnoord)
+  * Colin Raffel (@craffel)
+  * Jan Schlüter (@f0k)
+  * Søren Kaae Sønderby (@skaae)
+
+* extra contributors, in chronological order:
```
craffel: cc @takacsg84 @instagibbs @peterderivaz 

Do any of you want your full name in the contributors list here instead of just your github nickname?

f0k:
> Should be pretty simple to find out

Okay, found it out. The badge URL includes the DOI, so it's not possible to include it in the very first release (i.e., the very first tag). I hoped that the badge might be generated based on the project name.

CHANGES.rst
instagibbs: You can add my real name: Gregory Sanders  

Before I forget: I've become busy with other stuff. If someone just wants to poach the PreLU implementation go ahead and take it. Don't have time to finish.

f0k: Added the name. (Good idea to attach the diff comment to a line that's not going to change, so it stays visible!)

CHANGES.rst
```diff
@@ -0,0 +1,35 @@
...
+  * Diogo Moitinho de Almeida (@diogo149): MNIST example fixes
+  * Brian McFee (@bmcfee): MaxPool2DLayer fix
+  * Martin Thoma (@MartinThoma): documentation
+  * Jeffrey De Fauw (@JeffreyDF): documentation, ADAM fix
+  * Michael Heilman (@mheilman): NonlinearityLayer, lasagne.random
+  * Gregory Sanders (@instagibbs): documentation fix
+  * Jon Crall (@erotemic): check for non-positive input shapes
+  * Hendrik Weideman (@hjweide): set_all_param_values() test, MaxPool2DCCLayer fix
+  * Kashif Rasul (@kashif): ADAM simplification
+  * (@peterderivaz): documentation fix
```
benanne: Nice work tracking down all of this :p it's nice to have an overview!

benanne:
LGTM! Too bad about the badge, but we can add it straight after the release so it's not a huge issue I guess.

CHANGES.rst
f0k: > Nice work tracking down all of this :p

In future we should probably add them on the go!

CHANGES.rst
f0k: Added the other two names. They were actually easy to google.

f0k:
> we can add it straight after the release so it's not a huge issue I guess

It's just that it won't be visible on PyPI then. (Unless we change the README.rst after tagging, but before doing `python setup.py sdist upload`.)

I've set up Zenodo now. It's waiting for a tag.

craffel:
> Unless we change the README.rst after tagging, but before doing python setup.py sdist upload.)

Let's just do that - it seems like the only disadvantage then would be that the README at the tag would be slightly different than the README on pypi, right?

f0k:
> it seems like the only disadvantage then would be that the README at the tag would be slightly different than the README on pypi, right?

Yes, I think so. By the way, I've also found this release checklist: https://gist.github.com/audreyr/5990987

f0k:
I've merged this to master locally for testing, and the tests still pass, and the documentation still builds.

`python setup.py sdist` creates a file with the following contents:
```
./Lasagne-0.1/lasagne/nonlinearities.py
./Lasagne-0.1/lasagne/random.py
./Lasagne-0.1/lasagne/updates_old.py
./Lasagne-0.1/lasagne/layers/dnn.py
./Lasagne-0.1/lasagne/layers/recurrent.py
./Lasagne-0.1/lasagne/layers/embedding.py
./Lasagne-0.1/lasagne/layers/base.py
./Lasagne-0.1/lasagne/layers/helper.py
./Lasagne-0.1/lasagne/layers/corrmm.py
./Lasagne-0.1/lasagne/layers/shape.py
./Lasagne-0.1/lasagne/layers/noise.py
./Lasagne-0.1/lasagne/layers/__init__.py
./Lasagne-0.1/lasagne/layers/merge.py
./Lasagne-0.1/lasagne/layers/pool.py
./Lasagne-0.1/lasagne/layers/conv.py
./Lasagne-0.1/lasagne/layers/normalization.py
./Lasagne-0.1/lasagne/layers/input.py
./Lasagne-0.1/lasagne/layers/dense.py
./Lasagne-0.1/lasagne/layers/cuda_convnet.py
./Lasagne-0.1/lasagne/__init__.py
./Lasagne-0.1/lasagne/utils.py
./Lasagne-0.1/lasagne/init.py
./Lasagne-0.1/lasagne/updates.py
./Lasagne-0.1/lasagne/objectives.py
./Lasagne-0.1/lasagne/theano_extensions/padding.py
./Lasagne-0.1/lasagne/theano_extensions/__init__.py
./Lasagne-0.1/lasagne/theano_extensions/conv.py
./Lasagne-0.1/lasagne/conftest.py
./Lasagne-0.1/lasagne/regularization.py
./Lasagne-0.1/setup.cfg
./Lasagne-0.1/setup.py
./Lasagne-0.1/PKG-INFO
./Lasagne-0.1/README.rst
./Lasagne-0.1/Lasagne.egg-info/dependency_links.txt
./Lasagne-0.1/Lasagne.egg-info/not-zip-safe
./Lasagne-0.1/Lasagne.egg-info/PKG-INFO
./Lasagne-0.1/Lasagne.egg-info/SOURCES.txt
./Lasagne-0.1/Lasagne.egg-info/top_level.txt
./Lasagne-0.1/Lasagne.egg-info/requires.txt
```

What strikes me is that `CHANGES.rst` is missing (should it be in there?), and that `./Lasagne-0.1/lasagne/conftest.py` is in there (shouldn't we remove this if we don't include the tests anyway?). @dnouri, master of Python project management, what do you think about this?

f0k:
Seems we need a `MANIFEST.in`. Will add one.

f0k:
Okay, added a MANIFEST.in that will make the docs and tests and all *.rst and *.txt files in the root directory part of the source distribution. Set `include_package_data=False` so the tests are not installed along with the rest of the `lasagne` directory when doing `python setup.py install` (or any pip equivalent).

`python setup.py test` doesn't work, that would need integration similar to this: https://pytest.org/latest/goodpractises.html
I'm not sure what the purpose of `extras_require={"testing": ...}` is, and whether it would be fine to change it to `tests_require`.
Unless Daniel sheds some light on this, let's leave this for another release. I don't think it's too crucial. You can run the tests in the extracted source distribution fine, if you follow the instructions, `python setup.py test` would just be a nice bonus.

f0k:
@benanne: Shall I go ahead, merge this and create a tag? (Side question: Should the tag contain all of the changelog, i.e., the contributor lists? The tag commit message is used for the release notes on github: https://github.com/audreyr/cookiecutter/releases/tag/1.0.0)

benanne:
1. yes!
2. do whatever you think is best!

f0k:
Okay, merged. Will create the tag at 23:00 CEST! I will include the contributor list just for consistency, as I guess we should include whatever the changelog says for future releases. It's not really meaningful for the first release, but for future ones there will be an actual changelog.

JeffreyDF:
Yay! Well done everyone. :-)

f0k:
GitHub dated back the tag by five minutes! (Created it in the web interface. Maybe their servers are a bit behind.)

benanne:
yay

Yay!

benanne:
Is there anything I need to do to push it to PyPI?

f0k:
> Is there anything I need to do to push it to PyPI?

Yep, I think you're the only one who can, see https://github.com/Lasagne/Lasagne/issues/74#issuecomment-130845323!

kashif:
Thank you!

PR #381, opened 2015-08-13
Add Pypi badge to README.rst

craffel:


f0k:
I had thought about that, but I'm not sure if that isn't too much. Also it may take away the focus from the installation instructions further below (which suggest to install the latest version, and link to pip installation instructions for the stable version). Hmm...

craffel:
> I had thought about that, but I'm not sure if that isn't too much. Also it may take away the focus from the installation instructions further below (which suggest to install the latest version, and link to pip installation instructions for the stable version). Hmm...

I didn't realize it was possible to have too many badges :)  This makes sense though, feel free to ignore.

benanne:
I take it we're not adding this? If anyone feels strongly about it please speak up! Else we should close the PR.

ebenolson:
I'm not sure why not, it seems more useful than the "build passing" badge which will hopefully always be green...

f0k:
I don't feel strongly either way... my concern was that people would go straight to PyPI and do `pip install lasagne`, which won't work. However, the PyPI page lists the installation instructions from the README, and people doing `pip install lasagne` and `import lasagne` will be pointed to the installation guide on readthedocs (if they hadn't installed Theano already). So probably no harm would be done. As you please!

> it seems more useful than the "build passing" badge

Yes, that should probably be the one to remove if we need more space. It's nice to show that we use Travis, though.

benanne:
> Yes, that should probably be the one to remove if we need more space. It's nice to show that we use Travis, though.

Yeah, although the coverage badge already sort of implies that we're using continuous integration, so I guess it's the least essential one.

craffel:
Removing badges??? Am I the only one who realizes more badges is always better?  ;) But seriously, I don't think it's terribly important either way.  There are lots of badges we could or could not have.

f0k:
> Am I the only one who realizes more badges is always better? ;)

Maybe :) For aesthetic reasons I'd very much like to keep it a single badge line, so there's only room for one more!

f0k:
With the new github design, there's suddenly a lot more horizontal space for badges: https://github.com/blog/2085-a-new-look-for-repositories
Let's re-think or merge this in two weeks when the new design is rolled out completely :)

PR #382, opened 2015-08-13
Bump to next development version

f0k:
This should be merged before merging any code changes (at the very least because the documentation on readthedocs uses the `dev` suffix to check whether to link to the `v0.1` tag or the `master` branch on github for source code).

I'm not sure if `0.2.dev` is the right choice, though. We could also go for `0.1.1.dev`. And it seems that's not a valid version string anyway, I've previously seen some tool (pip?) complain and change it to `0.1.dev0`. So we could also go for `0.2.dev0` and slightly adapt `docs/conf.py` to cope with that.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -64,7 +64,7 @@ copyright = u'2014–2015, Lasagne contributors'
 #
 import lasagne
 # The short X.Y version.
-version = '.'.join(lasagne.__version__.split('.', 3)[:2]).rstrip('dev')
+version = '.'.join(lasagne.__version__.split('.', 2)[:2])
 # The full version, including alpha/beta/rc tags.
 release = lasagne.__version__
 
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ import os
 from setuptools import find_packages
 from setuptools import setup
 
-version = '0.1'
+version = '0.2.dev1'
 
 here = os.path.abspath(os.path.dirname(__file__))
 try:


f0k:
Bump... any opinions on this? Shall we go for `0.2.dev`, `0.2.dev0`, `0.1.1.dev`, `0.1.1.dev0`? @dnouri, any insights on the correct suffix? (We can't merge anything else before.)

benanne:
I'm fine with any of those.

dnouri:
I think at some point the consensus was to try and use [PEP 0440](https://www.python.org/dev/peps/pep-0440/) version strings.  Here's how their 'public version identifier' looks like:

`[N!]N(.N)*[{a|b|rc}N][.postN][.devN]`

So `0.2.dev0` would be a match.

f0k:
Although they don't suggest to use `dev0` except for normalizing a version string that only has `dev`. Let's start with `0.2.dev1` then. I'll update the PR.

docs/conf.py
```diff
@@ -64,7 +64,7 @@
 #
 import lasagne
 # The short X.Y version.
-version = '.'.join(lasagne.__version__.split('.', 3)[:2]).rstrip('dev')
+version = '.'.join(lasagne.__version__.split('.', 2)[:2])
```
f0k: Just noticed that the variant before handled version strings we didn't even have, such as `0.1dev`. Simplified it to just keep the X.Y version no matter what. I'm not really sure if `version` is used anywhere in the templates we use, most things seem to use `release` which is the full version string.

PR #383, opened 2015-08-14
Add cropping to merge layers

Britefury:
This PR adds automatic cropping to layers in lasagne/layers/merge.py.

This allows `ConcatLayer`, `ElemwiseMergeLayer` etc to automatically crop their inputs before use if the shapes don't match as required.
My use for this is in implementing segmentation algorithms such as [1]. While the same effect can be achieved using `SliceLayer`, manually computing the slice size can be a pain, so adding automatic cropping to the merge layers simplifies matters.

[1] U-Net: Convolutional Networks for Biomedical Image Segmentation, http://arxiv.org/abs/1505.04597

### DIFF
--- a/lasagne/layers/merge.py
+++ b/lasagne/layers/merge.py
@@ -4,11 +4,239 @@ from .base import MergeLayer
 
 
 __all__ = [
+    "autocrop",
+    "autocrop_array_shapes",
     "ConcatLayer",
     "concat",
     "ElemwiseMergeLayer",
     "ElemwiseSumLayer",
 ]
+
+
+def autocrop(inputs, cropping):
+    """
+    Crops the given input arrays.
+
+    Cropping takes a sequence of inputs and crops them per-axis in order to
+    ensure that their sizes are consistent so that they can be combined
+    in an element-wise fashion. If cropping is enabled for a specific axis,
+    the minimum size in that axis of all inputs is computed, and all
+    inputs are cropped to that size.
+
+    The per-axis cropping modes are:
+
+    `None`: this axis is not cropped, inputs are unchanged in this axis
+
+    `'lower'`: inputs are cropped choosing the lower portion in this axis
+    (`a[:crop_size, ...]`)
+
+    `'upper'`: inputs are cropped choosing the upper portion in this axis
+    (`a[-crop_size:, ...]`)
+
+    `'center'`: inputs are cropped choosing the central portion in this axis
+    (``a[offset:offset+crop_size, ...]`` where
+    ``offset = (a.shape[0]-crop_size)//2)``
+
+    Parameters
+    ----------
+    inputs : list of Theano expressions
+        The input arrays in the form of a list of Theano expressions
+
+    cropping : list of cropping modes
+        Cropping modes, one for each axis. If length of `cropping` is less
+        than the number of axes in the inputs, it is padded with `None`.
+        If `cropping` is None, `input` is returned as is.
+
+    Returns
+    -------
+    list of Theano expressions
+        each expression is the cropped version of the corresponding input
+
+    Example
+    -------
+    For example, given three inputs:
+
+    >>> import numpy
+    >>> import theano
+
+    >>> a = numpy.random.random((1, 2, 3, 4))
+    >>> b = numpy.random.random((5, 4, 4, 2))
+    >>> c = numpy.random.random((7, 1, 8, 9))
+
+    Cropping mode for each axis:
+
+    >>> cropping = [None, 'lower', 'center', 'upper']
+
+    Crop (note that the input arrays are converted to Theano vars first,
+    and that the results are converted back from Theano expressions to
+    numpy arrays by calling `eval()`)
+    >>> xa, xb, xc = autocrop([theano.shared(a), \
+                               theano.shared(b), \
+                               theano.shared(c)], cropping)
+    >>> xa, xb, xc = xa.eval(), xb.eval(), xc.eval()
+
+    They will be left as is in axis 0 and cropped in the other three,
+    choosing the lower, center and upper portions:
+
+    Axis 0: choose all, axis 1: lower 1 element,
+    axis 2: central 3 (all) and axis 3: upper 2
+    >>> (xa == a[:, :1, :3, -2:]).all()
+    True
+
+    Axis 0: choose all, axis 1: lower 1 element,
+    axis 2: central 3 starting at 0 and axis 3: upper 2 (all)
+    >>> (xb == b[:, :1, :3, -2:]).all()
+    True
+
+    Axis 0: all, axis 1: lower 1 element (all),
+    axis 2: central 3 starting at 2 and axis 3: upper 2
+    >>> (xc == c[:, :1, 2:5:, -2:]).all()
+    True
+    """
+    if cropping is None:
+        # No cropping in any dimension
+        return inputs
+    else:
+        # Get the number of dimensions
+        ndim = inputs[0].ndim
+        # Check for consistent number of dimensions
+        if not all(input.ndim == ndim for input in inputs):
+            raise ValueError("Not all inputs are of the same "
+                             "dimensionality. Got {0} inputs of "
+                             "dimensionalities {1}.".format(
+                                len(inputs),
+                                [input.ndim for input in inputs]))
+        # Get the shape of each input, where each shape will be a Theano
+        # expression
+        shapes = [input.shape for input in inputs]
+        # Convert the shapes to a matrix expression
+        shapes_tensor = T.as_tensor_variable(shapes)
+        # Min along axis 0 to get the minimum size in each dimension
+        min_shape = T.min(shapes_tensor, axis=0)
+
+        # Nested list of slices; each list in `slices` corresponds to
+        # an input and contains a slice for each dimension
+        slices_by_input = [[] for i in range(ndim)]
+
+        # If there are more dimensions than cropping entries, pad
+        # the cropping
+        cropping = list(cropping)
+        if ndim > len(cropping):
+            cropping = list(cropping) + \
+                         [None] * (ndim - len(cropping))
+
+        # For each dimension
+        for dim, cr in enumerate(cropping):
+            if cr is None:
+                # Don't crop this dimension
+                slice_all = slice(None)
+                for slices in slices_by_input:
+                    slices.append(slice_all)
+            else:
+                # We crop all inputs in the dimension `dim` so that they
+                # are the minimum found in this dimension from all inputs
+                sz = min_shape[dim]
+                if cr == 'lower':
+                    # Choose the first `sz` elements
+                    slc_lower = slice(None, sz)
+                    for slices in slices_by_input:
+                        slices.append(slc_lower)
+                elif cr == 'upper':
+                    # Choose the last `sz` elements
+                    slc_upper = slice(-sz, None)
+                    for slices in slices_by_input:
+                        slices.append(slc_upper)
+                elif cr == 'center':
+                    # Choose `sz` elements from the center
+                    for sh, slices in zip(shapes, slices_by_input):
+                        offset = (sh[dim] - sz) // 2
+                        slices.append(slice(offset, offset+sz))
+                else:
+                    raise ValueError(
+                        'Unknown crop mode \'{0}\''.format(cr))
+
+        return [input[slices] for input, slices in
+                zip(inputs, slices_by_input)]
+
+
+def autocrop_array_shapes(input_shapes, cropping):
+    """
+    Computes the shapes of the given arrays after auto-cropping is applied.
+
+    For more information on cropping, see the :func:`autocrop` function
+    documentation.
+
+    Parameters
+    ----------
+    input_shapes : the shapes of input arrays prior to cropping in
+        the form of a list of tuples
+
+    cropping : a list of cropping modes, one for each axis. If length of
+        `cropping` is less than the number of axes in the inputs, it is
+        padded with `None`. If `cropping` is None, `input_shapes` is returned
+        as is. For more information on their values and operation, see the
+        :func:`autocrop` documentation.
+
+    Returns
+    -------
+    list of tuples
+        each tuple is a cropped version of the corresponding input
+        shape tuple in `input_shapes`
+
+    For example, given three input shapes with 4 axes each:
+
+    >>> a = (1, 2, 3, 4)
+    >>> b = (5, 4, 4, 2)
+    >>> c = (7, 1, 8, 9)
+
+    Cropping mode for each axis:
+
+    >>> cropping = [None, 'lower', 'center', 'upper']
+
+    Apply:
+
+    >>> cropped_shapes = autocrop_array_shapes([a, b, c], cropping)
+    >>> cropped_shapes[0]
+    (1, 1, 3, 2)
+
+    >>> cropped_shapes[1]
+    (5, 1, 3, 2)
+
+    >>> cropped_shapes[2]
+    (7, 1, 3, 2)
+
+    Note that axis 0 remains unchanged, where all the others are cropped
+    to the minimum size in that axis.
+    """
+    if cropping is None:
+        return input_shapes
+    else:
+        # Check for consistent number of dimensions
+        ndim = len(input_shapes[0])
+        if not all(len(sh) == ndim for sh in input_shapes):
+            raise ValueError("Not all inputs are of the same "
+                             "dimensionality. Got {0} inputs of "
+                             "dimensionalities {1}.".format(
+                                len(input_shapes),
+                                [len(sh) for sh in input_shapes]))
+
+        result = []
+
+        # If there are more dimensions than cropping entries, pad
+        # the cropping
+        cropping = list(cropping)
+        if ndim > len(cropping):
+            cropping = list(cropping) + \
+                         [None] * (ndim - len(cropping))
+
+        for sh, cr in zip(zip(*input_shapes), cropping):
+            if cr is None:
+                result.append(sh)
+            elif cr in {'lower', 'center', 'upper'}:
+                result.append([min(sh)] * len(sh))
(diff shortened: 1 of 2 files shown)


f0k:
Nice idea! Seems very convenient. I'm just not sure whether it would be better to provide this functionality without having existing merge layers derive from a new base class. The alternative would be defining an `AutoCropLayer` to wrap any `MergeLayer` like this:
```python
class AutoCropLayer(MergeLayer):
    def __init__(self, incomings, layer, cropping=None, **kwargs):
        super(AutoCropLayer, self).__init__(incomings, **kwargs)
        self.layer = layer
        self.params = layer.params
        # process cropping as before
        # modify input_shapes as before
        return self.layer.get_output_shape_for(input_shapes)
        # modify inputs as before
        return self.layer.get_output_for(inputs, **kwargs)
```

Then you could use it with any `MergeLayer` like this:
```python
l1 = InputLayer(...)
l2 = InputLayer(...)
l3 = AutoCropLayer([l1, l2], ElemwiseSumLayer([l1, l2]), cropping=...)
```

Hmm, no, it wouldn't work, actually. The `ElemwiseSumLayer` would have to be created with `AutoCropLayer.output_shapes`. We'd need to do a hack like:
```python
l3 = AutoCropLayer([l1, l2], None, cropping=...)
l3.layer = ElemwiseSumLayer(l3.output_shapes)
```

That's not really nicer than what you have. The cleanest solution would be stacking the `ElemwiseSumLayer` on top of the `AutoCropLayer`, but that's not possible because the `AutoCropLayer` has to give a single output and a single output shape with the current API.

So, well, maybe your solution is the best we can have for now! Any other thoughts?

lasagne/layers/merge.py
```diff
@@ -11,7 +11,143 @@
...
+
+    They will be left as is in axis 0 and cropped in all others,
+    choosing the lower, center and upper portions:
+
+    a[:, :1, :3, -2:]   # Choose all, lower 1 element, 3 central (all)
+                        # and upper 2
+    b[:, :1, :3, -2:]   # Choose all, lower 1 element, 3 central starting at 0
+                        # and upper 2 (all)
+    c[:, :1, 2:5:, -2:] # Choose all, lower 1 element (all),
+                        # 3 central starting at 2 and upper 2 (all)
```
dnouri: Nice explanation!  But could benefit from being turned into a runnable doctest, to make sure all the numbers are correct, too.

Britefury:
@benanne Yes, the design issues are a little tricky here. An alternative I considered is to move the cropping functionality out of `MergeCropLayer` so that it is just a couple of functions that can be used by layers that need it. They could stay in `merge.py` or they could live in `base.py` or `helpers.py` perhaps.
@dnouri The docstring in  `MergeCropLayer` is now a runnable doctest.

benanne:
I think you meant to address @f0k :) But I agree that this is probably something that should be moved out of the base class. As it stands it adds a lot of complexity to it for a relatively uncommon use case.

Britefury:
@f0k @benanne I beg your pardon, yes, I did intend to address @f0k :)   I have also refactored the auto-cropping functionality; `MergeCropLayer` is gone and has been replaced by two functions. The layers that use it still take an extra `cropping` parameter to their constructors and apply auto-cropping in their `get_output` and `get_output_shape methods.

docs/modules/layers/merge.rst
```diff
@@ -5,6 +5,9 @@ Merge layers
 
 .. currentmodule:: lasagne.layers
 
+.. autoclass:: MergeCropLayer
+    :members:
+
```
f0k: That one needs to be removed again!

lasagne/layers/merge.py
```diff
@@ -11,6 +17,217 @@
...
+                               theano.shared(b), \
+                               theano.shared(c)], cropping)
+    >>> xa, xb, xc = xa.eval(), xb.eval(), xc.eval()
+
+    They will be left as is in axis 0 and cropped in the other three,
+    choosing the lower, center and upper portions:
+
+    Axis 0: choose all, axis 1: lower 1 element,
+    axis 2: central 3 (all) and axis 3: upper 2
+    >>> assert (xa == a[:, :1, :3, -2:]).all()
```
f0k: Instead of an assert, you can also write:
```
>>> (x< == a[:, :1, :3, -2:]).all()
True
```
That one simulates entering something into a prompt and getting back the result. doctests will then check whether it gets the expected result.

lasagne/layers/base.py
```diff
@@ -1,5 +1,7 @@
 from collections import OrderedDict
 
+from theano import tensor as T
+
```
f0k: That one also needs to be removed again. You didn't change anything else in `base.py` that requires this.

lasagne/layers/merge.py
```diff
@@ -11,6 +17,217 @@
...
+    list of Theano expressions
+        each expression is the cropped version of the corresponding input
+
+    Example
+    -------
+    For example, given three inputs:
+
+    >>> import numpy
+    >>> import theano
+    >>> from mock import Mock
```
f0k: And the last import should be removed, you didn't use it in the end.

lasagne/layers/merge.py
```diff
@@ -11,6 +17,217 @@
...
+
+    `CROP_CENTER` (`'center'`): inputs are cropped choosing the central
+    portion in this axis (`a[offset:offset+crop_size, ...]` where
+    `offset = (a.shape[0]-crop_size)//2)
+
+    Parameters
+    ----------
+    inputs : the input arrays in the form of a list of Theano expressions
+
+    cropping : a list of cropping modes, one for each axis. If length of
```
f0k: The first thing after the colon should be the type of variable, the explanation should be on the second line: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections

E.g.:
```
inputs : list of tensor variables
    The arrays to crop according to the cropping specification.
    All arrays must be of the same dimensionality.
```

lasagne/layers/merge.py
```diff
@@ -11,6 +17,217 @@
...
+    Axis 0: all, axis 1: lower 1 element (all),
+    axis 2: central 3 starting at 2 and axis 3: upper 2
+    >>> assert (xc == c[:, :1, 2:5:, -2:]).all()
+    """
+    if cropping is None:
+        # No cropping in any dimension
+        return inputs
+    else:
+        # Get the number of dimensions
+        ndim = inputs[0].ndim
```
f0k: It might be nice to add an explicit check for compatible dimensionalities here, otherwise it will fail at `T.as_tensor_variable` with some error that may be hard to understand (didn't try). Something like:
```python
if not all (input.ndim == ndim for input in inputs):
    raise ValueError("Not all inputs are of the same dimensionality. "
                     "Got %d inputs of dimensionalities %r." % (
                     len(inputs), [input.ndim for input in inputs]))
```

lasagne/layers/merge.py
```diff
@@ -11,6 +17,217 @@
...
+    cropping : a list of cropping modes, one for each axis. If length of
+        `cropping` is less than the number of axes in the inputs, it is
+        padded with `CROP_NONE`. If `cropping` is None, `input_shapes`
+        is returned as is. For more information on their values and
+        operation, see the `autocrop` documentation.
+
+    Returns
+    -------
+    list of tuples
+        each tuple is the
```
f0k: What a cliffhanger! ;)

lasagne/layers/merge.py
```diff
@@ -11,6 +17,217 @@
...
+
+        return [input[slices] for input, slices in
+                zip(inputs, slices_by_input)]
+
+
+def autocrop_array_shapes(input_shapes, cropping):
+    """
+    Computes the shapes of the given arrays after auto-cropping is applied.
+
+    For more information on cropping, see the `autocrop` function
```
f0k:
```
see the :func:`autocrop` function documentation
```

lasagne/layers/merge.py
```diff
@@ -11,6 +17,217 @@
...
+    documentation.
+
+    Parameters
+    ----------
+    input_shapes : the shapes of input arrays prior to cropping in
+        the form if a list of tuples
```
f0k: `if` -> `of` (but see [my previous comment](https://github.com/Lasagne/Lasagne/pull/383/files#r37176402))

lasagne/layers/merge.py
```diff
@@ -11,6 +17,217 @@
...
+    >>> c = (7, 1, 8, 9)
+
+    Cropping mode for each axis:
+
+    >>> cropping = [CROP_NONE, CROP_LOWER, CROP_CENTER, CROP_UPPER]
+
+    Apply:
+
+    >>> cropped_shapes = autocrop_array_shapes([a, b, c], cropping)
+    >>> assert cropped_shapes[0] == (1, 1, 3, 2)
```
f0k: Again, you could write instead:
```
>>> cropped_shapes[0]
(1, 1, 3, 2)
```

lasagne/layers/merge.py
```diff
@@ -11,6 +17,217 @@
...
+    to the minimum size in that axis.
+    """
+    if cropping is None:
+        return input_shapes
+    else:
+        result = []
+
+        # If there are more dimensions than cropping entries, pad
+        # the cropping
+        ndim = len(input_shapes[0])
```
f0k: Again, you might want to add a test for consistent dimensionalities here. This function won't fail if any of the shapes has a lower dimensionality, it will just return lower-dimensional output shapes then, but still that might be hard to debug later.

lasagne/layers/merge.py
```diff
@@ -24,18 +241,29 @@ class ConcatLayer(MergeLayer):
 
     axis : int
         Axis which inputs are joined over
+
+    cropping : None or [crop]
+        Cropping for each input axis. Cropping is described in the docstring
+        for `autocrop`
```
f0k:
```
for :func:`autocrop`. It is always disabled for `axis`.
```

lasagne/layers/merge.py
```diff
@@ -57,21 +285,28 @@ class ElemwiseMergeLayer(MergeLayer):
         updated value. Some possible merge functions are ``theano.tensor``:
         ``mul``, ``add``, ``maximum`` and ``minimum``.
 
+    cropping : None or [crop]
+        Cropping for each input axis. Cropping is described in the docstring
+        for `autocrop`
```
f0k: for :func:`autocrop`.

lasagne/layers/merge.py
```diff
@@ -97,6 +332,10 @@ class ElemwiseSumLayer(ElemwiseMergeLayer):
         is to be applied to all instances. By default, these will not
         be included in the learnable parameters of this layer.
 
+    cropping : None or [crop]
+        Cropping for each input axis. Cropping is described in the docstring
+        for `autocrop`
```
f0k:
```
for func:`autocrop`.
```

f0k:
> `MergeCropLayer` is gone and has been replaced by two functions.

Thanks, I like that! There are some minor problems with the docstrings and an additional check for matching dimensionalities would be cool, otherwise it's looking very good now.

Britefury:
@f0k Done.

lasagne/layers/merge.py
```diff
@@ -24,18 +265,29 @@ class ConcatLayer(MergeLayer):
 
     axis : int
         Axis which inputs are joined over
+
+    cropping : None or [crop]
+        Cropping for each input axis. Cropping is described in the docstring
+        for :func:`autocrop`. Cropping as always disabled for `axis`.
```
f0k: not "as", "is" :)

lasagne/layers/merge.py
```diff
@@ -11,6 +17,241 @@
...
+    Parameters
+    ----------
+    input_shapes : the shapes of input arrays prior to cropping in
+        the form of a list of tuples
+
+    cropping : a list of cropping modes, one for each axis. If length of
+        `cropping` is less than the number of axes in the inputs, it is
+        padded with `CROP_NONE`. If `cropping` is None, `input_shapes`
+        is returned as is. For more information on their values and
+        operation, see the `autocrop` documentation.
```
f0k:
```
:func:`autocrop`
```

lasagne/layers/merge.py
```diff
@@ -4,6 +4,12 @@
 
 
 __all__ = [
+    "CROP_NONE",
+    "CROP_LOWER",
+    "CROP_CENTER",
+    "CROP_UPPER",
```
f0k: It just occurred to me that these will end up directly in the `lasagne.layers` namespace now. Is it really necessary to define those constants? In all other places, we just use strings directly (e.g., for the different padding modes in convolutions). Defining global constants doesn't seem very Pythonic to me. Do others have any opinion? (Sorry for noticing this so late, I should have thought about it earlier!)

lasagne/layers/merge.py
```diff
@@ -11,6 +17,241 @@
...
+
+        # If there are more dimensions than cropping entries, pad
+        # the cropping
+        cropping = list(cropping)
+        if ndim > len(cropping):
+            cropping = list(cropping) + \
+                         [None] * (ndim - len(cropping))
+
+        for sh, cr in zip(zip(*input_shapes), cropping):
+            if cr == CROP_NONE:
```
f0k: A side effect of using constants is that `cr == CROP_NONE` looks perfectly valid, but actually you shouldn't compare to `None`: ["Comparisons to singletons like None should always be done with is or is not , never the equality operators."](https://www.python.org/dev/peps/pep-0008/#programming-recommendations)
So I'd prefer to use `cr is None` and replace all other occurrences of `CROP_*` with their `None` or `str` equivalent.

f0k:
Almost there! If you feel I'm asking too much, I can also lend you a hand and send a PR to your PR.

lasagne/layers/merge.py
dnouri: I think the only reason you should use `is None` is for readability; there should be one way to read it.  Other than that, `None == None` works perfectly fine.

As an example, consider flipping a boolean mask in numpy: `myarray == False` obviously can't be replaced with `myarray is False` (it can with `myarray *-1` but that's not the point).  So the rule isn't universal to begin with.  And not clear that `cr == CROP_NONE` reads any worse than `cr is CROP_NONE`.  (Though when he gets rid of the constants, it'll be better to write `cr is None`.)

f0k: > I think the only reason you should use `is None` is for readability

I always had the impression that there's some subtlety that can make `x == None` or `None in x` fail, but if it's just for readability, `cr == CROP_NONE` is better. (I still think those constants are more C-ic than Pythonic.)

ebenolson: I'm also not a fan of the constants. As for `is` vs `==`, while the justifications for the idiom are weak, I think the fact that it's recommended in PEP8 (and will give us a warning if ignored) are plenty reason not to contradict it.

Britefury:
@f0k Done.

lasagne/layers/merge.py
```diff
@@ -11,6 +13,232 @@
...
+
+    `'lower'`: inputs are cropped choosing the lower portion in this axis
+    (`a[:crop_size, ...]`)
+
+    `'upper'`: inputs are cropped choosing the upper portion in this axis
+    (`a[-crop_size:, ...]`)
+
+    `'center'`: inputs are cropped choosing the central portion in this axis
+    (`a[offset:offset+crop_size, ...]` where
+    `offset = (a.shape[0]-crop_size)//2)
```
f0k: You seem to miss the second apostrophe in that line. (And apart from that, it should be double apostrophes for code fragments. Single apostrophes are for referring to argument or variable names.) To be sure, run `make html` in the `docs` directory and open `_build/html/modules/layers/merge.html` in a browser to check whether it renders fine.

lasagne/layers/merge.py
```diff
@@ -11,6 +13,232 @@
...
+
+    Parameters
+    ----------
+    inputs : list of Theano expressions
+        The input arrays in the form of a list of Theano expressions
+
+    cropping : list of cropping modes
+        Cropping modes, one for each axis. If length of `cropping` is less
+        than the number of axes in the inputs, it is padded with `None`.
+        If `cropping` is None, `input_shapes` is returned as is.
```
f0k: Should be `inputs`, not `input_shapes`. (Again I'm sorry I didn't notice before. Apparently it either needs more eyes or multiple passes to do a proper review...)

lasagne/tests/layers/test_merge.py
```diff
@@ -4,17 +4,135 @@
...
+class TestAutocrop:
+    # Test internal helper methods of MergeCropLayer
+    def test_autocrop_array_shapes(self):
+        from lasagne.layers.merge import autocrop_array_shapes
+        crop0 = None
+        crop1 = [None, 'lower', 'center', 'upper']
+        # Too few crop modes; should get padded with None
+        crop2 = ['lower', 'upper']
+        # Invalid crop modes
+        crop_bad = ['lower', 'upper', 'bad', 'worse']
```
f0k: :)

f0k:
Very cool, thank you! Just two very (very very) minor things and it's perfect! Could you please squash your commits then? It should be (after everything is committed, but before it is pushed):
```
git checkout -b backup  # just to be sure
git checkout merge_cropping
git reset --soft 92622ed
git commit -m "Add cropping to merge layers"
git push --force origin merge_cropping
```
I'm good with merging then!

Britefury:
Okay, done.

f0k:
Great! Lets just wait for Travis and merge! (Probably tomorrow. Unless Sander beats me to it.)

benanne:
yay! I did something useful!

lasagne/layers/merge.py
dnouri: FTR, `cr == CROP_NONE` isn't going to give you a warning in any PEP8 tool because they're doing static analysis.

PR #385, opened 2015-08-17
Move base layer constructor descriptions to class docstrings

f0k:
For the two base layers, the constructors were described in the constructor docstring, while the convention throughout Lasagne is to describe them in the class docstring. This PR fixes this just for consistency (and to not mislead contributors which have a look at `Layer` to check how a layer docstring is to be set up).


benanne:
Good catch :)

craffel:
:+1:

f0k:
I regard this as a green-light for merging :)

PR #386, opened 2015-08-18
Add BiasLayer

### LINKED ISSUE #378: Add bias to NonlinearityLayer
For building an auto-encoder via #373, it would be useful to be able to drag out not only the nonlinearity, but also the bias from a layer into a separate layer (so the `UnLayer`/`InverseLayer` can be built to not invert the nonlinearity and bias, but be followed by its own bias and nonlinearity).

We could either add a `BiasLayer` for that, or regard it as part of the `NonlinearityLayer`. I think it wouldn't hurt to add it to the latter.

To be useful for different purposes, the bias should broadcast over all but the second dimension by default (that would be reasonable for dense and convolutional layers), but also be customizable to broadcast over a set of given axes (as in my [`BatchNormLayer` implementation](https://gist.github.com/f0k/f1a6bd3c8585c400c190)). This would also partly address the use case mentioned in #249.

f0k:
This moves `NonlinearityLayer` from `dense.py` to `special.py`, and adds a `BiasLayer`.

The implementation allows to specify which dimensions to share biases over. By default, it matches what the dense and convolutional layers do, so you can use the `BiasLayer` and `NonlinearityLayer` to pull out the bias and nonlinearity from a convolution layer over the following pooling layer for efficiency. I've also taken care to not create a fully-dimensional, broadcastable bias tensor, but to match the implementation of the dense and convolutional layers, which create a lower-dimensional tensor and dimshuffle it to the correct shape afterwards (i.e., in `get_output_for()`). This makes it even easier to steal an existing layer's bias and put it into a `BiasLayer`.

Resolves #378.

### DIFF
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -1,12 +1,14 @@
 import theano
 import theano.tensor as T
 
+from .. import init
 from .. import nonlinearities
 from .base import Layer, MergeLayer
 
 
 __all__ = [
     "NonlinearityLayer",
+    "BiasLayer",
     "InverseLayer",
     "TransformerLayer",
 ]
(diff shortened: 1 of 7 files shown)


lasagne/layers/special.py
```diff
@@ -1,15 +1,114 @@
...
+            # default: share biases over all but the second axis
+            shared_axes = (0,) + tuple(range(2, len(self.input_shape)))
+        elif isinstance(shared_axes, int):
+            shared_axes = (shared_axes,)
+        self.shared_axes = shared_axes
+
+        if b is None:
+            self.b = None
+        else:
+            # create bias parameter, ignoring all dimensions in self.axes
```
skaae: I think the comment should refer to self.shared_axis

skaae:
looks good to me.

lasagne/layers/special.py
f0k: Good catch. (Had it named `self.axes` first, and I thought I got all occurrences...)

lasagne/layers/special.py
```diff
@@ -1,15 +1,114 @@
...
+            if any(size is None for size in shape):
+                raise ValueError("BiasLayer needs specified input sizes for "
+                                 "all axes that biases are not shared over.")
+            self.b = self.add_param(b, shape, 'b', regularizable=False)
+
+    def get_output_for(self, input, **kwargs):
+        if self.b is not None:
+            bias_axes = iter(range(self.b.ndim))
+            pattern = ['x' if input_axis in self.shared_axes
+                       else next(bias_axes)
```
benanne: digging this `iter` `next` stuff, it's neat! :)

benanne:
Looks good! No substantial comments from me. The tests seem to cover all the bases as well. I guess we could add one checking that `output_shape` always equals the input shape, but maybe we're just testing the underlying `Layer` functionality then. This is good to merge as far as I'm concerned!

lasagne/layers/special.py
f0k: Yeah! You go and have fun with Lua, though ;) (SCNR.)

PR #387, opened 2015-08-19
WIP: Give RecurrentLayer W params better names

craffel:
Because of the way they are constructed, the weight matrices in a `RecurrentLayer` are both just called `layer_name.W` which is not terribly helpful.  This fixes that.  Are there any edge cases I'm not thinking of, e.g. would the `.name` attribute of the parameters ever not exist?

### DIFF
--- a/lasagne/layers/recurrent.py
+++ b/lasagne/layers/recurrent.py
@@ -551,16 +551,30 @@ class RecurrentLayer(CustomRecurrentLayer):
             input_shape = incoming
         else:
             input_shape = incoming.output_shape
+        # Retrieve the supplied name, if it exists; otherwise use ''
+        if 'name' in kwargs:
+            basename = kwargs['name'] + '.'
+            # Create a separate version of kwargs for the contained layers
+            # which does not include 'name'
+            layer_kwargs = dict((key, arg) for key, arg in kwargs.items()
+                                if key != 'name')
+        else:
+            basename = ''
+            layer_kwargs = kwargs
         # We will be passing the input at each time step to the dense layer,
         # so we need to remove the second dimension (the time dimension)
         in_to_hid = DenseLayer(InputLayer((None,) + input_shape[2:]),
                                num_units, W=W_in_to_hid, b=b,
-                               nonlinearity=None, **kwargs)
+                               nonlinearity=None,
+                               name=basename + 'input_to_hidden',
+                               **layer_kwargs)
         # The hidden-to-hidden layer expects its inputs to have num_units
         # features because it recycles the previous hidden state
         hid_to_hid = DenseLayer(InputLayer((None, num_units)),
                                 num_units, W=W_hid_to_hid, b=None,
-                                nonlinearity=None, **kwargs)
+                                nonlinearity=None,
+                                name=basename + 'hidden_to_hidden',
+                                **layer_kwargs)
 
         # Make child layer parameters intuitively accessible
         self.W_in_to_hid = in_to_hid.W
--- a/lasagne/tests/layers/test_recurrent.py
+++ b/lasagne/tests/layers/test_recurrent.py
@@ -100,6 +100,15 @@ def test_recurrent_init_val_error():
         l_rec = RecurrentLayer(InputLayer((2, 2, 3)), 5, hid_init=hid_init)
 
 
+def test_recurrent_name():
+    l_in = lasagne.layers.InputLayer((2, 3, 4))
+    layer_name = 'l_rec'
+    l_rec = lasagne.layers.RecurrentLayer(l_in, 4, name=layer_name)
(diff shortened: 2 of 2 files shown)


f0k:
> Are there any edge cases I'm not thinking of

Yes, the user could have passed a named variable for `W_hid_to_hid`, say, "recurrent weights", and you'd turn it into "recurrent weightW_hid_to_hid" now :)
Instead, you should give a name to the `DenseLayer`s you are constructing, so they will name their variables accordingly if they don't have a name yet. Something like:
```python
basename = '' if 'name' not in kwargs else (kwargs['name'] + '.')
in_to_hid = DenseLayer(..., name=basename + 'input_to_hidden')
hid_to_hid = DenseLayer(..., name=basename + 'hidden_to_hidden')
```
This would give you `layer_name.input_to_hidden.W` and `layer_name.hidden_to_hidden.W`, which would also match how they can be accessed through the `CustomRecurrentLayer`.

I agree that `layer_name.W_in_to_hid` and `layer_name.W_hid_to_hid` would be nicer, but then you'd need to create the weight parameters using `lasagne.utils.create_param` inside `RecurrentLayer` and pass them to the `DenseLayer`s:
```python
basename = '' if 'name' not in kwargs else (kwargs['name'] + '.')
W_in_to_hid = create_param(W_in_to_hid, (int(np.prod(input_shape[2:])), num_units), basename + 'W_in_to_hid')
b = create_param(b, (num_units,), basename + 'b')
in_to_hid = DenseLayer(...)
...
```
That sounds like a lot of effort just for the sake of getting good variable names, but it's still less hackish than the third alternative: Check if `W_in_to_hid` had a name before passing it to `DenseLayer`, and only apply your name hack if it hadn't.

Pick whichever you like best. I think I'd prefer the first. (And make sure to amend your commit rather than adding a new one and squashing it afterwards, to save electricity! :) )

craffel:
> Yes, the user could have passed a named SharedVariable for W_hid_to_hid, say, "recurrent weights", and you'd turn it into "recurrent weightW_hid_to_hid" now :)

Ah, sure, that's what I was worried about, but I wasn't thinking of a way to do it!

> Instead, you should give a name to the DenseLayers you are constructing, so they will name their variables accordingly if they don't have a name yet. Something like:

> ```Python
basename = '' if 'name' not in kwargs else (kwargs['name'] + '.')
in_to_hid = DenseLayer(..., name=basename + 'input_to_hidden')
hid_to_hid = DenseLayer(..., name=basename + 'hidden_to_hidden')
```
> This would give you layer_name.input_to_hidden.W and layer_name.hidden_to_hidden.W, which would also match how they can be accessed through the CustomRecurrentLayer.

Ok, I switched to this, with the minor caveat that I believe you need to delete the `'name'` entry from `kwargs` or else it will be passed in twice, because ``**kwargs` is passed to the `DenseLayer` constructors.  I implemented this, which ended up requiring a test for full converage.  Let me know what you think.

**Comment by f0k** [comment]
> Ok, I switched to this, with the minor caveat that I believe you need to delete the `'name'` entry from `kwargs` or else it will be passed in twice, because ``**kwargs` is passed to the `DenseLayer` constructors.

Yes, I saw that, too. The alternative would be not to pass on `**kwargs` to the `DenseLayer` constructors at all. It's actually not clear whether it should be passed on or not... this depends on what else might be added to the `Layer` base class. I'm leaning towards not passing it on, interpreting `**kwargs` to be meant for `RecurrentLayer` itself only. You can just as well flip a coin to decide on that, though.

**Comment by craffel** [comment]
> You can just as well flip a coin to decide on that, though.

I flipped a biased coin, and it landed in favor of leaving it how it is.

File: lasagne/layers/recurrent.py [hunk 1]
```diff
@@ -547,16 +547,24 @@ def __init__(self, incoming, num_units,
                  mask_input=None,
                  **kwargs):
         input_shape = incoming.output_shape
+        # Retrieve the supplied name, if it exists; otherwise use ''
+        if 'name' in kwargs:
+            basename = kwargs['name'] + '.'
+            del kwargs['name']
```
f0k: You shouldn't delete the name before calling the super constructor, though. Either don't pass `**kwargs` to `DenseLayer` at all, or create a copy of `kwargs` without `'name'` to be passed on to `DenseLayer`.

craffel: Oh, sure, sorry, I will fix that.

f0k:
> I flipped a biased coin, and it landed in favor of leaving it how it is.

Fair enough :) Although "leaving it how it is" was more work in this case. Looks clean, though!

f0k:
Sorry, needs a rebase after #390 was merged... git doesn't seem to be sure in what order things belong now.

craffel:
Rebased.

f0k:
Since there are no more comments and everything looks good to me (and to Travis), let's merge this!

PR #390, opened 2015-08-19
Fix RecurrentLayer type(incoming) == tuple error

craffel:
RecurrentLayer would throw an error when incoming was a tuple because it was trying to access incoming.output_shape.  This adds an explicit check whether incoming is a tuple, and handles it appropriately. Also adds a test for this special case because all of the RecurrentLayer tests used a InputLayer as incoming.


f0k:
Good catch! That's the price of not being able to call the super constructor in the beginning.

PR #392, opened 2015-08-21
Helper functions for making symmetrical autoencoders with tied params (WIP)

neosatrapahereje:
Hi!

I added a couple of helper methods for making symmetrical autoencoders using the InverseLayer. Everything is now in lasagne/layers/autoenc.py (including moving the  `InverseLayer` there). 

Have a nice weekend!


f0k:
Looks like a good start!

There are several lines not covered in the tests: https://coveralls.io/builds/3368850/source?filename=lasagne%2Flayers%2Fautoenc.py
When running the tests locally, make sure you get 100% coverage for all the files you've changed.

I'll review and add some more specific comments!

docs/modules/layers.rst
```diff
@@ -18,6 +18,7 @@
     layers/merge
     layers/embedding
     layers/special
+    layers/autoenc
```
f0k: I'd prefer to have *special* the last one in the list before the ones that need to be imported explicitly. *special* is the bin for everything falling through the previous categories. So please add *autoenc* before that!

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+
+    Note that if the layer to be inverted contains a nonlinearity
+    and/or a bias, the :class:`InverseLayer` will include the derivative
+    of that in its computation.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape.
+    layer : a :class:`Layer` instance or a tuple
```
f0k: "or a tuple" is wrong here, actually. It has to be a Layer instance, otherwise it won't work.

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+    >>> import lasagne
+    >>> from lasagne.layers import InputLayer, Conv2DLayer, DenseLayer
+    >>> from lasagne.layers import InverseLayer
+    >>> l_in = InputLayer((100, 3, 28, 28))
+    >>> l1 = Conv2DLayer(l_in, num_filters=16, filter_size=5)
+    >>> l2 = DenseLayer(l1, num_units=20)
+    >>> l_u = InverseLayer(l2, l1)  # As Deconv2DLayer
+    """
+    def __init__(self, incoming, layer, **kwargs):
+
```
f0k: Maybe you can remove this blank line while you're at it.

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+__all__ = [
+    "InverseLayer",
+    "build_autoencoder",
+    "unfold_bias_and_nonlinearity_layers",
+]
+
+
+class InverseLayer(MergeLayer):
+    """
```
f0k: It would be good to have a short first sentence in the beginning, since that's used for overviews such as the tables here: http://lasagne.readthedocs.org/en/latest/modules/layers.html

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def get_output_for(self, inputs, **kwargs):
+        input, layer_out, layer_in = inputs
+        return theano.grad(None, wrt=layer_in, known_grads={layer_out: input})
+
+
+def build_autoencoder(layer, nonlinearity='same', b='new'):
+    """
+    Given a :class:`Layer` instance, this function builds a
+    symmetric autoencoder with tied weights.
```
f0k: Again, it would be good to have a very short sentence so it fits the overview tables. Maybe: "Unfolds a stack of layers into a symmetric autoencoder with tied weights." (Okay, that's not much shorter, but it gives some more information in the same length.)

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+
+    Parameters
+    ----------
+    layer: a :class:`Layer` instance
+        The :class:`Layer` instance with respect to which a symmetric
+        autoencoder is built.
+    nonlinearity : 'same', list, callable, or None
+        The nonlinearity that is applied to the layer activations of theano
```
f0k: why "theano"?

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+        :class:`InverseLayer`. If None is provided, the layer will be linear.
```
f0k: "the layer" is confusing. You're creating a lot of layers, not a single one.

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+        If 'same', the :class:`InverseLayer` has the same nonlinearity as the
+        mirror :class:`Layer`.
```
f0k: That's confusing, you didn't explain that you're using `InverseLayer` at all. Either add that explanation upfront, or just say that "each decoder layer has the same nonlinearity as its corresponding encoder layer."

f0k: What happens if you give a list of nonlinearities? I can imagine, but you should state it here!

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+    b : 'new', list, Theano shared variable, numpy array, callable or None
+        An initializer for the biases. If 'new', new biases are initialized
+        using the default initialization. If a shared variable or a numpy array
```
f0k: What is "the default initialization"?

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+    layer: a :class:`Layer` instance
+        The :class:`Layer` instance with respect to which a symmetric
+        autoencoder is built.
+    nonlinearity : 'same', list, callable, or None
+        The nonlinearity that is applied to the layer activations of theano
+        :class:`InverseLayer`. If None is provided, the layer will be linear.
+        If 'same', the :class:`InverseLayer` has the same nonlinearity as the
+        mirror :class:`Layer`.
+    b : 'new', list, Theano shared variable, numpy array, callable or None
+        An initializer for the biases. If 'new', new biases are initialized
```
f0k: Better be explicit: "An initializer for the decoder biases."

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+        using the default initialization. If a shared variable or a numpy array
+        is provided, the shape must match the incoming shape, skipping those
+        axes the biases are shared over (see below for an example). If set to
+        ``None``, the layer will have no biases and pass through its input
+        unchanged.
```
f0k: It seems you copied some text from my `BiasLayer`, this doesn't work here. (There are no axes to be skipped, there is no example demonstrating this, and "the layer" is unspecified again as you're creating a lot of them, not a single one!)

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+
+    Returns
+    -------
+    layer: :class:`Layer` instance
+       The output :class:`Layer` of the symmetric autoencoder with
+       tied weights.
+
+    Examples
+    --------
+    >>> import lasagne
```
f0k: Unused import. You can just delete this line.

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+
+    if isinstance(b, (tuple, list)):
+        b_idx = 0
+
+    layer = unfold_bias_and_nonlinearity_layers(layer)
+    layers = get_all_layers(layer)
+    autoencoder_layers = [layer]
+
+    for layer in layers[::-1]:
+        if not isinstance(layer, InputLayer):
```
f0k: Instead of indenting everything underneath, it would be easier to just have a:
```
if isinstance(layer, InputLayer):
    continue
```
in the beginning! That makes it easier to understand (I had to scroll a long way down to check if there is an `else` case for this `if` statement).

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+            elif isinstance(layer, GaussianNoiseLayer):
+                a_layer = GaussianNoiseLayer(
+                    incoming=incoming,
+                    sigma=layer.sigma
+                )
+            elif isinstance(layer, BiasLayer):
+                if b is None:
+                    kwargs = dict(b=None)
+                elif b == 'new':
+                    kwargs = dict(b=init.Constant(0.))
```
f0k: Ahaa! If that's the slightly mysterious "default bias" you mentioned above, you can just drop the `b='new'` case and have the default be `b=init.Constant(0.)` in the signature. That's a lot clearer and does exactly the same!

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+                a_layer = DropoutLayer(
+                    incoming=incoming,
+                    p=layer.p,
+                    rescale=layer.rescale
+                )
+            elif isinstance(layer, GaussianNoiseLayer):
+                a_layer = GaussianNoiseLayer(
+                    incoming=incoming,
+                    sigma=layer.sigma
+                )
```
f0k: That's going to be hard to maintain. Whenever somebody adds another noise layer, we'll have to remember to add it here as well. Maybe we should give layers an optional `copy(incoming=...)` method that just recreates another instance of the same layer with the same parameters and a different incoming layer? All layers that don't have a `copy` method would be handled by the general `else:` case below. What would our design goals tell us to do? Of course this would just slightly change things from "having to think about adding it here" to "having to think about adding a copy() method", but the latter would be quite obvious when starting with an existing noise layer.
Hmm... `copy()` actually is a misleading name, `decoder()` would be better... but then we'd start putting all this functionality into the class hierarchy... that would be a lot cleaner, but too obtrusive. Those pesky design choices!

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+    return autoencoder_layers[-1]
+
+
+def unfold_bias_and_nonlinearity_layers(layer):
+    """
+    Given a :class:`Layer` instance representing a stacked network,
+    this function adds a :class:`NonlinearityLayer` instance in between
+    each layer with attribute nonlinearity, with the same nonlinearity,
+    while setting the nonlinearity of the original layer to the identity
+    function.
```
f0k: Again, that's too long for a first sentence, and too short for a second sentence because it misses the explanation for `BiasLayer`.

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+
+    This function is useful to build autoencoders with tied parameters.
```
f0k: That's not really obvious. Either remove this sentence or explain why!

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+
+    Returns
+    -------
+        layer: :class:`Layer` instance
+           The output :class:`Layer` of the symmetric autoencoder with
+           tied weights.
+
+    Examples
+    --------
+    >>> import lasagne
```
f0k: Unused import.

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+    >>> l_out = unfold_bias_and_nonlinearity_layers(l_out)
+    >>> all_layers = get_all_layers(l_out)
+    >>> isinstance(all_layers[0], InputLayer)
+    True
+    >>> isinstance(all_layers[1], DenseLayer)
+    True
+    >>> all_layers[1].nonlinearity == identity
+    True
+    >>> all_layers[1].b is None
+    True
```
f0k: That style is too long for an example, and difficult to read, too. What about something like: 
```
>>> [l.__class__.__name__ for l in get_all_layers(l_out)]
['InputLayer', 'DenseLayer', 'BiasLayer', NonlinearityLayer', ...]
>>> all(l.b is None if not isinstance(l, BiasLayer))
True
```
Just some style that does not check each and every layer's attributes separately.

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+    """
+    layers = get_all_layers(layer)
+    incoming = layers[0]
+
+    for ii, layer in enumerate(layers[1:]):
+        layer.input_layer = incoming
+        if hasattr(layer, 'b') and not isinstance(layer, BiasLayer):
+            b = layer.b
+            layer.b = None
```
f0k: That's not enough, you also need to remove it from their `params`:
```python
del layer.params[b]
```
It didn't hurt because `get_all_params` weeds out duplicates to deal with tied parameters already.

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+        else:
+            b = None
+
+        if hasattr(layer, 'nonlinearity') \
+           and not isinstance(layer, NonlinearityLayer):
```
f0k: Its nicer to use brackets than line continuation characters, and indentation with three spaces looks weird. Please do:
```python
if (hasattr(layer, 'nonlinearity')
        and not isinstance(layer, NonlinearityLayer)):
```

lasagne/layers/autoenc.py
```diff
@@ -0,0 +1,262 @@
...
+        axes the biases are shared over (see below for an example). If set to
+        ``None``, the layer will have no biases and pass through its input
+        unchanged.
+        See :func:`lasagne.utils.create_param` for more information.
+
+    Returns
+    -------
+    layer: :class:`Layer` instance
+       The output :class:`Layer` of the symmetric autoencoder with
+       tied weights.
```
f0k: It also has to return the code layer of the autoencoder. As you're changing the encoder in `unfold_bias_and_nonlinearity_layers()`, the `layer` the user put into the function is not the code layer any more, because it has got its bias and nonlinearity stripped off. You need to return the code layer and explain in the documentation that the `layer` given by the user is not valid any more.

lasagne/tests/layers/test_autoenc.py
```diff
@@ -0,0 +1,159 @@
...
+    assert isinstance(autoencoder_layers[3], NonlinearityLayer)
+    assert isinstance(autoencoder_layers[4], DenseLayer)
+    assert isinstance(autoencoder_layers[5], BiasLayer)
+    assert isinstance(autoencoder_layers[6], NonlinearityLayer)
+    assert isinstance(autoencoder_layers[7], NonlinearityLayer)
+    assert isinstance(autoencoder_layers[8], BiasLayer)
+    assert isinstance(autoencoder_layers[9], InverseLayer)
+    assert isinstance(autoencoder_layers[10], NonlinearityLayer)
+    assert isinstance(autoencoder_layers[11], BiasLayer)
+    assert isinstance(autoencoder_layers[12], InverseLayer)
```
f0k: Wait, that's not what I would expect the decoder to look like. It should be `InverseLayer` -> `BiasLayer` -> `NonlinearityLayer` -> `InverseLayer` -> `BiasLayer` -> `NonlinearityLayer`. Maybe you should re-think splitting up the creation of the auto-encoder into two functions, and deal with everything in one go. Otherwise `build_autoencoder` needs to deal with reordering things correctly.

lasagne/tests/layers/test_autoenc.py
```diff
@@ -0,0 +1,159 @@
...
+
+def test_build_autoencoder():
+
+    from lasagne.layers import InputLayer, DenseLayer, BiasLayer
+    from lasagne.layers import InverseLayer, NonlinearityLayer
+    from lasagne.layers import build_autoencoder
+    from lasagne.nonlinearities import identity
+    from lasagne.layers import get_all_layers, get_output
+
+    input_shape = tuple(np.random.randint(20, high=100, size=2))
```
f0k: What's the advantage of using a random input shape?

lasagne/tests/layers/test_autoenc.py
```diff
@@ -0,0 +1,159 @@
...
+    ae_output = get_output(l_ae)
+    W1 = l1.W.get_value()
+    W2 = l2.W.get_value()
+    ae_output_fun = theano.function([l_in.input_var], ae_output)
+    X_input = np.random.rand(*input_shape).astype(theano.config.floatX)
+
+    assert np.allclose(
+        ae_output_fun(X_input),
+        np.dot(np.dot(np.dot(np.dot(X_input, W1), W2), W2.T), W1.T),
+        atol=1e-3)
```
f0k: Without knowing the absolute numbers, this sounds like a pretty high `atol`. Can you tune `rtol` instead?

lasagne/tests/layers/test_autoenc.py
```diff
@@ -0,0 +1,159 @@
...
+    for p_b, p_a in zip(params_before, params_after):
+        assert np.allclose(p_b, p_a)
+
+    # Check that all layers are the right type,
+    # and that all nonlinearities are correct
+    all_layers = get_all_layers(l_out)
+    assert isinstance(all_layers[0], InputLayer)
+    assert isinstance(all_layers[1], DenseLayer)
+    assert all_layers[1].nonlinearity == identity
+    assert all_layers[1].b is None
```
f0k: Please also add `assert len(all_layers[1].params) == 1` to ensure that the bias is not part of that layer's learnable parameters any more. Same for the other `DenseLayer` below.

alexjc:
@neosatrapahereje Hi, nice work! Very interested in using this. Do you have any plans for this PR?

benanne:
It looks like this PR was abandoned, but note that you can use arbitrary Theano expressions as layer parameters in Lasagne since last month: http://benanne.github.io/2015/11/10/arbitrary-expressions-as-params.html

I don't know what you have in mind of course, but perhaps this suffices :)

alexjc:
Thanks @benanne, I will look into it. I'll be adding Lasagne-based auto encoders to `scikit-neuralnetwork` and I think that will be a good starting point!

neosatrapahereje:
@alexjc Thanks for your interest in these functions. @f0k helped me to get this code to the current state, and hopefully, will review this PR. Otherwise, you can use the methods directly from here:

https://gist.github.com/neosatrapahereje/69562642b92996fba408

f0k:
We've squashed and rebased this PR, so it's mergeable again. We just need to decide:

1. Do we want to move `InverseLayer` to `lasagne.layers.autoenc` along with the helper functions, or should we add the helper functions somewhere else?
2. Is the code okay? Personally, I find the `n_idx` and `b_idx` a bit confusing, and the code will do funny things if the input network is not just a "tower". Can we protect against that?

rmanor:
This PR seems very useful.
Are there any plants to commit this?
Thanks.

f0k:
> This PR seems very useful.
Are there any plants to commit this?

In hindsight I think this may be too specific / not general enough to be included. It only supports simple architectures of a single stack of layers. We could turn it into a Lasagne Recipe, though, for easier copy/pasting. For now, the easiest will be to copy/paste the functions from https://github.com/neosatrapahereje/Lasagne/blob/a61947c/lasagne/layers/autoenc.py into your code.

PR #394, opened 2015-08-24
Optimize TransformerLayer

f0k:
This PR does a few things to the spatial transform layer:

* For consistency, it renames the `input` constructor argument to `incoming` and adapts the docstring accordingly
* Also for consistency, it changes the exception messages a little
* It drops the `deterministic=False` keyword argument from `get_output_for`, it was never used
* It allows `downscale_factor` to be a two-element tuple and adapts the docstring and tests accordingly
* ~~If the image size is known at compile time, it constructs the meshgrid offline instead of doing it symbolically~~
* Shape and coordinate casting is done in a better order to avoid needless device/host transfers
* The custom `_repeat` function is replaced by `T.repeat`

Unluckily, speed improvements were minuscule. I've got something between 1% and 4% both for the forward and backward pass, mostly because of reordering the casts. Offline computation of the meshgrid hardly changes anything (but it definitely didn't get slower). If the batch size is known at compile time, the `base` index array in `_interpolate` could be precomputed as well, but that seemed to be a less common case than a fixed image size, so I didn't try that. After all, a more promising approach would be turning this into a custom kernel, possibly using `tex2D`, but then we'd also need to do a gradient Op and CPU Ops and a graph optimizer.

### DIFF
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -167,9 +167,10 @@ class TransformerLayer(MergeLayer):
 
     Parameters
     ----------
-    input : a :class:`Layer` instance
-        The input where the affine transformation is applied. This should
-        have convolution format, i.e. (num_batch, channels, height, width).
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape. The
+        output of this layer should be a 4D tensor, with shape
+        ``(batch_size, num_input_channels, input_rows, input_columns)``.
 
     localization_network : a :class:`Layer` instance
         The network that calculates the parameters of the affine
@@ -207,21 +208,21 @@ class TransformerLayer(MergeLayer):
     ... nonlinearity=None)
     >>> l_trans = lasagne.layers.TransformerLayer(l_in, l_loc)
     """
-    def __init__(
-            self, input, localization_network, downsample_factor=1, **kwargs):
+    def __init__(self, incoming, localization_network, downsample_factor=1,
+                 **kwargs):
         super(TransformerLayer, self).__init__(
-            [input, localization_network], **kwargs)
+            [incoming, localization_network], **kwargs)
         self.downsample_factor = downsample_factor
 
         input_shp, loc_shp = self.input_shapes
 
         if loc_shp[-1] != 6 or len(loc_shp) != 2:
             raise ValueError("The localization network must have "
-                             "output shape [num_batch, 6]")
+                             "output shape: (batch_size, 6)")
         if len(input_shp) != 4:
-            raise ValueError('The input network must have a 4-dimensional '
-                             'output shape:(batch_size, num_input_channels, '
-                             'input_rows, input_columns)')
+            raise ValueError("The input network must have a 4-dimensional "
+                             "output shape: (batch_size, num_input_channels, "
+                             "input_rows, input_columns)")
 
     def get_output_shape_for(self, input_shapes):
         shp = input_shapes[0]
@@ -229,7 +230,7 @@ class TransformerLayer(MergeLayer):
         return (shp[:2] + tuple(
             None if s is None else int(s/dsf) for s in shp[2:]))
 
-    def get_output_for(self, inputs, deterministic=False, **kwargs):
+    def get_output_for(self, inputs, **kwargs):
         # see eq. (1) and sec 3.1 in [1]
         input, theta = inputs
         return _transform(theta, input, self.downsample_factor)
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -3,6 +3,7 @@ import theano.tensor as T
 
 from .. import init
 from .. import nonlinearities
+from ..utils import as_tuple
 from .base import Layer, MergeLayer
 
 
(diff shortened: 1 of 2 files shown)


skaae:
Hi, I i'll review this. Hopefully tomorrow. 

-Søren

benanne:
If the mesh grid precomputation has such a small effect on performance, maybe it's better not to do this in the interest of reducing code complexity?

f0k:
> If the mesh grid precomputation has such a small effect on performance, maybe it's better not to do this in the interest of reducing code complexity?

Yeah... I should benchmark that on some other GPUs as well, or for larger images. Intuitively, it should help to precompute the grid once instead of recomputing it in each forward pass, but most of the computation time is spent in `GpuAdvancedSubtensor1` (advanced indexing of the input image) anyway.

For 64x3x256x256 input with a downsample factor of 1.5, precomputing the meshgrid is 1% faster both in the forward and in the backward pass on my desktop (GT 640).

benanne:
If you have a benchmark script I could try to run it on a Tesla K40, GTX 980, GTX 680.

lasagne/layers/special.py
```diff
@@ -207,46 +211,50 @@ class TransformerLayer(MergeLayer):
...
     """
-    def __init__(
-            self, input, localization_network, downsample_factor=1, **kwargs):
+    def __init__(self, incoming, localization_network, downsample_factor=1,
+                 **kwargs):
         super(TransformerLayer, self).__init__(
-            [input, localization_network], **kwargs)
-        self.downsample_factor = downsample_factor
+            [incoming, localization_network], **kwargs)
+        self.downsample_factor = as_tuple(downsample_factor, 2)
```
skaae: What happens when downsample_factor is an iterable?

skaae:
I like the clean up and the flexible downsampling, thanks. It all looks fine to me. If precalculating the meshgrid is only ~1% then i don't think its worth the extra code complexity.

lasagne/layers/special.py
f0k: "nothing" is not quite right, it will turn it into a tuple (which is important in case it's a generator) and check whether it's got the correct length. So it will do exactly the right thing :)

skaae: yeah, before the EDIT i thought that the input was repeated N times :)

f0k:
I've benchmarked on a K40 now. Before this PR:
```
fwd dynamic
0.449194908142
fwd fixed
0.449412107468
bwd dynamic
0.966952800751
bwd fixed
0.966949939728
```
Where `fwd` and `bwd` is for the forward and backward pass, and `dynamic` and `fixed` is for an unknown and known image size.

After this PR:
```
fwd dynamic
0.338356018066
fwd fixed
0.337163209915
bwd dynamic
0.844947814941
bwd fixed
0.844409227371
```

So for some reason the improvement (before/after this PR) is much more pronounced than on my GT640. The difference for a known image size is not. I'd be okay with undoing that change (and just leaving a note that we've tried). I'd still keep the test as it is now, i.e., test both for a fixed and a variable image shape.

skaae:
Yes. Do you know what caused the speedup then? Is it the different cast you changed?

benanne:
That speedup seems substantial enough to keep this enhancement :) But the fixed/dynamic one should probably be undone. Leaving a note is a good idea.

f0k:
> Do you know what caused the speedup then?

The cast and the `_repeat`, I guess. The latter was causing a CPU `dot`. I didn't time the steps separately on the K40, only on the GT640, where each brought a minor improvement only.

skaae:
I prefer merging without the fixed meshgrid, but everything else looks good!

benanne:
Why? A speedup of 10-20% seems to be pretty worthwhile. This is on the order of the 'precompute' optimization for the `LSTMLayer`, isn't it?

skaae:
Maybe I got the benchmark wrong. But isn't the speedup only ~1% when comparing fixed and dynamic meshgrid?

I meant that we should keep the other enchancements but undo the meshgrid changes. 

Similar to your comment above:

> That speedup seems substantial enough to keep this enhancement :) But the fixed/dynamic one should probably be undone. Leaving a note is a good idea.

benanne:
No, from the latest numbers it looks like the meshgrid speedup is in the 10-20% range on a K40 (which is arguably a more useful measurement than that on the GT640). The fixed/dynamic thing was another enhancement that does not seem to make a significant difference anywhere.

skaae:
ok. In that case i agree that we should keep the meshgrid enhancements.

f0k:
> Maybe I got the benchmark wrong. But isn't the speedup only ~1% when comparing fixed and dynamic meshgrid?

Yes!

> The fixed/dynamic thing was another enhancement that does not seem to make a significant difference anywhere.

No! That shows the difference between precomputing the meshgrid and not precomputing it (it's only precomputed for fixed input image sizes, otherwise it's created dynamically as before). The 10-20% speedup does not come from precomputing the meshgrid. I'll remove that one again.

benanne:
oh, I was confused then. Where does the 10-20% speedup come from then?

f0k:
https://github.com/Lasagne/Lasagne/pull/394#issuecomment-134610557 :)

benanne:
Right, I missed that the first time around!

f0k:
All right, removed the `_meshgrid_const` function and any other distinction of known or unknown input sizes. Performance stays the same. Ready to merge from my side.

skaae:
Looks good to me, its nice work.  Sorry for the slow replies, but I have been busy with teaching this week.

f0k:
No problem, thanks for reviewing! Merged it.

PR #395, opened 2015-08-24
Correct variable shape propagation for merge layers

f0k:
As noticed in https://groups.google.com/forum/#!topic/lasagne-users/Zqrj1lSm9nc, `ConcatLayer` does not support `None`s in the concatenation axis. While solving this, I noticed that we're always basing the output shape on the input shape of the first incoming layer, which might have more (or different) unknowns than other incoming layers. I've addressed this as well, both for `ConcatLayer` and `MergeLayer`, by taking the maximum over all input sizes for each input dimension -- this conveniently replaces any `None`s with known values if there are any (we can do that because we know the input shapes have to match). This also allows us to add more thorough error checks. The improved shape propagation and error checking is covered by tests.

### DIFF
--- a/lasagne/layers/merge.py
+++ b/lasagne/layers/merge.py
@@ -268,9 +268,24 @@ class ConcatLayer(MergeLayer):
 
     def get_output_shape_for(self, input_shapes):
         input_shapes = autocrop_array_shapes(input_shapes, self.cropping)
+        # Infer the output shape by grabbing, for each axis, the first
+        # input size that is not `None` (if there is any)
+        output_shape = [next((s for s in sizes if s is not None), None)
+                        for sizes in zip(*input_shapes)]
+
+        def match(shape1, shape2):
+            return (len(shape1) == len(shape2) and
+                    all(i == self.axis or s1 is None or s2 is None or s1 == s2
+                        for i, (s1, s2) in enumerate(zip(shape1, shape2))))
+
+        # Check for compatibility with inferred output shape
+        if not all(match(shape, output_shape) for shape in input_shapes):
+            raise ValueError("Mismatch: input shapes must be the same except "
+                             "in the concatenation axis")
+        # Infer output shape on concatenation axis and return
         sizes = [input_shape[self.axis] for input_shape in input_shapes]
-        output_shape = list(input_shapes[0])  # make a mutable copy
-        output_shape[self.axis] = sum(sizes)
+        concat_size = None if any(s is None for s in sizes) else sum(sizes)
+        output_shape[self.axis] = concat_size
         return tuple(output_shape)
 
     def get_output_for(self, inputs, **kwargs):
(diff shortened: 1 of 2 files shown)


lasagne/tests/layers/test_merge.py
```diff
@@ -122,9 +122,16 @@ def crop_layer_1(self):
                            cropping=['lower'] * 2)
 
     def test_get_output_shape_for(self, layer):
-        input_shapes = [(3, 2), (3, 5)]
-        result = layer.get_output_shape_for(input_shapes)
-        assert result == (3, 7)
+        assert layer.get_output_shape_for([(3, 2), (3, 5)]) == (3, 7)
+        assert layer.get_output_shape_for([(3, 2), (3, None)]) == (3, None)
```
f0k: This failed before.

lasagne/tests/layers/test_merge.py
```diff
@@ -122,9 +122,16 @@ def crop_layer_1(self):
...
+        assert layer.get_output_shape_for([(None, 2), (3, 5)]) == (3, 7)
```
f0k: This returned `(None, 7)` before.

benanne:
> this conveniently replaces any Nones with known values if there are any
Based on Travis's output it looks like this trick doesn't work anymore in Python 3 :(

f0k:
> Based on Travis's output it looks like this trick doesn't work anymore in Python 3 :(

Ah, how very annoying! The alternative would be using `functools.reduce` with a `lambda a, b: a or b` (`operator.or_` is bitwise). It would have been easier if `any` returned the first true-ish value instead of converting to `True`...

f0k:
Okay, fixed it. `next()` to the rescue :)

f0k:
Travis passes. Any more comments?

lasagne/layers/merge.py
```diff
@@ -312,9 +327,20 @@ def __init__(self, incomings, merge_function, cropping=None, **kwargs):
 
     def get_output_shape_for(self, input_shapes):
         input_shapes = autocrop_array_shapes(input_shapes, self.cropping)
-        if any(shape != input_shapes[0] for shape in input_shapes):
+        # Infer as much of the shape as possible, avoiding Nones
+        output_shape = tuple(next((s[i] for s in input_shapes
+                                   if s[i] is not None), None)
+                             for i in range(len(input_shapes[0])))
+
```
ebenolson: This might be a little cleaner as 

        output_shape = [next((s for s in shape if s is not None), None)
                        for shape in zip(*input_shapes)]

Although it still seems pretty obtuse to me, a for loop might be better for readability.

f0k: Good idea about swapping things with `zip` (although `shape` should be `sizes` then, as it's all sizes for a particular axis)! If I saw a `for` loop I'd be wondering why it's not a list comprehension. It would get quite long:
```python
output_shape = []
ndim = len(input_shapes[0])
for i in range(ndim):
    size = None
    for shape in input_shapes:
        if shape[i] is not None:
            size = shape[i]
            break
    output_shape.append(size)
```
I find it more difficult to read this, and I'd feel bad about having this twice. I'll change the code according to your suggestion and add a second comment line stating that we're simply grabbing the first non-None-size per dimension.

f0k:
Addressed @ebenolson's comment. Good to merge?

benanne:
LGTM!

PR #396, opened 2015-08-25
Pool1DLayer

hjweide:
I recently wanted to do mean pooling in one dimension.  I saw #320, but could not find something similar for only one dimension.  So I moved the functionality from MaxPool1DLayer to Pool1DLayer, and let MaxPool1DLayer inherit from Pool1DLayer in a similar way to how MaxPool2DLayer works.  Using `mode='average_exc_pad'` seems to be much slower than `mode='max'`, but this is probably to be expected.

If this is useful and correct, I'll go ahead and squash the commits.

### DIFF
--- a/docs/modules/layers.rst
+++ b/docs/modules/layers.rst
@@ -79,6 +79,7 @@
 
     MaxPool1DLayer
     MaxPool2DLayer
+    Pool1DLayer
     Pool2DLayer
     Upscale2DLayer
     GlobalPoolLayer
--- a/docs/modules/layers/pool.rst
+++ b/docs/modules/layers/pool.rst
@@ -11,6 +11,9 @@ Pooling layers
 .. autoclass:: MaxPool2DLayer
     :members:
 
+.. autoclass:: Pool1DLayer
+    :members:
+
 .. autoclass:: Pool2DLayer
     :members:
 
--- a/lasagne/layers/pool.py
+++ b/lasagne/layers/pool.py
@@ -9,6 +9,7 @@ from theano.tensor.signal import downsample
 __all__ = [
     "MaxPool1DLayer",
     "MaxPool2DLayer",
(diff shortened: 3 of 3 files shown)


f0k:
Perfect, thank you! I noticed this myself at some point, but didn't need it, so thanks for contributing! If you squash the two commits, I'll merge.

> Using `mode='average_exc_pad'` seems to be much slower than `mode='max'`, but this is probably to be expected.

Depends... do you have cuDNN? I didn't try, but would have expected the two modes to be of comparable performance in cuDNN. Not sure how Theano's own implementation fares, and whether it has a GPU version of average-pooling at all.

hjweide:
Done!  I don't have cuDNN installed, but I could install it to test the performance sometime.

f0k:
> Done!

Thanks, merging!

> I don't have cuDNN installed, but I could install it to test the performance sometime.

You definitely should if you're interested in performance.

> the reason for this might be that `theano/sandbox/cuda/dnn.py`'s `dnn_pool` doesn't seem to support pooling in one dimension.

`Pool1DLayer` uses `max_pool_2d`, which also implements a 2D pooling operation. `Pool1DLayer` just does the necessary reshaping to use `max_pool_2d` along one axis only. So you could just as well use `Pool2DDNNLayer` for this, or add a `Pool1DDNNLayer`.

However, if you have cuDNN installed, `max_pool_2d` (and thus `Pool1DLayer`) will be using cuDNN automatically, there's no need to use any `lasagne.layers.dnn` layer for that. We should probably document that a bit more clearly...

hjweide:
> However, if you have cuDNN installed, max_pool_2d (and thus Pool1DLayer) will be using cuDNN automatically, there's no need to use any lasagne.layers.dnn layer for that. We should probably document that a bit more clearly...

I didn't know this, but it's very useful, thanks!

Issue #11, opened 2014-09-11
Arbitrary expressions as Layer parameters


benanne:
Currently, `Layer` parameters are assumed to be Theano shared variables. That way we can call `get_value()` and `set_value()` to them, and pass them to `theano.function` in `updates`.

However, it would be cool if `Layer` parameters could be arbitrary Theano expressions. There are a few use cases for this:

* autoencoders with tied weights. If you have two layers `l1` and `l2`, you might want to do something like `l2 = nntools.layers.DenseLayer(l1, W=l1.W.T)`. This is currently not possible because l1.W.T is not a shared variable, but rather an expression.

* sometimes the domain of parameters is restricted, and you might want to reparameterize things. For example, you might want all values to be positive. In that case you could reparameterize as follows: `l2 = nntools.layers.DenseLayer(l1, W=T.exp(V))`, where V is a shared variable. This ensures that all values in W are positive.

However, this requires some modifications to `nntools.layers.get_all_params()`, because we need a way to get the actual shared variables containing parameter values, not just the Theano expressions built on top of them. Otherwise there's no way to update them.

Given an arbitary Theano expression, it is fairly easy to get a list of all shared variables that occur in it, by traversing the Theano graph. This isn't very 'clean' I suppose, but definitely possible.

However, that just gives us a list of all shared variables, and we have no way of knowing if all of those contain learnable parameters. Perhaps some of them should not be touched by the learning.

We could assume that all shared variables represent learnable parameters by default. This would usually be the case. But then we need to provide a way for the user to specify that a given variable is not to be touched (for example, this could be a variable that contains a binary mask that restricts some of the parameters to be zero). Perhaps an extra attribute on the `Layer` instances that lists all "non-trainable" shared variables.

--

Should we support arbitrary expressions as `Layer` parameters? If so, there will be some added complexity, but it might be worth it. If we do not support this, a new `Layer` subclass has to be implemented for every new parameterization (so to support autoencoders with tied weights, we would need to implement a `TransposedDenseLayer`, for example).

What do you guys think?

benanne:
I guess the current idea is that we are not supporting this - in other words, it is safe to assume that layer parameters are always shared variables in the code (and indeed, this assumption is already made in many parts of the code anyway). If users want to reparameterize a layer, the preferred method is to subclass it.

f0k:
[...]

Issue #310, opened 2015-06-23
Memory leak on unpickling layers


f0k:
My colleague just stumbled upon what seemed to be a memory leak that eats up his GPU memory. We've condensed it down to a minimum working example:
```python
import sys
import cPickle as pickle
import numpy as np
import theano
import lasagne

VARIANT = int(sys.argv[1]) if len(sys.argv) > 1 else 1

print "Free memory: %g MiB" % (theano.sandbox.cuda.mem_info()[0] / 1024**2)
if VARIANT == 1:
    x = theano.shared(np.ones((10000, 1000), dtype=np.float32) * 42)
elif VARIANT in (2, 3):
    x = lasagne.layers.DenseLayer((None, 10000), 1000,
            W=lasagne.init.Constant(42))
with open('/dev/shm/tmpfile', 'wb') as f:
    pickle.dump(x, f, protocol=-1)
for _ in range(10):
    with open('/dev/shm/tmpfile', 'rb') as f:
        x = pickle.load(f)
    if VARIANT == 3:
        import gc; gc.collect()
    print "Free memory: %g MiB" % (theano.sandbox.cuda.mem_info()[0] / 1024**2)
```

There are three variants:
1. Just pickle and repeatedly unpickle a shared variable
2. Pickle and repeatedly unpickle a `DenseLayer`
3. As 2, but call `gc.collect()` after unpickling

We get the following output:
```
$ python memleak.py 1
Using gpu device 0: GeForce GT 640
Free memory: 522 MiB
Free memory: 483 MiB
Free memory: 483 MiB
Free memory: 483 MiB
Free memory: 483 MiB
Free memory: 483 MiB
Free memory: 483 MiB
Free memory: 483 MiB
Free memory: 483 MiB
Free memory: 483 MiB
Free memory: 483 MiB
$ python memleak.py 2
Using gpu device 0: GeForce GT 640
Free memory: 522 MiB
Free memory: 444 MiB
Free memory: 406 MiB
Free memory: 368 MiB
Free memory: 329 MiB
Free memory: 291 MiB
Free memory: 253 MiB
Free memory: 215 MiB
Free memory: 176 MiB
Free memory: 138 MiB
Free memory: 100 MiB
$ python memleak.py 3
Using gpu device 0: GeForce GT 640
Free memory: 522 MiB
Free memory: 483 MiB
Free memory: 483 MiB
Free memory: 483 MiB
Free memory: 483 MiB
Free memory: 483 MiB
[...]

PR #404, opened 2015-09-04
Allow arbitrary parameter expressions

f0k:
First stab at #11, to continue the discussion based on actual code.

This gets the basic case working:
```
In [3]: l1 = lasagne.layers.DenseLayer((100, 10), 20)

In [4]: l2 = lasagne.layers.DenseLayer(l1, 10, W=l1.W.T)

In [5]: lasagne.layers.get_all_params(l2)
Out[5]: [b, W, b]

In [6]: l1.params

Out[6]: OrderedDict([(W, set(['trainable', 'regularizable'])), (b, set(['trainable']))])

In [7]: l2.params
Out[7]: OrderedDict([(W.T, set(['trainable', 'regularizable'])), (b, set(['trainable']))])
```
It will work just as fine with `T.exp(l1.W)`, something adding up two shared variables, or something involving no shared variables at all (e.g., a constant weight matrix).

If there are no objections to the way it's implemented, I can go ahead and update the documentation and tests, as well as the recurrent layers that now have a specific switch for supporting arbitrary expressions for `hid_init`.

Todo:
* [x] update documentation and tests
* [x] add helpful error message to `lasagne.updates.*` in case you're not passing shared variables only
* ~~update the recurrent layer to not distinguish between parameter expressions and parameter variables in the constructor, because the base class supports both now~~
* [x] update all the docstrings of all the layers because the parameters can be expressions now

### DIFF
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -50,15 +50,20 @@ class Layer(object):
 
     def get_params(self, **tags):
         """
-        Returns a list of all the Theano variables that parameterize the layer.
+        Returns a list of Theano shared variables that parameterize the layer.
 
-        By default, all parameters that participate in the forward pass will be
-        returned (in the order they were registered in the Layer's constructor
-        via :meth:`add_param()`). The list can optionally be filtered by
-        specifying tags as keyword arguments. For example, ``trainable=True``
-        will only return trainable parameters, and ``regularizable=True``
-        will only return parameters that can be regularized (e.g., by L2
-        decay).
+        By default, all shared variables that participate in the forward pass
+        will be returned (in the order they were registered in the Layer's
+        constructor via :meth:`add_param()`). The list can optionally be
+        filtered by specifying tags as keyword arguments. For example,
+        ``trainable=True`` will only return trainable parameters, and
+        ``regularizable=True`` will only return parameters that can be
+        regularized (e.g., by L2 decay).
+
+        If any of the layer's parameters was set to a Theano expression instead
+        of a shared variable, the shared variables involved in that expression
+        will be returned rather than the expression itself. Tag filtering
+        considers all variables within an expression to be tagged the same.
 
         Parameters
         ----------
@@ -92,7 +97,7 @@ class Layer(object):
             result = [param for param in result
                       if not (self.params[param] & exclude)]
 
-        return result
+        return utils.collect_shared_vars(result)
 
     def get_output_shape_for(self, input_shape):
         """
@@ -149,56 +154,54 @@ class Layer(object):
 
     def add_param(self, spec, shape, name=None, **tags):
         """
-        Register and initialize a Theano shared variable containing parameters
-        associated with the layer.
+        Register and possibly initialize a parameter tensor for the layer.
 
-        When defining a new layer, this method can be used in the constructor
+        When defining a layer class, this method is called in the constructor
         to define which parameters the layer has, what their shapes are, how
         they should be initialized and what tags are associated with them.
+        This allows layer classes to transparently support parameter
+        initialization from numpy arrays and callables, as well as setting
+        parameters to existing Theano shared variables or Theano expressions.
 
-        All parameter variables associated with the layer can be retrieved
-        using :meth:`Layer.get_params()`.
+        All registered parameters are stored along with their tags in the
+        ordered dictionary :attr:`Layer.params`, and can be retrieved with
+        :meth:`Layer.get_params()`, optionally filtered by their tags.
 
         Parameters
         ----------
-        spec : Theano shared variable, numpy array or callable
-            an initializer for this parameter variable. This should initialize
-            the variable with an array of the specified shape. See
-            :func:`lasagne.utils.create_param` for more information.
+        spec : Theano shared variable, expression, numpy array or callable
+            initial value, expression or initializer for this parameter.
+            See :func:`lasagne.utils.create_param` for more information.
 
         shape : tuple of int
             a tuple of integers representing the desired shape of the
-            parameter array.
+            parameter tensor.
 
         name : str (optional)
-            the name of the parameter variable. This will be passed to
-            ``theano.shared`` when the variable is created. If ``spec`` is
-            already a shared variable, this parameter will be ignored to avoid
-            overwriting an existing name. If the layer itself has a name,
-            the name of the parameter variable will be prefixed with it and it
-            will be of the form 'layer_name.param_name'.
+            a descriptive name for the parameter variable. This will be passed
+            to ``theano.shared`` when the variable is created, prefixed by the
+            layer's name if any (in the form ``'layer_name.param_name'``). If
+            ``spec`` is already a shared variable or expression, this parameter
+            will be ignored to avoid overwriting an existing name.
 
         **tags (optional)
-            tags associated with the parameter variable can be specified as
-            keyword arguments.
-
-            To associate the tag ``tag1`` with the variable, pass
+            tags associated with the parameter can be specified as keyword
+            arguments. To associate the tag ``tag1`` with the parameter, pass
             ``tag1=True``.
 
             By default, the tags ``regularizable`` and ``trainable`` are
-            associated with the parameter variable. Pass
-            ``regularizable=False`` or ``trainable=False`` respectively to
-            prevent this.
+            associated with the parameter. Pass ``regularizable=False`` or
+            ``trainable=False`` respectively to prevent this.
 
         Returns
         -------
-        Theano shared variable
-            the resulting parameter variable
+        Theano shared variable or Theano expression
+            the resulting parameter variable or parameter expression
 
         Notes
         -----
-        It is recommend to assign the resulting parameter variable to an
-        attribute of the layer, so it can be accessed easily, for example:
+        It is recommended to assign the resulting parameter variable/expression
+        to an attribute of the layer for easy access, for example:
 
         >>> self.W = self.add_param(W, (2, 3), name='W')  #doctest: +SKIP
         """
@@ -206,7 +209,7 @@ class Layer(object):
         if name is not None:
             if self.name is not None:
                 name = "%s.%s" % (self.name, name)
-
+        # create shared variable, or pass through given variable/expression
         param = utils.create_param(spec, shape, name)
         # parameters should be trainable and regularizable by default
         tags['trainable'] = tags.get('trainable', True)
--- a/lasagne/layers/dense.py
+++ b/lasagne/layers/dense.py
@@ -29,16 +29,16 @@ class DenseLayer(Layer):
     num_units : int
         The number of units of the layer
 
-    W : Theano shared variable, numpy array or callable
-        An initializer for the weights of the layer. If a shared variable or a
-        numpy array is provided the shape should  be (num_inputs, num_units).
-        See :meth:`Layer.create_param` for more information.
+    W : Theano shared variable, expression, numpy array or callable
+        Initial value, expression or initializer for the weights.
+        These should be a matrix with shape ``(num_inputs, num_units)``.
+        See :func:`lasagne.utils.create_param` for more information.
 
-    b : Theano shared variable, numpy array, callable or None
-        An initializer for the biases of the layer. If a shared variable or a
-        numpy array is provided the shape should be (num_units,).
-        If None is provided the layer will have no biases.
-        See :meth:`Layer.create_param` for more information.
+    b : Theano shared variable, expression, numpy array, callable or ``None``
+        Initial value, expression or initializer for the biases. If set to
+        ``None``, the layer will have no biases. Otherwise, biases should be
+        a 1D array with shape ``(num_units,)``.
+        See :func:`lasagne.utils.create_param` for more information.
 
     nonlinearity : callable or None
         The nonlinearity that is applied to the layer activations. If None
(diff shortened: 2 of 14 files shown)


f0k:
Bump... anybody willing to have a look?

> If there are no objections to the way it's implemented, I can go ahead and update the documentation and tests, as well as the recurrent layers that now have a specific switch for supporting arbitrary expressions for `hid_init`.

benanne:
Yes, sorry! It's on my to do list, but I'm moving this weekend. Will try to get around to it ASAP.

skaae:
I can take a look in the evening.

f0k:
Ah, all right! Don't worry, just wanted to make sure it wasn't forgotten. Good luck with moving!

skaae:
I think this looks good.

For RNNs we could consider allowing `hid_init` be an input layer like mask? Then you could easily construct encoder/decoders without calling `get_output` multiple times.

Its not a huge issue though.

lasagne/utils.py
```diff
@@ -76,6 +76,29 @@ def as_theano_expression(input):
...
+    Parameters
+    ----------
+    expressions : Theano expression or iterable of Theano expressions
+        The expressions to collect shared variables from.
+
+    Returns
+    -------
+    list of Theano shared variables
+        All shared variables the given expression(s) depend on, in the order
+        found by a left-recursive depth-first search. If some expressions are
```
benanne: Is there any particular reason you went with this search strategy? Just asking because it's mentioned explicitly. I guess it's just the easiest to implement?

lasagne/utils.py
```diff
@@ -236,11 +259,11 @@ def create_param(spec, shape, name=None):
 
     Parameters
     ----------
-    spec : numpy array, Theano shared variable, or callable
+    spec : numpy array, Theano expression, or callable
         Either of the following:
 
         * a numpy array with the initial parameter values
-        * a Theano shared variable representing the parameters
+        * a Theano expression representing the parameters
```
benanne: I think that's a good idea, since that is the most common use case. Maybe something like "usually this is a Theano shared variable."

lasagne/layers/base.py
```diff
@@ -92,7 +92,7 @@ def get_params(self, **tags):
             result = [param for param in result
                       if not (self.params[param] & exclude)]
 
-        return result
+        return utils.collect_shared_vars(result)
```
benanne: I like how this is a one-line change now :)

benanne:
LGTM! Just some comments, but nothing that needs to change as far as I can see. This is probably something we need to test thoroughly though, to ensure that we don't break any existing use cases.

benanne:
Oh, three more things I thought of:

- the doc string for `add_param()` needs updating to indicate that expressions are now also accepted (but I guess you were planning on getting to that anyway).

- regarding parameter 'tagging', will everything still behave correctly when the keys for the tag dictionary can be arbitrary expressions? Are those hashable? I suppose so?

- in the issue I suggested that we should take precautions to ensure that people get meaningful error messages when they accidentally pass expressions where shared variables are expected (e.g. in `lasagne.updates`). Any concrete ideas on how to implement that without littering multiline checks all over the code?

lasagne/utils.py
f0k: > I guess it's just the strategy that theano.gof.graph.inputs() implements

Yes, exactly, I copied it from there. I thought we should mention that there's a particular order. I can add that it's in topological order, like: "All shared variables the given expression(s) depend on, in topological order (as found by a left-recursive depth-first search)."

lasagne/layers/base.py
f0k: (Although we could also change line 81 above to not listify `result`, for unmeasurably improved performance. Maybe I'll still do that.)

benanne: It's probably more readable as it is :)

lasagne/utils.py
benanne: sure!

f0k:
> the doc string for `add_param()` needs updating

Yes, and the `get_params()` docstrings should probably mention what happens for parameter expressions -- for sure that needs to be explained *somewhere*.

> will everything still behave correctly when the keys for the tag dictionary can be arbitrary expressions? Are those hashable? I suppose so?

Yes, they're hashable, theano needs that internally.

> in the issue I suggested that we should take precautions to ensure that people get meaningful error messages when they accidentally pass expressions where shared variables are expected (e.g. in `lasagne.updates`). Any concrete ideas on how to implement that without littering multiline checks all over the code?

Not really. There could be a utility function like `ensure_all_shared()` for the check that also raises the exception, but we'd still need to call it in every `lasagne.updates.*` function. It seems that's the only place we need it, though.

benanne:
>  It seems that's the only place we need it, though, so we could also include it in get_or_compute_grads() that's called by every lasagne.updates.* function already.

Sounds like a plan :)

lasagne/utils.py
f0k: I take that back, every order is a topological order since shared variables obviously don't have any parents :) I'll just leave it as it is.

f0k:
Updated documentation and tests. Had to add a `reversed()` in `collect_shared_vars()` to make it respect the order of input expressions (this is due to how the underlying Theano graph traversal routine is implemented, and the `reverse` doesn't hurt as much as implementing the traversal routine ourselves).

Todo:
* [ ] add helpful error message to `lasagne.updates.*` in case you're not passing shared variables only
* [ ] update the recurrent layer to not distinguish between parameter expressions and parameter variables in the constructor, because the base class supports both now
* [ ] update all the docstrings of all the layers because the parameters can be expressions now

benanne:
Looking good! When this is merged I'm thinking about writing a brief blog post about some of the possibilities this will offer (low-rank parameters, non-negativity constraints on parameters, easy parameter sharing etc.). Perhaps a bunch of examples like this would be useful in the docs as well.

f0k:
Sorry for the lack of progress, I'm experimenting and didn't need it so far :)

> I'm thinking about writing a brief blog post

Sure, count me in if you'd like a co-author.

By the way, low-rank dense layers work really well with just popping a linear, biasless dense layer in front of the normal dense layer. [MEAN-NORMALIZED STOCHASTIC GRADIENT FOR LARGE-SCALE DEEP LEARNING](http://www.mirlab.org/conference_papers/International_Conference/ICASSP%202014/papers/p180-wiesler.pdf) made me a bit scared that you'd need fancy learning methods or batch normalization for that, but I found it to work just fine with sgd+nesterov.

> Perhaps a bunch of examples like this would be useful in the docs as well.

Somewhere I suggested that we'd move the existing `Layers` and `Creating custom layers` into a `Library walk-through` chapter, and add more pages to that covering the updates, regularization, and whatnot. We could also add parameter sharing and other tricks there! It's a lot of prose to write, though.

benanne:
Yeah, I don't think I'll have time for that anytime soon myself. There's already a brief section on parameter sharing ( http://lasagne.readthedocs.org/en/latest/user/layers.html#parameter-sharing ), I guess we could rename and extend that.

f0k:
> I guess we could rename and extend that.

Ah, yes, that would be a good spot for now!

f0k:
Update:

1. `lasagne.updates.get_or_compute_grads()` now checks whether all params are shared variables. I'm not completely happy with this, but adding another function called in every `lasagne.updates.*` function seemed stupid as well. Note that with this change you can't use `get_or_compute_grads()` to compute gradients for things that are not shared variables, and that you can't use `sgd()` or anything else to compute updates for things that are not shared variables (but except for `sgd()`, all of them expect shared variables anyway, as they use `get_value()` to check the shape for creating accumulators).
2. The recurrent layers bypassed `create_param()` if `hid_init` or `cell_init` was a `TensorVariable`, and expected it to include the batch dimension if and only if it was a `TensorVariable`. Now that `create_param()` accepts any Theano expressions, I figured a better way to distinguish the two cases was via the dimensionality -- if it has enough dimensions to include the batch dimension, it probably includes the batch dimension, and if it has one dimension less, it will be replicated via a dot product against a vector of ones (as before). This required some more changes:
   (a) `create_param()` currently cannot be told to accept either a parameter of `(num_units)` or of `(batchsize, num_units)`. I've solved this by calling it a second time with the second shape if the first call failes (for which I changed `create_param()` to raise a `ValueError`, not a `RuntimeError`, before these were mixed).
   (b) `batchsize` is not usually known in a recurrent layer, only `num_units` is. I've solved this by changing `create_param()` to also accept `None` in its shape specification.
   Note that all this makes the recurrent layers incompatible with previous versions, because in previous versions `hid_init` was initialized to have a batch size of 1 for some reason, while now I initialize it to not include the batch size at all (otherwise we can't use `ndim` to know whether it needs replication over the batch or not).
  And again, I'm not completely happy with this. Maybe it would be nicer to change `create_param` to accept a set of alternative shapes that are also acceptable? The original, primary shape would be passed to a callable if a callable was given, and it would not be allowed to contain `None`s, but all other shapes may be underspecified and are only used to check whether a given shared variable, Theano expression or numpy array is accepted.

Any input or questions are welcome!

f0k:
> By the way, low-rank dense layers work really well with just popping a linear, biasless dense layer in front of the normal dense layer.

A bit off-topic here, but I just saw that the authors of [DeepFont](http://arxiv.org/abs/1507.03196) train a low-rank dense layer by replacing it with a low-rank approximation after every update (Equation 3 on page 6). Fun.

benanne:
It's interesting that training low-rank layers directly seems to be difficult with the current methods. It bothers me that the parameterization affects the efficiency of learning so much, but I guess that's the way it is. In speech, training a big fully connected layer and then replacing it with a low-rank version is quite popular as well (they call this a 'bottleneck', which gives you 'bottleneck features').

benanne:
1. I think that's a good trade-off. It's the best place to put it, and it'll probably prevent some long-winded bug hunts in the near future :)

2. Since I haven't used the recurrent layers I find it hard to reason about the implications of this, so I'd like it if @craffel or @skaae (or someone else who knows the codebase) comments on this.

f0k:
> I think that's a good trade-off. It's the best place to put it, and it'll probably prevent some long-winded bug hunts in the near future :)

Maybe we should solve #310 in Theano's `theano.updates.OrderedUpdates` and then use that one in `lasagne.updates`. It doesn't accept anything else but shared variables as keys, so that would remove the need of artificially constraining `get_or_compute_grads()`.

> Since I haven't used the recurrent layers I find it hard to reason about the implications of this

There's a general design question involved as well: To allow a layer to support different shapes for a particular parameter, we either need to extend `create_params()` to support multiple shapes, or have it called multiple times. What's better?
And to have it support shapes that are not completely known at construction time, we also need to extend it to allow `None`s in the specification. That's mostly useful in combination with symbolic parameter expressions, though, not so much with shared variables and not at all with weight initializers from `lasagne.init`. Again, would it be better to support this for secondary shapes only?

> so I'd like it if @craffel or @skaae (or someone else who knows the codebase) comments on this

Me too! @craffel, @skaae, what do you think of my proposed change to `hid_init`, i.e., distinguishing the two cases (including the batch dimension or not) via the dimensionality?

craffel:
Your changes look reasonable to me; using `hid_init` as a `TensorVariable` is kind of an edge-case as it is.  AFAIK, @skaae is the only one using this.

f0k:
> Your changes look reasonable to me; using `hid_init` as a `TensorVariable` is kind of an edge-case as it is.

Sure, but the change now requires the standard case to have `hid_init` be one dimension smaller than before (instead of adding a `1` as the first dimension -- we should have thought about that earlier). This will break compatibility with saved parameters of existing models. The alternative would be adding another keyword argument that tells whether `hid_init` and/or `cell_init` shall include the batch size or not, but there are so many keyword arguments already...

craffel:
Yes, like I said, I don't think breaking compatibility is bad because I don't know of anyone using this except @skaae. It's an uncommon use-case.  He should probably give the go-ahead though.

f0k:
> Yes, like I said, I don't think breaking compatibility is bad because I don't know of anyone using this except @skaae.

Maybe you're still misreading me -- this will break compatibility for everyone *except* @skaae.

craffel:
Oops, I see what you're saying now, sorry.  Is there any reason not to require the hidden state to always be 2D, where the first dimension is either 1 (meaning "use the same initialization for all samples in the batch", which is the use-case the majority of the time), or `n_batch` (meaning "use a different initialization for each sample in the batch", arguably an edge-case)?  In the case of `n_batch = 1`, if we dotted it against a matrix of shape `(n_batch, 1)`, it would be harmless.  Sorry if you discussed this above and I missed it.

f0k:
> Is there any reason not to require the hidden state to always be 2D, where the first dimension is either 1 (meaning "use the same initialization for all samples in the batch", which is the use-case the majority of the time), or `n_batch` (meaning "use a different initialization for each sample in the batch", arguably an edge-case)?

Yes, we cannot distinguish these two cases, unless we add an extra keyword to the constructor. All we can reliably check for a given `hid_init` is `ndim`. Maybe an extra constructor keyword would be better to distinguish the cases. If we add this, we could use it to exactly specify which axes the initial values are to be shared over -- e.g., for convolutional recurrent layers, one may want to have a single value per channel, not per spatial position. (That won't play nicely with the dotting trick any more, though.) Hmm, this PR is getting out of hands... maybe I should leave the recurrent layer untouched for now?

craffel:
> Yes, we cannot distinguish these two cases, unless we add an extra keyword to the constructor. All we can reliably check for a given hid_init is ndim. Maybe an extra constructor keyword would be better to distinguish the cases. If we add this, we could use it to exactly specify which axes the initial values are to be shared over -- e.g., for convolutional recurrent layers, one may want to have a single value per channel, not per spatial position. (That won't play nicely with the dotting trick any more, though.) Hmm, this PR is getting out of hands... maybe I should leave the recurrent layer untouched for now?

Oh right, because we won't know whether to dot against a ones matrix in `get_output` unless we know the shape from the get-go, which we won't in general if `hid_init` is a `TensorVariable`.  I am personally ok with an extra kwarg.  Having lots of kwargs is not a big deal to me, as long as their defaults are sensible and their name/docstring is understandable.

f0k:
Took out the changes to recurrent layers, they were getting too much to tackle in one go. Changed all relevant layer docstrings to accept expressions as an alternative to shared variables, numpy arrays or callables. Fixed some documentation errors on the way.

lasagne/layers/dense.py
```diff
@@ -115,19 +115,18 @@ class NINLayer(Layer):
...
         dimension beyond the 2nd.
 
-    W : Theano shared variable, numpy array or callable
-        An initializer for the weights of the layer. If a shared variable or a
-        numpy array is provided the shape should be (num_inputs, num_units),
-        where num_units is the size of the 2nd. dimension of the input.
+    W : Theano shared variable, expression, numpy array or callable
+        Initial value, expression or initializer for the weights.
+        These should be a matrix with shape ``(num_inputs, num_units)``,
+        where ``num_inputs`` is the size of the second dimension of the input.
```
f0k: > num_units is the size of the 2nd. dimension of the input

This was wrong, wasn't it?

benanne: in what sense? That period definitely shouldn't have been there.

f0k: > in what sense?

`num_units` vs. `num_inputs`

benanne: oh.... right :)

f0k:
Any other comments? Is this good to merge?

benanne:
I went through the diff just now and everything looks good to me. I didn't bother to check the entire code base for docstrings you may have forgotten to update, but that wouldn't really be a showstopper anyway. Goot to merge as far as I'm concerned!

f0k:
Okay, thanks! Will merge in the afternoon if there are no more comments.

PR #408, opened 2015-09-07
floatX cast spatial transformer width and height

diogo149:
this avoids errors when warn_float64 is set, since int / float = float64

### DIFF
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -243,8 +243,9 @@ def _transform(theta, input, downsample_factor):
     theta = T.reshape(theta, (-1, 2, 3))
 
     # grid of (x_t, y_t, 1), eq (1) in ref [1]
-    out_height = T.cast(height / downsample_factor[0], 'int64')
-    out_width = T.cast(width / downsample_factor[1], 'int64')
+    floatX = theano.config.floatX
+    out_height = T.cast(height.astype(floatX) / downsample_factor[0], 'int64')
+    out_width = T.cast(width.astype(floatX) / downsample_factor[1], 'int64')
     grid = _meshgrid(out_height, out_width)
 
     # Transform A x (x_t, y_t, 1)^T -> (x_s, y_s)


f0k:
But we don't care that it's float64 in between, we want that part to stay on CPU no matter what. I fear it might copy the shape to the GPU if we cast it to float32, then divide it by the scaling factor on GPU and copy it back to CPU -- I specifically removed the cast in #394 for that reason. I guess you shouldn't put too much emphasis on eliminating all float64 warnings!
Another idea: Does the warning also disappear if you cast the shape to float64? This wouldn't have any unintended side effects (except that it needlessly does a floating point operation in case you've got an integer scale factor).

diogo149:
Alright, the rationale kind of makes sense (: It seems weird to me that Theano couldn't do scalar arithmetic on the CPU, but sounds good.

When warn_float64 is set, casting anything to float64 triggers it.

f0k:
> It seems weird to me that Theano couldn't do scalar arithmetic on the CPU, but sounds good.

It could, but I *think* anything float32 may be copied over. We'd need to check if that's the case. If not, we can also merge your change, so the warning disappears (although it's a bit silly doing so just for the warning -- if float32 scalar operations are performed on CPU, maybe it shouldn't warn about a float64 scalar operation).

f0k:
> We'd need to check if that's the case.

Okay, float32 scalars are not turned into tensors and copied over (would have been silly in the first place anyway):
```
In [1]: import theano
Using gpu device 0: GeForce GT 640 (CNMeM is disabled)

In[2]: T = theano.tensor

In [3]: x = T.matrix()

In [4]: s = x.shape[0]

In [5]: theano.printing.debugprint(theano.function([s], (s / 1.5).astype('int64')))
Elemwise{Composite{Cast{int64}((i0 * i1))}} [@A] ''   0
 |TensorConstant{0.666666666667} [@B]
 |<TensorType(int64, scalar)> [@C]

In [6]: theano.printing.debugprint(theano.function([s], (s.astype('float32') / 1.5).astype('int64')))
Elemwise{Composite{Cast{int64}((i0 * Cast{float32}(i1)))}} [@A] ''   0
 |TensorConstant{0.666666686535} [@B]
 |<TensorType(int64, scalar)> [@C]
```
Note that `/ 1.5` gets turned into a multiplication with two thirds, but for the second case (that casts the shape to float32 first), the two thirds are less precise, because they're single precision. Fun.

When the downsample factor is an integer, the float32 cast makes things a little ugly:
```
In [22]: theano.printing.debugprint(theano.function([s], (s / 2).astype('int64')))                    
Elemwise{int_div,no_inplace} [@A] ''   0
 |<TensorType(int64, scalar)> [@B]
 |TensorConstant{2} [@C]

In [23]: theano.printing.debugprint(theano.function([s], (s.astype('float32') / 2).astype('int64')))
Elemwise{Composite{Cast{int64}((i0 * Cast{float32}(i1)))}} [@A] ''   0
 |TensorConstant{0.5} [@B]
 |<TensorType(int64, scalar)> [@C]
```
The float32 cast turns a simple integer division into a floating point multiplication with two type casts. Probably that won't ever lead to numerical problems with usual input sizes and sampling factors, but I'm leaning towards keeping things as they are. If we get reports of users getting confused by the float64 warning, we can think about merging it. (Unless anybody else has a different opinion?)

benanne:
sounds good, I also think we shouldn't be working around these warnings if we don't need to (especially since they are not enabled by default).

PR #409, opened 2015-09-08
Added an option for scaling of the tanh nonlinearity.

Botev:
I'm not sure how popular this is, but we do it a lot in my lab. It's a very synthetic change.
If it is accepted can implement similar for the sigmoid. Also usefull if you need directly your output to be in range different than [-1,1]

### DIFF
--- a/lasagne/nonlinearities.py
+++ b/lasagne/nonlinearities.py
@@ -46,18 +46,82 @@ def softmax(x):
 # tanh
 def tanh(x):
     """Tanh activation function :math:`\\varphi(x) = \\tanh(x)`
-
     Parameters
     ----------
     x : float32
         The activation (the summed, weighted input of a neuron).
-
     Returns
     -------
     float32 in [-1, 1]
         The output of the tanh function applied to the activation.
     """
     return theano.tensor.tanh(x)
+
+
+# scaled tanh
+class ScaledTanH(object):
+    """A scaled tanh :math:`\\varphi(x) = \\alpha \\tanh(\\beta x)`
+
+    This is a modified tanh function which allows to rescale both the input and
+    the output of the activation.
+
+    Scaling the input down will result in decreasing the maximum slope of the
+    tanh and as a result it will be in the "linear" mode in a larger interval
+    of the input space. Scaling the input up would increase the maximum slope
+    of the tanh and thus make bring it closer to a step function.
+
+    Scaling the output variable will make the output interval larger as well.
+
+    Parameters
+    ----------
+    x : float32
+        The activation (the summed, weighted input of a neuron).
+
+    scale_in : float32
+        The scale parameter :math:`\\beta` for the input
+
+    scale_out : float32
+        The scale parameter :math:`\\alpha` for the output
+
+    Returns
+    -------
+    float32 in `[-:math:\\alpha,:math:\\alpha]`
+        The output of the scaled tanh function applied to the activation.
+
+    Suggested values:
+        1, 1 - Standard tanh
+        1.7159, 0.6666 - Suggested in [1]
+        0.5, 2.4  - If the input is a random normal variable the output will
+        have a 0 mean and variance 1.
+        0.5, 2.27 - If the input is a uniform normal variable the output will
+        have a 0 mean and variance 1.
+        1, 1.6    -  If the input is a random normal variable the output will
+        have a 0 mean and variance 1.
+        1, 1.48   - If the input is a uniform normal variable the output will
+        have a 0 mean and variance 1.
+
+
+    References
+    ----------
+    .. [1] LeCun, Yann A., et al. (2012):
+       Efficient Backprop,
+       http://link.springer.com/chapter/10.1007/978-3-642-35289-8_3
+    .. [2] Masci, Jonathan, et al. (2011):
+       Stacked Convolutional Auto-Encoders for Hierarchical Feature Extraction,
+       http://link.springer.com/chapter/10.1007/978-3-642-21735-7_7
+    """
+
+    def __init__(self, scale_in=1, scale_out=1):
+        self.scale_in = scale_in
+        self.scale_out = scale_out
+
+    def __call__(self, x):
+        if self.scale_in != 1:
+            x = theano.tensor.mul(x, self.scale_in)
+        if self.scale_out != 1:
+            return theano.tensor.mul(self.scale_out, theano.tensor.tanh(x))
+        else:
+            return theano.tensor.tanh(x)
 
 
 # rectify
--- a/lasagne/tests/test_nonlinearities.py
+++ b/lasagne/tests/test_nonlinearities.py
@@ -22,20 +22,33 @@ class TestNonlinearities(object):
     def tanh(self, x):
         return np.tanh(x)
 
+    def scaled_tanh(self, x):
+        return np.tanh(x)
+
+    def scaled_tanh_p(self, x):
+        return 2.27 * np.tanh(0.5 * x)
+
     def softmax(self, x):
         return (np.exp(x).T / np.exp(x).sum(-1)).T
 
     @pytest.mark.parametrize('nonlinearity',
                              ['linear', 'rectify',
                               'leaky_rectify', 'sigmoid',
-                              'tanh', 'softmax',
-                              'leaky_rectify_0'])
+                              'tanh', 'scaled_tanh',
+                              'softmax', 'leaky_rectify_0',
+                              'scaled_tanh_p'])
     def test_nonlinearity(self, nonlinearity):
         import lasagne.nonlinearities
 
         if nonlinearity == 'leaky_rectify_0':
             from lasagne.nonlinearities import LeakyRectify
             theano_nonlinearity = LeakyRectify(leakiness=0)
+        elif nonlinearity == 'scaled_tanh':
+            from lasagne.nonlinearities import ScaledTanH
+            theano_nonlinearity = ScaledTanH()
+        elif nonlinearity == 'scaled_tanh_p':
+            from lasagne.nonlinearities import ScaledTanH
+            theano_nonlinearity = ScaledTanH(scale_in=0.5, scale_out=2.27)
         else:
             theano_nonlinearity = getattr(lasagne.nonlinearities,
                                           nonlinearity)


f0k:
It's a bit difficult to use this way, though, you'd need a lambda or functools.partial:
```python
layer = DenseLayer(..., nonlinearity=lambda x: lasagne.nonlinearities.tanh(x, 0.5, 1.5))
```
It would be easier to have a callable class like for `LeakyRectify`, and have it separate from the original `tanh`, maybe `ScaledTanh`. And then you'd also need tests, coveralls doesn't forgive any uncovered line for our project since we're at 100% coverage.

> I'm not sure how popular this is

That's the question, though. I'm not sure if this is useful to have in Lasagne -- maybe that's just something to keep in your personal toolbox. Is there any literature backing this?

Botev:
Mmm I'm not sure it will be easy to find literature on it, most of all is what we got from our experiments. 
The theoretical motivation is explained briefly in the comment, as if you want the scales of the objective function to be similar in all layers you don't want to contract your space. We have considered actually doing a more indepth study on this, but never got to there to publish something. Also one can view it as a simpler variant of Batch Normalization, which is something I found out very recently. I've added it now as a class. Gona have to check how to make tests etc... But well if you thing it is not needed feel free to reject the pull request.

benanne:
Section 4.4 in this (quite dated, but still mostly relevant) tech report by LeCun et al. mentions it: http://yann.lecun.com/exdb/publis/pdf/lecun-98b.pdf

f0k:
> Section 4.4 in this (quite dated, but still mostly relevant) tech report by LeCun et al. mentions it

Ah, yes, I've been citing that paper for another trick but didn't remember this one. Does that mean we'll include it in Lasagne? (With a paper reference I'd feel much better about it, at least.)

> The theoretical motivation is explained briefly in the comment, as if you want the scales of the objective function to be similar in all layers you don't want to contract your space.

But isn't that covered by scaling the weights appropriately (see the [`Glorot`](http://lasagne.readthedocs.org/en/latest/modules/init.html#lasagne.init.Glorot) and [`He`](http://lasagne.readthedocs.org/en/latest/modules/init.html#lasagne.init.He) initializers in Lasagne for code and paper references)? Can't you scale the weights instead of scaling the input and output of the nonlinearity? Or does it make the interactions between weight scale and transfer function easier to handle? How do you typically initialize weights in your lab?

benanne:
> But isn't that covered by scaling the weights appropriately (see the Glorot and He initializers in Lasagne for code and paper references)? Can't you scale the weights instead of scaling the input and output of the nonlinearity? 

No, changing the weight initialization by incorporating the scale factor will have the same effect for the forward pass, but the scale of the gradients is unaffected by this, whereas with the alternative tanh formulation, the scale of the gradient is more well-behaved. Supposedly :)

Botev:
So scaling the weights at a single time step (or let's say initialization) would be equivalent in terms of ouptut, but you need to keep that constant in place. Let me try to describe the problem without a blackboard. So if we have 2 layer network with `h=tanh(W_1 x+b)' and the second layer is `h_2=tanh(W_2 h + b)`. No if I rescale `W_2` by `C` this is equivalent to rescaling the output of `h` by `C`. This is true. But however, this means that the landscapes in the parameter space of W_2 and W_1 are going to be very different (e.g. if C=2 in general the values of W_2 would need to be in the twice larger values of W_1). This is bad for 2 reasons. First, any L2 or L1 loss in general discards this information. Unless you want to weight each layer separately, but than this gets even more complicates as you have more hyperparameters. Secondly, and which is the reason I use it for, is because the objective function space get's better conditioned. Here I'm sure you've seen the pictures with the convex circle bowl  the one which is squished in one direction. Essentially, having to scale up `W_2` to get the same output values means that the objective will have this kind bad conditioning where on each subsequent layer your parameters would have to be larger in order to get good outputs. Thus, rescaling the non linearity instead by a constant tries to address this problem. If you want to rescale parameters you should have something like `W_2 = W_2hat * C` to use where C is fixed and your true parameters to be `W_2hat`. Than they will be equivalent.

f0k:
> but the scale of the gradients is unaffected by this

Ah, sure. The gradient wrt. inputs would be correctly scaled, but the gradient wrt. weights wouldn't change.

> If you want to rescale parameters you should have something like `W_2 = W_2hat * C` to use where C is fixed and your true parameters to be `W_2hat`. Than they will be equivalent.

That will be possible soon, but changing the nonlinearity would be a lot easier. I'm fine with adding a scaled tanh then.

I'm unsure about the explanation and the default parameters, though -- should we just follow LeCun instead, or can you provide some details on how the factors were chosen in your lab?

f0k:
How/why does `py.test` fail on your machine? Did you follow [the instructions](http://lasagne.readthedocs.org/en/latest/user/development.html#development-setup)? I guess fixing this would make it a lot easier :)

Botev:
These ones are just derived by passing a random normal or uniform with mean 0 and variance 1 trough the `tanh` and emeprically measuring the correct constant (in fact for the uniform you can find analytical formula). 

As for the testing, so I did everything as the instruction. I can verify that my install is editable (as I already used it in other repository). Now you must excuse me cause earlier I was developing in Matlab and Rust so I'm not too familiar with the Python test chain. Currently I get:
`py.test: error: unrecognized arguments: --cov=lasagne --cov-report=term-missing --pep8`
Any idea how to fix it (if I happen to contribute to Lasagne again)?

benanne:
Looks like you need to install pytest-cov and pytest-pep8 (both installable via pip).

Botev:
Thanks got it working now :) 
Btw I saw there were a few threads for the BatchNormalization. What was the final outcome for that?

Botev:
Just while reading I came to this: http://link.springer.com/chapter/10.1007%2F978-3-642-21735-7_7, where in section 3 is said: `σ is an activation function (we used the scaled hyperbolic tangent in all our experiments)` as a second reference for this being used in practice.

f0k:
> These ones are just derived by passing a random normal or uniform with mean 0 and variance 1 trough the tanh and emeprically measuring the correct constant (in fact for the uniform you can find analytical formula).

Still there are multiple solutions (= combinations of alpha and beta) that give an output of mean 0 and variance 1, and you gave them for constraining beta=1 or beta=0.5. Is there any reason why beta shouldn't be something different? LeCun arrived at his combination of alpha and beta by constraining scaled_tanh(1) = 1 and retaining the variance, it seems. Hmm, no, checking myself, LeCun's recommendation of `tanh(2./3*x)*1.7159` does not retain the variance of a normal. What does it retain, actually?
I'm still a bit unsure about which factors we should have for the default `scaled_tanh`, and I think the docstring needs to explain a bit clearer how the numbers were arrived at. I also don't agree with "This tries to achieve what Batch Normalization does, but in a more simple version and with less computation." -- I think this goes too far, the objective of batch normalization is to counter the change in mean/variance caused by the network weights moving away from their well-controlled initialization, not only to counter the static influence of the nonlinearity.
So, in short, I'm fine with adding this feature, but we should reach consensus on what parameters to use for the default `scaled_tanh`, or whether we shouldn't have a default `scaled_tanh` at all.

> as a second reference for this being used in practice

Nice find! They don't mention the parameters they use, though.

> Btw I saw there were a few threads for the BatchNormalization. What was the final outcome for that?

The final outcome was that I'd file a PR for an implementation based on [my gist](https://gist.github.com/f0k/f1a6bd3c8585c400c190) with some changes we found by comparing alternative implementations. Haven't gotten around to that yet, though.

Botev:
I agree that the statement about BN was slightly over. Probably a more correct one is to say that it is aiming to make the input distributions to each layer to be on same scale, given that the weight parameters were drawn from the same distribution, which could be beneficial for deep networks. Now for the choice of beta, well that is something debatable. I personally, thing that `beta=1` is ok, however other people here prefer `0.5` because the slope of the sigmoidal is smaller, which gives it higher range for "linear" mode. I do not have too much emperical evidence for or against the other, so I can't claim anything. Potentially, if I get the K80's to work soon, cause we have some problems, I intend to do some evaluation of different optimization methods + such tricks in order to gain more evidence what is good and what makes no difference. Maybe than I can give some more sensible feedback.

Otherwise for fixed `beta` I can give some derivation for what can be done analytically and give to higher precision the correct constants on the sampling side.

Also, I need in fact to implement BN, so I was only thinking at the moment how to incorporate the mean and variance as being able to change state... Do you mind if I incorporate your code and maybe do some alternation?

f0k:
> I do not have too much emperical evidence for or against the other, so I can't claim anything.

So maybe we shouldn't have a default `scaled_tanh` until we found some plausible choice for `beta`? Right now the only choice backed by literature is LeCun's, but it doesn't retain the variance of a normal or uniform distribution. I'm wary of adding something to the library that's neither obviously a good choice nor backed by literature (`ScaledTanh` sure is, but `scaled_tanh` isn't).

> Do you mind if I incorporate your code and maybe do some alternation?

Not sure what you mean? Feel free to download the code and use it for your experiments, but don't include it in this PR. I see that your PR is from your `master` branch, so make sure to use a different branch if you need to change anything in your clone of Lasagne (it's usually easier to just copy the extra code somewhere outside of the repository anyway, you don't need to change your Lasagne clone to be able to use batch normalization).

Botev:
If I drop down the default values would that be ok for a merge as currently I there is no backing up by literature of correct values.

lasagne/nonlinearities.py
```diff
@@ -46,12 +46,10 @@ def softmax(x):
 # tanh
 def tanh(x):
     """Tanh activation function :math:`\\varphi(x) = \\tanh(x)`
-
     Parameters
     ----------
     x : float32
         The activation (the summed, weighted input of a neuron).
-
```
f0k: Please leave those empty lines in place!

f0k:
> If I drop down the default values would that be ok for a merge as currently I there is no backing up by literature of correct values.

Yes, dropping the default `scaled_tanh` would be good, and setting the default values in `ScaledTanh` to `scale_x=1, scale_out=1`. Maybe renaming `scale_x` to `scale_in` because it's not immediately clear what `x` refers to unless you're looking at the signature of `__call__` (i.e., when you use `ScaledTanh` in your code, there's no `x`).
The docstring should still be changed, but I can send you a PR for a better one if you don't have the time to work on it. Just let me know when you're done otherwise (github doesn't send notifications for commits on PRs, only for comments).

Botev:
Ok so done that. Feel free to write a better comment and give me a PR.

Also where would be appropriate to ask if you are interested in adding some stuff to the library (without making 10 pull requests). And still thumbs up for the library very nice and easy to use!

f0k:
> Ok so done that.

Nice, thank you! Could you also address https://github.com/Lasagne/Lasagne/pull/409#discussion_r39526174 and then squash your commits into one? (Commit everything, then do `git reset --soft 400b7d8; git commit --amend; git push --force`.)

> Feel free to write a better comment and give me a PR.

I'll do that after you squashed things (it's much easier than squashing afterwards).

> Also where would be appropriate to ask if you are interested in adding some stuff to the library (without making 10 pull requests)

Our [mailing list](https://groups.google.com/forum/#!forum/lasagne-users )! It's called lasagne-users, but that includes anything users do with the library, even extending it (we don't have a separate lasagne-dev list). We're always happy about additional contributors, even if we may seem a bit picky with pull requests (to maintain 100% test coverage and keep a good style of code and documentation).

Botev:
Closed in favour of Lasagne/Lasagne#430

PR #410, opened 2015-09-10
Migrate Travis config to container-based infrastructure.

dnouri:
Also remove use of Conda.  The new pip with a wheels cache should be
just as fast.

Gets rid of the Travis warning "This job ran on our legacy infrastructure..."

It's a work in progress because I still have to see that it's actually fast enough.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,25 @@
 language: python
+sudo: false
 python:
   - "2.7"
   - "3.4"
+addons:
+  apt:
+    packages:
+    - libblas-dev
+    - liblapack-dev
+    - gfortran
 before_install:
-  - sudo apt-get update -qq
-  - sudo apt-get install -qq libatlas-dev libatlas-base-dev liblapack-dev gfortran
-  - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
-  - chmod +x miniconda.sh
-  - ./miniconda.sh -b
-  - export PATH=/home/travis/miniconda/bin:$PATH
-  - conda update --yes conda
+  - pip install -U pip
 install:
-  - travis_retry conda install --yes python=$TRAVIS_PYTHON_VERSION pip numpy scipy
-  - travis_retry pip install -r requirements-dev.txt
+  - travis_wait travis_retry pip install -r requirements-dev.txt
   - travis_retry pip install python-coveralls
   - travis_retry python setup.py dev
 script: py.test --runslow --cov-config=.coveragerc-nogpu
 after_success:
   - coveralls
-cache: apt
+cache:
+  - apt
+  - directories:
+    - $HOME/.cache/pip
+    - $HOME/.theano


.travis.yml
```diff
@@ -1,21 +1,25 @@
...
   - "3.4"
+addons:
+  apt:
+    packages:
+    - libblas-dev
+    - liblapack-dev
+    - gfortran
 before_install:
-  - sudo apt-get update -qq
-  - sudo apt-get install -qq libatlas-dev libatlas-base-dev liblapack-dev gfortran
```
f0k: `libatlas-dev` and `libatlas-base-dev` are replaced by `libblas-dev`? That would be a slower BLAS library, wouldn't it? What about `libopenblas-dev`?

.travis.yml
```diff
@@ -1,21 +1,25 @@
...
   - travis_retry python setup.py dev
 script: py.test --runslow --cov-config=.coveragerc-nogpu
 after_success:
   - coveralls
-cache: apt
+cache:
+  - apt
+  - directories:
+    - $HOME/.cache/pip
+    - $HOME/.theano
```
f0k: Cool!

f0k:
> It's a work in progress because I still have to see that it's actually fast enough.

The way to check it would probably to trigger another run, right? And check that both installing the requirements and running the tests is a lot faster? Feel free to either close/reopen the PR, amend the commit or add another commit to try.

dnouri:
So the Theano cache makes a big difference: with Python 2 it's now down from 15 minutes to 5 minutes.  Python 3 is still wayy slower at 14 minutes (from ~22).  Need to find some time to see check what's up with that.

Regarding the BLAS library: this is a combination that I've found to work.  Trying to go back to ATLAS may be worthwhile, but we'll have to remember to purge the Travis cache when we do that, since the cached scipy wheels will no longer work.

f0k:
> So the Theano cache makes a big difference: with Python 2 it's now down from 15 minutes to 5 minutes.

Cool! We probably don't really need to look into ATLAS/OpenBLAS then. I wouldn't expect a large improvement.

> Python 3 is still wayy slower at 14 minutes (from ~22). Need to find some time to see check what's up with that.

What about profiling the tests once? Do you have the time to trigger that, in a second commit? https://pytest.org/latest/example/simple.html#profiling-test-duration

dnouri:
Added `--duration=10`.  Here's the results for Python 3:

```
190.84s call     lasagne/tests/test_examples.py::test_example[recurrent]
97.85s call     lasagne/tests/test_examples.py::test_example[mnist]
19.77s call     lasagne/tests/layers/test_merge.py::TestAutocrop::test_crop_inputs
12.16s call     lasagne/tests/layers/test_recurrent.py::test_lstm_precompute
11.89s call     lasagne/tests/layers/test_recurrent.py::test_gru_precompute
11.67s call     lasagne/tests/layers/test_recurrent.py::test_gru_unroll_scan_bck
11.64s call     lasagne/tests/layers/test_recurrent.py::test_gru_unroll_scan_fwd
11.22s call     lasagne/tests/layers/test_recurrent.py::test_lstm_unroll_scan_fwd
10.55s call     lasagne/tests/layers/test_recurrent.py::test_lstm_bck
10.09s call     lasagne/tests/layers/test_recurrent.py::test_gru_bck
```

And for Python 2:

```
115.21s call     lasagne/tests/test_examples.py::test_example[recurrent]
87.68s call     lasagne/tests/test_examples.py::test_example[mnist]
13.49s call     lasagne/updates.py::lasagne.updates
9.14s call     lasagne/tests/layers/test_merge.py::TestAutocrop::test_crop_inputs
5.73s call     lasagne/tests/layers/test_recurrent.py::test_gru_unroll_scan_bck
5.54s call     lasagne/tests/layers/test_recurrent.py::test_lstm_precompute
5.23s call     lasagne/tests/layers/test_recurrent.py::test_gru_precompute
5.15s call     lasagne/tests/layers/test_recurrent.py::test_gru_unroll_scan_fwd
4.92s call     lasagne/tests/layers/test_recurrent.py::test_lstm_unroll_scan_fwd
4.78s call     lasagne/tests/layers/test_recurrent.py::test_lstm_bck
```

f0k:
Thanks! Unfortunately it shows that everything is slower, there's no single offender as I had hoped for. I don't have a plausible theory on what's wrong either. It could have been the compilation (e.g., maybe it takes longer to compile against `Python.h` of Python 3 for some reason), but that has been ruled out now that the compilation results are cached and it's still slower than Python 2. Very weird...

I guess we should just remove `--duration=10` again and merge it? It's still saving a lot of processor time, good job! We can look into OpenBLAS or ATLAS any time later.

dnouri:
Removed `--duration=10` again.

f0k:
Great, merging! Thanks for looking into this!

PS: For some reason, the Python 3 build is down to 10 minutes now! Yay!

benanne:
Nice work :)

PR #415, opened 2015-09-16
Conv1d slow compilation fix

ebenolson:
Attempts to address the bug discussed [here](https://groups.google.com/forum/#!topic/lasagne-users/XR-jmU1pMWc). 

The changes to `theano_extensions.conv` improved the optimization speed dramatically, but I think it's still triggering some underlying problem in Theano, as the compilation time increases nonlinearly with each additional layer.

The change to `Conv1DLayer` seems to be an effective workaround, but I need to profile and see how the execution speed compares.

### DIFF
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -249,7 +249,7 @@ class Conv1DLayer(Layer):
                                       filter_shape=self.get_W_shape(),
                                       border_mode='full')
             shift = (self.filter_size[0] - 1) // 2
-            conved = conved[:, :, shift:input.shape[2] + shift]
+            conved = conved[:, :, shift:-shift]
         else:
             # no padding needed, or explicit padding of input needed
             if self.pad == 'full':
--- a/lasagne/theano_extensions/conv.py
+++ b/lasagne/theano_extensions/conv.py
@@ -37,7 +37,11 @@ def conv1d_sc(input, filters, image_shape=None, filter_shape=None,
     conved = T.nnet.conv2d(input_sc, filters_sc, image_shape=image_shape_sc,
                            filter_shape=filter_shape_sc,
                            subsample=(1, subsample[0]))
-    return conved[:, :, 0, :]  # drop the unused dimension
+    try:
+        conved = conved.dimshuffle(0, 1, 3)  # drop the unused dimension
+    except ValueError:
+        conved = T.addbroadcast(conved, 2).dimshuffle(0, 1, 3)
+    return conved
 
 
 def conv1d_mc0(input, filters, image_shape=None, filter_shape=None,
@@ -64,7 +68,11 @@ def conv1d_mc0(input, filters, image_shape=None, filter_shape=None,
         input_mc0, filters_mc0, image_shape=image_shape_mc0,
         filter_shape=filter_shape_mc0, subsample=(1, subsample[0]),
         border_mode=border_mode)
-    return conved[:, :, 0, :]  # drop the unused dimension
+    try:
+        conved = conved.dimshuffle(0, 1, 3)  # drop the unused dimension
+    except ValueError:
+        conved = T.addbroadcast(conved, 2).dimshuffle(0, 1, 3)
+    return conved
 
 
 def conv1d_mc1(input, filters, image_shape=None, filter_shape=None,
@@ -91,7 +99,11 @@ def conv1d_mc1(input, filters, image_shape=None, filter_shape=None,
         input_mc1, filters_mc1, image_shape=image_shape_mc1,
         filter_shape=filter_shape_mc1, subsample=(subsample[0], 1),
         border_mode=border_mode)
-    return conved[:, :, :, 0]  # drop the unused dimension
+    try:
+        conved = conved.dimshuffle(0, 1, 2)  # drop the unused dimension
+    except ValueError:
+        conved = T.addbroadcast(conved, 3).dimshuffle(0, 1, 2)
+    return conved
 
 
 def conv1d_unstrided(input, filters, image_shape, filter_shape,


benanne:
I'm uncomfortable with making our code more verbose just to work around a bug in Theano. The most desirable solution would be to fix this in Theano so we don't have to make things more convoluted on our side.

lasagne/layers/conv.py
```diff
@@ -249,7 +249,7 @@ def get_output_for(self, input, input_shape=None, **kwargs):
                                       filter_shape=self.get_W_shape(),
                                       border_mode='full')
             shift = (self.filter_size[0] - 1) // 2
-            conved = conved[:, :, shift:input.shape[2] + shift]
+            conved = conved[:, :, shift:-shift]
```
benanne: I haven't done the math, but I'm unsure whether this amounts to the same thing for every possible length. Have you? I wonder if there is maybe a reason that I wrote `input.shape[2] + shift` instead of `-shift` in the first place (but there might not be).

ebenolson: It won't work for even kernel sizes, but I don't think we were trying to support 'same' for those anyway.

The original way makes more sense to me, but it seems to really confuse Theano.

benanne: In that case we should probably make it explicit and have it raise an exception when the filter size is even.

f0k: > It won't work for even kernel sizes, but I don't think we were trying to support 'same' for those anyway.

I think we do. For the cases we don't, it's explicitly mentioned in the code (but I'm not sure about the documentation, we should document that and give a warning if we can).

As an alternative fix, what about using `input_shape[2] or input.shape[2]` instead of `input.shape[2]`? That will use the non-symbolic shape if it's not `None` (or `0`, but I think we don't need to distinguish that case). Maybe this would work just as well? If it does, I guess `Conv2DLayer` would need the same?

lasagne/theano_extensions/conv.py
```diff
@@ -37,7 +37,11 @@ def conv1d_sc(input, filters, image_shape=None, filter_shape=None,
     conved = T.nnet.conv2d(input_sc, filters_sc, image_shape=image_shape_sc,
                            filter_shape=filter_shape_sc,
                            subsample=(1, subsample[0]))
-    return conved[:, :, 0, :]  # drop the unused dimension
+    try:
+        conved = conved.dimshuffle(0, 1, 3)  # drop the unused dimension
+    except ValueError:
+        conved = T.addbroadcast(conved, 2).dimshuffle(0, 1, 3)
+    return conved
```
f0k: Maybe add a comment that this is faster than the original, more straight-forward solution? Otherwise somebody in the future might file a PR "simplify conv1D code" that reverts it :)

f0k: Also I think it wouldn't hurt to be a little more explicit:
```python
# drop the unused dimension (if needed, mark it broadcastable before)
if not conved.broadcastable[2]:
    conved = T.addbroadcast(coved, 2)
conved = conved.dimshuffle(0, 1, 3)
```

ebenolson: Yeah, I'm not sure if this is even worth keeping since it wasn't a complete fix... I need to profile and see if it executes any faster (the generated graph does look a bit simpler).

lasagne/layers/conv.py
ebenolson: I'm not sure how to even define a 'same' convolution for even kernels... it's not covered by the tests, and the current implementations fais also. 

Raising warnings / fixing docs should be done though.

ebenolson: > As an alternative fix, what about using input_shape[2] or input.shape[2] instead of input.shape[2]? That will use the non-symbolic shape if it's not None (or 0, but I think we don't need to distinguish that case). Maybe this would work just as well? If it does, I guess Conv2DLayer would need the same?

I think this could work, I'll try profiling both ways later.

f0k: > I'm not sure how to even define a 'same' convolution for even kernels...

We've stated in the documentation how we define it:
```
'same' pads with half the filter size on both sides (one less on the second side for an even filter size). When stride=1, this results in an output size equal to the input size.
```

> it's not covered by the tests, and the current implementations fail also.

Oh yes, that's something were our coverage reporting fails, because it only checks whether each line is touched. We should test it, and have convolution implementations that don't support it raise a `NotImplementedError`.

In Theano's GPUCorrMM, I've added a padding mode `"half"` that always pads with half the filter size.

> I think this could work, I'll try profiling both ways later.

Alternatively, you could have a `left = (self.filter_size[0] // 2) or None` and `right = -(self.filter_size[1] - left) or None`, then do `conved = conved[:, :, left:right, :]`, something like that? I think I planned to do that in #374, but then I remember it turned out to be more complicated than anticipated -- or maybe that was for the Conv2DDNNLayer et al?

ebenolson: Ah yes, this discussion is coming back to me now... I was basing my opinion on a vague memory of [this comment](https://github.com/Lasagne/Lasagne/blob/102203f8ef23ae573bb20783cbe7cf208e9d1f8c/lasagne/layers/dnn.py#L262), but it seems like at some point that changed. 

I can start a new PR to update the tests and see what else needs to be fixed, but probably won't have time today.

f0k: > but it seems like at some point that changed.

It's still the same, but it's only in `Conv2D??*Layer`, not in the original `Conv2DLayer`. To support it in `Conv2DDNNLayer` et al., we'd need to pad a little more than needed and crop the result, but right now we don't even remember `pad="same"` outside of the constructor. The easiest fix would be raising a `NotImplementedError` for even filter sizes.

nouiz:
Just to let know that we merged in Theano some change to speed up compilation related to cudnn. I also merged this week a few other compilation speed up again regression on that. Maybe an update of Theano would fix the speed issue.

ebenolson:
Hi @nouiz, I still experience this problem with the latest Theano. I've opened [an issue there](https://github.com/Theano/Theano/issues/3412), as I don't know enough to debug it further.

nouiz:
Just to let know that I have a PR https://github.com/Theano/Theano/pull/3477 with a fix for the crash and the slowdown. The use case reported in Theano is 300x faster to optimizer.

PR #416, opened 2015-09-16
Change MNIST example to pull dataset from LeCun's website

### LINKED ISSUE #414: MNIST example: Pull data from original website
As discussed in https://github.com/Lasagne/Recipes/pull/21, it's not a terribly good idea to have the MNIST example depend on the pickled MNIST dataset from deeplearning.net: (a) the website is down sometimes, causing Travis to fail for any pull requests submitted during that time, (b) asking users to download and unpickle a pickle is not a good practice ([anything could happen](http://www.cs.uic.edu/~s/musings/pickle.html) on unpickling a pickle).

We should change the MNIST example to download and load the data from Yann's original website: http://yann.lecun.com/exdb/mnist/
It's relatively easy to read using a combination of `gzip.open`, `np.fromfile` and `np.reshape`, you just need the correct dtype, offset (to skip the header) and shape.

I also wrote a generic MNIST format reader some time which can be used as a reference (it can be simplified a lot since we can just hard-code the shape and dtype):
```python
def read_mnistfile(fn):
    with gzip.open(fn, 'rb') as f:
        # read header
        zeros = struct.unpack("<H", f.read(2))
        magic, ndim = struct.unpack("<2B", f.read(2))
        # read matrix shape
        shape = list(struct.unpack(">%dL" % ndim, f.read(ndim * 4)))
        # create matrix
        types = {\
                0x08: np.uint8,
                0x09: np.int8,
                0x0B: np.int16,
                0x0C: np.int32,
                0x0D: np.float32,
                0x0E: np.float64
        }
        dtype = types.get(magic, None)
        if dtype is None:
            raise ValueError("This is not a valid MNIST matrix file.")
        mat = np.frombuffer(f.read(), dtype).reshape(shape)
    return mat
```

f0k:
This changes the MNIST example to download and use the original MNIST dataset as provided on Yann LeCun's website. This makes the tests independent of whether deeplearning.net is online, and it removes the possibility of somebody sneaking in a malicious pickle file at deeplearning.net that executes arbitrary code. Closes #414.

(If anybody is wondering, the data is exactly equivalent to what we had before, at least according to `np.allclose`.)


skaae:
LGTM. The conversion from binary is pretty elegant :)

examples/mnist.py
```diff
@@ -30,48 +30,52 @@
...
-    y_train = y_train.astype(np.uint8)
-    y_val = y_val.astype(np.uint8)
-    y_test = y_test.astype(np.uint8)
+
+    def load_mnist_images(filename):
+        if not os.path.exists(filename):
+            download(filename)
+        # Read the inputs in Yann LeCun's binary format.
+        with gzip.open(filename, 'rb') as f:
+            data = np.frombuffer(f.read(), np.uint8, offset=16)
```
f0k: By the way, if you're wondering, `f.seek(16); np.fromfile(f, np.uint8)` doesn't work. `fromfile` expects a filename or an open file, a file-like object such as a `GzipFile` is not accepted. But it's not a lot of data to copy around anyway.

Issue #417, opened 2015-09-16
Docs failing to build


Kaixhin:
Was just checking on the documentation and realised it hasn't been building properly [for weeks](https://readthedocs.org/projects/lasagne/builds/). [This](https://readthedocs.org/projects/lasagne/builds/3263592/) is the last build to have passed, and [this](https://readthedocs.org/projects/lasagne/builds/3268724/) is the first build to have failed.

benanne:
The badge is still green for whatever reason. What happened 3 weeks ago? Seems like `lasagne` fails to import.

f0k:
> Was just checking on the documentation and realised it hasn't been building properly for weeks.

Oh!

> The badge is still green for whatever reason.

Not surprising, the image is static: http://img.shields.io/badge/docs-latest-brightgreen.svg
Is there a dynamic version somewhere?

> What happened 3 weeks ago?

The first build that failed is the one that merged #393. That one only changed a single word in a docstring, this can't be the reason. Did readthedocs change anything?

f0k:
> Did readthedocs change anything?

I diffed the two log files, and there are some changes.

Completed build:
```
Successfully installed Jinja2-2.7.2 Markdown-2.6.2 MarkupSafe-0.18 PyYAML-3.10 Pygments-2.0.2 alabaster-0.7.6 babel-2.0 backports.ssl-match-hostname-3.4.0.2 certifi-2015.4.28 click-5.1 commonmark-0.5.4 docutils-0.11 livereload-2.4.0 mkdocs-0.14.0 mock-1.0.1 nilsimsa-0.3.7 pillow-2.3.0 pytz-2012rc0 readthedocs-sphinx-ext-0.5.4 recommonmark-0.1.1 requests-2.2.1 setuptools-3.3 six-1.5.2 snowballstemmer-1.2.0 sphinx-1.3.1 sphinx-rtd-theme-0.1.8 tornado-4.2.1 virtualenv-13.1.0
```
Failed build:
```
[...]

PR #418, opened 2015-09-17
Fix Read the Docs badge

Kaixhin:
Does what it says. Make detecting https://github.com/Lasagne/Lasagne/issues/417 a little easier.


f0k:
:+1:! But maybe we should merge this after the problem is fixed :) I'll have a look later.

f0k:
Problem fixed, merging! Thanks a lot!

PR #412, opened 2015-09-14
Add Docker images to docs

Kaixhin:
I've added information on the Docker images to the docs, which closes https://github.com/Lasagne/Lasagne/issues/411.

I have also successfully run the test suite on both CPU and CUDA containers without errors.

PR #419, opened 2015-09-17
Fix readthedocs build

### LINKED ISSUE #417: Docs failing to build

f0k:
This hopefully fixes #417 (and it fixes markup errors introduced in #412, so we're down to zero Sphinx warnings again).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,13 +12,14 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-# import sys
+import sys
 import os
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 # sys.path.insert(0, os.path.abspath('.'))
+sys.path.insert(0, os.path.abspath('..'))  # make lasagne available for import
 
 # -- General configuration ------------------------------------------------
 
--- a/docs/user/installation.rst
+++ b/docs/user/installation.rst
@@ -231,9 +231,9 @@ Docker images
 =============
 
 Instead of manually installing Theano and Lasagne on your machines as described
-above, you may want to use a pre-made `Docker<https://www.docker.com/whatisdocker>`_
-image: `Lasagne Docker (CPU)<https://hub.docker.com/r/kaixhin/lasagne/>`_ or
-`Lasagne Docker (CUDA)<https://hub.docker.com/r/kaixhin/cuda-lasagne/>`_. These
+above, you may want to use a pre-made `Docker <https://www.docker.com/whatisdocker>`_
+image: `Lasagne Docker (CPU) <https://hub.docker.com/r/kaixhin/lasagne/>`_ or
+`Lasagne Docker (CUDA) <https://hub.docker.com/r/kaixhin/cuda-lasagne/>`_. These
 are updated on a weekly basis with bleeding-edge builds of Theano and Lasagne.
 Examples of running bash in a Docker container are as follows:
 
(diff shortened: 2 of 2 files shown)


f0k:
The only way to know whether it's good is to try!

f0k:
Nah, doesn't work: https://readthedocs.org/projects/lasagne/builds/3331105/
It gets past the `import lasagne` line now, but then `__init__.py` can't set the `__version__` because Lasagne is not actually installed:
```
  File "/home/docs/checkouts/readthedocs.org/user_builds/lasagne/checkouts/latest/lasagne/__init__.py", line 28, in <module>
    __version__ = pkg_resources.get_distribution("Lasagne").version
  File "/home/docs/checkouts/readthedocs.org/user_builds/lasagne/envs/latest/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 551, in get_distribution
    dist = get_provider(dist)
  File "/home/docs/checkouts/readthedocs.org/user_builds/lasagne/envs/latest/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 431, in get_provider
    return working_set.find(moduleOrReq) or require(str(moduleOrReq))[0]
  File "/home/docs/checkouts/readthedocs.org/user_builds/lasagne/envs/latest/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 952, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/home/docs/checkouts/readthedocs.org/user_builds/lasagne/envs/latest/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 839, in resolve
    raise DistributionNotFound(req, requirers)
DistributionNotFound: The 'Lasagne' distribution was not found and is required by the application
```

It looks like readthedocs used to install lasagne before running Sphinx, and now it ceased to do that. Any idea why?

f0k:
Looking at the documentation again, this time at https://read-the-docs.readthedocs.org/en/latest/builds.html#understanding-what-s-going-on:
> If you have the Use Virtualenv option enabled, we will run setup.py install on your package, installing it into a virtual environment. You can also define additional packages to install with the Requirements File option.
> When we build your documentation, we run sphinx-build -b html . _build/html, where html would be replaced with the correct backend. We also create man pages and pdf’s automatically based on your project.

Do we have that option enabled? @benanne is the only one who's got access to this, right?

Another thing further down:
> If you’re having trouble getting your version to build, try wiping out the existing build/environment files. On your version list page /projects/[project]/versions there is a “Wipe” button that will remove all of the files associated with your documentation build, but not the documentation itself.

@benanne, could you also try that?

benanne:
The virtualenv option is disabled. I seem to vaguely recall that I couldn't get this to work when we started to use readthedocs. But I might be misremembering. I'll try to enable it and see what happens.

The second thing doesn't seem applicable, you've clearly identified the problem: it's not installing Lasagne, it's trying to import it from a clone (which used to work, but no longer does because of the pkg_resources stuff). but I guess I can try it anyway :)

f0k:
> The second thing doesn't seem applicable, you've clearly identified the problem: it's not installing Lasagne, it's trying to import it from a clone

Although I don't see how it was ever going to try importing from a clone before this very PR. I really don't understand why it was ever working properly. (I've signed up on readthedocs as well now, maybe you can make me a co-owner or something -- I don't know how they structure their projects.)

f0k:
Thanks, it works! https://readthedocs.org/projects/lasagne/builds/3331150/

I'm going to revert the `sys.path` addition, as it's not working anyway and only leads to a more cryptic error message if you don't install lasagne before building the docs.

PR #422, opened 2015-09-19
Adding an LSTM Text Generation Example

keskarnitish:
This adds an example which trains a 2 layered LSTM on a user-provided text file and uses that to generate text. It is partly based on [Andrej Karpathy's blog](karpathy.github.io/2015/05/21/rnn-effectiveness/) and his accompanying file [min-char-rnn.py](https://gist.github.com/karpathy/d4dee566867f8291f086) . When I playing around with [Keras] (keras.io), I noticed it had a similar example and wrote one which uses Lasagne. I realize that Lasagne already has an example for RNNs (The addition problem), but I felt this might be an interesting addition. I have commented it in attempts to make it self-explanatory and have tested this on both CPU and GPU.


skaae:
This look like a nice addition :) 

We keep the larger examples in https://github.com/Lasagne/Recipes/tree/master/examples

Can you move the pull request there and we'll review it?

keskarnitish:
Sure! I'll close this request and start a new one in Lasagne/Recipes.

skaae:
Great thanks.

moscow25:
That's awesome! So glad you're doing this. Thanks @keskarnitish!

PR #423, opened 2015-09-21
Pass kwargs to CustomRecurrentLayer child layers

### LINKED ISSUE #421: Inconsistent passing of **kwargs to CustomRecurrentLayer's child layers
`CustomRecurrentLayer` doesn't pass `kwargs` to `hidden_to_hidden` when calling `get_output`:
https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/recurrent.py#L387
It does for in_to_hid when `precompute_input` is `True`:
https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/recurrent.py#L370
but not when `precompute_input` is `False`:
https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/recurrent.py#L394

Intuitively, I'd think it should always pass `**kwargs` to its child layers, but I have a vague feeling there may be a case when the user wouldn't want it to.  Is there?  If not, I will fix this and make it pass them.

craffel:
Resolves #421

### DIFF
--- a/lasagne/tests/layers/test_recurrent.py
+++ b/lasagne/tests/layers/test_recurrent.py
@@ -1,11 +1,13 @@
 import pytest
+
 from lasagne.layers import RecurrentLayer, LSTMLayer, CustomRecurrentLayer
-from lasagne.layers import InputLayer, DenseLayer, GRULayer, Gate
+from lasagne.layers import InputLayer, DenseLayer, GRULayer, Gate, Layer
 from lasagne.layers import helper
 import theano
 import theano.tensor as T
 import numpy as np
 import lasagne
+from mock import Mock
 
 
 def test_recurrent_return_shape():
@@ -832,3 +834,40 @@ def test_unroll_none_input_error():
 
     with pytest.raises(ValueError):
         GRULayer(l_in, 5, unroll_scan=True)
+
+
+def test_CustomRecurrentLayer_child_kwargs():
+    in_shape = (2, 3, 4)
+    n_hid = 5
+    # Construct mock for input-to-hidden layer
+    in_to_hid = Mock(
+        Layer,
+        output_shape=(in_shape[0]*in_shape[1], n_hid),
+        input_shape=(in_shape[0]*in_shape[1], in_shape[2]),
+        input_layer=InputLayer((in_shape[0]*in_shape[1], in_shape[2])))
+    # These two functions get called, need to return dummy values for them
+    in_to_hid.get_output_for.return_value = T.matrix()
+    in_to_hid.get_params.return_value = []
+    # As above, for hidden-to-hidden layer
+    hid_to_hid = Mock(
+        Layer,
+        output_shape=(in_shape[0], n_hid),
+        input_shape=(in_shape[0], n_hid),
+        input_layer=InputLayer((in_shape[0], n_hid)))
+    hid_to_hid.get_output_for.return_value = T.matrix()
+    hid_to_hid.get_params.return_value = []
+    # Construct a CustomRecurrentLayer using these Mocks
+    l_rec = lasagne.layers.CustomRecurrentLayer(
+        InputLayer(in_shape), in_to_hid, hid_to_hid)
+    # Call get_output with a kwarg, should be passd to in_to_hid and hid_to_hid
+    helper.get_output(l_rec, foo='bar')
+    # Retrieve the arguments used to call in_to_hid.get_output_for
+    args, kwargs = in_to_hid.get_output_for.call_args
+    # Should be one argument - the Theano expression
+    assert len(args) == 1
+    # One keywould argument - should be 'foo' -> 'bar'
+    assert kwargs == {'foo': 'bar'}
+    # Same as with in_to_hid
+    args, kwargs = hid_to_hid.get_output_for.call_args
+    assert len(args) == 1
+    assert kwargs == {'foo': 'bar'}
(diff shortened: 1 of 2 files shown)


f0k:
Looks good, could you add a test as well? Probably just whip up a `CustomRecurrentLayer` with a `Mock(Layer, output_shape=some_shape, input_layer=some_input_layer)` instance for `in_to_hid` and `hid_to_hid`, call `get_output_for(foo='bar')` on the recurrent layer and assert that the two mocks had `get_output_for` called with that keyword argument. That is, just make it quick and easy, do the minimum possible to satisfy the algorithm in `helper.get_output()` :)

craffel:
Sorry, I won't get around to making tests til after my next paper deadline (next week).

craffel:
Sorry for the delay on this. 

I'm new to Mock, and I'm not seeing how to get it to work here.  Using
```Python
def test_CustomRecurrentLayer_child_kwargs():
    in_shape = (2, 3, 4)
    n_hid = 5
    in_to_hid_in = InputLayer((in_shape[0]*in_shape[1], in_shape[2]))
    in_to_hid = Mock(
        Layer,
        output_shape=(in_shape[0]*in_shape[1], n_hid),
        input_shape=(in_shape[0]*in_shape[1], in_shape[2]),
        input_layer=in_to_hid_in)
    hid_to_hid_in = InputLayer((in_shape[0], n_hid))
    hid_to_hid = Mock(
        Layer,
        output_shape=(in_shape[0], n_hid),
        input_shape=(in_shape[0], n_hid),
        input_layer=hid_to_hid_in)
    l_rec = lasagne.layers.CustomRecurrentLayer(
        InputLayer(in_shape), in_to_hid, hid_to_hid)
    helper.get_output(l_rec, foo='bar')
    in_to_hid.get_output_for.assert_called_with(foo='bar')
    hid_to_hid.get_output_for.assert_called_with(foo='bar')
```

the test fails with

```Python
            # Reshape back to (seq_len, batch_size, trailing dimensions...)
>           trailing_dims = tuple(input.shape[n] for n in range(1, input.ndim))
E           TypeError: range() integer end argument expected, got Mock.
```

The `Mock` object needs to work so that `helper.get_output(mock_object, input, **kwargs)` returns something which has an attribute `ndim` which is an `int`, but it doesn't, it just returns a `Mock`.  I tried it with the `Mock` using a `DenseLayer` as a spec too, and that had the same issue.  So I think I can only make it work if I re-define `in_to_hid.get_output_for` within the test, or something, which seems like a bad idea.  Is there any simpler way to use `Mock` where you just decorate a method of a class instance so you can check whether the method was called with a given call spec?

f0k:
> Is there any simpler way to use Mock where you just decorate a method of a class instance so you can check whether the method was called with a given call spec?

You can't simply decorate a class method, I think, but you can tell the mocked functions what to return: `in_to_hid.get_output_for.return_value = something`. This way you can tell them to return something else than a `Mock`, and still use `assert_called_with()`.

craffel:
Ok, added a test.  It ended up being kind of long-winded, because a lot of the child layer's functionality gets used in the course of a `get_output` call of the parent layer.  Also, it seemed better to use `call_args` instead of 'assert_called_with', because `get_output_for` is also called with a Theano expression, which would be silly to try to reconstruct.

lasagne/tests/layers/test_recurrent.py
```diff
@@ -1,12 +1,13 @@
...
 from lasagne.layers import RecurrentLayer, LSTMLayer, CustomRecurrentLayer
-from lasagne.layers import InputLayer, DenseLayer, GRULayer, Gate
+from lasagne.layers import InputLayer, DenseLayer, GRULayer, Gate, Layer
 from lasagne.layers import helper
 import theano
 import theano.tensor as T
 import numpy as np
 import lasagne
-
+from mock import Mock
```
f0k: You removed a newline here. pep8 is offended! (It requires two empty lines in front of every function definition, including the first one.)

lasagne/tests/layers/test_recurrent.py
```diff
@@ -832,3 +833,44 @@ def test_unroll_none_input_error():
...
+    # Call get_output with a kwarg, should be passd to in_to_hid and hid_to_hid
+    helper.get_output(l_rec, foo='bar')
+    # Retrieve the arguments used to call in_to_hid.get_output_for
+    args, kwargs = in_to_hid.get_output_for.call_args
+    # Should be one argument - the Theano expression
+    assert len(args) == 1
+    # One keywould argument - should be 'foo' -> 'bar'
+    assert len(kwargs) == 1
+    assert 'foo' in kwargs
+    assert kwargs['foo'] == 'bar'
```
f0k: Can't you just test `assert kwargs == {'foo': 'bar'}`?

f0k:
It's a pain that Mocks tend up getting so complicated if they need to replicate a lot of behaviour... I don't know a better way, though, and the test looks good except for my comments. I'm fine with merging if you address them (and squash commits)!

lasagne/tests/layers/test_recurrent.py
craffel: The reason I did it with three asserts is that if `'foo'` is not in `kwargs` the test will fail with a `KeyError` rather than a test failure. So, it's more descriptive this way.

lasagne/tests/layers/test_recurrent.py
craffel: Oops.  The tests have lots of PEP-8/pyflakes warnings on my system, so I overlooked this in the midst of the others.  Fixed it.

craffel:
> I'm fine with merging if you address them (and squash commits)!

Ok, done.

f0k:
Perfect, merging! Thank you!

Issue #141, opened 2015-02-16
Batch Normalization


skaae:
Moved from https://github.com/benanne/Lasagne/issues/133

diogo149:
Something relevant that I referenced in https://github.com/benanne/Lasagne/issues/136: "If batch normalization mean/variance counts as a param, then SGD update code won't work with it. If it doesn't count as a param, users relying on get_all_params for saving model state will have their code break (because the mean/variance won't be saved)."

ebattenberg:
> Something relevant that I referenced in #136: "If batch normalization mean/variance counts as a param, then SGD update code won't work with it. If it doesn't count as a param, users relying on get_all_params for saving model state will have their code break (because the mean/variance won't be saved)."

I thought the mean/variance were computed for each batch, so they would be intermediate variables and not shared variables/params.  At test time though, they become parameters, but not learned parameters, more like arguments to the function computing the outputs.  

The beta and gamma parameters (from the paper) that transform the normalized outputs *are* learned parameters though.

diogo149:
I did mean at test time, sorry

ebattenberg:
I think that the mean/variance of each unit across the entire training set are pretty specialized parameters in that they have to be computed iteratively layer-by-layer before test time and then communicated to the network at test time.  Because this is different than what is done during training, I'm not sure it makes sense to support this case in the general model serialization code (unless you want to write your own version of model serialization that appends the mean/variance to the serialized model).  Though I haven't dealt with an actual implementation of Batch Normalization yet.

benanne:
That's true, but you also want to get a validation estimate during training so you can track progress. The way they propose to do this in the paper is by using a moving average of the per-batch estimates from training (I'm guessing they mean an exponential moving average which can be implemented more memory-efficiently). But I suppose that's not the only way to do it.

If we do decide to add support code for either of these things (validation estimation, test time estimation), I definitely think it should be separate from our `get_params` infrastructure, which is for learnable parameters only.

I guess this shows that `get_params` is a bit overloaded: it is used both for getting the learnable parameters of a layer, and for getting its state. It seems like those two things do not necessarily overlap completely in all cases. But I don't think this is a strong enough use case to split it up into two separate methods, since they do actually overlap for almost all other use cases.

diogo149:
[...]

PR #430, opened 2015-09-23
Scaled tanh nonlinearity

f0k:
This extends PR #409 with better documentation (and correct markup) and slightly simplified code. If this is merged, it closes #409.
(I originally wanted to send @Botev a PR to his PR, but since no further work is required on top, I figured it would be easier to send a new PR to Lasagne instead.)

/edit: I also swapped the meaning of alpha and beta, since I found it extremely confusing that they were not matching the order of scale_in and scale_out in the constructor parameters. For a fact, Alexander got confused by this as well and reported LeCun's suggested factors in a different order than the other ones.

### DIFF
--- a/lasagne/nonlinearities.py
+++ b/lasagne/nonlinearities.py
@@ -46,10 +46,12 @@ def softmax(x):
 # tanh
 def tanh(x):
     """Tanh activation function :math:`\\varphi(x) = \\tanh(x)`
+
     Parameters
     ----------
     x : float32
         The activation (the summed, weighted input of a neuron).
+
     Returns
     -------
     float32 in [-1, 1]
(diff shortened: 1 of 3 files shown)


Botev:
Cool. Sorry for the confusion in the docs. Happy it finally made it in.

PS: Btw, I just noticed that the BN operator was merged in the main Theano branch Theano/Theano#3410 2 days ago. Maybe should have a formal BN layer now?

f0k:
> Maybe should have a formal BN layer now?

That one wasn't holding us back. It's mostly that there's no clean interface to include the running averages, but I guess we'll use my default_updates trick (https://gist.github.com/f0k/f1a6bd3c8585c400c190), I just have to do a PR.

skaae:
I tested your BN for reproducing the ladder network. It seems to work fine. The only thing i changed was an option to collect the statistics by running the entire dataset through the network in a single batch. This works fine for MNIST but will not work for larger datasets. The benefit is that I get the correct statistics.

f0k:
We should discuss that in #141, but...

> The only thing i changed was an option to collect the statistics by running the entire dataset through the network in a single batch. This works fine for MNIST but will not work for larger datasets.

That's good, it should be possible to turn this into a function that iterates over the dataset in mini-batches to collect those statistics. I've got something in place for my implementation, but results were worse than with the running averages collected during training, so I guess there's something wrong. If you get good statistics with a single-batch approach, we've got at least something to compare to.

f0k:
Back to topic, any comments on this PR? Is it good to merge?

Botev:
How would this be more beneficial in terms of computation (Running trough the mini batches for statistics). This would mean that either:
a) You run trough all mini batches to collect statistics, discarding all intermediate results (considering for any deep architecture). Thus this would be the equivalent of an extra full forward pass on the NN for each single mini batch.
b) You run trough all mini batches to collect statistics + you keep all intermediate results. Although this would be optimal in terms of speed the memory requirement will blow up too much for larger models. 
Am I missing something on how you guys collect these statistics (considering they change on every mini batch gradient update)?

f0k:
Our discussion was just about re-collecting the statistics over the full training set *after* training, instead of using the exponential moving averages collected *during* training to have something for the validation set. But we're off topic here, please have a look at #141 and post there if you have any questions / ideas!

f0k:
> Back to topic, any comments on this PR? Is it good to merge?

lasagne/nonlinearities.py
```diff
@@ -60,6 +60,87 @@ def tanh(x):
     return theano.tensor.tanh(x)
 
 
+# scaled tanh
+class ScaledTanH(object):
```
ebenolson: Why `ScaledTanH` instead of `ScaledTanh`?

lasagne/nonlinearities.py
```diff
@@ -60,6 +60,87 @@ def tanh(x):
...
+    Scaling the input down will result in decreasing the maximum slope of the
+    tanh and as a result it will be in the linear regime in a larger interval
+    of the input space. Scaling the input up will increase the maximum slope
+    of the tanh and thus bring it closer to a step function.
+
+    Scaling the output changes the output interval to :math:`[-\\beta,\\beta]`.
+
+    Parameters
+    ----------
+    scale_in : float32
```
ebenolson: why `float32` instead of `float`?

lasagne/tests/test_nonlinearities.py
```diff
@@ -22,20 +22,33 @@ def sigmoid(self, x):
     def tanh(self, x):
         return np.tanh(x)
 
+    def scaled_tanh(self, x):
+        return np.tanh(x)
+
+    def scaled_tanh_p(self, x):
```
ebenolson: what does `_p` stand for? parametrized?

lasagne/nonlinearities.py
```diff
@@ -60,6 +60,87 @@ def tanh(x):
...
+       http://link.springer.com/chapter/10.1007/978-3-642-21735-7_7,
+       http://people.idsia.ch/~ciresan/data/icann2011.pdf
+    """
+
+    def __init__(self, scale_in=1, scale_out=1):
+        self.scale_in = scale_in
+        self.scale_out = scale_out
+
+    def __call__(self, x):
+        if self.scale_in != 1:
```
ebenolson: how much does this improve the efficiency vs just

    return self.scale_out * theano.tensor.tanh(x * self.scale_in)

ebenolson:
couple of nitpicks

lasagne/nonlinearities.py
f0k: It's camel case for Scaled Tangens Hyperbolicus, I assume. But I think `ScaledTanh` looks more natural... maybe we should support both?

lasagne/nonlinearities.py
f0k: I guess that was copy/pasted from the other docstrings here written by Martin Thoma. At some point, we need to figure out how to call different things and unify all the docstrings in the library accordingly. That's outside the scope of this PR.

lasagne/nonlinearities.py
f0k: Not at all, it's optimized away:
```
import theano
>>> x = theano.tensor.matrix('x')
>>> x.dtype
float32
>>> (1.0*x*1.0).dtype
float32
>>> theano.printing.debugprint(1.0*x*1.0)
Elemwise{mul,no_inplace} [@A] ''   
 |Elemwise{mul,no_inplace} [@B] ''   
 | |DimShuffle{x,x} [@C] ''   
 | | |TensorConstant{1.0} [@D]
 | |<TensorType(float32, matrix)> [@E]
 |DimShuffle{x,x} [@F] ''   
   |TensorConstant{1.0} [@D]
>>> theano.printing.debugprint(theano.function([x], 1.0*x*1.0))
DeepCopyOp [@A] ''   0
 |<TensorType(float32, matrix)> [@B]
```
I'll replace it by the one-liner.

lasagne/tests/test_nonlinearities.py
f0k: @Botev?

f0k:
Changed implementation to one-liner (https://github.com/Lasagne/Lasagne/pull/430#discussion_r40545331), added `ScaledTanh` alias for people reading it as `scaled tanh()` instead of `scaled tangens hyperbolicus` -- both are valid interpretations, I guess.
Will leave the test and the docstring argument type unchanged, but I'm always happy about nitpicking! :)

lasagne/tests/test_nonlinearities.py
Botev: The full thing should stand for `scaled_tanh_parametrized` as it is a test for the parametrized non standard `tanh`.

lasagne/nonlinearities.py
Botev: This I think is a matter of convention, but I just looked around in different tools and literature and I think @ebenolson is probably right and `ScaledTanh` is the much more widely used option.

lasagne/nonlinearities.py
Botev: Ah sorry for that it was written in my first days in using Lasagne/Theano where I did not know what is happening.

PR #431, opened 2015-09-23
fix slicing axis with size None

skaae:
Fix a but where the SliceLayer would fail if one tried to slice a dimension with size None.

This would fail:
```Python
import lasagne
import theano.tensor as T
from lasagne.layers import *
import numpy as np
sym_x = T.matrix()
l_in = InputLayer((None, 10), sym_x)
l_slice = SliceLayer(l_in, axis=0, indices=slice(0, 50))
l_out = ConcatLayer([l_slice, l_slice], axis=0)
output = lasagne.layers.get_output(l_out)
print output.eval({sym_x :np.ones((100, 10), dtype='float32')}).shape
```

### DIFF
--- a/lasagne/layers/shape.py
+++ b/lasagne/layers/shape.py
@@ -375,9 +375,11 @@ class SliceLayer(Layer):
         output_shape = list(input_shape)
         if isinstance(self.slice, int):
             del output_shape[self.axis]
-        else:
+        elif input_shape[self.axis] is not None:
             output_shape[self.axis] = len(
                 range(*self.slice.indices(input_shape[self.axis])))
+        else:
+            output_shape[self.axis] = None
         return tuple(output_shape)
 
     def get_output_for(self, input, **kwargs):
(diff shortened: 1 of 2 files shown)


lasagne/layers/shape.py
```diff
@@ -375,9 +375,11 @@ def get_output_shape_for(self, input_shape):
         output_shape = list(input_shape)
         if isinstance(self.slice, int):
             del output_shape[self.axis]
-        else:
+        elif input_shape[self.axis] is not None:
             output_shape[self.axis] = len(
                 range(*self.slice.indices(input_shape[self.axis])))
+        else:
+            output_shape[self.axis] = None
```
f0k: You could omit the last two lines, `output_shape[self.axis]` is already None if `input_shape[self.axis]` is None. (But you can answer "explicit is implicit" and I'm with it :). As you prefer.)

skaae: yes, didn't think of that. I think i still prefer the explicit version.

skaae:
Can this be merged? For one of my projects I can avoid a workaround when this is merged.

f0k:
Ah yes, sure! Was waiting for Travis yesterday and then didn't think of it again. Thanks, merging!

skaae:
thanks

PR #428, opened 2015-09-22
More objectives

Botev:
So I wanted to add accuracy objective (so that it can be made more general of calculating validation and test statistics such as just passing a handle) as well as hinge loss objectives (I've seen them used often in the NLP papers). Additionally, changed the rectifier to use the theano implementation, because as of the current state they are the same.

PR #433, opened 2015-09-23
Add recently added objectives to docs; remove unnecessary restrictions

f0k:
We overlooked in #428 that it didn't add the new functions to the Sphinx docs. I'm doing this here, changing some bits and pieces for better consistency (including removal of the restrictions for the binary hinge loss and accuracy, to be consistent with the binary cross-entropy).


f0k:
Any comments?

ebenolson:
looks good to me

f0k:
Thank you for reviewing, merging!

PR #435, opened 2015-09-24
MNIST example: Show how to save and load param values

f0k:
I had a commented line in the MNIST example that shows how to save parameters. However, it inadvertently saved all the parameters as a single numpy array of dtype='O' that had all the other numpy arrays as its elements. Although it works, this was not what I meant to suggest to users.

This PR changes the commented line to save the parameters as separate arrays, and adds two more lines that show how to re-read them (in correct order).


f0k:
Any comments?

benanne:
works for me.

f0k:
Good, thanks!

PR #438, opened 2015-09-26
Remove conv same pad for even filter_size

ebenolson:
Currently the docs say that `pad='same'` will work for both odd and even `filter_size`, but this is not tested and fails on all implementation except `Conv2DLayer`.

I don't think this case is worth supporting:

1. It's not possible to avoid asymmetric edge effects
2. I think it's uncommon in practice
3. It requires ugly and inefficient code to support

Therefore, this PR raises `NotImplementedError` in all conv layers for this case. 

It also changes how `Conv1DLayer` implements same padding to avoid the slow compilation bug. 
This supersedes and will close https://github.com/Lasagne/Lasagne/pull/415 if merged.

### DIFF
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -253,8 +253,8 @@ class Conv1DLayer(Layer):
                                       image_shape=input_shape,
                                       filter_shape=self.get_W_shape(),
                                       border_mode='full')
-            shift = (self.filter_size[0] - 1) // 2
-            conved = conved[:, :, shift:input.shape[2] + shift]
+            crop = self.filter_size[0] // 2
+            conved = conved[:, :, crop:-crop or None]
         else:
             # no padding needed, or explicit padding of input needed
             if self.pad == 'full':
@@ -468,10 +468,10 @@ class Conv2DLayer(Layer):
                                       image_shape=input_shape,
                                       filter_shape=self.get_W_shape(),
                                       border_mode='full')
-            shift_x = (self.filter_size[0] - 1) // 2
-            shift_y = (self.filter_size[1] - 1) // 2
-            conved = conved[:, :, shift_x:input.shape[2] + shift_x,
-                            shift_y:input.shape[3] + shift_y]
+            crop_x = self.filter_size[0] // 2
+            crop_y = self.filter_size[1] // 2
+            conved = conved[:, :, crop_x:-crop_x or None,
+                            crop_y:-crop_y or None]
         else:
             # no padding needed, or explicit padding of input needed
             if self.pad == 'full':
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -123,9 +123,9 @@ class Conv1DLayer(Layer):
         is equivalent to computing the convolution wherever the input and the
         filter overlap by at least one position.
 
-        ``'same'`` pads with half the filter size on both sides (one less on
-        the second side for an even filter size). When ``stride=1``, this
-        results in an output size equal to the input size.
+        ``'same'`` pads with half the filter size (rounded down) on both sides.
+        When ``stride=1`` this results in an output size equal to the input
+        size. Even filter size is not supported.
 
         ``'valid'`` is an alias for ``0`` (no padding / a valid convolution).
 
@@ -196,6 +196,11 @@ class Conv1DLayer(Layer):
         self.stride = as_tuple(stride, 1)
         self.untie_biases = untie_biases
         self.convolution = convolution
+
+        if pad == 'same':
+            if self.filter_size[0] % 2 == 0:
+                raise NotImplementedError(
+                    '`same` padding requires odd filter size.')
 
         if pad == 'valid':
             self.pad = (0,)
@@ -326,9 +331,9 @@ class Conv2DLayer(Layer):
         is equivalent to computing the convolution wherever the input and the
         filter overlap by at least one position.
 
-        ``'same'`` pads with half the filter size on both sides (one less on
-        the second side for an even filter size). When ``stride=1``, this
-        results in an output size equal to the input size.
+        ``'same'`` pads with half the filter size (rounded down) on both sides.
+        When ``stride=1`` this results in an output size equal to the input
+        size. Even filter size is not supported.
 
         ``'valid'`` is an alias for ``0`` (no padding / a valid convolution).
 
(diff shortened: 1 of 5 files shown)


benanne:
I'm okay with this in principle. I haven't reviewed the code changes though, someone else should look at it in more detail :)

lasagne/layers/conv.py
```diff
@@ -248,8 +253,8 @@ def get_output_for(self, input, input_shape=None, **kwargs):
                                       image_shape=input_shape,
                                       filter_shape=self.get_W_shape(),
                                       border_mode='full')
-            shift = (self.filter_size[0] - 1) // 2
-            conved = conved[:, :, shift:input.shape[2] + shift]
+            crop = self.filter_size[0] // 2
+            conved = conved[:, :, crop:-crop]
```
f0k: Should be `conved[:, :, crop:-crop or None]`, otherwise it will fail for `self.filter_size[0] == 1`!

ebenolson: Thanks! I should probably add that case to the tests..

lasagne/layers/conv.py
```diff
@@ -123,9 +123,9 @@ class Conv1DLayer(Layer):
         is equivalent to computing the convolution wherever the input and the
         filter overlap by at least one position.
 
-        ``'same'`` pads with half the filter size on both sides (one less on
-        the second side for an even filter size). When ``stride=1``, this
-        results in an output size equal to the input size.
+        ``'same'`` pads with half the filter size (rounded down) on both sides,
+        this results in an output size equal to the input size. Even filter
+        size is not supported.
```
f0k: I specifically wrote `When stride=1`, because it only results in an output size equal to the input size if there's no striding. Would be nice if you could retain this!

ebenolson: Good point, will fix that.

lasagne/layers/conv.py
```diff
@@ -326,9 +331,9 @@ class Conv2DLayer(Layer):
         is equivalent to computing the convolution wherever the input and the
         filter overlap by at least one position.
 
-        ``'same'`` pads with half the filter size on both sides (one less on
-        the second side for an even filter size). When ``stride=1``, this
-        results in an output size equal to the input size.
+        ``'same'`` pads with half the filter size (rounded down) on both sides,
+        this results in an output size equal to the input size. Only odd filter
+        sizes are allowed.
```
f0k: Again, please retain `when stride==1` (commenting here so it's not forgotten).

lasagne/layers/corrmm.py
```diff
@@ -73,9 +73,9 @@ class Conv2DMMLayer(MMLayer):
         is equivalent to computing the convolution wherever the input and the
         filter overlap by at least one position.
 
-        ``'same'`` pads with half the filter size on both sides (one less on
-        the second side for an even filter size). When ``stride=1``, this
-        results in an output size equal to the input size.
+        ``'same'`` pads with half the filter size (rounded down) on both sides,
+        this results in an output size equal to the input size. Even filter
+        size is not supported.
```
f0k: Again, please retain `when stride=1` (commenting here so it's not forgotten).

lasagne/layers/conv.py
```diff
@@ -458,8 +468,8 @@ def get_output_for(self, input, input_shape=None, **kwargs):
                                       image_shape=input_shape,
                                       filter_shape=self.get_W_shape(),
                                       border_mode='full')
-            shift_x = (self.filter_size[0] - 1) // 2
-            shift_y = (self.filter_size[1] - 1) // 2
+            shift_x = self.filter_size[0] // 2
+            shift_y = self.filter_size[1] // 2
             conved = conved[:, :, shift_x:input.shape[2] + shift_x,
                             shift_y:input.shape[3] + shift_y]
```
f0k: Shouldn't this use `crop` as well? Or is the compilation problem limited to Conv1DLayer?

lasagne/layers/cuda_convnet.py
```diff
@@ -87,9 +87,9 @@ class Conv2DCCLayer(CCLayer):
         is equivalent to computing the convolution wherever the input and the
         filter overlap by at least one position.
 
-        ``'same'`` pads with half the filter size on both sides (one less on
-        the second side for an even filter size). When ``stride=1``, this
-        results in an output size equal to the input size.
+        ``'same'`` pads with half the filter size (rounded down) on both sides,
+        this results in an output size equal to the input size. Even filter
+        size is not supported.
```
f0k: Again, please retain `when stride=1` (commenting here so it's not forgotten).

lasagne/layers/dnn.py
```diff
@@ -175,9 +174,9 @@ class Conv2DDNNLayer(DNNLayer):
         is equivalent to computing the convolution wherever the input and the
         filter overlap by at least one position.
 
-        ``'same'`` pads with half the filter size on both sides (one less on
-        the second side for an even filter size). When ``stride=1``, this
-        results in an output size equal to the input size.
+        ``'same'`` pads with half the filter size (rounded down) on both sides,
+        this results in an output size equal to the input size. Even filter
+        size is not supported.
```
f0k: Again, please retain `when stride=1` (commenting here so it's not forgotten).

f0k:
I'm fine as well except for a few comments. Thanks for tackling this!

lasagne/layers/conv.py
ebenolson: It seems to be...

I may change anyway though for consistency (and I think the generated graph comes out simpler).

f0k: +1 for consistency! Here it'd probably be simpler to use the `crop_x:-crop_x or None, crop_y:-crop_y or None` trick then...

f0k:
Didn't see your update (there's no notification for commits). Looks good, just the commits could be grouped nicer. If you're a perfectionist, you could do:
```
git reset HEAD^^
git add -p  # add all changes that remove support for pad=same with even filter size
git commit -m 'raise NotImplementedError for pad=same and even filter size, add tests'
git commit -am 'alternative method for Conv1DLayer and Conv2DLayer with pad=same'
```
But I'd also be fine with merging as it is. Let me know.

ebenolson:
thanks, made that grouping.

f0k:
Perfect! Thanks a lot!

Issue #440, opened 2015-09-26
Restrictions in TransformerLayer


rmanor:
Hi,

Are there any plans to add optional restrictions to the TransformerLayer?
e.g., do only translation and scaling in one axis, etc.
Thanks.

benanne:
I don't know if there are plans to provide explicit support for it, but you could simply add a custom layer that adequately restrict the 6 numbers that describe the transformation matrix. Maybe providing a few helper functions/layers for common cases (only rotations, only translations etc.) could be useful.

skaae:
I don't think we plan to support that. 

@benanne: Is it easy to clamp outputs to a specific value? Maybe we could provide a layer for that?

benanne:
Well, it's easy enough to write a custom layer. For example, you could have a layer that takes 1 single value as input (a rotation angle) and outputs the corresponding 6 numbers describing the transformation matrix. All you need is a class with two methods, both of which are probably 1-2 lines long. I'd say that's already easy enough and we don't need to complicate the `TransformerLayer` code to support this.

skaae:
> I'd say that's already easy enough and we don't need to complicate the TransformerLayer code to support this.

agreed. I thought we could create a KeepConstantLayer that could clamp the outputs of the localization network. 
Something like
```Python
loc_constant = KeepConstant(loc, indices=slice(0,3))
transformer  = TransformerLayer(incoming=img, localization_network=loc_constant)
```

benanne:
That seems like a bit of hacky solution though? It doesn't really cover all the bases. You don't necessarily want to keep certain values constant, sometimes the constraints take more complex forms (as in the case of rotations). So I feel like it makes more sense to have a neural net that outputs the parameters of whatever restricted set of transformations you've chosen, and then a simple custom layer that maps these onto the 6 values that describe general affine transformations.

skaae:
ok. I havent really looked at how you constrain to e.g. rotations.

ebenolson:
There seem to be a number of cases where you'd like to just tweak the output of a layer in some small way. What about adding something like this?

    class LambdaLayer(lasagne.layers.Layer):
        def __init__(self, incoming, function, output_shape=None, **kwargs):
            super(LambdaLayer, self).__init__(incoming, **kwargs)

            if output_shape is not None:
                if hasattr(output_shape, '__call__'):
                    self.get_output_shape_for = output_shape
                else:
                    self.input_shape = output_shape
            
            self.function = function
        
[...]

PR #441, opened 2015-09-27
add LambdaLayer

ebenolson:
This provides some boilerplate to make it easier to apply simple operations without needing to define a custom Layer class. There's been a few occasions where I think this could be useful, most recently https://github.com/Lasagne/Lasagne/issues/440.

Since everyone loves a good naming discussion, `LambdaLayer` is probably not that great, since using a lambda expression is optional... `FunctionLayer` or `ExpressionLayer` are my other ideas.

### DIFF
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -10,6 +10,7 @@ from .base import Layer, MergeLayer
 __all__ = [
     "NonlinearityLayer",
     "BiasLayer",
+    "ExpressionLayer",
     "InverseLayer",
     "TransformerLayer",
     "ParametricRectifierLayer",
@@ -109,6 +110,71 @@ class BiasLayer(Layer):
             return input + self.b.dimshuffle(*pattern)
         else:
             return input
+
+
+class ExpressionLayer(Layer):
+    """
+    This layer provides boilerplate for a custom layer that applies a
+    simple transformation to the input.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape.
+
+    function : callable
+        A function to be applied to the output of the previous layer.
+
+    output_shape : None, callable, tuple, or 'auto'
+        Specifies the output shape of this layer. If a tuple, this fixes the
+        output shape for any input shape (the tuple can contain None if some
+        dimensions may vary). If a callable, it should return the calculated
+        output shape given the input shape. If None, the output shape is
+        assumed to be the same as the input shape. If 'auto', an attempt will
+        be made to automatically infer the correct output shape.
+
+    Notes
+    -----
+    An :class:`ExpressionLayer` that does not change the shape of the data
+    (i.e., is constructed with the default setting of ``output_shape=None``)
+    is functionally equivalent to a :class:`NonlinearityLayer`.
+
+    Examples
+    --------
+    >>> from lasagne.layers import InputLayer, ExpressionLayer
+    >>> l_in = InputLayer((32, 100, 20))
+    >>> l1 = ExpressionLayer(l_in, lambda X: X.mean(-1), output_shape='auto')
+    >>> l1.output_shape
+    (32, 100)
+    """
+    def __init__(self, incoming, function, output_shape=None, **kwargs):
+        super(ExpressionLayer, self).__init__(incoming, **kwargs)
+
+        if output_shape is None:
+            self._output_shape = None
+        elif output_shape == 'auto':
+            self._output_shape = 'auto'
+        elif hasattr(output_shape, '__call__'):
+            self.get_output_shape_for = output_shape
+        else:
+            self._output_shape = tuple(output_shape)
+
+        self.function = function
+
+    def get_output_shape_for(self, input_shape):
+        if self._output_shape is None:
+            return input_shape
+        elif self._output_shape is 'auto':
+            input_shape = (0 if s is None else s for s in input_shape)
+            X = theano.tensor.alloc(0, *input_shape)
+            output_shape = self.function(X).shape.eval()
+            output_shape = tuple(s if s else None for s in output_shape)
+            return output_shape
+        else:
+            return self._output_shape
+
+    def get_output_for(self, input, **kwargs):
+        return self.function(input)
 
 
 class InverseLayer(MergeLayer):
--- a/docs/modules/layers.rst
+++ b/docs/modules/layers.rst
@@ -152,6 +152,7 @@
 
     NonlinearityLayer
     BiasLayer
+    ExpressionLayer
     InverseLayer
     TransformerLayer
     ParametricRectifierLayer
--- a/docs/modules/layers/special.rst
+++ b/docs/modules/layers/special.rst
@@ -11,6 +11,9 @@ Special-purpose layers
 .. autoclass:: BiasLayer
    :members:
 
+.. autoclass:: ExpressionLayer
+   :members:
+
 .. autoclass:: InverseLayer
     :members:
 
--- a/lasagne/tests/layers/test_special.py
+++ b/lasagne/tests/layers/test_special.py
@@ -4,6 +4,123 @@ import pytest
 import theano
 
 
+class TestExpressionLayer:
+    @pytest.fixture
+    def ExpressionLayer(self):
+        from lasagne.layers.special import ExpressionLayer
+        return ExpressionLayer
+
+    @pytest.fixture
+    def input_layer(self):
+        from lasagne.layers import InputLayer
+        return InputLayer((2, 3, 4, 5))
+
+    @pytest.fixture
+    def input_layer_nones(self):
+        from lasagne.layers import InputLayer
+        return InputLayer((1, None, None, 5))
+
+    def np_result(self, func, input_layer):
+        X = np.random.uniform(-1, 1, input_layer.output_shape)
+        return X, func(X)
+
+    @pytest.mark.parametrize('func',
+                             [lambda X: X**2,
+                              lambda X: X.mean(-1),
+                              lambda X: X.sum(),
+                              ])
+    def test_tuple_shape(self, func, input_layer, ExpressionLayer):
+        from lasagne.layers.helper import get_output
+
+        X, expected = self.np_result(func, input_layer)
+        layer = ExpressionLayer(input_layer, func, output_shape=expected.shape)
+        assert layer.get_output_shape_for(X.shape) == expected.shape
+
+        output = get_output(layer, X).eval()
+        assert np.allclose(output, expected)
+
+    @pytest.mark.parametrize('func',
+                             [lambda X: X**2,
+                              lambda X: X.mean(-1),
+                              lambda X: X.sum(),
+                              ])
+    def test_callable_shape(self, func, input_layer, ExpressionLayer):
+        from lasagne.layers.helper import get_output
+
+        X, expected = self.np_result(func, input_layer)
+
+        def get_shape(input_shape):
+            return func(np.empty(shape=input_shape)).shape
+
+        layer = ExpressionLayer(input_layer, func, output_shape=get_shape)
+        assert layer.get_output_shape_for(X.shape) == expected.shape
+
+        output = get_output(layer, X).eval()
+        assert np.allclose(output, expected)
+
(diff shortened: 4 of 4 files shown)


benanne:
Cool! I'm not sure if I like the 'automatic shape' functionality though. This seems a bit too magic and error-prone. The way it's setup it'll just interpret some random data as a float array and try to infer the shape by doing the computation on this array. That makes me feel queasy :) For starters, that array is almost guaranteed to be full of NaNs, which could have unintended side effects.

Overall I think we shouldn't be using that much magic here. If the `output_shape` is not specified, I think it should just default to assuming that the output shape is the same as the input shape. This is also what is done for `Layer` subclasses that do not override `get_output_shape_for()`.

ebenolson:
yeah, that's probably for the best. I like the magic but it will cause delays for compilation and potentially large spikes in memory consumption... will update tomorrow.

lasagne/layers/special.py
```diff
@@ -15,6 +17,54 @@
...
+    def __init__(self, incoming, function, output_shape=None, **kwargs):
+        super(LambdaLayer, self).__init__(incoming, **kwargs)
+
+        if output_shape is None:
+            self._output_shape_memo = {}
+        else:
+            if hasattr(output_shape, '__call__'):
+                self.get_output_shape_for = output_shape
+            else:
+                self.get_output_shape_for = lambda _: tuple(output_shape)
```
f0k: This is not pickable; we shouldn't use `lambda` in any of the classes we provide. (If a user chooses to pass a `lambda` expression for `function` or `output_shape`, it's their responsibility.)

f0k:
> The way it's setup it'll just interpret some random data as a float array and try to infer the shape by doing the computation on this array.

One could use zeros instead of random data.

> potentially large spikes in memory consumption

One could use `theano.tensor.alloc` instead of a shared variable. `function(X).shape.eval()` allows Theano to infer the shape without doing any actual computation, if the Ops support it.

The remaining problem is that we don't know if the function expects a float array at all, it might expect integers, for example.

What about treating `None` as the default case of a layer that just does an element-wise transformation, not changing the shape?

lasagne/layers/special.py
```diff
@@ -1,13 +1,15 @@
 import theano
 import theano.tensor as T
+import numpy as np
 
 from .. import init
 from .. import nonlinearities
-from ..utils import as_tuple
+from ..utils import as_tuple, floatX
```
f0k: now unused.

ebenolson:
> One could use theano.tensor.alloc instead of a shared variable. function(X).shape.eval() allows Theano to infer the shape without doing any actual computation, if the Ops support it.

This sounds interesting, might be nice to include as an option if `output_shape = 'auto'` or something like that. 

I pushed a new version that removes the automatic calculation and the lambda function.

benanne:
> What about treating `None` as the default case of a layer that just does an element-wise transformation, not changing the shape?

As I said, that's what I would prefer. Also for consistency with the case of subclassing `Layer`.

f0k:
> As I said, that's what I would prefer.

Ah yes. Forgot about this between reading your comment and reading the code :)

> This sounds interesting, might be nice to include as an option if `output_shape = 'auto'` or something like that.

Sounds good, I had the same in mind (but it still has the problem of not knowing the dtype to use).

> Since everyone loves a good naming discussion, `LambdaLayer` is probably not that great, since using a lambda expression is optional...

That's not much of a problem to me, but that name is geared towards Python, not towards machine learning -- it won't make a lot of sense to somebody coming here from university or from a different framework. `FunctionLayer` or `CustomLayer` would be okay to me.

benanne:
Out of all those I like `FunctionLayer` best. `CustomLayer` might get confusing because we sometimes refer to user-made `Layer` subclasses as "custom `Layer` classes".

craffel:
Anything holding this up from being merged?  FWIW I like `ExpressionLayer` the most, as you're providing a theano *expression*, not a function.  But I don't care much either way, I just want to use it!

bmcfee:
> FWIW I like ExpressionLayer

:+1:

ebenolson:
I have a nagging feeling there was a bug I needed to fix... need to look it over again

f0k:
> Anything holding this up from being merged? FWIW I like `ExpressionLayer` the most, as you're providing a theano *expression*, not a function.

But you're not providing a Theano expression, you're providing a *function* (or callable) that takes a Theano expression and returns a Theano expression. So I'd still prefer `FunctionLayer`. @bmcfee, what's your reasoning?

craffel:
> But you're not providing a Theano expression, you're providing a function (or callable) that takes a Theano expression and returns a Theano expression. So I'd still prefer FunctionLayer. @bmcfee, what's your reasoning?

Right, but `function` is pretty overloaded because of its parallel definition in `theano` as something different.  `ExpressionLayer` communicates that the function supplied should accept and return an expression (which is what I meant by "providing").  Again, I don't really care one way or the other, I'd rather get this merged sooner than have it be named `ExpressionLayer`!

f0k:
> Right, but `function` is pretty overloaded because of its parallel definition in `theano` as something different.

Ah. `CallableLayer`?

craffel:
Well, lots of things are callable, but `Expression` is pretty specific (in this context).

As a side argument, I think of the layer name as specifying what it adds to your network (e.g. `DenseLayer` adds a dense connection).  Using this layer doesn't add a function, or a callable, it adds an arbitrary Theano expression.

BUT I DIGRESS!  Call it whatever you want.

f0k:
> As a side argument, I think of the layer name as specifying what it adds to your network (e.g. `DenseLayer` adds a dense connection).

Okay, I hadn't seen it this way, but it makes sense. Granted, `CallableLayer` sounds as if the layer itself was callable, not as if you'd provide a callable to its constructor. Well then, as it could add anything, what about `AnyLayer`? Otherwise I'm more or less convinced that `ExpressionLayer` is better than `FunctionLayer`.

> BUT I DIGRESS!

I still think thoughtful and consistent naming of components is very important for a library.

craffel:
> Well then, as it could add anything, what about AnyLayer

Any is maybe too generic.  `np.any` and `T.any` also have a pretty specific, and different, meaning.

> I still think thoughtful and consistent naming of components is very important for a library. Thanks for sharing your thoughts!

I agree, but I also think any of the proposed names are fine, we probably won't come to a consensus, and reaching a consensus is less important than just merging the layer.

benanne:
Names are important :) Thanks for the insightful discussion. I think I prefer `ExpressionLayer` (and the motivation for it) as well.

ebenolson:
Seems like `ExpressionLayer` has the consensus - can someone give this a review?

lasagne/layers/special.py
```diff
@@ -15,6 +16,57 @@
...
+
+        self.function = function
+
+    def get_output_shape_for(self, input_shape):
+        try:
+            output_shape = self._output_shape
+            if output_shape == 'auto':
+                input_shape = (0 if s is None else s for s in input_shape)
+                X = theano.tensor.alloc(0, *input_shape)
+#                output_shape = tuple(self.function(X).shape.eval())
```
craffel: Looks like some cruft here.

lasagne/layers/special.py
```diff
@@ -15,6 +16,57 @@
...
+    function : callable
+        A function to be applied to the output of the previous layer.
+
+    output_shape : None, callable, tuple, or 'auto'
+        Specifies the output shape of this layer. If a tuple, this fixes the
+        output shape for any input shape (the tuple can contain None if some
+        dimensions may vary). If a callable, it should return the calculated
+        output shape given the input shape. If None, the output shape is
+        assumed to be the same as the input shape. If 'auto', an attempt will
+        be made to automatically infer the correct output shape.
```
craffel: When will `output_shape='auto'` fail?

lasagne/layers/special.py
```diff
@@ -15,6 +16,57 @@
...
+            output_shape = self._output_shape
+            if output_shape == 'auto':
+                input_shape = (0 if s is None else s for s in input_shape)
+                X = theano.tensor.alloc(0, *input_shape)
+#                output_shape = tuple(self.function(X).shape.eval())
+                output_shape = self.function(X).shape.eval()
+                output_shape = tuple(s if s else None for s in output_shape)
+            return output_shape
+        except AttributeError:
+            return input_shape
```
f0k: I'm not sure about placing the `except AttributeError` so far from the line that's meant to cause it. I'd prefer either:
```python
try:
    output_shape = self._output_shape
except AttributeError:
    return input_shape
else:
    ...
```
or setting `self._output_shape` to `None` in the constructor if `output_shape` is `None`. That way, `_output_shape` is always defined when `get_output_shape_for` wasn't overwritten and can simply be checked for.

lasagne/layers/special.py
```diff
@@ -15,6 +16,57 @@
...
+
+    def get_output_shape_for(self, input_shape):
+        try:
+            output_shape = self._output_shape
+            if output_shape == 'auto':
+                input_shape = (0 if s is None else s for s in input_shape)
+                X = theano.tensor.alloc(0, *input_shape)
+#                output_shape = tuple(self.function(X).shape.eval())
+                output_shape = self.function(X).shape.eval()
+                output_shape = tuple(s if s else None for s in output_shape)
```
f0k: What case is this for? Is it possible that some elements are zero? Why should they be replaced by `None` then?

lasagne/layers/special.py
f0k: I think an example would be nice. Probably one that changes the shape, because otherwise you could just use a `NonlinearityLayer`. Maybe that should also be mentioned as a Note?

craffel: A good example: Computing the mean of the input across one axis, e.g. `function` would be something like `lambda x: T.mean(X, axis=-1)`.

f0k:
Looks good except for our comments! Very nice, clean and complete tests.

lasagne/layers/special.py
ebenolson: > When will output_shape='auto' fail?

for example, this will give a `NotImplementedError`:

    W = theano.shared(np.zeros((1, 1, 3, 3)).astype('float32'))
    def f(X):
        return T.nnet.conv2d(X, W)
    l2 = ExpressionLayer(l, f, output_shape='auto')

ebenolson:
Thanks, I think I've addressed all comments and I also added it to sphinx. Anything else before I squash?

lasagne/layers/special.py
craffel: Ah, makes sense.

f0k: > Ah, makes sense.

Really? Where does the `NotImplementedError` come from? Shouldn't Theano always be able to figure out the shape in some way if you call `eval()` on it?

f0k:
> Anything else before I squash?

In https://github.com/Lasagne/Lasagne/pull/441#discussion_r44554708, I wondered (not very prominently) if there should be a "Notes" section explaining how `ExpressionLayer` relates to `NonlinearityLayer`. Something like:
```
An :class:`ExpressionLayer` that does not change the shape of the data (i.e., is constructed
with the default setting of ``output_shape=None``) is functionally equivalent to a
:class:`NonlinearityLayer`.
```
Alternatively, this could be mentioned in the description of the `output_shape` parameter. Otherwise it might leave people wondering which one to use.

And I think I'd place `ExpressionLayer` after `NonlinearityLayer` and `BiasLayer` (in the file and in sphinx) because I guess they'd be used more often, but I don't care much -- as you prefer!

lasagne/layers/special.py
ebenolson: alloc gives an `int` tensor, and Theano won't convolve it with the `float` W.

ebenolson:
Sure, I just went with latest first.. I'll make those changes now.

f0k:
Perfect, thanks! I think you can squash already so we don't need to wait for Travis again.

ebenolson:
Ok, should be ready now. Thanks for reviewing.

f0k:
Looks good to merge for me! I'll leave it open until tomorrow in case anybody else wants to have a look before merging. We're not adding layers on a daily basis :)

f0k:
Merging, thanks again!

benanne:
Great, nice work!

PR #444, opened 2015-09-29
Correct handling of borders in TransformerLayer

### LINKED ISSUE #442: TransformerLayer doesn't handle edge pixels properly
The interpolation in TransformerLayer averages two adjacent pixels to get the interpolated value, and multiplies it by a coefficient that is the distance of the new coordinate from the interpolated coordinates.

The adjacent pixels of the edge pixels are clipped to the same value, and so the coefficient can be zero (in the code it comes down to x = x0 = x1).

The solution I am using now is to scale the coordinates to [0, width - 2] and [0, height - 2] so there are always two adjacent pixels.

I read online that a common solution for edges is take only one closest pixel.

f0k:
Fixes #442. While this passes through an image almost unchanged for the identity transform values [1, 0, 0, 0, 1, 0], there are still some discrepancies due to mapping coordinates from [0, size] to [-1, 1] and back. I'll try to fix this by changing the meshgrid to return coordinates in [0, size].

### DIFF
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -272,27 +272,25 @@ def _interpolate(im, x, y, out_height, out_width):
     height_f = T.cast(height, theano.config.floatX)
     width_f = T.cast(width, theano.config.floatX)
 
-    # scale indices from [-1, 1] to [0, width/height].
-    x = (x + 1) / 2 * width_f
-    y = (y + 1) / 2 * height_f
+    # clip coordinates to [-1, 1]
+    x = T.clip(x, -1, 1)
+    y = T.clip(y, -1, 1)
 
-    # Clip indices to ensure they are not out of bounds.
-    max_x = width_f - 1
-    max_y = height_f - 1
-    x0 = T.clip(x, 0, max_x)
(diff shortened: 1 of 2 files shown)


f0k:
> I'll try to fix this by changing the meshgrid to return coordinates in [0, size].

Hmm, no. That's not as easy as I thought at first glance (it's easy to get rid of the rounding errors for the identity transform, but not easy to keep all other transforms working the same :) ).

skaae:
Do you want it reviewed now or later?

f0k:
You can review it! I think you can't easily change the meshgrid to avoid the mapping to [-1,1] and back. It would be cool, but this can wait for another PR.

skaae:
Looks good to me. I checked the transformations and they look fine and the black pixels at the border are gone.

f0k:
Okay, I'll merge it then. Just note that this will break compatibility with old models. (Which is fine because the layer wasn't include in any release, but which is bad because you already did some work with it. You'll need to record/tag the Lasagne commit you used for your paper, if you didn't do that anyway.)

PR #445, opened 2015-09-29
Avoid empty placeholder tuples in recurrent layers

### LINKED ISSUE #443: Scan: float64 warnings when non_sequences contains  "()"
For example:

```
theano.scan(..., non_sequences=[..., ()], ...)
```
This gives float64 warnings.

This is used [here](https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/recurrent.py#L1037) for example.

See https://github.com/Theano/Theano/issues/3345

skaae:
Fixes #443.

Fixes float64 error when recurrent layer step functions were using empty placeholder lists. I rewrote all the recurrent layers to use *args.

### DIFF
--- a/lasagne/layers/recurrent.py
+++ b/lasagne/layers/recurrent.py
@@ -940,10 +940,7 @@ class LSTMLayer(MergeLayer):
 
         # Create single recurrent computation step function
         # input_n is the n'th vector of the input
-        def step(input_n, cell_previous, hid_previous, W_hid_stacked,
-                 W_cell_to_ingate, W_cell_to_forgetgate,
-                 W_cell_to_outgate, W_in_stacked, b_stacked):
-
+        def step(input_n, cell_previous, hid_previous, *args):
             if not self.precompute_input:
                 input_n = T.dot(input_n, W_in_stacked) + b_stacked
 
@@ -963,8 +960,8 @@ class LSTMLayer(MergeLayer):
 
             if self.peepholes:
                 # Compute peephole connections
-                ingate += cell_previous*W_cell_to_ingate
-                forgetgate += cell_previous*W_cell_to_forgetgate
+                ingate += cell_previous*self.W_cell_to_ingate
+                forgetgate += cell_previous*self.W_cell_to_forgetgate
 
             # Apply nonlinearities
             ingate = self.nonlinearity_ingate(ingate)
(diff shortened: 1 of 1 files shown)


lasagne/layers/recurrent.py
```diff
@@ -1359,11 +1341,8 @@ def step(input_n, hid_previous, W_hid_stacked, W_in_stacked,
             hid = (1 - updategate)*hid_previous + updategate*hidden_update
             return hid
 
-        def step_masked(input_n, mask_n, hid_previous, W_hid_stacked,
-                        W_in_stacked, b_stacked):
-
-            hid = step(input_n, hid_previous, W_hid_stacked, W_in_stacked,
-                       b_stacked)
+        def step_masked(input_n, mask_n, hid_previous, *args):
+            hid = step(input_n, hid_previous, args)
```
f0k: It works because you're not using it, but you should pass `*args` instead of `args` (just in case the original step function decides to use `*args` in future).

lasagne/layers/recurrent.py
```diff
@@ -976,20 +973,14 @@ def step(input_n, cell_previous, hid_previous, W_hid_stacked,
...
-        def step_masked(input_n, mask_n, cell_previous, hid_previous,
-                        W_hid_stacked, W_cell_to_ingate, W_cell_to_forgetgate,
-                        W_cell_to_outgate, W_in_stacked, b_stacked):
-
-            cell, hid = step(input_n, cell_previous, hid_previous,
-                             W_hid_stacked, W_cell_to_ingate,
-                             W_cell_to_forgetgate, W_cell_to_outgate,
-                             W_in_stacked, b_stacked)
+        def step_masked(input_n, mask_n, cell_previous, hid_previous, *args):
+            cell, hid = step(input_n, cell_previous, hid_previous, args)
```
f0k: It works because you're not using it, but you should pass `*args` instead of `args` (just in case the original step function decides to use `*args` in future).

skaae: Thanks. Fixed

f0k:
Thanks! Could you please squash this right away?

skaae:
Done

f0k:
Okay, thanks! Merging and hoping that `theano.scan` will never require `fn` to use the variables passed in `*args` instead of the ones that were passed to `non_sequences` (for now it seems they are the same, otherwise the strict mode would complain).

PR #446, opened 2015-09-30
ParametricRectifyLayer

Botev:
So I need to add tests. Plus additionally I'm not sure if I'm not missing with the shape of the parameter, if someone can give me a feedback if there are any hidden corners when it won't broadcast correctly.

### DIFF
--- a/docs/modules/layers/special.rst
+++ b/docs/modules/layers/special.rst
@@ -17,3 +17,8 @@ Special-purpose layers
 .. autoclass:: TransformerLayer
     :members:
 
+.. autoclass:: ParametricRectifierLayer
+    :members:
+
+.. autofunction:: prelu
+
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -12,6 +12,8 @@ __all__ = [
     "BiasLayer",
     "InverseLayer",
     "TransformerLayer",
+    "ParametricRectifierLayer",
+    "prelu",
 ]
 
 
@@ -352,3 +354,114 @@ def _meshgrid(height, width):
     ones = T.ones_like(x_t_flat)
     grid = T.concatenate([x_t_flat, y_t_flat, ones], axis=0)
     return grid
+
+
+class ParametricRectifierLayer(Layer):
+    """
+    lasagne.layers.ParametricRectifierLayer(incoming,
+    alpha=init.Constant(0.25), shared_axes='auto', **kwargs)
+
+    A layer that applies parametric rectify nonlinearity to its input
+    following [1]_ (http://arxiv.org/abs/1502.01852)
+
+    Equation for the parametric rectifier linear unit:
+    :math:`\\varphi(x) = \\max(x,0) + \\alpha \\min(x,0)`
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
+
+    alpha : Theano shared variable, expression, numpy array or callable
+        Initial value, expression or initializer for the alpha values. The
+        shape must match the incoming shape, skipping those axes the alpha
+        values are shared over (see the example below).
+        See :func:`lasagne.utils.create_param` for more information.
+
+    shared_axes : 'auto', 'all', int or tuple of int
+        The axes along which the parameters of the rectifier units are
+        going to be shared. If ``'auto'`` (the default), share over all axes
+        except for the second - this will share the parameter over the
+        minibatch dimension for dense layers, and additionally over all
+        spatial dimensions for convolutional layers. If ``'all'``, share over
+        all axes, which corresponds to a single scalar parameter.
+
+    **kwargs
+        Any additional keyword arguments are passed to the `Layer` superclass.
+
+     References
+    ----------
+    .. [1] K He, X Zhang et al. (2015):
+       Delving Deep into Rectifiers: Surpassing Human-Level Performance on
+       ImageNet Classification,
+       http://link.springer.com/chapter/10.1007/3-540-49430-8_2
+
+    Notes
+    -----
+    The alpha parameter dimensionality is the input dimensionality minus the
+    number of axes it is shared over, which matches the same convention as
+    the :class:`BiasLayer`.
+
+    >>> layer = ParametricRectifierLayer((20, 3, 28, 28), shared_axes=(0, 3))
+    >>> layer.alpha.get_value().shape
+    (3, 28)
+    """
+    def __init__(self, incoming, alpha=init.Constant(0.25), shared_axes='auto',
+                 **kwargs):
+        super(ParametricRectifierLayer, self).__init__(incoming, **kwargs)
+        if shared_axes == 'auto':
+            self.shared_axes = (0,) + tuple(range(2, len(self.input_shape)))
+        elif shared_axes == 'all':
+            self.shared_axes = tuple(range(len(self.input_shape)))
+        elif isinstance(shared_axes, int):
+            self.shared_axes = (shared_axes,)
+        else:
+            self.shared_axes = shared_axes
+
+        shape = [size for axis, size in enumerate(self.input_shape)
+                 if axis not in self.shared_axes]
+        if any(size is None for size in shape):
+            raise ValueError("ParametricRectifierLayer needs input sizes for "
+                             "all axes that alpha's are not shared over.")
+        self.alpha = self.add_param(alpha, shape, name="alpha",
+                                    regularizable=False)
+
+    def get_output_for(self, input, **kwargs):
+        axes = iter(range(self.alpha.ndim))
+        pattern = ['x' if input_axis in self.shared_axes
+                   else next(axes)
+                   for input_axis in range(input.ndim)]
+        alpha = self.alpha.dimshuffle(pattern)
+        return theano.tensor.nnet.relu(input, alpha)
+
+
+def prelu(layer, **kwargs):
+    """
+    Convenience function to apply parametric rectify to a given layer's output.
+    Will set the layer's nonlinearity to identity if there is one and will
+    apply the parametric rectifier instead.
+
+    Parameters
+    ----------
+    layer: a :class:`Layer` instance
+        The `Layer` instance to apply the parametric rectifier layer to;
+        note that it will be irreversibly modified as specified above
+
+    **kwargs
+        Any additional keyword arguments are passed to the
+        :class:`ParametericRectifierLayer`
+
+    Examples
+    --------
+    Note that this function modifies an existing layer, like this:
+    >>> from lasagne.layers import InputLayer, DenseLayer, prelu
+    >>> layer = InputLayer((32, 100))
+    >>> layer = DenseLayer(layer, num_units=200)
+    >>> layer = prelu(layer)
+
+    In particular, :func:`prelu` can *not* be passed as a nonlinearity.
+    """
+    nonlinearity = getattr(layer, 'nonlinearity', None)
+    if nonlinearity is not None:
+        layer.nonlinearity = nonlinearities.identity
+    return ParametricRectifierLayer(layer, **kwargs)
--- a/lasagne/tests/layers/test_special.py
+++ b/lasagne/tests/layers/test_special.py
@@ -224,3 +224,108 @@ def test_transform_identity():
     thetas = floatX(np.tile([1, 0, 0, 0, 1, 0], (batchsize, 1)))
     outputs = layer.get_output_for([constant(inputs), constant(thetas)]).eval()
     np.testing.assert_allclose(inputs, outputs, rtol=1e-6)
+
+
+class TestParametricRectifierLayer:
+    @pytest.fixture
+    def ParametricRectifierLayer(self):
+        from lasagne.layers.special import ParametricRectifierLayer
+        return ParametricRectifierLayer
+
+    @pytest.fixture
+    def init_alpha(self):
+        # initializer for a tensor of unique values
+        return lambda shape: (np.arange(np.prod(shape)).reshape(shape)) \
+            / np.prod(shape)
+
+    def test_alpha_init(self, ParametricRectifierLayer, init_alpha):
+        input_shape = (None, 3, 28, 28)
+        # default: alphas only over 2nd axis
+        layer = ParametricRectifierLayer(input_shape, alpha=init_alpha)
+        alpha = layer.alpha
+        assert layer.shared_axes == (0, 2, 3)
+        assert alpha.get_value().shape == (3, )
+        assert np.allclose(alpha.get_value(), init_alpha((3, )))
+
+        # scalar alpha
+        layer = ParametricRectifierLayer(input_shape, alpha=init_alpha,
+                                         shared_axes='all')
+        alpha = layer.alpha
+        assert layer.shared_axes == (0, 1, 2, 3)
+        assert alpha.get_value().shape == ()
+        assert np.allclose(alpha.get_value(), init_alpha((1,)))
+
+        # alphas shared over the 1st axis
+        layer = ParametricRectifierLayer(input_shape, alpha=init_alpha,
+                                         shared_axes=0)
+        alpha = layer.alpha
+        assert layer.shared_axes == (0,)
+        assert alpha.get_value().shape == (3, 28, 28)
+        assert np.allclose(alpha.get_value(), init_alpha((3, 28, 28)))
+
+        # alphas shared over the 1st and 4th axes
+        layer = ParametricRectifierLayer(input_shape, alpha=init_alpha,
+                                         shared_axes=(0, 3))
+        alpha = layer.alpha
+        assert layer.shared_axes == (0, 3)
+        assert alpha.get_value().shape == (3, 28)
+        assert np.allclose(alpha.get_value(), init_alpha((3, 28)))
+
+    def test_undefined_shape(self, ParametricRectifierLayer):
+        with pytest.raises(ValueError):
+            ParametricRectifierLayer((None, 3, 28, 28), shared_axes=(1, 2, 3))
+
+    def test_get_output_for(self, ParametricRectifierLayer, init_alpha):
+        input_shape = (3, 3, 28, 28)
+        # random input tensor
+        input = np.random.randn(*input_shape).astype(theano.config.floatX)
+
+        # default: alphas shared only along 2nd axis
+        layer = ParametricRectifierLayer(input_shape, alpha=init_alpha)
+        alpha_v = layer.alpha.get_value()
+        expected = np.maximum(input, 0) + np.minimum(input, 0) * \
+            alpha_v[None, :, None, None]
+        assert np.allclose(layer.get_output_for(input).eval(), expected)
+
+        # scalar alpha
+        layer = ParametricRectifierLayer(input_shape, alpha=init_alpha,
+                                         shared_axes='all')
+        alpha_v = layer.alpha.get_value()
+        expected = np.maximum(input, 0) + np.minimum(input, 0) * alpha_v
+        assert np.allclose(layer.get_output_for(input).eval(), expected)
+
+        # alphas shared over the 1st axis
+        layer = ParametricRectifierLayer(input_shape, alpha=init_alpha,
+                                         shared_axes=0)
+        alpha_v = layer.alpha.get_value()
+        expected = np.maximum(input, 0) + np.minimum(input, 0) * \
+            alpha_v[None, :, :, :]
+        assert np.allclose(layer.get_output_for(input).eval(), expected)
+
+        # alphas shared over the 1st and 4th axes
+        layer = ParametricRectifierLayer(input_shape, shared_axes=(0, 3),
+                                         alpha=init_alpha)
+        alpha_v = layer.alpha.get_value()
+        expected = np.maximum(input, 0) + np.minimum(input, 0) * \
+            alpha_v[None, :, :, None]
+        assert np.allclose(layer.get_output_for(input).eval(), expected)
+
+    def test_prelu(self, init_alpha):
+        import lasagne
+        input_shape = (3, 28)
+        input = np.random.randn(*input_shape).astype(theano.config.floatX)
+
+        l_in = lasagne.layers.input.InputLayer(input_shape)
+        l_dense = lasagne.layers.dense.DenseLayer(l_in, num_units=100)
+        l_prelu = lasagne.layers.prelu(l_dense, alpha=init_alpha)
+        output = lasagne.layers.get_output(l_prelu, input)
+
+        assert l_dense.nonlinearity == lasagne.nonlinearities.identity
+
+        W = l_dense.W.get_value()
+        b = l_dense.b.get_value()
+        alpha_v = l_prelu.alpha.get_value()
+        expected = np.dot(input, W) + b
+        expected = np.maximum(expected, 0) + \
+            np.minimum(expected, 0) * alpha_v
+        assert np.allclose(output.eval(), expected)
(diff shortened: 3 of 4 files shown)


lasagne/layers/special.py
```diff
@@ -354,3 +356,63 @@ def _meshgrid(height, width):
...
+        See :meth:`Layer.create_param` for more information.
+    """
+    def __init__(self, incoming, alpha=init.Uniform((0.001, 0.1)), axes=None,
+                 **kwargs):
+        super(ParametricRectifierLayer, self).__init__(incoming, **kwargs)
+        if axes is None:
+            # default: the parameters are same across only second axis
+            # this will do correct thing for both dense operations
+            # and convolution
+            axes = (1, )
```
f0k: It's easier to just have `axes=1` in the definition then, instead of defaulting `axes=None` and checking it here. I guess you copied that from the `BiasLayer` or `BatchNormLayer`, where the default depended on the input dimensionality.

lasagne/layers/special.py
```diff
@@ -354,3 +356,63 @@ def _meshgrid(height, width):
...
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
+
+    a : Theano shared variable, numpy array or callable
+        An initializer for the weights of the layer. If a shared variable or a
+        numpy array is provided the shape should  be incoming.shape[1:]).
+        See :meth:`Layer.create_param` for more information.
```
f0k: `axes` is missing.

lasagne/layers/special.py
```diff
@@ -354,3 +356,63 @@ def _meshgrid(height, width):
...
+        elif isinstance(axes, int):
+            axes = (axes,)
+        shape = [1, ] * len(self.input_shape)
+        for axis in axes:
+            shape[axis] = self.input_shape[axis]
+        self.broadcast = tuple(set(range(len(self.input_shape))) - set(axes))
+        if any(size is None for size in shape):
+            raise ValueError("ParametricRectifierLayer needs specified input "
+                             "sizes for all dimensions/axes which constitute "
+                             "the same parameters.")
```
f0k: I don't understand this sentence (the last part, actually: "all dimensions/axes which constitute the same parameters").

lasagne/tests/layers/test_special.py
```diff
@@ -17,12 +17,12 @@ def layer_vars(self, NonlinearityLayer, dummy_input_layer):
         layer = NonlinearityLayer(
             dummy_input_layer,
             nonlinearity=nonlinearity,
-            )
+        )
```
f0k: Not your business! :) Please don't commit any whitespace changes in unrelated code. (Some editors do that automatically, reconfigure your editor or just make sure not to commit such changes.)

lasagne/tests/layers/test_special.py
```diff
@@ -17,12 +17,12 @@ def layer_vars(self, NonlinearityLayer, dummy_input_layer):
...
 
         return {
             'nonlinearity': nonlinearity,
             'layer': layer,
-            }
+        }
```
f0k: Same here.

lasagne/tests/layers/test_special.py
```diff
@@ -34,7 +34,7 @@ def test_init_none_nonlinearity(self, NonlinearityLayer,
         layer = NonlinearityLayer(
             dummy_input_layer,
             nonlinearity=None,
-            )
+        )
```
f0k: Same here.

lasagne/tests/layers/test_special.py
```diff
@@ -170,43 +170,135 @@ def test_transform_errors():
 
 
 def test_transform_downsample():
+    import lasagne
+    downsample = (0.7, 2.3)
```
f0k: I think you messed up something... that was from an old commit of mine.

f0k: Ah, no, I see, that was from an old commit of Søren, where the indentation was wrong. Then it's just a case of "not your business" again :)

lasagne/layers/special.py
f0k: And it seems the code is exactly opposite to the comment -- in your case, `axes` defines the axes the parameters are *not* shared over, but learned separately. In this case, `axes=None` could be used as a synonym for `axes=()`, i.e., having a scalar `alpha`.

lasagne/layers/special.py
```diff
@@ -354,3 +356,63 @@ def _meshgrid(height, width):
...
+
+
+def parametric_relu(layer, **kwargs):
+    """
+    Convenience function to apply parametric rectify to a given layer's output.
+    Will set the layer's nonlinearity to identity if there is one and will
+    apply the parametric rectifier instead.
+    @param layer: The `Layer` instance to apply the parametric relu; note that
+        it will be irreversibly modified as specified above
+    @return: A `ParametricRectifierLayer` instance stacked on the given `layer`
```
f0k: Should document `**kwargs` (like here: https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/conv.py#L166-L167), and follow the numpydoc style. I guess you copied this from my BatchNormLayer, which uses the Javadoc style.

lasagne/layers/special.py
```diff
@@ -354,3 +356,63 @@ def _meshgrid(height, width):
...
     return grid
+
+
+class ParametricRectifierLayer(Layer):
+    """
+    lasagne.layers.ParametricRectifierLayer(incoming,
+    alpha=init.Uniform((0.001, 0.1)), **kwargs)
+
+    A layer that applies parametric rectify nonlinearity to its input
+    following K He, X Zhang et al. (http://arxiv.org/abs/1502.01852):
```
f0k: The citation should follow the usual form (put `[1]_` here, then put the citation in a `References` section).

lasagne/layers/special.py
```diff
@@ -354,3 +356,63 @@ def _meshgrid(height, width):
...
+    :math:`\\varphi(y_i) = \\max(x_i,0) + \\alpha_i \\min(x_i,0)`
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
+
+    a : Theano shared variable, numpy array or callable
```
f0k: In the code it's called `alpha`, not `a`!

lasagne/layers/special.py
```diff
@@ -354,3 +356,63 @@ def _meshgrid(height, width):
...
+            # default: the parameters are same across only second axis
+            # this will do correct thing for both dense operations
+            # and convolution
+            axes = (1, )
+        elif isinstance(axes, int):
+            axes = (axes,)
+        shape = [1, ] * len(self.input_shape)
+        for axis in axes:
+            shape[axis] = self.input_shape[axis]
+        self.broadcast = tuple(set(range(len(self.input_shape))) - set(axes))
```
f0k: I'd prefer to store `self.axes` (because that's what was given in the constructor) and handle the broadcasting in `get_output_for()`. I think I'd also prefer `alpha` to have fewer dimensions and add them via `dimshuffle` instead of setting them to `1` and adding them via broadcasting. See `BiasLayer` or this prior implementation of a PReLULayer: https://github.com/Lasagne/Lasagne/pull/247/files#diff-bce2cd455789587cf49cbb4b404b5a3dR15
I like the idea of having `axes` specify the axes to untie over, rather than the axes to share over. This seems easier.

lasagne/layers/special.py
Botev: Ah here I did not know how to express myself in english and I knew it was messed up. Could help me. Basically what I want to say is that axes will give all dimensions on which there will be a different parameter (e.g. they won't be shared along the axis).

lasagne/layers/special.py
Botev: Correct, the comment was from my first commit of this, later I changed it. I think agree that `axes=1` for default cause that does the Dense and Conv nets implementation correct. I'll correct it.

Botev:
Can you just explain me, how to restart the branch at the end of the original Lasagne repo master branch last commit (e.g. it is something in the lines of git reset --hard ??? if I remember correctly from previous PR, but I do not know how to get the ????). I'll do that since all the errors from my editor gona be pain in the ass to fix otherwise ( as well as the changes on the transform layer commit)

f0k:
> Can you just explain me, how to restart the branch at the end of the original Lasagne repo master branch last commit

Something like `git fetch upstream master; git reset --hard upstream/master` might work. Instead of a hard reset, you can also do a plain reset: `git reset upstream/master` and then use `git add -p` for an interactive staging session where it asks you for each change whether you want to include it or not.

f0k:
Made some comments (inadvertently on the commit instead of the diff).

Apart from that, there's a merge conflict -- can you rebase your PR? This would be:
```
git fetch upstream master
git rebase upstream/master
# git will tell you there are conflicts, edit the files to fix them, look for ">>>" markers
# then run "git add" for the files you fixed, and "git rebase --continue"
git push --force
```

I think the shortcut function could be renamed to "prelu", that term is well-known enough.

And lastly, the new layer and the helper function need to be added to the documentation as well, namely to `docs/layers.rst` and `docs/layers/special.rst`! Otherwise they won't show up in the Sphinx documentation on readthedocs.

lasagne/layers/special.py
```diff
@@ -354,3 +355,95 @@ def _meshgrid(height, width):
...
+    alpha : Theano shared variable, numpy array or callable
+        An initializer for the weights of the layer. If a shared variable or a
+        numpy array is provided the shape should  match the `incoming` shape
+        in all dimension provided by the `axes` parameter and 1 in all others.
+        See :meth:`Layer.create_param` for more information.
+
+    axes : 'auto', 'all', int or tuple of int
+        The axes along which the parameters of the rectifier units are
+        going to be shared. If ``'auto'`` (the default),
+        share over all axes except for the second: this will share biases over
```
f0k: If you copy/paste, you should still adapt the text :) The ParametricRectifierLayer doesn't have any biases. Also the keyword was `shared_axes` in `BiasLayer`, it should be the same here if the meaning is the same.

lasagne/layers/special.py
```diff
@@ -354,3 +355,95 @@ def _meshgrid(height, width):
...
+        the minibatch dimension for dense layers, and additionally over all
+        spatial dimensions for convolutional layers. If ``'all'`` share over
+        all axes which corresponds to a single scalar parameter
```
f0k: Good idea! Could just use a comma before "share", and a full stop at the end of the sentence.

lasagne/layers/special.py
```diff
@@ -6,12 +6,13 @@
 from ..utils import as_tuple
 from .base import Layer, MergeLayer
 
-
```
f0k: why this?

lasagne/layers/special.py
```diff
@@ -6,12 +6,13 @@
...
 __all__ = [
     "NonlinearityLayer",
     "BiasLayer",
     "InverseLayer",
     "TransformerLayer",
+    "ParametricRectifierLayer",
+    "parametric_relu"
```
f0k: Should end with a comma. The idea is that you can always add or insert items without touching existing lines.

lasagne/layers/special.py
```diff
@@ -354,3 +355,95 @@ def _meshgrid(height, width):
...
+
+    Parameters
+    ----------
+    layer: a :class:`Layer` instance
+        The `Layer` instance to apply the parametric rectifier layer to;
+        note that it will be irreversibly modified as specified above
+
+    **kwargs
+        Any additional keyword arguments are passed to the
+        `ParametricRectifierLayer` class.
```
f0k: The correct markup is:
```
... to the :class:`ParametericRectifierLayer` class.
```
Otherwise it won't be linked to it in the generated documentation.

Botev:
Someone to review?

docs/modules/layers/special.rst
```diff
@@ -17,3 +17,8 @@ Special-purpose layers
 .. autoclass:: TransformerLayer
     :members:
 
+.. autoclass:: ParametricRectifierLayer
+    :memmbers:
```
ebenolson: extra m

lasagne/layers/special.py
```diff
@@ -352,3 +353,96 @@ def _meshgrid(height, width):
...
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
+
+    alpha : Theano shared variable, numpy array or callable
+        An initializer for the weights of the layer. If a shared variable or a
+        numpy array is provided the shape should  match the `incoming` shape
+        subindexed by the negation of the `shared_axes` set.
```
ebenolson: I feel like there must be a simpler way to say this.. I see @f0k commented earlier on (y, z) vs (1, y, z), but I think "the shape should be broadcastable to match the `incoming` shape" is much more understandable.

lasagne/tests/layers/test_special.py
```diff
@@ -224,3 +224,118 @@ def test_transform_identity():
...
+        input = np.random.randn(*input_shape).astype(theano.config.floatX)
+
+        # default: alphas shared only along 2nd axis
+        layer = ParametricRectifierLayer(input_shape, alpha=init_alpha)
+        alpha_v = layer.alpha.get_value()
+        expected = np.copy(input)
+        for i in range(3):
+            a = alpha_v[i]
+            expected[:, i, :, :] = np.maximum(input[:, i, :, :], 0) + \
+                np.minimum(input[:, i, :, :], 0) * a
```
ebenolson: not really important, but these could be written more compactly with broadcasting:

    expected = np.maximum(input, 0) + np.minimum(input, 0) * alpha_v[None, :, None, None]

lasagne/layers/special.py
f0k: "broadcastable to match the `incoming` shape" is not generally correct, though. Why don't you just copy/paste my docstring from the `BiasLayer`, or a part of it?

lasagne/layers/special.py
f0k: This still needs to be fixed! We don't want random whitespace changes (at the best they make commit diffs longer than needed, and at the worst they cause merge conflicts with other PRs). Sorry for the hassle.

Botev:
Should be done.

lasagne/layers/special.py
```diff
@@ -6,12 +6,13 @@
...
 from .base import Layer, MergeLayer
 
-
 __all__ = [
     "NonlinearityLayer",
     "BiasLayer",
     "InverseLayer",
     "TransformerLayer",
+    "ParametricRectifierLayer",
+    "prelu"
```
f0k: This still needs to end with a comma (as I already mentioned a week ago: https://github.com/Lasagne/Lasagne/pull/446#discussion_r41035676).

lasagne/tests/layers/test_special.py
f0k: @Botev: This is still open. Eben's line replaces all of your `expected = np.copy(input)` and the following `for` loop. That's much easier to read, and a better template in case anybody copies this for another test!

lasagne/tests/layers/test_special.py
```diff
@@ -224,3 +224,117 @@ def test_transform_identity():
...
+        # alphas shared over the 1st and 4th axes
+        layer = ParametricRectifierLayer(input_shape, shared_axes=(0, 3),
+                                         alpha=init_alpha)
+        alpha_v = layer.alpha.get_value()
+        expected = np.copy(input)
+        for i in range(3):
+            for j in range(28):
+                a = alpha_v[i, j]
+                expected[:, i, j, :] = np.maximum(input[:, i, j, :], 0) + \
+                    np.minimum(input[:, i, j, :], 0) * a
```
f0k: Same here, you can just write `expected = np.maximum(input, 0) + np.minimum(input, 0) * alpha_v[None, :, :, None]` instead of the loop.

lasagne/layers/special.py
```diff
@@ -352,3 +353,95 @@ def _meshgrid(height, width):
...
+            self.shared_axes = (shared_axes,)
+        else:
+            self.shared_axes = shared_axes
+
+        shape = [size for axis, size in enumerate(self.input_shape)
+                 if axis not in self.shared_axes]
+        if any(size is None for size in shape):
+            raise ValueError("ParametricRectifierLayer needs specified input "
+                             "sizes for all dimensions/axes which constitute "
+                             "the same parameters.")
```
f0k: `constitute the same parameters` is still unclear, can't you just copy the error message from the `BiasLayer`?

f0k:
> Should be done.

These ones are still open:
https://github.com/Lasagne/Lasagne/pull/446/files#r41427611
https://github.com/Lasagne/Lasagne/pull/446/files#r41636499
https://github.com/Lasagne/Lasagne/pull/446/files#r41404971 (although you had a fix in https://github.com/Botev/Lasagne/commit/abb4d1f858fd9c36b55d6c86ce5ebab23591d10e#commitcomment-13688141, but that was wrong and it's not included in the PR diff, probably it is screwed up by some merges in between -- maybe try to squash all commits and strongly avoid merges in PR branches!)
https://github.com/Lasagne/Lasagne/pull/446/files#r41637156
https://github.com/Lasagne/Lasagne/pull/446/files#r41638254

Please carefully read and address all comments in the PR. Bonus points if you comment on the ones you addressed, that makes it easy to track which ones were addressed and which ones are marked "comment on an outdated diff" for unrelated changes. Thanks!

Botev:
I still don't understand why sometimes Git makes me do this merges. If you check them out there are 0 changed files? But still Git insists that my branch is somehow behind? This happened when I needed to reset my local changes back to the repo. Sorry about it, but did not figure it out how to force git not to do it.

f0k:
> I still don't understand why sometimes Git makes me do this merges.

When the upstream repo and yours have diverged, it tells you that they have diverged and suggests to do a `git pull` to merge in the remote changes. You need to remember what you did and either do a `git push --force` instead to overwrite the remote branch, or a `git pull --rebase` to replay your changes on top of the remote changes in case you want to include them. Both will give you a clean history without merges.

f0k:
> If you check them out there are 0 changed files?

Yes, but the merge commit has two parents, and those lead to chains of commits that actually did something.

Botev:
So if this one passes I think it should have now fixed all the issue with merges and addressed the ones that were commented on.

lasagne/layers/special.py
```diff
@@ -352,3 +354,94 @@ def _meshgrid(height, width):
...
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
+
+    alpha : Theano shared variable, numpy array or callable
+        An initializer for the alpha parameters. If a shared variable or numpy
+        array is provided, the shape must match the incoming shape, skipping
+        those axes the alpha's are shared over (see below for an example).
```
craffel: It says "see below for an example", but I don't see an example.  Sorry if I'm missing it.

Botev:
That line was added because I've deleted it earlier as you mention in https://github.com/Lasagne/Lasagne/pull/446#discussion_r41427611 
In the original repo there are exactly 2 lines before and after __all__ definitions.
Also In my commit there are no spaces on those lines (at least my editor says so).

f0k:
> That line was added because I've deleted it earlier

Ah, sorry. It's okay then.

lasagne/layers/special.py
f0k: Okay, last request: With #404, the text needs to change a bit. Copied from the new `BiasLayer` and adapted for your needs:
```
    alpha : Theano shared variable, expression, numpy array or callable
        Initial value, expression or initializer for the alpha values. The
        shape must match the incoming shape, skipping those axes the alpha
        values are shared over (see the example below).
        See :func:`lasagne.utils.create_param` for more information.
```
Note that this also fixes `alpha's`, please just copy it verbatim.

f0k:
Looks good except for one last thing! Could you fix this and then squash all your commits into one? Do you know how?

Botev:
I guess this was what I was suppose to do: http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

f0k:
Yes! Except that the commit message should be something like "Add ParametricRectifyLayer". And you're still missing my last request: https://github.com/Lasagne/Lasagne/pull/446#discussion_r42125511
If you implement that and do "git add", you can simply do "git commit --amend" afterwards to amend your existing commit, and edit the commit message in one go. Then force-push again and I think we're finally good to merge (although I will re-read everything to make sure none of your earlier commits went missing).

Botev:
Done!

f0k:
Fine, looks good to merge to me! I'll leave it open for comments until the evening.

craffel:
Made two minor docstring comments.  Also, the usage of `prelu` is little unusual, right?  E.g. rather than
```Python
layer = lasagne.layers.DenseLayer(incoming, n_hidden, nonlinearity=lasagne.layers.prelu)
```
it's
```Python
layer = lasagne.layers.DenseLayer(incoming, n_hidden)
layer = lasagne.layers.prelu(layer)
```
right?  If so, I'd suggest throwing in that latter code block as an example in `prelu`'s docstring.

f0k:
> If so, I'd suggest throwing in that latter code block as an example in `prelu`'s docstring.

Good idea. It lives in `lasagne.layers` rather than `lasagne.nonlinearities`, so I hope only few users will be tempted to try `nonlinearity=lasagne.layers.prelu`.

Botev:
So ready?

f0k:
Sorry, there's no notification for new commits to a PR, only for new comments.

> So ready?

Colin's other point (https://github.com/Lasagne/Lasagne/pull/446#issuecomment-148748536) would still be good to address so people don't use it wrongly. Add an example block to the `prelu()` docstring, like:
```
Examples
--------
Note that this function modifies an existing layer, like this:
>>> from lasagne.layers import InputLayer, DenseLayer, prelu
>>> layer = InputLayer((32, 100))
>>> layer = DenseLayer(layer, num_units=200)
>>> layer = prelu(layer)
In particular, `prelu` can *not* be passed as a nonlinearity.
```
Please let us know when you're done, and when amending the commit, please leave the commit message at "Add ParametricRectifyLayer". Changing it to "finalizes docs based on ..." doesn't make sense if it's the only commit in a PR!

Botev:
Added that to the doc string.

lasagne/layers/special.py
```diff
@@ -352,3 +354,113 @@ def _meshgrid(height, width):
...
+        :class:`ParametericRectifierLayer`
+
+    Examples
+    --------
+    Note that this function modifies an existing layer, like this:
+    >>> from lasagne.layers import InputLayer, DenseLayer, prelu
+    >>> layer = InputLayer((32, 100))
+    >>> layer = DenseLayer(layer, num_units=200)
+    >>> layer = prelu(layer)
+    In particular, func:`prelu` can *not* be passed as a nonlinearity.
```
f0k: Sorry, this needs an empty line between the last `>>>` and the `In particular` -- my bad. Also it should be `:func:`, not `func:` -- also my bad. Cheers!

Botev:
Hope done now

f0k:
Yep, Travis passes! I'm fine with merging (there's a typo in the commit comment and another one in the docstring formula, but whatever). Any other comments?

Botev:
Fixed and that. No more comments.

f0k:
> Fixed and that.

Actually, you replaced the previous commit comment typo with another, but that's unimportant. Merging!

(Sorry for the delay, I'm at a conference and had to prepare for that.)

moscow25:
Yay! Thank Jan. Have been waiting to try the parametric ReLUs.

PR #449, opened 2015-10-01
Fix bug in LSTMLayer where peephole to outgate was added before nonlinearity

### LINKED ISSUE #448: Issue with LSTM output
I noticed LSTM layer output is a little bit off. 

Even with sigmoid activation I would sometimes get 1 < output or output < 0. 

The value of the cell to output gate was being added after the activation function when peepholes are enabled.

At line 970~ in recurrent.py
```python
            outgate = self.nonlinearity_outgate(outgate)

            # Compute new cell value
            cell = forgetgate*cell_previous + ingate*cell_input

            if self.peepholes:
                outgate += cell*self.W_cell_to_outgate
```

should be changed to:
```python
            # Compute new cell value
            cell = forgetgate*cell_previous + ingate*cell_input

            if self.peepholes:
                outgate += cell*self.W_cell_to_outgate
	
            outgate = self.nonlinearity_outgate(outgate)
```

I'm pretty sure there's a way to pull or push or something but I don't know how. No real experience with Github. I hope I helped anyway.

skaae:
Fixes  #448: peepholes being added after outgate nonlin

### DIFF
--- a/lasagne/layers/recurrent.py
+++ b/lasagne/layers/recurrent.py
@@ -967,13 +967,13 @@ class LSTMLayer(MergeLayer):
             ingate = self.nonlinearity_ingate(ingate)
             forgetgate = self.nonlinearity_forgetgate(forgetgate)
             cell_input = self.nonlinearity_cell(cell_input)
-            outgate = self.nonlinearity_outgate(outgate)
 
             # Compute new cell value
             cell = forgetgate*cell_previous + ingate*cell_input
 
             if self.peepholes:
                 outgate += cell*self.W_cell_to_outgate
+            outgate = self.nonlinearity_outgate(outgate)
 
             # Compute new hidden unit activation
             hid = outgate*self.nonlinearity(cell)


skaae:
This is ready for review.

benanne:
Thanks! Looks okay to me :)

Is there any way to test for things like this that's not too cumbersome? I guess it's pretty challenging, but it would be good to figure out how we can prevent logical errors like this one slipping by the radar.

skaae:
Mayb calculating correct values for a single step?

f0k:
Could you please give the PR a real title that describes what it changes, not just "Fixes ..."? Makes it easier to see what's going on when browsing the PRs (also in retrospect). Other than that, it looks good!

> Is there any way to test for things like this that's not too cumbersome? I guess it's pretty challenging, but it would be good to figure out how we can prevent logical errors like this one slipping by the radar.

I think we can only find this by reviewing, and for the recurrent layers the only two RNN experts wrote the code, so there was no RNN expert left for reviewing it.

> Mayb calculating correct values for a single step?

How would you have computed the correct values? Wouldn't you have done the same mistake then? We can just add some sanity check such as ensuring the output of recurrent layers with tanh nonlinearities always stays within (-1,1), for all kinds of different configurations... but that'd be fishing in the dark.

skaae:
> How would you have computed the correct values? Wouldn't you have done the same mistake then?

Hopefully not :) At least the equation in the help file correct so it was "just" not correctly implemented.

skaae:
I changed the title

f0k:
OK, I'm fine with merging, but we should send a notification to the mailing list that `LSTMLayer` changes in an incompatible way if `peepholes=True` (which is the default) -- not in the sense that end user code needs to be changed, but that previously saved models will behave differently.

PR #451, opened 2015-10-02
Recurrent layers: Also allow grad_clipping=0 to disable gradient clipping

f0k:
Right now, to disable gradient clipping in recurrent layers, you need to set `grad_clipping=False`. This is tested for in the code via `grad_clipping is False`, so any other falseish value will not work. This PR changes the implementation to also have `grad_clipping=0` (or any falseish value) disable gradient clipping. Right now, `grad_clipping=0` completely disables backpropagation through the recurrent layer -- I don't think there's an important use case for this, and I know at least one person who would have expected `grad_clipping=0` to mean the same as `grad_clipping=False` and wondered why their networks wouldn't learn anything.


skaae:
Looks good.

benanne:
Good call.

PR #453, opened 2015-10-04
added Upscale1DLayer implementation

kylemcdonald:
This adds Upscale1DLayer, based on the Upscale2DLayer implementation. I've trained with it and it seems to work correctly.

It would be nice to have something like UpscaleNDLayer with a mode parameter that could be 'nearest' (current behavior), 'bilinear', or something else, similar to Pool1DLayer or Pool2DLayer. Though I'm not sure if there is any literature supporting the possibility of other kinds of upscaling.

### DIFF
--- a/docs/modules/layers.rst
+++ b/docs/modules/layers.rst
@@ -81,6 +81,7 @@
     MaxPool2DLayer
     Pool1DLayer
     Pool2DLayer
+    Upscale1DLayer
     Upscale2DLayer
     GlobalPoolLayer
     FeaturePoolLayer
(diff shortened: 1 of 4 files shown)


kylemcdonald:
hmm according to travis this is failing, sorry -- i'll make sure i did this right.

kylemcdonald:
ok! that was a bit of a learning experience, but i think i did everything to spec. except the "check for an issue before adding a feature" :) please let me know if there's anything else i can do to make this helpful.

f0k:
Cool, this looks perfect, and you even squashed it into a single commit already! I'm fine with merging!

benanne:
Nice work, thanks!

f0k:
Merged, thanks again :+1:

PR #455, opened 2015-10-05
Clarify CustomRecurrentLayer input_to_hidden docs

### LINKED ISSUE #452: Docs of the CustomRecurrentLayer
In the docs the examples for the `CustomRecurrentLayer` and the docs don't match. Particularly - ``
```
input_to_hidden : :class:`lasagne.layers.Layer`
        :class:`lasagne.layers.Layer` instance which connects input to the
        hidden state (:math:`f_i`).  This layer may be connected to a chain of
        layers, which must end in a :class:`lasagne.layers.InputLayer` with the
        same input shape as `incoming`. 
```
However, in the examples to `incoming` has shape `(n_batch, n_steps, n_in)` while the bottom `InputLayer` of the `input_to_hidden` has shape `(None, n_in)`. The same applies an the example of convolutional recurrence where the `incoming` has shape `(n_batch, n_steps, n_channels, width, height)`, while the bottom of the `input_to_hidden` has shape `(None, n_channels, width, height)`. From this it seems that the `InputLayer` of the `input_to_hidden` must have shape `incoming.shape[1:]`.

Or is it intentional that it takes `None` at the start so that you flat out the batch and the input length?

craffel:
Resolves #452.   Someone (@skaae @f0k @Botev) should double-check my logic.

### DIFF
--- a/lasagne/layers/recurrent.py
+++ b/lasagne/layers/recurrent.py
@@ -107,7 +107,11 @@ class CustomRecurrentLayer(MergeLayer):
         :class:`lasagne.layers.Layer` instance which connects input to the
         hidden state (:math:`f_i`).  This layer may be connected to a chain of
         layers, which must end in a :class:`lasagne.layers.InputLayer` with the
-        same input shape as `incoming`.
+        same input shape as `incoming`, except for the first dimension: When
+        ``precompute_input == True`` (the default), it must be
+        ``incoming.output_shape[0]*incoming.output_shape[1]`` or ``None``; when
+        ``precompute_input == False``, it must be ``incoming.output_shape[0]``
+        or ``None``.
     hidden_to_hidden : :class:`lasagne.layers.Layer`
         Layer which connects the previous hidden state to the new state
         (:math:`f_h`).  This layer may be connected to a chain of layers, which


lasagne/layers/recurrent.py
```diff
@@ -106,8 +106,13 @@ class CustomRecurrentLayer(MergeLayer):
     input_to_hidden : :class:`lasagne.layers.Layer`
         :class:`lasagne.layers.Layer` instance which connects input to the
         hidden state (:math:`f_i`).  This layer may be connected to a chain of
-        layers, which must end in a :class:`lasagne.layers.InputLayer` with the
-        same input shape as `incoming`.
+        layers, which must end in a :class:`lasagne.layers.InputLayer`.
+        ``input_to_hidden.input_shape[1:]`` should always be the same as
```
f0k: That needs to be formulated differently... `input_to_hidden.input_shape[1:]` is the input shape of the last layer of `input_to_hidden`, while we're interested in the input shape of the aforementioned `InputLayer`.

f0k: Maybe just start this off with: `... InputLayer with the same input shape as incoming, except for the first dimension: This must either be None, or ... (when precompute_input=True), or ... (when precompute_input=False).`

craffel: Ok, changed.

lasagne/layers/recurrent.py
```diff
@@ -107,7 +107,12 @@ class CustomRecurrentLayer(MergeLayer):
         :class:`lasagne.layers.Layer` instance which connects input to the
         hidden state (:math:`f_i`).  This layer may be connected to a chain of
         layers, which must end in a :class:`lasagne.layers.InputLayer` with the
-        same input shape as `incoming`.
+        same input shape as incoming, except for the first dimension: When
+        ``precompute_input == True`` (the default),
+        ``input_to_hidden.input_shape[0]`` must be
```
f0k: Hey, that's the same mistake as before. You can't refer to this as `input_to_hidden.input_shape`, because that would be the input to the last layer of `input_to_hidden`, which could be a deep network. Just say "it", which then refers to "except for the first dimension" of "the same input shape" of the ":class:`lasagne.layers.InputLayer`". Also wouldn't it be easier to name `None` only once instead of twice? (https://github.com/Lasagne/Lasagne/pull/455#discussion_r41174756)

craffel: Oops, I only fixed it in one place, sorry.  The wording of naming `None` once is confusing to me.  If you want to change it or make other fixes, maybe it would be faster for you to make them then squash/merge?  This is a pretty small change.

f0k: > If you want to change it or make other fixes, maybe it would be faster for you to make them then squash/merge?

Hmm, I don't like pushing something to master that's not reviewed... but we can discuss the formulation here instead of going via commits. So my first proposal was:
```
with the same input shape as `incoming`, except for the first dimension:
That must be ``None``,
or ``incoming.output_shape[0]*incoming.output_shape[1]`` if ``precompute_input=True``,
or ``incoming.output_shape[0]`` if ``precompute_input=False``.
```

If that reads confusing for you, what about putting `None` in the end:
```
with the same input shape as `incoming`, except for the first dimension:
That must be ``incoming.output_shape[0]*incoming.output_shape[1]``
if ``precompute_input=True`` (the default), or ``incoming.output_shape[0]``
if ``precompute_input=False``, or simply left unspecified as ``None``.
```

lasagne/layers/recurrent.py
```diff
@@ -107,7 +107,11 @@ class CustomRecurrentLayer(MergeLayer):
         :class:`lasagne.layers.Layer` instance which connects input to the
         hidden state (:math:`f_i`).  This layer may be connected to a chain of
         layers, which must end in a :class:`lasagne.layers.InputLayer` with the
-        same input shape as `incoming`.
+        same input shape as incoming, except for the first dimension: When
+        ``precompute_input == True`` (the default), it must be
+        ``incoming.output_shape[0]*incoming.output_shape[1]`` or `None`; when
+        ``precompute_input == False``, it must be ``incoming.output_shape[0]``
+        or `None`.
```
f0k: Ah sorry, didn't see you committed in between. This formulation is also fine with me! There are just two markup errors: "incoming" should have single backticks because it refers to the argument, and "None" should have double backticks because it's Python code.

craffel: > There are just two markup errors: "incoming" should have single backticks 

Oops, fixed.

> "None" should have double backticks because it's Python code. We've got markup errors like these all over the place, but let's get this right while we're at it!

Hm, then there are a few other places where that needs to be fixed too:
https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/recurrent.py#L150
https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/recurrent.py#L525
and in other files
https://github.com/Lasagne/Lasagne/blob/b22500e736bf1b8f773a5e11257c272880aa6dd6/lasagne/layers/input.py#L25
https://github.com/Lasagne/Lasagne/blob/b22500e736bf1b8f773a5e11257c272880aa6dd6/lasagne/layers/input.py#L28
https://github.com/Lasagne/Lasagne/blob/b22500e736bf1b8f773a5e11257c272880aa6dd6/lasagne/layers/input.py#L43
https://github.com/Lasagne/Lasagne/blob/7e39d52075772ef9351f9d9face40c1a0a955491/lasagne/layers/pool.py#L43
etc.

Perhaps we need to do a docstring audit?  Which convention would you prefer I use in this PR?

f0k: > Perhaps we need to do a docstring audit?

Yes, at some point we definitely need that. We also need to unify the parameter type conventions used throughout the docs. The downside is that this will touch a lot of code, so this should probably be done at a time where there are no open PRs.

> Which convention would you prefer I use in this PR?

I'd prefer the one with double backticks. Single backticks are for variables, double backticks are for code: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections
But we can also leave that for a future audit, `recurrent.py` uses both single backticks and no backticks already. Unifying this is out of scope for this PR anyway.

craffel: Ok, should be fixed.

f0k: Great, thanks! :+1:

f0k:
Merging, no need to wait for Travis again. Thank you!

PR #456, opened 2015-10-07
KL divergence 

botev:
Following on https://groups.google.com/forum/#!topic/lasagne-users/w5ReeOIIzSM I'm attempting to implement a KL divergence measure for parametric expressions of the distributions. Started with a Gaussian and wanted to see what you guys think. 

Also if someone could give me a hint how to properly address the case where the full covariance matrix is given how to calculate efficiently (per example point) `mu^T Sigma mu`. In such case the `mu` will be a matrix of size `(batch_size, dim)` while `Sigma` should be `(batch_size, dim, dim)`. I have no idea how to paralelize this? In the other cases it should work as it is (possibly im still missing something).

Further this could be implemented for number of different distributions.

### DIFF
--- a/lasagne/objectives.py
+++ b/lasagne/objectives.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 """
 Provides some minimal help with building loss expressions for training or
 validating a neural network.
@@ -76,8 +78,6 @@ This gives a loss expression good for monitoring validation error.
 """
 
 import theano.tensor.nnet
-
-from lasagne.layers import get_output
 
 __all__ = [
     "binary_crossentropy",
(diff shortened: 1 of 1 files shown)


skaae:
Is this still relevant? We also have something similar in parmesan:

https://github.com/casperkaae/parmesan/blob/master/parmesan/distributions.py

f0k:
> Is this still relevant?

The current proposal tries to be very generic, but then just implements the KL divergence for Gaussian distributions with a convoluted and undocumented way of passing the parameters.

> We also have something similar in parmesan

This looks more like it, yes! There'd still be two different cases we should support (diag-cov against diag-cov, full-cov against full-cov), and yours just handles the special case diag-cov against standard normal (mean=0, var=1).

But the general question is: Do we need this in Lasagne? Are there enough use cases that warrant inclusion? For now I doubt so. (I only needed it once, and then I just wrote down the cost function directly.)

skaae:
> But the general question is: Do we need this in Lasagne? Are there enough use cases that warrant inclusion? For now I doubt so. (I only needed it once, and then I just wrote down the cost function directly.)

Parmesan has it for variational autoencoders where it is relevant. But Lasagne doesn't support VAE's so I don't think it should be included.

f0k:
Closing because it's not general enough, and we couldn't agree we need this in Lasagne. Thanks anyway!

PR #460, opened 2015-10-12
Check that Theano is recent enough

### LINKED ISSUE #454: rectify use theano.tensor.nnet.relu which is not exist
Hi, When I execute the code layers.get_output_for (In line 185, lasagne.layers.helper.py ), the procedure will execute the function rectify, here it encapsulate theano.tensor.nnet.relu which is not exist.
 How should I fix this problem?
We see that in the lasagne.layers.Conv2DCCLayer.__init__, there is default setting nonlinearity = nonlinearities.rectify, also in the function, there is branch depends on if the nonliearity is None
which seems inconsistence

f0k:
This adds an additional check to `__init__.py` to not only test whether Theano is importable, but also test whether it is recent enough to support what's needed by Lasagne. The most recent feature we actually use is `theano.tensor.nnet.relu`. There are some commits between the addition of that feature to Theano and the commit we've got in our `requirements.txt`, but it's unlikely that anybody uses this and even if so, it wouldn't cause any problems -- the main goal is to catch users who merely installed the latest Theano release.
Corrects the link to the installation instructions, and clarifies that `pip install theano` is not the solution.
This closes #454.

### DIFF
--- a/lasagne/__init__.py
+++ b/lasagne/__init__.py
@@ -3,16 +3,21 @@ Tools to train neural nets in Theano
 """
 
 try:
+    install_instr = """
+
+Please make sure you install a recent enough version of Theano. Note that a
+simple 'pip install theano' will usually give you a version that is too old
+for Lasagne. See the installation docs for more details:
+http://lasagne.readthedocs.org/en/latest/user/installation.html#theano"""
     import theano
 except ImportError:  # pragma: no cover
-    raise ImportError("""Could not import Theano.
-
-Please make sure you install a recent enough version of Theano.  See
-section 'Install from PyPI' in the installation docs for more details:
-http://lasagne.readthedocs.org/en/latest/user/installation.html#install-from-pypi
-""")
+    raise ImportError("Could not import Theano." + install_instr)
 else:
+    if not hasattr(theano.tensor.nnet, 'relu'):  # pragma: no cover
+        raise ImportError("Your Theano version is too old." + install_instr)
+    del install_instr
     del theano
+
 
 from . import nonlinearities
 from . import init


f0k:
Well, this is silly. pep8 fails with `E402 module level import not at top of file`. We're allowed to do `try: import theano; except: raise Something` before importing all the lasagne submodules, but we're neither allowed to do `if not hasattr(theano....)`, nor are we allowed to define a string variable for the long text message that's to be shared between "you don't have Theano, please do ..." and "your Theano is too old, please do ...".

Shall we disable E402 for `__init__.py`, or move the `if ...` check to the end and duplicate the string? We can't move the `try: import theano` to the end because we need to run that before importing any submodule.

ebenolson:
you could also put the check into `checktheano.py` and import that.

dnouri:
I think you can also add a `# noqa` to the line you want the pep8 checker to ignore.

f0k:
> you could also put the check into `checktheano.py` and import that.

But in that file I still wouldn't be allowed to define a string before `try: import theano`. Oh wait, it seems I am.

> I think you can also add a # noqa to the line you want the pep8 checker to ignore.

That works, but requires to add `#noqa` to every `from . import ...` line.

I'll try to go the `_check_theano.py` route.

f0k:
There was an easier solution: Moving everything inside the `try: except: else:` block. pep8 doesn't complain about anything we do there.

f0k:
Merging later this afternoon if there are no objections. If the check ever gets more complicated, we can still move it to an extra file.

nouiz:
Do you use THeano batch normalization? If so, it was added after relu, so you should check for that one.

f0k:
> Do you use THeano batch normalization? If so, it was added after relu, so you should check for that one.

Thanks for your attention :) No, we don't have batch normalization included yet, I should do that... Our "requirements.txt" lists Theano.git@15c90dd3, and the latest feature we use is the `relu()` function we moved there from Lasagne.

The plan would be to update the version check whenever we need a new feature from Theano, and update our `requirements.txt` to mention a Theano version or git commit which includes this feature.

PR #465, opened 2015-10-16
Added softplus non-linearity

wuaalb:


lasagne/tests/test_nonlinearities.py
```diff
@@ -16,6 +16,9 @@ def leaky_rectify(self, x):
     def leaky_rectify_0(self, x):
         return self.rectify(x)
 
+    def softplus(self, x):
+        return np.log(1 + np.exp(x))
```
f0k: `np.log1p(np.exp(x))` should be more accurate for very small `x`. Not that it matters for the test, but somebody might look here and copy it. If you change that, please amend your commit and force-push rather than adding a new commit. Thank you!

f0k:
Thanks, looks good except for one minor nitpick!

wuaalb:
OK, updated.. thanks!

f0k:
Perfect, I'm fine with merging!

f0k:
Oh, this slipped off the radar. Merging, thanks again!

Issue #429, opened 2015-09-23
CustomRecurrentLayer with DropoutLayer in hid_to_hid fails to compile


f0k:
From the mailing list: https://groups.google.com/forum/#!topic/lasagne-users/8E1iR04_nK8

The following code fails:
```python
import theano
import lasagne
n_batch, n_steps, n_in = (2, 3, 4)
n_hid = 5
l_in = lasagne.layers.InputLayer((n_batch, n_steps, n_in))
l_in_hid = lasagne.layers.DenseLayer(
    lasagne.layers.InputLayer((None, n_in)), n_hid)
l_hid_hid = lasagne.layers.DropoutLayer(lasagne.layers.DenseLayer(
    lasagne.layers.InputLayer((None, n_hid)), n_hid))
l_rec = lasagne.layers.CustomRecurrentLayer(l_in, l_in_hid, l_hid_hid)
out = lasagne.layers.get_output(l_rec)
fn = theano.function([l_in.input_var], out)
```
The exception is:
```
MissingInputError: ("An input of the graph, used to compute dot(<TensorType(float32, matrix)>, HostFromGpu.0), was not provided and not given a value.Use the Theano flag exception_verbosity='high',for more information on this error.", <TensorType(float32, matrix)>)
```
With exception_verbosity='high', it is:
```
theano.gof.fg.MissingInputError: A variable that is an input to the graph was neither provided as an input to the function nor given a value. A chain of variables leading from this input to an output is [<TensorType(float32, matrix)>, dot.0, Elemwise{add,no_inplace}.0, Elemwise{add,no_inplace}.0, Elemwise{mul,no_inplace}.0, Elemwise{true_div,no_inplace}.0, Shape.0, Elemwise{Cast{int32}}.0, GPU_mrg_uniform{CudaNdarrayType(float32, matrix),no_inplace}.0]. This chain may not be unique
Backtrace when the variable is created:
  File "/raid/user/jan/install/Lasagne-git/lasagne/layers/recurrent.py", line 451, in get_output_for
    strict=True)[0]
```
That line is: https://github.com/Lasagne/Lasagne/blob/fe2608c9c675a437dd5bbcb577e6f09817fc57bc/lasagne/layers/recurrent.py#L451

It is the call to `theano.scan`.

The error disappears when disabling the dropout layer (by setting `p=0`). Is this a Theano bug, or is there something special to do when using a symbolic RNG within scan?

f0k:
[...]

PR #467, opened 2015-10-16
Add BatchNormLayer

### LINKED ISSUE #141: Batch Normalization

f0k:
Finally started integrating my batch normalization implementation in Lasagne. Changes from [the implementation in my gist](https://gist.github.com/f0k/f1a6bd3c8585c400c190):
* Adding epsilon before taking the square root of the variance
* Change parameters to have reduced dimensionality rather than broadcastable singleton dimensions, to match with `BiasLayer` (which in turn was made to match `DenseLayer`, `Conv2DLayer` etc.)

TODO:
* [x] add tests
* [x] fix LocalResponseNormalization2DLayer docstring while we're at it

Resolves #141.

### DIFF
--- a/lasagne/layers/normalization.py
+++ b/lasagne/layers/normalization.py
@@ -162,11 +162,6 @@ class BatchNormLayer(Layer):
         Coefficient for the exponential moving average of batch-wise means and
         standard deviations computed during training; the closer to one, the
         more it will depend on the last batches seen
-    mode : {'low_mem', 'high_mem'}
-        Specify which batch normalization implementation to use: ``'low_mem'``
-        avoids storing intermediate representations and thus requires less
-        memory, while ``'high_mem'`` can reuse representations for the backward
-        pass and is thus 5-10% faster.
     beta : Theano shared variable, expression, numpy array, callable or None
         Initial value, expression or initializer for :math:`\\beta`. Must match
         the incoming shape, skipping all axes in `axes`. Set to ``None`` to fix
@@ -272,19 +267,17 @@ class BatchNormLayer(Layer):
 
     def get_output_for(self, input, deterministic=False, **kwargs):
         input_mean = input.mean(self.axes)
-        input_std = T.sqrt(input.var(self.axes) + self.epsilon)
+        input_inv_std = T.inv(T.sqrt(input.var(self.axes) + self.epsilon))
 
         # Decide whether to use the stored averages or mini-batch statistics
         use_averages = kwargs.get('batch_norm_use_averages',
                                   deterministic)
         if use_averages:
             mean = self.mean
-            std = None
             inv_std = self.inv_std
         else:
             mean = input_mean
-            std = input_std
-            inv_std = None
+            inv_std = input_inv_std
 
         # Decide whether to update the stored averages
         update_averages = kwargs.get('batch_norm_update_averages',
@@ -299,11 +292,12 @@ class BatchNormLayer(Layer):
                                            self.alpha * input_mean)
             running_inv_std.default_update = ((1 - self.alpha) *
                                               running_inv_std +
-                                              self.alpha / input_std)
+                                              self.alpha * input_inv_std)
             # and make sure they end up in the graph without participating in
             # the computation (this way their default_update will be collected
             # and applied, but the computation will be optimized away):
-            mean += 0 * running_mean + 0 * running_inv_std
+            mean += 0 * running_mean
+            inv_std += 0 * running_inv_std
 
         # prepare dimshuffle pattern inserting broadcastable axes as needed
         param_axes = iter(range(input.ndim - len(self.axes)))
@@ -315,15 +309,10 @@ class BatchNormLayer(Layer):
         beta = 0 if self.beta is None else self.beta.dimshuffle(pattern)
         gamma = 1 if self.gamma is None else self.gamma.dimshuffle(pattern)
         mean = mean.dimshuffle(pattern)
-        std = 1 if std is None else std.dimshuffle(pattern)
-        inv_std = 1 if inv_std is None else inv_std.dimshuffle(pattern)
+        inv_std = inv_std.dimshuffle(pattern)
 
         # normalize
-        # normalized = (input - mean) * (gamma * inv_std / std) + beta
-        # where either inv_std == 1 or std == 1, depending on which one is used
-        normalized = T.nnet.batch_normalization(input, gamma=gamma * inv_std,
-                                                beta=beta, mean=mean, std=std,
-                                                mode=self.mode)
+        normalized = (input - mean) * (gamma * inv_std) + beta
         return normalized
 
 
@@ -343,8 +332,7 @@ def batch_norm(layer, **kwargs):
         irreversibly modified as specified above
     **kwargs
         Any additional keyword arguments are passed on to the
-        :class:`BatchNormLayer` constructor. Especially note the `mode`
-        argument, which controls a memory usage to performance tradeoff.
+        :class:`BatchNormLayer` constructor.
 
     Returns
     -------
--- a/lasagne/layers/normalization.py
+++ b/lasagne/layers/normalization.py
@@ -122,7 +122,7 @@ class BatchNormLayer(Layer):
     lasagne.layers.BatchNormLayer(incoming, axes='auto', epsilon=1e-4,
     alpha=0.1, mode='low_mem',
     beta=lasagne.init.Constant(0), gamma=lasagne.init.Constant(1),
-    mean=lasagne.init.Constant(0), var=lasagne.init.Constant(1), **kwargs)
+    mean=lasagne.init.Constant(0), inv_std=lasagne.init.Constant(1), **kwargs)
 
     Batch Normalization
 
@@ -139,7 +139,9 @@ class BatchNormLayer(Layer):
     mean and variance of the current input mini-batch :math:`x`, and during
     testing, they are replaced with average statistics over the training
     data. Consequently, this layer has four stored parameters: :math:`\\beta`,
-    :math:`\\gamma`, and the averages :math:`\\mu` and :math:`\\sigma^2`.
+    :math:`\\gamma`, and the averages :math:`\\mu` and :math:`\\sigma^2`
+    (nota bene: instead of :math:`\\sigma^2`, the layer actually stores
+    :math:`1 / \\sqrt{\\sigma^2 + \\epsilon}`, for compatibility to cuDNN).
     By default, this layer learns the average statistics as exponential moving
     averages computed during training, so it can be plugged into an existing
     network without any changes of the training procedure (see Notes).
@@ -179,9 +181,10 @@ class BatchNormLayer(Layer):
         Initial value, expression or initializer for :math:`\\mu`. Must match
         the incoming shape, skipping all axes in `axes`.
         See :func:`lasagne.utils.create_param` for more information.
-    var : Theano shared variable, expression, numpy array, or callable
-        Initial value, expression or initializer for :math:`\\sigma^2`. Must
-        match the incoming shape, skipping all axes in `axes`.
+    inv_std : Theano shared variable, expression, numpy array, or callable
+        Initial value, expression or initializer for :math:`1 / \\sqrt{
+        \\sigma^2 + \\epsilon}`. Must match the incoming shape, skipping all
+        axes in `axes`.
         See :func:`lasagne.utils.create_param` for more information.
     **kwargs
         Any additional keyword arguments are passed to the :class:`Layer`
@@ -232,7 +235,7 @@ class BatchNormLayer(Layer):
     """
     def __init__(self, incoming, axes='auto', epsilon=1e-4, alpha=0.1,
                  mode='low_mem', beta=init.Constant(0), gamma=init.Constant(1),
-                 mean=init.Constant(0), var=init.Constant(1), **kwargs):
+                 mean=init.Constant(0), inv_std=init.Constant(1), **kwargs):
         super(BatchNormLayer, self).__init__(incoming, **kwargs)
 
         if axes == 'auto':
@@ -264,22 +267,24 @@ class BatchNormLayer(Layer):
                                         trainable=True, regularizable=True)
         self.mean = self.add_param(mean, shape, 'mean',
                                    trainable=False, regularizable=False)
-        self.var = self.add_param(var, shape, 'var',
-                                  trainable=False, regularizable=False)
+        self.inv_std = self.add_param(inv_std, shape, 'inv_std',
+                                      trainable=False, regularizable=False)
 
     def get_output_for(self, input, deterministic=False, **kwargs):
         input_mean = input.mean(self.axes)
-        input_var = input.var(self.axes)
+        input_std = T.sqrt(input.var(self.axes) + self.epsilon)
 
         # Decide whether to use the stored averages or mini-batch statistics
         use_averages = kwargs.get('batch_norm_use_averages',
                                   deterministic)
         if use_averages:
             mean = self.mean
-            var = self.var
+            std = None
+            inv_std = self.inv_std
         else:
             mean = input_mean
-            var = input_var
+            std = input_std
+            inv_std = None
 
         # Decide whether to update the stored averages
         update_averages = kwargs.get('batch_norm_update_averages',
@@ -288,17 +293,17 @@ class BatchNormLayer(Layer):
             # Trick: To update the stored statistics, we create memory-aliased
             # clones of the stored statistics:
             running_mean = theano.clone(self.mean, share_inputs=False)
-            running_var = theano.clone(self.var, share_inputs=False)
+            running_inv_std = theano.clone(self.inv_std, share_inputs=False)
             # set a default update for them:
             running_mean.default_update = ((1 - self.alpha) * running_mean +
                                            self.alpha * input_mean)
-            running_var.default_update = ((1 - self.alpha) * running_var +
-                                          self.alpha * input_var)
+            running_inv_std.default_update = ((1 - self.alpha) *
+                                              running_inv_std +
+                                              self.alpha / input_std)
             # and make sure they end up in the graph without participating in
             # the computation (this way their default_update will be collected
             # and applied, but the computation will be optimized away):
-            mean += 0 * running_mean
-            var += 0 * running_var
+            mean += 0 * running_mean + 0 * running_inv_std
 
         # prepare dimshuffle pattern inserting broadcastable axes as needed
         param_axes = iter(range(input.ndim - len(self.axes)))
@@ -310,12 +315,14 @@ class BatchNormLayer(Layer):
         beta = 0 if self.beta is None else self.beta.dimshuffle(pattern)
         gamma = 1 if self.gamma is None else self.gamma.dimshuffle(pattern)
         mean = mean.dimshuffle(pattern)
-        std = T.sqrt(var + self.epsilon).dimshuffle(pattern)
+        std = 1 if std is None else std.dimshuffle(pattern)
+        inv_std = 1 if inv_std is None else inv_std.dimshuffle(pattern)
 
         # normalize
-        # normalized = (input - mean) * (gamma / std) + beta
-        normalized = T.nnet.batch_normalization(input, gamma=gamma, beta=beta,
-                                                mean=mean, std=std,
+        # normalized = (input - mean) * (gamma * inv_std / std) + beta
+        # where either inv_std == 1 or std == 1, depending on which one is used
+        normalized = T.nnet.batch_normalization(input, gamma=gamma * inv_std,
+                                                beta=beta, mean=mean, std=std,
                                                 mode=self.mode)
         return normalized
 
--- a/lasagne/layers/normalization.py
+++ b/lasagne/layers/normalization.py
@@ -59,30 +59,35 @@ class LocalResponseNormalization2DLayer(Layer):
     Aggregation is purely across channels, not within channels,
     and performed "pixelwise".
 
-    Input order is assumed to be `BC01`.
-
-    If the value of the ith channel is :math:`x_i`, the output is
+    If the value of the :math:`i` th channel is :math:`x_i`, the output is
 
     .. math::
-
-        x_i = \frac{x_i}{ (k + ( \alpha \sum_j x_j^2 ))^\beta }
+        x_i = \\frac{x_i}{ (k + ( \\alpha \\sum_j x_j^2 ))^\\beta }
 
     where the summation is performed over this position on :math:`n`
     neighboring channels.
 
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape. Must
+        follow *BC01* layout, i.e., ``(batchsize, channels, rows, columns)``.
+    alpha : float scalar
+        coefficient, see equation above
+    k : float scalar
+        offset, see equation above
+    beta : float scalar
+        exponent, see equation above
+    n : int
+        number of adjacent channels to normalize over, must be odd
+
+    Notes
+    -----
     This code is adapted from pylearn2. See the module docstring for license
     information.
     """
 
     def __init__(self, incoming, alpha=1e-4, k=2, beta=0.75, n=5, **kwargs):
-        """
-        :parameters:
-            - incoming: input layer or shape
-            - alpha: see equation above
-            - k: see equation above
-            - beta: see equation above
-            - n: number of adjacent channels to normalize over.
-        """
         super(LocalResponseNormalization2DLayer, self).__init__(incoming,
                                                                 **kwargs)
         self.alpha = alpha
--- a/lasagne/layers/normalization.py
+++ b/lasagne/layers/normalization.py
@@ -1,9 +1,10 @@
 # -*- coding: utf-8 -*-
 
 """
-
-This file contains code from pylearn2, which is covered by the following
-license:
+The :class:`LocalResponseNormalization2DLayer
+<lasagne.layers.LocalResponseNormalization2DLayer>` implementation contains
+code from `pylearn2 <http://github.com/lisa-lab/pylearn2>`_, which is covered
+by the following license:
 
 
 Copyright (c) 2011--2014, Université de Montréal
@@ -35,12 +36,19 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 """
 
+import theano
 import theano.tensor as T
+
+from .. import init
+from .. import nonlinearities
 
 from .base import Layer
 
+
 __all__ = [
     "LocalResponseNormalization2DLayer",
+    "BatchNormLayer",
+    "batch_norm",
 ]
 
 
@@ -62,7 +70,8 @@ class LocalResponseNormalization2DLayer(Layer):
     where the summation is performed over this position on :math:`n`
     neighboring channels.
 
-    This code is adapted from pylearn2.
+    This code is adapted from pylearn2. See the module docstring for license
+    information.
     """
 
     def __init__(self, incoming, alpha=1e-4, k=2, beta=0.75, n=5, **kwargs):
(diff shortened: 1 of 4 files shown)


skaae:
Very cool.

Botev:
Someone was talking in a previous conversation about after training to have the option to collect for inference the mean/std on minibatches from the full set. Would that be an option?

benanne:
Cool! I believe Theano recently added a batch norm op (see https://github.com/Theano/Theano/blob/a60bf55d7f5259d1c78c93d4d7b129c1d237860b/theano/tensor/nnet/bn.py ), shouldn't we consider using that instead of reimplementing it? Or do we need more flexibility?

f0k:
> I believe Theano recently added a batch norm op [...], shouldn't we consider using that instead of reimplementing it?

That's what I do: https://github.com/f0k/Lasagne/commit/6aabb4a5bb3276d94fe54d7efe21c4785b30476d#diff-eb58ddcf5382bc22630fc8e034702878R228
It just replaces the `(inputs - mean) / std * gamma + beta` by a composite Op, which saves memory at the cost of performance. At my intervention it can be disabled by passing `mode="high_mem"`, though, and then it's the same as not using it.

> Someone was talking in a previous conversation about after training to have the option to collect for inference the mean/std on minibatches from the full set. Would that be an option?

To do it as in the paper, you need to collect those statistics over the training set one layer at a time, that's something you'll need to do outside the layer (and you may even want to use double precision accumulators for that). We can document that in a Recipe, but I don't think the layer can help with that.

We could discuss adding some more keyword arguments to `get_output_for()`, though. Right now `deterministic=False` uses the minibatch-dependent statistics and updates the running means, and `deterministic=True` uses the running means. Actually, the former is not really indeterministic, it just happens to match with how we usually obtain the training and test predictions. We could have some extra keyword arguments like `batch_norm_update_averages` and `batch_norm_use_averages` to override the default behaviour.

Two other thoughts:

1. In #429, we thought about allowing users to pass an `updates` dictionary to `get_output`/`get_output_for` where layers can register additional updates. That's needed for recurrent layers involving nondeterminism, and we could use it for the `BatchNormLayer` as well, instead of the `default_updates` trick. That's something we can change later, though (as long as we change it before releasing Lasagne 0.2).
2. After training, it's possible to merge mean/std into gamma/beta, to make evaluation a little faster. Shall we add some support for that or leave it up to the user? Actually, gamma/beta can often even be merged into the weights of the previous layer...

f0k:
> To do it as in the paper, you need to collect those statistics over the training set one layer at a time, that's something you'll need to do outside the layer (and you may even want to use double precision accumulators for that). We can document that in a Recipe, but I don't think the layer can help with that.

We could provide a way to generate updates for the statistics that compute a normal average (using Welford's algorithm) instead of an exponential moving average. That wouldn't be useful as a `get_output_for()` keyword argument, though, as that would affect all batch normalization layers in a network at once. So if anything, it should probably become a helper function that takes a single BatchNormLayer and returns an appropriate update dictionary.

> We could discuss adding some more keyword arguments to `get_output_for()`, though.

Did that. Updating and using the exponential moving averages can now be enabled and disabled independently. Not sure how useful this is, but at least it allows to disable maintaining the exponential moving averages if you're not going to use them for some reason, and it feels a bit cleaner.

> After training, it's possible to merge mean/std into gamma/beta, to make evaluation a little faster. Shall we add some support for that or leave it up to the user? Actually, gamma/beta can often even be merged into the weights of the previous layer...

Incidentally, somebody implemented exactly that for Torch: https://github.com/e-lab/torch-toolbox/blob/master/BN-absorber/BN-absorber.lua. A helper function merging all batch normalization layers of a network into their predecessors would be fun.

lasagne/layers/normalization.py
```diff
@@ -101,3 +109,234 @@ def get_output_for(self, input, **kwargs):
...
+    validation. This can be achieved by passing ``deterministic=False``.
+    For validation, [1]_ normalize each input mini-batch by the stored
+    statistics. This can be achieved by passing ``deterministic=True``.
+
+    For more fine-grained control, ``batch_norm_update_averages`` can be passed
+    to update the exponential moving averages (``True``) or not (``False``),
+    and ``batch_norm_use_averages`` can be passed to use the exponential moving
+    averages for normalization (``True``) or normalize each mini-batch by its
+    own statistics (``False``). These settings override ``deterministic``.
+
```
skaae: Is the usecase for `batch_norm_update_averages` if one want to collec the correct statistic in a layer wise fashion? Then disabling collection of averages may give a little speed up?

f0k: If you want to collect the statistics for the current layer, you will only compute its input (with `deterministic=True`, which sets `batch_norm_update_averages=False` for the previous layers), but this was possible before. What's possible with the split is (a) update *and* use the exponential averages (didn't work well when I tried, though), (b) neither use nor update the exponential averages (if you don't plan to do any validation in between, then you don't need the exponential averages), (c) continue training a network with dropouts and everything while using the exponential averages and holding them fixed (to slightly fine-tune a network afterwards, to let the network adapt to the switch from dynamic training statistics to fixed running averages). Not sure if anything of this is useful, but it's possible.

DediGadot:
I'm getting the following error while trying to use the new code:


  File "/home/dedigadot/Envs/env3/local/lib/python2.7/site-packages/Lasagne-0.2.dev1-py2.7.egg/lasagne/layers/normalization.py", line 179, in __init__
    if axis not in self.shared_axes]

AttributeError: 'BatchNormLayer' object has no attribute 'shared_axes'

lasagne/layers/normalization.py
DediGadot: I am getting the following error while trying to use the code:

  File "/home/dedigadot/Envs/env3/local/lib/python2.7/site-packages/Lasagne-0.2.dev1-py2.7.egg/lasagne/layers/normalization.py", line 179, in __init__
    if axis not in self.shared_axes]

AttributeError: 'BatchNormLayer' object has no attribute 'shared_axes'


And also, in lines:
        beta = beta.dimshuffle(pattern)
        gamma = gamma.dimshuffle(pattern)

Souldn't they be:
        self.beta = beta.dimshuffle(pattern)
        self.gamma = gamma.dimshuffle(pattern)

f0k: > I am getting the following error while trying to use the code:

Sorry, this is still completely untested, I just took [my old implementation](https://gist.github.com/f0k/f1a6bd3c8585c400c190) and adapted it. Fixed that bug now.

> Souldn't they be:
> self.beta = beta.dimshuffle(pattern)
> self.gamma = gamma.dimshuffle(pattern)

No, they should be:
```python
beta = self.beta.dimshuffle(pattern)
gamma = self.gamma.dimshuffle(pattern)
```
Fixed that as well. Thanks for reporting!

DediGadot: Great, thank you. Sorry for the mistake in specifying the second issue.

DediGadot: Also:
line 222: axes=None --> axes='auto'
add 'import theano' at the begining of the file
line 338: remove 'lasagne.'

lasagne/layers/normalization.py
```diff
@@ -101,3 +109,234 @@ def get_output_for(self, input, **kwargs):
...
+        # normalize
+        # normalized = (input - mean) * (gamma / std) + beta
+        normalized = T.nnet.batch_normalization(input, gamma=gamma, beta=beta,
+                                                mean=mean, std=std,
+                                                mode=self.mode)
+        return self.nonlinearity(normalized)
+
+
+def batch_norm(layer, mode='low_mem'):
+    """
```
Botev: Maybe allow for `**kwargs` to be passed down to the `BatchNormLayer` from the function?

lasagne/layers/normalization.py
f0k: Addressed! Sorry for the delay, I'm at a conference and had to prepare for that. Possible remaining bugs will be found and addressed when I resume work and write the tests :)

lasagne/layers/normalization.py
f0k: Good idea, did that instead of only allowing `mode` to be passed on.

f0k:
Ready to review!

f0k:
Rebased to include the updated `requirements.txt`, the former Theano version was too old. Travis should pass now, ready to review!

f0k:
Anyone willing/capable to have a look? Cheers!

skaae:
yes. I think I can do it tonight.

lasagne/layers/normalization.py
```diff
@@ -101,3 +115,233 @@ def get_output_for(self, input, **kwargs):
...
+
+    See also
+    --------
+    batch_norm : Convenience function to apply batch normalization to a layer
+
+    References
+    ----------
+    .. [1]: Ioffe, Sergey and Szegedy, Christian (2015):
+            Batch Normalization: Accelerating Deep Network Training by Reducing
+            Internal Covariate Shift. http://arxiv.org/abs/1502.03167.
```
skaae: I think the documentation is well written. 
Can you add an example? 
 When I have helped people with Lasagne they often forget `trainable=True` when they collect the parameter. Normally that is not a problem but for batchnormalization it is. I think we could avoid some questions with a small example.

lasagne/layers/normalization.py
```diff
@@ -101,3 +115,233 @@ def get_output_for(self, input, **kwargs):
             scale += self.alpha * input_sqr[:, i:i+ch, :, :]
         scale = scale ** self.beta
         return input / scale
+
+
+class BatchNormLayer(Layer):
+    """
+    lasagne.layers.BatchNormLayer(incoming, axes='auto', epsilon=1e-4,
+    alpha=0.5, nonlinearity=None, mode='low_mem',
```
skaae: `alpha=0.5` seems a bit high to me? I checked Torch and they use 0.1

lasagne/layers/normalization.py
```diff
@@ -101,3 +115,233 @@ def get_output_for(self, input, **kwargs):
...
+        standard deviations computed during training; the closer to one, the
+        more it will depend on the last batches seen
+    nonlinearity : callable or None
+        The nonlinearity that is applied to the layer activations. If ``None``
+        is provided, the layer will be linear (this is the default).
+    mode : {'low_mem', 'high_mem'}
+        Specify which batch normalization implementation to use: ``'low_mem'``
+        avoids storing intermediate representations and thus requires less
+        memory, while ``'high_mem'`` can reuse representations for the backward
+        pass and is thus 5-10% faster.
```
skaae: Don't you need escape percent?

lasagne/layers/normalization.py
```diff
@@ -101,3 +115,233 @@ def get_output_for(self, input, **kwargs):
...
+        shape = [size for axis, size in enumerate(self.input_shape)
+                 if axis not in self.axes]
+        if any(size is None for size in shape):
+            raise ValueError("BatchNormLayer needs specified input sizes for "
+                             "all axes not normalized over.")
+        self.beta = self.add_param(beta, shape, 'beta',
+                                   trainable=True, regularizable=True)
+        self.gamma = self.add_param(gamma, shape, 'gamma',
+                                    trainable=True, regularizable=False)
+        self.mean = self.add_param(mean, shape, 'mean',
```
skaae: I think regularizable should be False for beta and True for gamma?

skaae:
I looks very good! 

two minor suggestions: 

1. I would change the default value for alpha to something smaller 
2. I think beta should have regularizeable=False and gamma should have regularizeable=True

lasagne/layers/normalization.py
benanne: This is a bit of a wart really, considering how common batchnorm is becoming. Maybe we should think about how we can make this more intuitive. Perhaps `get_params()` should not be returning these batchnorm parameters by default.

skaae: Cant we make trainable a required param? Same goes for deterministic? I tend to forget them

benanne: No, because in many cases specifying them is not necessary, and this would add unnecessary cognitive overhead to those cases.

As an example, in many nets, all parameters are `trainable`, so then it's nice to not have to worry about this distinction. I think this is where the "restraint" part of Lasagne's design philosophy applies :) (see http://lasagne.readthedocs.org/en/latest/user/development.html )

The same argument can be made for nets without dropout layers.

skaae: I think its a matter of taste :) I think i would prefer to get an error if trainable is not specified rather than having to figure out some unspecified batchnorm error.

benanne: I agree that we should handle this better, but making the specification of `trainable` a requirement is not the right solution. It goes directly against the design principles of the library. It will add complexity (however minor, but these things always creep up on you) for use cases where the user should be able to be oblivious to the existence of parameter tags.

You could argue that these principles are a matter of taste, which is true, but I think that's beside the point.

benanne: Having thought about this a bit more, perhaps we could do something similar to Theano's `warn_float64` configuration option: http://deeplearning.net/software/theano/library/config.html#config.warn_float64

i.e. allow users to set a flag so that a warning is raised when `get_all_params` is called without `trainable` set, and even for `get_output` without `deterministic` set.

I think this would have to be off by default though, otherwise we are again confronting people with library features that they might not be using / be interested in, which we should strive to avoid where possible.

The problem is then that there is no such thing as a Lasagne configuration file (and there probably shouldn't be?), so then it's a question of setting these flags programmatically at the top of your script, or something... which could easily be forgotten. That brings us back to our original problem. So yeah, this is tricky! 

Overall, I like the idea of using opt-in warnings to shield against this type of mistakes.

skaae: Isn't it possible to have `get_params` default to trainable=True. That is what you want most cases. I think that is also what the users expect.

lasagne/layers/normalization.py
f0k: Hmm, good question. When I did my first experiments, there was no other implementation yet. It would probably be good to use similar values as other libraries just in case people switch between them.

* Torch has 0.1: https://github.com/torch/nn/blob/eb13695/BatchNormalization.lua#L46
* Caffe has 0.01: https://github.com/BVLC/caffe/blob/39f69fbe520d90d30fc342e5d69153ec9534d2bc/src/caffe/proto/caffe.proto#L476
* TensorFlow: Can't find anything about running mean/variance: https://tensorflow.googlesource.com/tensorflow/+/master/tensorflow/core/kernels/batch_norm_op.cc
* Keras has 0.9, which means 0.1: https://github.com/fchollet/keras/blob/e839a4b/keras/layers/normalization.py#L19
* MXNet has 0.9, which probably means 0.1: http://mxnet.readthedocs.org/en/latest/python/symbol.html#mxnet.symbol.BatchNorm

I'll change it to 0.1 then.

We *could* also rename it to `momentum` and invert the meaning, defaulting it to 0.9, but I think that's a bit misleading: In `lasagne.updates.momentum`, velocities are updated as `x := momentum * x + new_x`, while here the formula would be `x := momentum * x + (1 - momentum) * new_x` (i.e., the new value is properly scaled here while it's not scaled in `lasagne.updates`).

lasagne/layers/normalization.py
f0k: No, that would be in LaTeX :) This renders fine.

lasagne/layers/normalization.py
f0k: Oh yes, good catch. I'll change that.

lasagne/layers/normalization.py
```diff
@@ -101,3 +115,233 @@ def get_output_for(self, input, **kwargs):
...
+    ----------
+    .. [1]: Ioffe, Sergey and Szegedy, Christian (2015):
+            Batch Normalization: Accelerating Deep Network Training by Reducing
+            Internal Covariate Shift. http://arxiv.org/abs/1502.03167.
+    """
+    def __init__(self, incoming, axes='auto', epsilon=1e-4, alpha=0.5,
+                 nonlinearity=None, mode='low_mem',
+                 beta=init.Constant(0), gamma=init.Constant(1),
+                 mean=init.Constant(0), var=init.Constant(1), **kwargs):
+        super(BatchNormLayer, self).__init__(incoming, **kwargs)
```
f0k: I don't think so. I'm not actually sure whether it should have a `nonlinearity` at all.

In hindsight, I think the best solution would have been not to have a `nonlinearity` keyword in any of the layers, and make the nonlinearities Layer classes as well. That would make it much more straightforward to support parameterizable nonlinearities, and to insert batch normalization between a linear transformation and its nonlinearity. It's too late for that, though.

I added a `nonlinearity` keyword here because usually, you will need a nonlinearity right after the batch normalization, but maybe it should be removed completely? I'd find it confusing to have the `BatchNormLayer` apply a linear rectifier by default, as that's not really the purpose of batch normalization.

lasagne/layers/normalization.py
f0k: > Isn't it possible to have `get_params` default to trainable=True. That is what you want most cases.

But then how do you save the model parameters into a file? `get_all_param_values()` calls `get_all_params()`, so it would be affected by that. If you make `trainable=True` a default argument, there's no way to get *all* parameters at once any more.

What happens if you forget `trainable=True` for collecting the training parameters? Is there a way for us to ensure it will spit out a somewhat decipherable error message later?

f0k:
Thanks for reviewing so far! Anything else to be changed?

benanne:
Will try to have a look this weekend, but don't wait for me if you think this is good to merge.

skaae:
@benanne: Maybe you can look at https://github.com/Lasagne/Lasagne/pull/498 instead?
This is a suggestion :)

benanne:
I've been keeping a list this past week and I have 4 PRs to look at over the weekend, including those two :)

skaae:
ouch. Which ones? I can probably help with some of them, but I don't feel qualified to review conv3d.

benanne:
Actually the other two seem to have been merged in the meantime, so that's good :) Just these two then. Hopefully tomorrow!

lasagne/layers/normalization.py
benanne: We should spend some time thinking about this, I think it would be valuable. But I guess it will be tricky since the tags are not part of the parameters themselves, but rather the layers that contain them.

benanne:
> We could discuss adding some more keyword arguments to get_output_for(), though. Right now deterministic=False uses the minibatch-dependent statistics and updates the running means, and deterministic=True uses the running means. Actually, the former is not really indeterministic, it just happens to match with how we usually obtain the training and test predictions. We could have some extra keyword arguments like batch_norm_update_averages and batch_norm_use_averages to override the default behaviour.

I think the most pragmatic approach here is to overload `deterministic` here, even if it's not 100% correct. I've always interpreted it as "disable any sources of stochasticity in the network", and I guess batchnorm qualifies as a source of stochasticity when you randomly sample the minibatch that a given example happens to be part of.

The main downside to this is that it makes it harder to selectively disable only dropout or only batchnorm. But I'm not sure if that's a use case we should consider. In that case I guess you should just rebuild the network with/without the specified layers.

> In #429, we thought about allowing users to pass an updates dictionary to get_output/get_output_for where layers can register additional updates. That's needed for recurrent layers involving nondeterminism, and we could use it for the BatchNormLayer as well, instead of the default_updates trick. That's something we can change later, though (as long as we change it before releasing Lasagne 0.2).

I haven't really decided what I think of this. It's maybe kind of odd to have to pass in an object to a method that is then modified in-place, since we don't really do this anywhere else. But it is probably the cleanest way to implement this. `default_updates` is a bit magical and could potentially lead to hard-to-track-down bugs. Explicit is better than implicit.

> After training, it's possible to merge mean/std into gamma/beta, to make evaluation a little faster. Shall we add some support for that or leave it up to the user? Actually, gamma/beta can often even be merged into the weights of the previous layer...

I wouldn't prioritize this right now. More of a "nice to have" :) I don't think Lasagne is typically used in use cases where inference speed is critical, and I imagine it doesn't make a huge difference. Does it?

> Not sure how useful this is, but at least it allows to disable maintaining the exponential moving averages if you're not going to use them for some reason

That's valuable I guess.

> A helper function merging all batch normalization layers of a network into their predecessors would be fun.

It would! I've also been thinking of a helper function that merges all model parameters into a single vector (like Torch's getParameters()). Could perhaps be useful when used in combination with scipy.optimize, or other "external" optimizers.

> In hindsight, I think the best solution would have been not to have a nonlinearity keyword in any of the layers, and make the nonlinearities Layer classes as well. That would make it much more straightforward to support parameterizable nonlinearities, and to insert batch normalization between a linear transformation and its nonlinearity. It's too late for that, though.

I think we've had this discussion before, but the main motivation for this is to have the concept of a `Layer` in Lasagne match the concept of a layer as it is understood by most neural network practitioners as closely as possible. That requires making the nonlinearity part of certain `Layer` classes.

> I added a nonlinearity keyword here because usually, you will need a nonlinearity right after the batch normalization, but maybe it should be removed completely? I'd find it confusing to have the BatchNormLayer apply a linear rectifier by default, as that's not really the purpose of batch normalization.

Yeah, I think it's okay to remove it here.

lasagne/layers/normalization.py
```diff
@@ -101,3 +115,233 @@ def get_output_for(self, input, **kwargs):
...
+    mean=lasagne.init.Constant(0), var=lasagne.init.Constant(1), **kwargs)
+
+    Batch Normalization
+
+    This layer implements batch normalization of its inputs, following [1]_:
+
+    .. math::
+        y = \\frac{x - \\mu}{\\sqrt{\\sigma^2 + \\epsilon}} \\gamma + \\beta
+
+    That is, the input is normalized to zero mean and unit variance, and then
```
benanne: Perhaps this is a bit confusing: the assumption that this normalization happens across the batch dimension and the spatial dimensions is implicit, but maybe it should be mentioned explicitly here (and not just in the argument list). It is also possible to do ZMUV normalization spatially only, this is sometimes referred to as global contrast normalization.

lasagne/layers/normalization.py
```diff
@@ -101,3 +115,233 @@ def get_output_for(self, input, **kwargs):
...
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
+    axes : 'auto', int or tuple of int
+        The axis or axes to normalize over. If ``'auto'`` (the default),
+        normalize over all axes except for the second: this will normalize over
+        the minibatch dimension for dense layers, and additionally over all
+        spatial dimensions for convolutional layers.
+    epsilon : scalar
+        Small constant :math:`\\epsilon` added to the variance before taking
+        the square root and dividing by it, to avoid numeric problems
```
benanne: should be "numerical", I think

f0k:
> I think the most pragmatic approach here is to overload `deterministic` here, even if it's not 100% correct.

I did that, but allowed behaviour to be overridden with `batch_norm_update_averages` and `batch_norm_use_averages` (I'd be fine with shorter names if you can come up with anything).

> It's maybe kind of odd to have to pass in an object to a method that is then modified in-place, since we don't really do this anywhere else. But it is probably the cleanest way to implement this.

It's got to be part of `get_output_for()`, and since we can't easily change the API to have it return an update dictionary along with the output, it's the easiest way to have a modifiable parameter. I agree that it's probably odd if you're not used to it.

> default_updates is a bit magical and could potentially lead to hard-to-track-down bugs. Explicit is better than implicit.

So should we change this to use an `updates` argument right from the start? We'd be free to change this any time later before releasing 0.2, but it might be good to come up with a final solution right away.

> I think we've had this discussion before, [...]

I know, and I agreed with your argument back then. At that point there were no parameterized nonlinearities or batch normalization, though, and now I'd have done things differently.

> Yeah, I think it's okay to remove it here.

Actually, when I wrote the `BatchNormLayer` back in March, there was no `NonlinearityLayer` in Lasagne -- I guess that was the only reason I gave it a `nonlinearity` keyword argument. I'll definitely remove it!

lasagne/layers/normalization.py
```diff
@@ -101,3 +115,233 @@ def get_output_for(self, input, **kwargs):
...
+            running_var = theano.clone(self.var, share_inputs=False)
+            # set a default update for them:
+            running_mean.default_update = ((1 - self.alpha) * running_mean +
+                                           self.alpha * input_mean)
+            running_var.default_update = ((1 - self.alpha) * running_var +
+                                          self.alpha * input_var)
+            # and make sure they end up in the graph without participating in
+            # the computation (this way their default_update will be collected
+            # and applied, but the computation will be optimized away):
+            mean += 0 * running_mean
```
benanne: This is... interesting :) But it's pretty cool that this works. Have you verified in any way that the computation is indeed optimized away? It's not always clear which optimizations are applied and which aren't...

lasagne/layers/normalization.py
f0k: Ah, right. I'll add a sentence or two explaining what it's normalizing over.

> It is also possible to do ZMUV normalization spatially only, this is sometimes referred to as global contrast normalization.

Hmm, this reminds me that when you're not normalizing over the batch dimension, you don't need running averages. We might want to explain this in the `Notes` section.

lasagne/layers/normalization.py
f0k: Right, thanks.

lasagne/tests/layers/test_normalization.py
```diff
@@ -140,3 +142,117 @@ def test_normalization(self, input_data, input_layer, layer):
...
+    def test_undefined_shape(self, BatchNormLayer):
+        # should work:
+        BatchNormLayer((64, None, 3), axes=(1, 2))
+        # should not work:
+        with pytest.raises(ValueError) as exc:
+            BatchNormLayer((64, None, 3), axes=(0, 2))
+        assert 'needs specified input sizes' in exc.value.args[0]
+
+
+def test_batch_norm_makro():
```
benanne: macro should be spelt with c I think

benanne:
> I did that, but allowed behaviour to be overridden with batch_norm_update_averages and batch_norm_use_averages (I'd be fine with shorter names if you can come up with anything).

This is great :)

> So should we change this to use an updates argument right from the start? We'd be free to change this any time later before releasing 0.2, but it might be good to come up with a final solution right away.

I'm okay with leaving this as-is for now, but it's up to you.

lasagne/layers/normalization.py
f0k: > Have you verified in any way that the computation is indeed optimized away?

Not exactly for this, but I did a test with a manually-constructed graph that included a `+ 0*x` where `x` was a shared variable with default updates, and it worked as intended (a `theano.printing.debugprint` of the graph included the updates, but did not include the `0*x`).

lasagne/tests/layers/test_normalization.py
f0k: Oh, right. Seems that was German.

lasagne/layers/normalization.py
benanne: okay, fair enough :)

lasagne/layers/normalization.py
```diff
@@ -101,3 +115,233 @@ def get_output_for(self, input, **kwargs):
...
+        gamma = self.gamma.dimshuffle(pattern)
+        mean = mean.dimshuffle(pattern)
+        std = T.sqrt(var + self.epsilon)
+        std = std.dimshuffle(pattern)
+
+        # normalize
+        # normalized = (input - mean) * (gamma / std) + beta
+        normalized = T.nnet.batch_normalization(input, gamma=gamma, beta=beta,
+                                                mean=mean, std=std,
+                                                mode=self.mode)
```
jonhoo: Shouldn't this be `T.nnet.bn.batch_normalization` (note the `.bn.`)? At least that's how it's mentioned in the [Theano docs](http://deeplearning.net/software/theano/library/tensor/nnet/bn.html).

jonhoo:
The [Theano docs](http://deeplearning.net/software/theano/library/tensor/nnet/bn.html) for `batch_normalization` claim that this operation should also run on the GPU. However, I'm not seeing this on my machine. The network takes much longer to train, and the GPU seems to be idle much of the time. Is this expected behavior?

lasagne/layers/normalization.py
f0k: It's imported into the `theano.tensor.nnet` namespace, though, and I think this is where it's intended to be used from. I don't know why they chose to have it documented as `theano.tensor.nnet.bn.batch_normalization`, maybe this wasn't even on purpose. Thanks for checking, though!

f0k:
> The network takes much longer to train, and the GPU seems to be idle much of the time. Is this expected behavior?

That the network takes longer to train -- yes. That the GPU is idle much of the time -- no. Can you do a profile run? (`THEANO_FLAGS=device=gpu,floatX=float32,profile=1 CUDA_LAUNCH_BLOCKING=1 python yourscript.py`) Are there any suspicious CPU Ops that aren't there with batch normalization disabled?

jonhoo:
Ah, seems I was just being silly. Never mind!

f0k:
All right, finally got around addressing @benanne's and @skaae's comments. Also made `beta` and `gamma` optional because Alec Radford et al. reported that this setup is useful for something (Appendix of their [DCGAN paper](http://arxiv.org/abs/1511.06434)).

I've added this as an additional commit so you can review the changeset separately. When you're happy, I'll squash it and let Travis run (tests pass locally and the docs render fine).

benanne:
ok, I'm at NIPS now so I might not be able to take a good look at it until next week. Sorry about that. But if someone else wants to wrap this up that's fine by me too.

skaae:
I think the batch normalization layer might need to support a masking like the recurrent layers. 

Say that the batchnormalizaton layer sits after some recurrent layer wich uses a mask, then the collected statistics will be wrong because the last hidden state is repeated in the recurrent layer. 

I think the running mean update rules could be changed to
```Python
if not masked:
            running_mean.default_update = ((1 - self.alpha) * running_mean +
                                           self.alpha * input_mean)
            running_var.default_update = ((1 - self.alpha) * running_var +
                                          self.alpha * input_var)
else:   # dont change running stats when masked
            running_mean.default_update = (
                       mask*((1 - self.alpha) * running_mean + self.alpha * input_mean)  + 
                       (1-mask)*running_mean)
            running_var.default_update = (
                       mask*((1 - self.alpha) * running_var + self.alpha * input_var) + 
                       (1-mask)*running_var )
```

kshmelkov:
@skaae Do you think it is really relevant now? Taking into account that there is no consensus on how to apply BN to recurrent layers: http://arxiv.org/abs/1510.01378 I would suggest to introduce it in another PR.

f0k:
> Do you think it is really relevant now? Taking into account that there is no consensus on how to apply BN to recurrent layers: http://arxiv.org/abs/1510.01378

There's pretty good evidence on how to do it (in a paper from Tuesday): http://arxiv.org/abs/1512.02595 (Section 3.2)
We should definitely add masks so you can apply batch normalization on padded sequences. Also we should possibly add another shortcut string for `axes` that does sequence-wise normalization for RNNs (`axes=(0,1)`), or at least document that the current `'auto'` mode will do something terribly wrong for RNNs (because it assumes anything past the second dimension are spatial dimensions the normalization factors should be shared over).

> I would suggest to introduce it in another PR.

Yes, I'd still hold that off for another PR. It would be more important to get this one merged. I don't think we'd need to break compatibility with what we have here.

skaae:
Thats fine. When will you merge?
Den 11/12/2015 11.45 skrev "Jan Schlüter" <notifications@github.com>:

f0k:
> When will you merge?

When my latest commit has been reviewed!

skaae:
> When my latest commit has been reviewed!

Good point :) It looks good to me.

benanne:
Looks good to me as well! Why did you add [skip ci] though? It looks like there are some substantial changes to the tests. It would be good to have them run, just in case.

f0k:
> Why did you add [skip ci] though?

Because I'll be squashing before merging anyway, which will trigger Travis (didn't want to run it twice to conserve energy). As noted, I just kept the commit separately to make it easier to review the last changeset.

> Looks good to me as well!

Thanks, I'll squash then and wait for Travis to sign it off!

f0k:
Hmm, seems cuDNN v4 integrates batch normalization? https://developer.nvidia.com/cudnn

benanne:
Sweet :) Here's a Theano branch that should work with cudnn v4: https://github.com/carriepl/Theano/tree/v4

Might give it a try sometime!

f0k:
> Here's a Theano branch that should work with cudnn v4

That branch "only" supports the changes in the convolution interface: https://github.com/Theano/Theano/pull/3773

There's an important twist in the cuDNN batch normalization, by the way. Instead of computing the running average of the variance, they compute the running average of `1 / sqrt(variance + eps)`, misnamed as InvVariance:
```
/* Output in training mode, input in inference. Is the moving average
   of 1 / sqrt( epsilon + variance[x] ) */
void                               *resultRunningInvVariance,
```
We might want to change the batch normalization layer to do this as well, to make it easier to switch to the cuDNN version. That would finally give us a slight benefit for waiting so long with merging it, compared to all other libraries :)
This will have an effect on test-time inference (because our current `1 / sqrt(running_avg(variance) + eps)` is not the same as `running_avg(1 / sqrt(variance + eps))`), but probably it's neither for the worse nor for the better.

benanne:
I suppose consistency is good :)

ebenolson:
Is this ready to merge or does it still need another review?

benanne:
I guess the question is whether we want to follow the same conventions as cuDNN, which will make it easier to swap different implementations. In that case some minor changes are still required. I think it would be a good idea to do this, so people don't end up getting different results if they change implementations.

lasagne/layers/normalization.py
```diff
@@ -101,3 +115,259 @@ def get_output_for(self, input, **kwargs):
...
+    This introduces batch normalization right before its nonlinearity:
+
+    >>> from lasagne.layers import get_all_layers
+    >>> [l.__class__.__name__ for l in get_all_layers(l2)]
+    ['InputLayer', 'DenseLayer', 'BatchNormLayer', 'NonlinearityLayer']
+    """
+    nonlinearity = getattr(layer, 'nonlinearity', None)
+    if nonlinearity is not None:
+        layer.nonlinearity = nonlinearities.identity
+    if hasattr(layer, 'b'):
```
felixlaumon: If `layer` already has `b` set to None, the line below will throw an exception since `layer.params` will not include the bias. Perhaps the condition should check `layer.b == None` too?

DediGadot:
I think we should merge this as soon as possible, regardless of future cuDNN related API changes. The fact that Lasagne doesn't yet have an in-place BN implementation is a severe limitation.

benanne:
I agree. Let's change it so we follow cuDNN's conventions, but let's maybe leave actually wrapping the cuDNN implementation for a later PR.

f0k:
> Let's change it so we follow cuDNN's conventions, but let's maybe leave actually wrapping the cuDNN implementation for a later PR.

Yes, that was my plan, I just got held up by Christmas. I'll address that as soon as I've got enough time.

lasagne/layers/normalization.py
f0k: Good catch! I'll extend the `if` clause with `and layer.b is not None` and add a test.

benanne:
No worries :) Thanks for all the hard work you've been putting in!

f0k:
All right, changed the implementation to store `inv_std` instead of `var`. This makes the docstring and the code a bit more ugly. I've written the code such that it can still use `T.nnet.batch_normalization()` during training, supplying `mean`, `std`, `gamma` and `beta` as before. During testing, it also uses `T.nnet.batch_normalization()`, setting `std=1` and `gamma=gamma*inv_std`. This way it can still benefit from the memory savings in `mode='low_mem'`, and perform like the equivalent vanilla Theano code in `mode='high_mem'`. This is all done in good faith that `T.nnet.batch_normalization()` provides a benefit even if one of its argument is set to a scalar `1`.

The main benefit of the added ugliness is that we could add a `DNNBatchNormLayer` later, with identical parameterization. If we stick to the previous implementation, `DNNBatchNormLayer` would have a different parameter from `BatchNormLayer` (i.e., `inv_std` instead of `var`), but it would be possible to convert between the two. I currently don't expect that Theano will provide a generic batch normalization Op that could be used by `BatchNormLayer` to seamlessly support CPU, GPU and cuDNN -- this would be a much stronger argument for using `inv_std` instead of `var`. And I currently don't expect that computing the running average of `inv_std` is better than the running average of `var` -- again, this would be a much stronger argument for making the code more ugly.

Conclusion: I'm not very happy about the ugly code, and tend to think that different interfaces for a cuDNN- and non-cuDNN version wouldn't be so bad after all...

benanne:
Right, the "either `std` or `inv_std` is 1" stuff is a bit tricky I guess. Is the original implementation closer to what's described in the paper (or to what the authors of the paper did originally, in case they didn't mention this detail)? That would be another good argument for reverting it. At the moment I guess I don't really care either way.

f0k:
> Right, the "either std or inv_std is 1" stuff is a bit tricky I guess.

Note that this is not the only possible solution, we could also go for `inv_std` completely and not use `T.nnet.batch_normalization()` any more. I just figured we shouldn't drop this possible advantage.

> Is the original implementation closer to what's described in the paper (or to what the authors of the paper did originally, in case they didn't mention this detail)?

Yes, in the paper it seems they always work with the variance, not the inverse standard deviation. And in the initial cuDNN v4 implementation they also used the variance, but switched to the inverse standard deviation later (this was (and possibly is) still visible from some inconsistencies in the documentation of the release candidate). It's not clear why, maybe it just was to improve performance for the inference step.

f0k:
> Note that this is not the only possible solution, we could also go for `inv_std` completely and not use `T.nnet.batch_normalization()` any more. I just figured we shouldn't drop this possible advantage.

Implemented this now. Testing some variations, I found `T.nnet.batch_normalization()` is useful for the basic case only. As soon as you invert the standard deviation beforehands or set `beta=0` or `gamma=1` (constant, not trainable), it's too slow compared to the vanilla Theano implementation.
This being said, the basic case is probably still the one that's used the most. So now we have three variants:

* 8818d9c: The original batch normalization layer with `var` and `T.nnet.batch_normalization`
* 82103d4: An updated version with `inv_std` and `T.nnet.batch_normalization`
* de97c0a: An updated version with `inv_std` and no `T.nnet.batch_normalization`

I personally liked the first version best, but then we cannot become compatible with an upcoming `BatchNormDNNLayer` (if NVIDIA sticks to the interface of cuDNN v4 RC1). This would create problems down the road, making it difficult to save/load parameters between models that can and cannot use cuDNN -- we would need helper functions to convert the batch normalization parameters of a network from or to cuDNN.

benanne:
ugh, I can't believe this turned into such a mess :) I don't have a clear view anymore of the advantages and disadvantages of all the options. Are you saying that striving for cuDNN v4 compatibility is desirable after all?

f0k:
> ugh, I can't believe this turned into such a mess :) I don't have a clear view anymore of the advantages and disadvantages of all the options.

Yeah, sorry. At first it seemed like a good idea to just add in cuDNN v4 compatibility since we got to know the interface before we merging it. Let me recap the advantages/disadvantages then:

* 8818dc9: Simplest code, can make use of the reduced memory implementation in `T.nnet.batch_normalization()` at a loss of performance. Not compatible to cuDNN v4 because it stores a running average of the variance, while cuDNN stores a running average of the inverse standard deviation (that is, `1 / sqrt(var + epsilon)`).
* 82103d4: Compatible to cuDNN v4. Does some extra stretch to still use `T.nnet.batch_normalization()` during training, just like the previous version.
* de97c0a: Compatible to cuDNN v4. Drops support for `T.nnet.batch_normalization()` for cleaner code.

Assuming we add a `BatchNormDNNLayer` (which I guess we should, assuming it brings some speed benefits), the second and third option would allow simply swapping a `BatchNormLayer` for a `BatchNormDNNLayer` just like `Conv2DLayer` and `Conv2DDNNLayer` don't (due to the `flip_filters` difference). It would also allow `BatchNormDNNLayer` to be derived from `BatchNormLayer`, just overwriting `get_output_for()` and enforcing some restrictions on accepted values for the `axes` constructor parameter and accepted input dimensionalities.
In theory, the second and third option would also allow a solution in Theano that automatically uses the cuDNN implementation when possible. This would mean quite some work on the Theano side, though, and I'm not sure this is going to happen because it will be easy enough to use cuDNN manually.

If we merge 8818d9c and add a `BatchNormDNNLayer` later, there are basically two options to allow moving a cuDNN model to CPU:
1. Add an `inverted_stddev` option to `BatchNormLayer` that changes the implementation to compute and store the inverted standard deviation instead of the variance. This will probably make the code even uglier. In terms of compatibility, we'd be close to the situation of the convolutional layers (except that a `BatchNormDNNLayer` couldn't be used to emulate a `BatchNormLayer`, only vice versa).
2. Add two helper functions: `batch_norm_to_cudnn` and `batch_norm_from_cudnn` that take a network and modify the value of all shared variables holding the variance or inverse standard deviation of a batch normalization layer to hold the opposite.

> Are you saying that striving for cuDNN v4 compatibility is desirable after all?

If cuDNN indeed provides a faster implementation, then yes, I think it would be good to be compatible to this and live with the disadvantages for the non-cuDNN implementation that hardly anybody will use afterwards.

benanne:
So are we hedging our bets on the fact that the cuDNN implementation will indeed be faster? It would be a bit silly if it wasn't :)

f0k:
> So are we hedging our bets on the fact that the cuDNN implementation will indeed be faster?

Yes... should we verify that first? Aww, I shouldn't have looked at cuDNN v4!

> It would be a bit silly if it wasn't :)

Silly for us or silly for Nvidia? I think the library aims to provide all primitives required to build modern neural networks, not all of them must be faster than the Theano equivalent (e.g., the activation functions probably perform about the same).

benanne:
That would be odd, I thought the intention of cuDNN is to provide optimized implementations of these primitives. If a combination of more primitive operations is faster, then why bother adding it to cuDNN in the first place?

f0k:
> That would be odd, I thought the intention of cuDNN is to provide optimized implementations of these primitives.

Well, ultimately this is going to happen, but I don't think everything added to cuDNN will necessarily be substantially faster than, say, vanilla Theano right from the start.

> If a combination of more primitive operations is faster, then why bother adding it to cuDNN in the first place?

Because a) those simpler primitives may be difficult enough to implement efficiently (e.g., for batch normalization, you need to be able to compute the mean and variance) and b) grouping them creates the opportunity to improve the underlying implementation in future. (I wouldn't expect the primitive operations to be faster, though, I just wouldn't be surprised if they were about on par.)
My understanding was that cuDNN does not only target feature-complete libraries like Caffe or Torch or Theano (where new features only make sense if they improve over existing implementations), but also developers who just have a GPU and cuBLAS, cuFFT, cuSparse [...] from where it's still a long way to build a neural network implementation.

Anyway... do we merge de97c0a?

benanne:
Sure, sounds good to me.

f0k:
Well then... it's merged. If cuDNN v4 turns out to change their interface for the final version, we'll roll back to 8818d9c, though :)

To anybody who trained models with any of the old batch normalization implementations, note that the layer stores `inv_std` instead of `var` now.
To convert old models, you will need to update their `inv_std` parameters after `set_all_param_values()`:
```python
set_all_param_values(network, my_loaded_values)
# convert from old batchnorm implementation to new one:
for layer in get_all_layers(network):
    if hasattr(layer, 'inv_std'):
        var = layer.inv_std.get_value()
        inv_std = (1 / np.sqrt(var + layer.epsilon)).astype(var.dtype)
        layer.inv_std.set_value(inv_std)
```

If this turns out to affect a lot of users and a lot of existing models, we can still add a flag to the `BatchNormLayer` that allows to switch back to the old behaviour of computing and storing the variance instead of the inverted standard deviation.

benanne:
Yay! Excellent work :)

lasagne/layers/normalization.py
```diff
@@ -101,3 +115,254 @@ def get_output_for(self, input, **kwargs):
...
+    batch_norm : Convenience function to apply batch normalization to a layer
+
+    References
+    ----------
+    .. [1] Ioffe, Sergey and Szegedy, Christian (2015):
+           Batch Normalization: Accelerating Deep Network Training by Reducing
+           Internal Covariate Shift. http://arxiv.org/abs/1502.03167.
+    """
+    def __init__(self, incoming, axes='auto', epsilon=1e-4, alpha=0.1,
+                 mode='low_mem', beta=init.Constant(0), gamma=init.Constant(1),
```
cancan101: What ever became of the `low_mem` setting? It does not seem to do anything.

f0k: It was left in from 8818d9c by accident, I removed it in #613. Now that cuDNN v5 changed the API to match what we had in 8818d9c, it may come back (once it's in Theano, see Theano/Theano/pull/4237), but then it won't have an effect if you use cuDNN.

cxy19941228:
how to Add BatchNormLayer  to my lasagne module  ???
thanks

rafaelvalle:
@cxy19941228 please ask such question on the users forums.
[this](http://stackoverflow.com/questions/37624279/how-to-properly-add-and-use-batchnormlayer/) provides an explanation how to use a BatchNormLayer or BatchNorm.

f0k:
> how to Add BatchNormLayer to my lasagne module ???

Install the bleeding-edge versions of Theano and Lasagne: http://lasagne.readthedocs.io/en/latest/user/installation.html#bleeding-edge-version

PR #480, opened 2015-10-30
mnist.py: Exit with error when model name isn't recognised.

darrengarvey:
mnist.py expects the first command line argument to the a model
name. If the model isn't recognised there is an UnboundedLocal
error as `network` is used before assignment.

Since there's nothing that can be done in this case, exit with
error letting the user pick a valid model.

### DIFF
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -249,6 +249,7 @@ def main(model='mlp', num_epochs=500):
         network = build_cnn(input_var)
     else:
         print("Unrecognized model type %r." % model)
+        return
 
     # Create a loss expression for training, i.e., a scalar objective we want
     # to minimize (for our multi-class problem, it is the cross-entropy loss):


examples/mnist.py
```diff
@@ -249,6 +249,7 @@ def main(model='mlp', num_epochs=500):
         network = build_cnn(input_var)
     else:
         print("Unrecognized model type %r." % model)
+        sys.exit(1)
```
f0k: Oh yes, that was missing. I'd prefer to just have a `return` here. If someone imports the file and calls `main()`, they wouldn't want their program to exit if `model` was unrecognized. Do you agree? Could you amend your commit to do a `return` instead of `sys.exit(1)`? (Edit the file, do `git add mnist.py; git commit --amend; git push --force`.) Thanks!

darrengarvey:
Thanks for the review @f0k. Yeah I agree that a `return` is a less brutal / kinder approach. PR updated!

f0k:
Sorry, have been busy -- thanks for updating it, merging!

PR #484, opened 2015-11-06
Update mnist.py

vikaschawlavc:


benanne:
Could you clarify why you think this change is necessary? It looks like `gzip` is already imported where it is needed, but perhaps I'm missing something :)

vikaschawlavc:
It was giving error on my system without that line. :) Maybe i was doing something wrong..!!

benanne:
That's weird, as far as I can see the only function that uses the module also imports it in the function body. So this shouldn't happen. Could you provide the error message + backtrace? Thanks!

f0k:
Tried the unmodified mnist.py with python 2.7 and 3.4, there were no problems with a missing `gzip` module or anything. Closing this issue. Please feel free to reopen if you can reproduce the error and post a full traceback!

Issue #463, opened 2015-10-15
installation from master fails in ipython3 with UnicodeDecodeError


matthias-k:
I try to install Lasange in python3 (in a docker container running `jupyter/scipy-notebook`) from master, but the installation fails with a UnicodeDecodeError when trying to read `CHANGES.rst`. To reproduce:

```
% docker run -t -i jupyter/scipy-notebook /bin/bash                        
root@80f2ce016b94:/home/jovyan/work# pip install --upgrade https://github.com/Lasagne/Lasagne/archive/master.zip
Collecting https://github.com/Lasagne/Lasagne/archive/master.zip
  Downloading https://github.com/Lasagne/Lasagne/archive/master.zip (176kB)
    100% |################################| 180kB 1.9MB/s 
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 20, in <module>
      File "/tmp/pip-m2x57j8q-build/setup.py", line 10, in <module>
        CHANGES = open(os.path.join(here, 'CHANGES.rst')).read()
      File "/opt/conda/lib/python3.4/encodings/ascii.py", line 26, in decode
        return codecs.ascii_decode(input, self.errors)[0]
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 243: ordinal not in range(128)
    
    ----------------------------------------
```

daikazoku:
I had the same problem yesterday (but not 2-3 days ago). I fixed it by setting an UTF-8 locale on my system.

f0k:
Hmm, that actually doesn't look very good anyway, we don't open the files for reading only and we don't close them:
```python
here = os.path.abspath(os.path.dirname(__file__))
try:
    README = open(os.path.join(here, 'README.rst')).read()
    CHANGES = open(os.path.join(here, 'CHANGES.rst')).read()
except IOError:
    README = CHANGES = ''
```

A possible fix for your case might be:
```python
here = os.path.abspath(os.path.dirname(__file__))
try:
    with open(os.path.join(here, 'README.rst'), 'rb') as f:
        README = f.read()
    with open(os.path.join(here, 'CHANGES.rst'), 'rb') as f:
        CHANGES = f.read()
except IOError:
    README = CHANGES = ''
```

This will read the files in binary mode and thus won't try to decode anything, but the `long_description` will be of type `bytes` instead of `str` then -- is that a problem? Can you try?

The alternative would be:
```python
[...]

PR #487, opened 2015-11-09
Avoid dependency on pkg_resources for lasagne.__version__

### LINKED ISSUE #477: get the __version__ differently
Right now, the `__version__` is retrieved by communicating with the packaging system. However, when using lasagne outside of the packaging system, this is quite annoying.
```
  File "/home/jdgrave/python_modules/lasagne/__init__.py", line 33, in <module>
    __version__ = pkg_resources.get_distribution("Lasagne").version
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 344, in get_distribution
    if isinstance(dist,Requirement): dist = get_provider(dist)
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 223, in get_provider
    return working_set.find(moduleOrReq) or require(str(moduleOrReq))[0]
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 725, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib/python2.7/dist-packages/pkg_resources.py", line 628, in resolve
    raise DistributionNotFound(req)
pkg_resources.DistributionNotFound: Lasagne
```
Not only is it crash prone, if I am correct it is also unreliable (because the Lasange installed might not be the Lasagne I am accessing).
I am juggling with a ton of libraries in different versions, so I install these libraries in a python_modules folder. Most of the packages do this fine (numpy, scipy, theano), but lasange doesn't play nicely right now.

```python
import pkg_resources
__version__ = pkg_resources.get_distribution("Lasagne").version
del pkg_resources
```

f0k:
Instead of using `pkg_resources.get_distribution()` to set `lasagne.__version__` to whatever version of Lasagne is currently installed, this PR hard-codes `lasagne.__version__` to a plain string and reads that string from `setup.py` to avoid maintaining it in two places.

Resolves #477, closes #478.

The less adventurous alternative would be hard-coding it both in `__init__.py` and in `setup.py`. Any preference?

### DIFF
--- a/lasagne/__init__.py
+++ b/lasagne/__init__.py
@@ -29,6 +29,4 @@ from . import updates
 from . import utils
 
 
-import pkg_resources
-__version__ = pkg_resources.get_distribution("Lasagne").version
-del pkg_resources
+__version__ = "0.2.dev1"
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,17 @@
 import os
+import re
 from setuptools import find_packages
 from setuptools import setup
 
-version = '0.2.dev1'
-
 here = os.path.abspath(os.path.dirname(__file__))
 try:
+    # obtain version string from __init__.py
+    init_py = open(os.path.join(here, 'lasagne', '__init__.py')).read()
+    version = re.search('__version__ = "(.*)"', init_py).groups()[0]
+except Exception:
+    version = ''
+try:
+    # obtain long description from README and CHANGES
     README = open(os.path.join(here, 'README.rst')).read()
     CHANGES = open(os.path.join(here, 'CHANGES.rst')).read()
 except IOError:


benanne:
Looks okay to me. There's a failing PEP8 check though :)

f0k:
Oops. Corrected (moved the `__version__ = ` assignment to the end, `python setup.py -V` still works as intended). The only question is whether there's any situation in which `setup.py` cannot access `lasagne/__init__.py` for reading -- but we're relying on it being able to read `README.rst` and `CHANGES.rst` already.

@dnouri, our packaging expert, any comment on whether this solution is fine?

PS: We should address #463 afterwards.

dnouri:
:+1:

f0k:
Thanks for the thumbs up! Felt confident to merge now.

PR #490, opened 2015-11-11
Fix default argument is mutable

sieben:


f0k:
Thanks, but did that bite you anywhere? We've got mutable arguments elsewhere (e.g., `W=GlorotUniform()` in many constructors in `lasagne.layers.*`), but like here, there's no practical reason anybody would try to modify them. And in fact, for the case you fixed, the `tags` argument can't be modified by user code because it's passed on as `**tags`. So I'd rather leave it the way it is because it results in a clearer signature.

f0k:
(Just to be clear, I'm happy to discuss this, please don't feel put off!)

sieben:
@f0k No don't worry :) It's just that my IDE (pycharm) suggested it was a bad practice (http://docs.python-guide.org/en/latest/writing/gotchas/) but if you think that's unnecessary no problem :)

PR #497, opened 2015-11-12
Change learn_init to default to False in GRULayer

### LINKED ISSUE #496: GRULayer - learn_init=True inconsistent with the rest
I guess it might be a typo, or there is some reason for this?

craffel:
Resolves #496.


f0k:
I'm fine with merging. Do you want to post an `[interface change] ..." mail to the mailing list afterwards or should I? It will affect all users experimenting with GRUs and not specifying `learn_init` explicitly.

craffel:
Sure, I figured we'd need to do an interface change email.  I can do it.

f0k:
All right, merged it! Spread the news as soon as you're ready :)

PR #498, opened 2015-11-12
Add Conv3DDNNLayer

f0k:
This adds a `Conv3DDNNLayer` class for cuDNN-based 3D convolution. It also refactors the convolution tests to reduce some of the code duplication in setting up the test cases (there's still a lot of duplication, but that's harder to tackle).

In a follow-up PR, I'd like to refactor all convolutional layers (all dimensionalities and all implementations) to be derived from a common `BaseConvLayer` that contains most of the code. This will make it much easier to add additional checks such as matching input and convolution dimensionality.

### DIFF
--- a/lasagne/layers/dnn.py
+++ b/lasagne/layers/dnn.py
@@ -17,6 +17,7 @@ __all__ = [
     "Pool2DDNNLayer",
     "MaxPool2DDNNLayer",
     "Conv2DDNNLayer",
+    "Conv3DDNNLayer",
 ]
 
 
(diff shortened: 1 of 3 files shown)


benanne:
Sounds like a good plan. In the meantime I will try to have an in-depth look at this within the next few days.

f0k:
> Sounds like a good plan.

The only trouble is that some convolutional layers are already derived from a `Layer` subclass such as `MMLayer`, `DNNLayer`, ... those would need to become a mix-in class then. They're not used for anything except for `isinstance` checks (and maybe not even that).

f0k:
Anybody up for a review? It's mostly a copy of Conv2DDNNLayer with some numbers incremented by one, plus some refactoring of the tests. Cheers!

benanne:
> The only trouble is that some convolutional layers are already derived from a Layer subclass such as MMLayer, DNNLayer, ... those would need to become a mix-in class then. They're not used for anything except for isinstance checks (and maybe not even that).

Indeed, I don't think these base classes are used for anything. We can probably just get rid of them, unless there are implications that I'm not thinking of?

lasagne/layers/dnn.py
```diff
@@ -314,3 +315,191 @@ def get_output_for(self, input, **kwargs):
...
+    lasagne.layers.Conv3DDNNLayer(incoming, num_filters, filter_size,
+    stride=(1, 1, 1), pad=0, untie_biases=False,
+    W=lasagne.init.GlorotUniform(), b=lasagne.init.Constant(0.),
+    nonlinearity=lasagne.nonlinearities.rectify, flip_filters=False,
+    **kwargs)
+
+    3D convolutional layer
+
+    Performs a 3D convolution on its input and optionally adds a bias and
+    applies an elementwise nonlinearity.  This is an alternative implementation
```
benanne: I guess "This is an alternative implementation" doesn't make much sense considering that it's the only one right now :)

lasagne/layers/dnn.py
```diff
@@ -314,3 +315,191 @@ def get_output_for(self, input, **kwargs):
...
+
+    pad : int, iterable of int, 'full', 'same' or 'valid' (default: 0)
+        By default, the convolution is only computed where the input and the
+        filter fully overlap (a valid convolution). When ``stride=1``, this
+        yields an output that is smaller than the input by ``filter_size - 1``.
+        The `pad` argument allows you to implicitly pad the input with zeros,
+        extending the output size.
+
+        A single integer results in symmetric zero-padding of the given size on
+        all borders, a tuple of two integers allows different symmetric padding
```
benanne: three integers, not two :)

lasagne/tests/layers/test_conv.py
```diff
@@ -7,120 +7,117 @@
...
     numpy array
     """
+    if n is None:
+        n = input.ndim - 2
     if pad not in ['valid', 'same', 'full']:
-        pad = as_tuple(pad, 2, int)
-        input = np.pad(input,
-                       ((0, 0), (0, 0), (pad[0], pad[0]), (pad[1], pad[1])),
-                       mode='constant')
+        pad = as_tuple(pad, n)
```
benanne: should this not be `pad = as_tuple(pad, n, int)`?

lasagne/tests/layers/test_conv.py
```diff
@@ -7,120 +7,117 @@
...
-        shift_x = (kernel.shape[2] - 1) // 2
-        shift_y = (kernel.shape[3] - 1) // 2
-        output = output[:, :, shift_x:input.shape[2] + shift_x,
-                        shift_y:input.shape[3] + shift_y]
+        shift = tuple((k - 1) // 2 for k in kernel.shape[2:])
+        slices = [slice(None), slice(None)]
+        slices += [slice(s, s + i) for s, i in zip(shift, input.shape[2:])]
+        output = output[slices]
+
+    stride = as_tuple(stride, n)
```
benanne: `as_tuple(stride, n, int)` here as well?

lasagne/tests/layers/test_conv.py
```diff
@@ -287,7 +284,7 @@ def test_init_none_nonlinearity_bias(self, Conv2DImpl, DummyInputLayer):
         assert layer.b is None
 
     def test_invalid_pad(self, Conv2DImpl, DummyInputLayer):
-        input_layer = DummyInputLayer((1, 2, 3))
+        input_layer = DummyInputLayer((1, 2, 3, 3))
```
benanne: Wait, how did this pass before if the input layer only had 3 axes? Or is it because it was a `DummyInputLayer`?

benanne:
Looking good, just a few minor remarks :)

lasagne/layers/dnn.py
f0k: I figured we'd forget changing this when we add a standard Conv3DLayer... but yeah, I'll just remove "alternative" :)

lasagne/layers/dnn.py
f0k: This was a test :) (Well, maybe not. Good catch!)

lasagne/tests/layers/test_conv.py
f0k: Ah, yes, that's better. Changed.

lasagne/tests/layers/test_conv.py
f0k: Good idea.

lasagne/tests/layers/test_conv.py
f0k: It's magic! Or maybe it's just because none of the convolutional layers actually bothers to check its input shape in the constructor, so it gets to the point where it runs `self.pad = as_tuple(pad, 2, int)`, which raises the `TypeError` tested for.

Checking the input dimensionality is one of the things I'd like to add after refactoring the convolutional layers to be derived from a common base class.

benanne: Sounds like a plan!

benanne:
Ready to merge as far as I'm concerned, unless you want someone else to take a look as well.

f0k:
Addressed your comments, thanks for reviewing! Let me know if you're fine with merging.

> I think this unified interface is the right way to go, I can't think of any good reason to split them up into three separate classes (with lots of code duplication as a result).

There wouldn't be any code duplication. My idea was to move everything into the base class (a single class for all dimensionalities) except for the convolution operation, which would be solved be polymorphism then. In Theano we don't have a `convNd()` function, so if we had a single `ConvLayer`, it'd end up including a lot of `if` cases in `get_output_for()` to distinguish the dimensionalities (where "a lot" means "three" ;) ). Plus, forcing users to explicitly choose between `Conv1DLayer`, `Conv2DLayer` and `Conv3DLayer` will avoid masking errors such as forgetting to specify a `channels` dimension for single-channel inputs.

benanne:
I guess you're right, there aren't many use cases where you would want to switch code from 1D to 2D or whatever, so it's not a problem to have different interfaces for this. Also, if you really want to do this, you only have to change one character :)

PR #499, opened 2015-11-12
init docs

### LINKED ISSUE #468: Glorot gains for Leaky_rectify and very_leaky_rectify
Should we add the correct Glorot gains for `leaky_rectify` and `very_leaky_rectify`. That would be

```Python
if gain == 'rectify':
    gain = np.sqrt(2)
elif nonlin == 'leaky_rectify':
    gain = np.sqrt(2/(1+0.01**2))
elif nonlin == 'very_leaky_rectify':
    gain = np.sqrt(2/(1+(1/3)**2))
```

Maybe also add a note to the doc that the correct  init for `leaky_rectify` with `alpha` leakiness is `np.sqrt(2/(1+alpha**2))`

skaae:
fixes #468 
Also add toc to inits. Makes it easier to read. Let me know what you think? 

- use ‘relu’ keyword in example. I think that is the most typical usecase?
- add autosummary to init docs
- add leaky relu formula to glorot
- moves some references that does not work in autosummary

### DIFF
--- a/lasagne/init.py
+++ b/lasagne/init.py
@@ -5,7 +5,8 @@ Examples
 --------
 >>> from lasagne.layers import DenseLayer
 >>> from lasagne.init import Constant, GlorotUniform
->>> l1 = DenseLayer((100,20), num_units=50, W=GlorotUniform(), b=Constant(0.0))
+>>> l1 = DenseLayer((100,20), num_units=50,
+...                 W=GlorotUniform('relu'), b=Constant(0.0))
 """
 
 import numpy as np
@@ -100,9 +101,9 @@ class Uniform(Initializer):
 
 
 class Glorot(Initializer):
-    """Glorot weight initialization [1]_.
+    """Glorot weight initialization.
 
-    This is also known as Xavier initialization.
+    This is also known as Xavier initialization [1]_.
 
     Parameters
     ----------
@@ -111,9 +112,11 @@ class Glorot(Initializer):
         constructor to sample from a distribution with a given standard
         deviation.
     gain : float or 'relu'
-        Scaling factor for the weights. Set this to 1.0 for linear and sigmoid
-        units, to 'relu' or sqrt(2) for rectified linear units. Other transfer
-        functions may need different factors.
+        Scaling factor for the weights. Set this to ``1.0`` for linear and
+        sigmoid units, to 'relu' or ``sqrt(2)`` for rectified linear units, and
+        to ``sqrt(2/(1+alpha**2)`` for leaky rectified linear units with
+        leakiness ``alpha``. Other transfer functions may need different
+        factors.
     c01b : bool
         For a :class:`lasagne.layers.cuda_convnet.Conv2DCCLayer` constructed
         with ``dimshuffle=False``, `c01b` must be set to ``True`` to compute
@@ -128,8 +131,8 @@ class Glorot(Initializer):
 
     Notes
     -----
-    For a :class:`DenseLayer`, if ``gain='relu'`` and ``initializer=Uniform``,
-    the weights are initialized as
+    For a :class:`DenseLayer <lasagne.layers.DenseLayer>`, if ``gain='relu'``
+    and ``initializer=Uniform``, the weights are initialized as
 
     .. math::
        a &= \\sqrt{\\frac{6}{fan_{in}+fan_{out}}}\\\\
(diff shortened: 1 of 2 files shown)


lasagne/init.py
```diff
@@ -5,7 +5,7 @@
 --------
 >>> from lasagne.layers import DenseLayer
 >>> from lasagne.init import Constant, GlorotUniform
->>> l1 = DenseLayer((100,20), num_units=50, W=GlorotUniform(), b=Constant(0.0))
+>>> l1 = DenseLayer((100,20), num_units=50, W=GlorotUniform('relu'), b=Constant(0.0))
```
f0k: Line too long! For a multiline doctest, the format is:
```python
>>> l1 = DenseLayer(something something something
...                 something something)
```

lasagne/init.py
```diff
@@ -100,9 +100,9 @@ def sample(self, shape):
 
 
 class Glorot(Initializer):
-    """Glorot weight initialization [1]_.
+    """Glorot weight initialization.
 
-    This is also known as Xavier initialization.
+    This is also known as Xavier initialization[1]_.
```
f0k: +1 for moving the reference (to fix the autosummary), but you're missing a blank before the `[1]_`.

lasagne/init.py
```diff
@@ -141,6 +141,15 @@ class Glorot(Initializer):
        \\sigma &= \\sqrt{\\frac{2}{fan_{in}+fan_{out}}}\\\\
        W &\sim N(0, \\sigma)
 
+    For leaky ReLU ``gain`` can be calculated as
+
+    .. math::
+       gain = \\sqrt{\\frac{2}{1+\\alpha^2}}
+
+    where :math:`\\alpha` is the leakyness. For
```
f0k: leakyness -> leakiness

lasagne/init.py
skaae: noob mistanke. sorry about that :)

lasagne/init.py
```diff
@@ -193,10 +202,10 @@ def __init__(self, gain=1.0, c01b=False):
 
 
 class He(Initializer):
-    """He weight initialization [1]_.
+    """He weight initialization.
 
     Weights are initialized with a standard deviation of
-    :math:`\\sigma = gain \\sqrt{\\frac{1}{fan_{in}}}`.
+    :math:`\\sigma = gain \\sqrt{\\frac{1}{fan_{in}}}`. [1]_
```
skaae: Moved the dot here as well.

lasagne/init.py
```diff
@@ -141,6 +141,15 @@ class Glorot(Initializer):
...
        W &\sim N(0, \\sigma)
 
+    For leaky ReLU ``gain`` can be calculated as
+
+    .. math::
+       gain = \\sqrt{\\frac{2}{1+\\alpha^2}}
+
+    where :math:`\\alpha` is the leakyness. For
+    ``leaky_rectify`` and
+    ``very_leaky_rectify`` :math:`\\alpha` is  0.01 and 0.333 respectively.
```
f0k: Can you correctly cross-reference those? Should be:
```
:func:`leaky_rectify <lasagne.nonlinearities.leaky_rectify>`
```
or shorter (never tried that):
```
:func:`~lasagne.nonlinearities.leaky_rectify`
```

skaae: Great. Tried to do cross-reference but i couldn't figure out how to reference a function.

f0k:
Actually, wouldn't it be easier to just give the formula in the description of the `gain` parameter? We currently have:
```
     Scaling factor for the weights. Set this to 1.0 for linear and sigmoid
     units, to 'relu' or sqrt(2) for rectified linear units. Other transfer
     functions may need different factors.
```
This could just become:
```
    Scaling factor for the weights. Set this to ``1.0`` for linear and sigmoid
    units, to 'relu' or ``sqrt(2)`` for rectified linear units, and to
    ``sqrt(2/(1+alpha**2)`` for leaky rectified linear units with leakiness
    ``alpha``. Other transfer functions may need different factors.
```
That's shorter than your "Notes" paragraph and could be copied over to all initializers that have a gain factor.

skaae:
Yes thats fine with me. Is the gain the same for Glorot and He?

skaae:
+ for orthogonal?

f0k:
It's the same for all three. While you're at it, can you reference http://arxiv.org/abs/1312.6120 in the orthogonal initialization?

skaae:
yes

f0k:
Cool! Looks good to merge for me!

f0k:
Since there are no objections, I'll merge it. Thanks again!

Issue #425, opened 2015-09-21
Provide RNN "containers"


f0k:
In https://github.com/Lasagne/Recipes/pull/23#issuecomment-141979978, @skaae brought up the idea of splitting up the recurrent layers into layers performing a single computation step and a container layer that handles the recurrence (i.e., sets up the `scan` function or explicitly unrolls it).

The use case was to provide a way to efficiently sample from an RNN that was trained to generate time sequences (e.g., text) -- the most efficient way would be to tap into the scan function, but that's impossible the way it's currently encapsulated.
Another advantage could be that it reduces the amount of code duplication (see #356), and maybe it would even make it easier to equip LSTMs or GRUs with custom connections (#357)?

As far as I've understood, the following current code:
```
layer = lasagne.layers.InputLayer(shape=(N_BATCH, MAX_LENGTH, 2))
layer = lasagne.layers.RecurrentLayer(layer, N_HIDDEN)
layer = lasagne.layers.SliceLayer(layer, -1, 1)
layer = lasagne.layers.DenseLayer(layer, num_units=1, nonlinearity=lasagne.nonlinearities.tanh)
```
Could be turned into:
```
layer = lasagne.layers.InputLayer(shape=(N_BATCH, MAX_LENGTH, 2))
# single time step
rnn_layer = lasagne.layers.InputLayer((N_BATCH, 2))
rnn_layer = lasagne.layers.VanillaRecurrentStep(rnn_layer, N_HIDDEN)
layer = lasagne.layers.Recurrence(layer, step=rnn_layer)
layer = lasagne.layers.SliceLayer(layer, -1, 1)
layer = lasagne.layers.DenseLayer(layer, num_units=1, nonlinearity=lasagne.nonlinearities.tanh)
```
Where we'd have different `Step` layers, and one of them could be sampling from a multivariate distribution, for the character generation use case.
`Recurrence` could have a keyword argument telling whether it persists its recurrent states, and another telling whether it feeds back on its own output (that would be needed for character generation).

Disclaimer: I've just started looking into RNNs and am not sure what's the best way to break down the code yet -- I'm just opening this issue so there is a place to discuss the idea for those who are into RNNs with possible participation by a future me :)

benanne:
For someone (like me) whose experience with RNNs is limited, the first (current) approach does seem a lot more natural, but I appreciate that some might need / expect more flexibility.

[...]

PR #501, opened 2015-11-13
Add kwarg to recurrent layers to return only final output

### LINKED ISSUE #471: Recurrent layers: "Keep final time step" possibly suboptimal
Recurrent layers are frequently used in a way that only their final output for a time sequence needs to be propagated. There's an optimization for Theano's scan function that recognizes the case:
```python
outputs, updates = theano.scan(...)
keep = outputs[-1]
```
In Lasagne, this use case is currently supported by adding a SliceLayer on top of a RecurrentLayer:
```python
layer = RecurrentLayer(...)
layer = SliceLayer(layer, -1, axis=1)
```
However, the recurrent layers shuffle the scan output from (timesteps, batchsize, features) to (batchsize, timesteps, features), so Theano will never encounter a subtensor(scan(...)) sequence, but a subtensor(dimshuffle(scan(...))) sequence. We need to check if that case is optimized as well.

Lasagne's recurrent layers also accept an `unroll_scan` keyword argument, which unrolls the scan into a graph that ends in a `T.stack()` for stacking all the time steps. Again, if there's a SliceLayer on top to only keep the last output, Theano will see a subtensor(dimshuffle(stack(...))) sequence. We need to check if that case is optimized into keeping the last time step only, or whether it will actually stack all the time steps first.

Both is complicated by the fact that for layers processing the sequences in reverse order, there will also be a subtensor(::-1) between the recurrent layer's output and the SliceLayer. We also need to check whether those cases are optimized.

For the first case, Theano would need an optimization subtensor(dimshuffle(...)) -> dimshuffle(subtensor(...)), but maybe it has this. For the second case, it would need an optimization subtensor(dimshuffle(stack(...))) -> ..., but maybe it has this already. For reverse layers, Theano would need to be able to turn x[::-1][0] into x[-1] before applying the other optimizations.

Alternatively, Lasagne's recurrent layers could get a `final_state_only` keyword argument that tells them to keep the final state only and constructs the appropriate graph. That would solve all problems at once, but only for Lasagne.

craffel:
Resolves #471.  I opted for `return_final`; it has the same meaning as `only_return_final` but is shorter to type.

### DIFF
--- a/examples/recurrent.py
+++ b/examples/recurrent.py
@@ -113,30 +113,24 @@ def main(num_epochs=NUM_EPOCHS):
     # We're using a bidirectional network, which means we will combine two
     # RecurrentLayers, one with the backwards=True keyword argument.
     # Setting a value for grad_clipping will clip the gradients in the layer
+    # Setting only_return_final=True makes the layers only return their output
+    # for the final time step, which is all we need for this task
     l_forward = lasagne.layers.RecurrentLayer(
         l_in, N_HIDDEN, mask_input=l_mask, grad_clipping=GRAD_CLIP,
         W_in_to_hid=lasagne.init.HeUniform(),
         W_hid_to_hid=lasagne.init.HeUniform(),
-        nonlinearity=lasagne.nonlinearities.tanh)
+        nonlinearity=lasagne.nonlinearities.tanh, only_return_final=True)
     l_backward = lasagne.layers.RecurrentLayer(
         l_in, N_HIDDEN, mask_input=l_mask, grad_clipping=GRAD_CLIP,
         W_in_to_hid=lasagne.init.HeUniform(),
         W_hid_to_hid=lasagne.init.HeUniform(),
-        nonlinearity=lasagne.nonlinearities.tanh, backwards=True)
-    # The objective of this task depends only on the final value produced by
-    # the network.  So, we'll use SliceLayers to extract the LSTM layer's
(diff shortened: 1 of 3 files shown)


craffel:
Writing this PR reminded me of how much code duplication we have and how badly we need #425

f0k:
I suggested `only_return_final` because to me `return_final=False` suggests it would omit the final state, and `return_final=True` suggests it would include it (as it does now). If three words are too long, what about `only_final` or `final_only`?

craffel:
Ok, changed to `only_return_final`.

f0k:
Looks good, but could you please update the example in the same commit? It shouldn't use the `SliceLayer(..., -1)` any longer. Cheers!

craffel:
> Looks good, but could you please update the example in the same commit? It shouldn't use the SliceLayer(..., -1) any longer. Cheers!

Ok, done.

f0k:
Coveralls found a line not covered: https://coveralls.io/builds/4167558/source?filename=lasagne%2Flayers%2Frecurrent.py#L910

f0k:
Looks good to merge now! If there are no further comments, I'll merge it tomorrow.
PS: Sorry for the slow/incremental review.

craffel:
> PS: Sorry for the slow/incremental review.

No worries, thanks as always for your thoroughness.

f0k:
Merged, thanks again!

PR #502, opened 2015-11-14
fix adadelta docs for moving average of update

diogo149:
There appeared to be a minor error. Before the change, the equations for r_t and s_t were identical.


f0k:
Thanks, your docstring change matches Algorithm 1 in [the paper](http://arxiv.org/abs/1212.5701), and Algorithm 1 matches the Python implementation. Merging!

PR #505, opened 2015-11-17
Top-k accuracy

### LINKED ISSUE #504: Getting top-k accuracy
Is there a straightforward way of having a Lasagne CNN (similar to the MNIST example) give me top-`k` accuracy instead of just exact match accuracy?

f0k:
This adds a `top_k` parameter to `lasagne.objectives.categorical_accuracy()` to allow computing the top-k accuracy.
If anybody knows a better way of expressing the top-k accuracy in a formula, please comment! If you think omitting the formula would be better than the one I have, also let me know.

Closes #504.

### DIFF
--- a/lasagne/objectives.py
+++ b/lasagne/objectives.py
@@ -291,7 +291,7 @@ def multiclass_hinge_loss(predictions, targets, delta=1):
 def binary_accuracy(predictions, targets, threshold=0.5):
     """Computes the binary accuracy between predictions and targets.
 
-    .. math:: L_i = \\mathbb{I}(t_i == p_i)
+    .. math:: L_i = \\mathbb{I}(t_i = \mathbb{I}(p_i \\ge \\alpha))
 
     Parameters
     ----------
@@ -325,7 +325,7 @@ def binary_accuracy(predictions, targets, threshold=0.5):
 def categorical_accuracy(predictions, targets):
     """Computes the categorical accuracy between predictions and targets.
 
-    .. math:: L_i = \\mathbb{I}(t_i == p_i)
+    .. math:: L_i = \\mathbb{I}(t_i = \\operatorname{argmax}_c p_{i,c})
 
     Parameters
     ----------
--- a/lasagne/objectives.py
+++ b/lasagne/objectives.py
@@ -322,10 +322,15 @@ def binary_accuracy(predictions, targets, threshold=0.5):
     return theano.tensor.eq(predictions, targets)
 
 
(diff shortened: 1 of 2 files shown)


benanne:
Nice! Looks like the test is failing on Travis, does it pass locally?

f0k:
Oops, the second test was missing the `top_k=3` param. It passed locally, but that was pure luck then. I'll figure out a random seed that ensures the test will not pass when computing the accuracy with `top_k=1`.

f0k:
If Travis passes, this is ready to be reviewed!

benanne:
I'm wondering about the motivation for adding it to the existing `categorical_accuracy` function instead of adding a new one? I'm not necessarily opposed to it, just curious. People unfamiliar with this concept might be puzzled why our function for computing accuracy is so complicated.

f0k:
> I'm wondering about the motivation for adding it to the existing `categorical_accuracy` function instead of adding a new one? I'm not necessarily opposed to it, just curious.

Because the first steps (support for one-hot-encoded targets, dimensionality checks) and most of the docstring (everything except for the `top_k` parameter) are identical, and the `top_k=1` case is identical to the standard accuracy.

> People unfamiliar with this concept might be puzzled why our function for computing accuracy is so complicated.

But if we copy the full function only to change the last two lines, people might be puzzled why we duplicated so much code ;) They're probably puzzled over the convolutional and recurrent layers as well.
Would it help if I label the two `if` cases with `# plain categorical accuracy` and `# top-k accuracy`?

benanne:
Fair enough :) Yeah, the extra comments might improve readability a bit, but it's not a huge deal.

lasagne/objectives.py
```diff
@@ -291,7 +291,7 @@ def multiclass_hinge_loss(predictions, targets, delta=1):
 def binary_accuracy(predictions, targets, threshold=0.5):
     """Computes the binary accuracy between predictions and targets.
 
-    .. math:: L_i = \\mathbb{I}(t_i == p_i)
+    .. math:: L_i = \\mathbb{I}(t_i == \mathbb{I}(p_i \\ge \\alpha))
```
ebenolson: Having `==` in a math formula looks a bit strange to me.

f0k: I thought so as well, but left it as it is. I'll replace with `=`.

f0k:
Addressed Sander's and Eben's comments.

f0k:
Is this good to merge?

benanne:
Go for it :)

PR #506, opened 2015-11-19
Allow PadLayer to accept None dimensions

craffel:
This makes it so that the `PadLayer` can accept shapes with `None` in them past their "batch" dimension, e.g.

```Python
import numpy as np
import lasagne

input_shape = (None, 3, None, 40)
dummy_input = np.random.standard_normal((100, 3, 30, 40)).astype(theano.config.floatX)

l_in = lasagne.layers.InputLayer(input_shape)
l_pad = lasagne.layers.PadLayer(l_in, width=4)
l_conv = lasagne.layers.Conv2DLayer(l_pad, 16, (3, 3))
print lasagne.layers.get_output(l_conv).eval({l_in.input_var: dummy_input}).shape
```

Without this change, this would result in 
```Python
lasagne/layers/shape.py in get_output_shape_for(self, input_shape)
    327             except TypeError:
    328                 l = r = w
--> 329             output_shape[k + self.batch_ndim] += l + r
    330         return tuple(output_shape)
    331 

TypeError: unsupported operand type(s) for +=: 'NoneType' and 'int'
```

After this change, it results in `(100, 16, 36, 46)`.

### DIFF
--- a/lasagne/layers/shape.py
+++ b/lasagne/layers/shape.py
@@ -322,11 +322,14 @@ class PadLayer(Layer):
             widths = self.width
 
         for k, w in enumerate(widths):
-            try:
-                l, r = w
-            except TypeError:
-                l = r = w
-            output_shape[k + self.batch_ndim] += l + r
+            if output_shape[k + self.batch_ndim] is None:
+                continue
+            else:
+                try:
+                    l, r = w
+                except TypeError:
+                    l = r = w
+                output_shape[k + self.batch_ndim] += l + r
         return tuple(output_shape)
 
     def get_output_for(self, input, **kwargs):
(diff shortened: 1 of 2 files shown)


benanne:
Looks good, thanks! I think we should probably add some tests for the cases where the `width` specification is an iterable of ints / iterable of tuples (and not just an integer), in combination with `None` dimensions. That seems like a use case that could easily break :)

f0k:
Good catch. We should probably review all `get_output_shape_for()` implementations to be sure they support `None`s wherever possible.

craffel:
> Looks good, thanks! I think we should probably add some tests for the cases where the width specification is an iterable of ints / iterable of tuples (and not just an integer), in combination with None dimensions. That seems like a use case that could easily break :)

Ok, added a few more tests, does that cover what you were thinking about?

> We should probably review all get_output_shape_for() implementations to be sure they support Nones wherever possible.

For sure.

benanne:
Thanks for adding the additional test cases :) Good to merge for me.

f0k:
For me as well. Will you merge or should I?

benanne:
Done!

Issue #457, opened 2015-10-08
InputLayer does not check shape


f0k:
`Layer.__init__()` checks whether its incoming shape is all-positive, but `InputLayer.__init__()` lacks this. It would be good to have it check for a valid shape as well, to fail as early as possible.

PR #515, opened 2015-11-26
add negative shape check to inputLayer

skaae:
Fix https://github.com/Lasagne/Lasagne/issues/457

Add test for negative shape in input Layer. 
The `str(name)` is a bit ugly but the inputLayer might have a `None` name as far as i read the code?

### DIFF
--- a/lasagne/layers/input.py
+++ b/lasagne/layers/input.py
@@ -49,6 +49,12 @@ class InputLayer(Layer):
     """
     def __init__(self, shape, input_var=None, name=None, **kwargs):
         self.shape = shape
+        if any(d is not None and d <= 0 for d in self.shape):
+            raise ValueError((
+                "Cannot create InputLayer with a non-positive shape "
+                "dimension. shape=%r, self.name=%r") % (
+                    self.shape, name))
+
         ndim = len(shape)
         if input_var is None:
             # create the right TensorType for the given number of dimensions
--- a/lasagne/tests/layers/test_input.py
+++ b/lasagne/tests/layers/test_input.py
@@ -31,3 +31,11 @@ class TestInputLayer:
 
         with pytest.raises(ValueError):
             InputLayer((3, 2), input_var)
+
+    def test_nonpositive_input_dims_raises_value_error(self):
+        from lasagne.layers import InputLayer
+        with pytest.raises(ValueError):
+            InputLayer(shape=(None, -1, -1))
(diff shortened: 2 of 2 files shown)


lasagne/layers/input.py
```diff
@@ -49,6 +49,12 @@ class InputLayer(Layer):
     """
     def __init__(self, shape, input_var=None, name=None, **kwargs):
         self.shape = shape
+        if any(d is not None and d <= 0 for d in self.shape):
+            raise ValueError((
+                "Cannot create InputLayer with a non-positive shape "
+                "dimension. shape=%r, self.name=%s") % (
+                    self.shape, str(name)))
```
f0k: Can you please just follow the code from the base Layer constructor? It uses `self.name=%r` and uses `name` directly (not wrapped in `str()` -- note that this doesn't make a difference when using `%s` anyway, it implicitly calls `str()` on the argument to be formatted). Cheers!

lasagne/tests/layers/test_input.py
```diff
@@ -31,3 +31,8 @@ def test_bad_shape_fails(self):
 
         with pytest.raises(ValueError):
             InputLayer((3, 2), input_var)
+
+    def test_negative_shape_fails(self):
+        from lasagne.layers.input import InputLayer
+        with pytest.raises(ValueError):
+            InputLayer((3, -2))
```
f0k: Should check for zeros as well, not just for negative shapes. This test can basically be copy-pasted from `TestLayer.test_nonpositive_input_dims_raises_value_error()` in `test_base.py`.

lasagne/layers/input.py
skaae: yes. I thought that would fail, but it seems to work fine.

f0k: `%r` also has the advantage that you can still distinguish between a layer that has no name and a layer of the name `"None"`. Not that it would matter really, but it reads better.

lasagne/tests/layers/test_input.py
skaae: Looking at the tests. I'm not sure what Mock(...) does?

f0k: > I'm not sure what Mock(...) does?

https://docs.python.org/3/library/unittest.mock.html

f0k:
Thanks, looks good now! Merging.

PR #517, opened 2015-11-30
slices_by_input was wrong

trustswz:
if input length > 4, the code will break due to this line of code

previous code takes the dim of the input not the number of the inputs


benanne:
Thanks! We should probably add a test for this as well to prevent regressions. Do you reckon you could add one to this PR? If not, is anyone else willing to do it?

trustswz:
I won't be able to do it today. Happy to let anyone else adding the unit test! Something like crop_test(crop_0, [x0, x1, x2, x0, x1, x2],
                  [x0, x1, x2, x0, x1, x2])?

PR #518, opened 2015-12-01
Add Exponential Linear Unit (ELU)

f0k:
Adds the exponential linear unit (http://arxiv.org/abs/1511.07289).

Note: This isn't optimized for performance yet. It's possible that there's a way to express it more efficiently in Theano, but for optimum performance this may need to be implemented as a Theano Op.

### DIFF
--- a/docs/modules/nonlinearities.rst
+++ b/docs/modules/nonlinearities.rst
@@ -13,6 +13,7 @@
    LeakyRectify
    leaky_rectify
    very_leaky_rectify
+   elu
    softplus
    linear
    identity
@@ -31,5 +32,6 @@ Detailed description
    :members:
 .. autofunction:: leaky_rectify
 .. autofunction:: very_leaky_rectify
+.. autofunction:: elu
 .. autofunction:: softplus
 .. autofunction:: linear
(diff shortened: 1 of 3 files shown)


benanne:
Looks good :) In the paper they add an alpha parameter that determines the negative value it converges to for x -> -infinity. For all experiments they set this to 1, but perhaps it would be useful to provide this functionality as well?

f0k:
> For all experiments they set this to 1, but perhaps it would be useful to provide this functionality as well?

I thought I could duck out by mentioning this in the Notes. Supporting an alpha value requires setting this up as a class with a shortcut for `alpha=1`. Is it worth it? And is there any clever way of making things faster? I don't see a similar shortcut as for the leaky rectifier, possibly the only way would be to implement it as a Theano Op, complete with CPU and GPU implementation.

f0k:
Hmm, is there any reason why Travis wouldn't run on this? Is "(ELU)" a special marker for something?

benanne:
Yeah, I guess it might not be worth the hassle seeing as this is still a relatively unproven development. Once more people start using this in their papers we can consider providing full support I guess. In the meantime it's quite trivial to implement for those that really need it.

Strange about Travis. Might be a one time occurrence, or maybe their config format changed or something.

PR #521, opened 2015-12-02
Fix relu weight init doc

### LINKED ISSUE #520: Relu weight init doc problem?
From the [doc](http://lasagne.readthedocs.org/en/latest/modules/init.html#lasagne.init.Glorot) or [source](https://github.com/Lasagne/Lasagne/blob/master/lasagne/init.py#L103), it says:

```{python}
For a :class:`DenseLayer <lasagne.layers.DenseLayer>`, if ``gain='relu'``
    and ``initializer=Uniform``, the weights are initialized as
    .. math::
       a &= \\sqrt{\\frac{6}{fan_{in}+fan_{out}}}\\\\
       W &\sim U[-a, a]
```
But from what I read from the souce, the formula is true for a sigmoid activator. In case of Relu, it should be `U[-sqrt(2)a, sqrt(2)a]`, as the gain is `sqrt(2)`.

Am I right?

colinfang:
Fixes #520

### DIFF
--- a/lasagne/init.py
+++ b/lasagne/init.py
@@ -135,7 +135,7 @@ class Glorot(Initializer):
     and ``initializer=Uniform``, the weights are initialized as
 
     .. math::
-       a &= \\sqrt{\\frac{6}{fan_{in}+fan_{out}}}\\\\
+       a &= \\sqrt{\\frac{12}{fan_{in}+fan_{out}}}\\\\
        W &\sim U[-a, a]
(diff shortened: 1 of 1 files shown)


benanne:
Well-spotted :) Thanks for the correction! Good to merge as far as I'm concerned.

f0k:
Checked this as well, it's correct. Merging, thank you!

Issue #337, opened 2015-07-21
API: Multiple outputs in Layer.get_output_for()


f0k:
In the context of recurrent neural networks, we've found two use cases that would need layers to be able to return multiple outputs:

1. @skaae wanted to return the hidden state along with the output for recurrent layers (I don't exactly recall why, please help me out, but I remember that there was no alternative to returning multiple outputs from a single `get_output_for()` call)
2. When training on batches of sequences of different length, it's necessary to mask out items that are to be ignored (or give the sequence lengths in another way). Currently, #294 solves this by adding a `mask` keyword argument to `get_output_for()`, but this only works for networks that have a single input layer only, or that have the same sequence lengths for all inputs. For more flexibility, the mask should be paired with the inputs as they are propagated through the network. This would require both the `InputLayer` and recurrent layers to return multiple outputs in `get_output_for()` (or, instead of having `InputLayer` returning multiple outputs itself, have a `MergeLayer` subclass that returns all its inputs instead of actually merging them).

While we don't have to solve this now, we should figure out whether it's possible to add this functionality later on without breaking the API. If so, we can postpone it until after the first release, otherwise we probably shouldn't.

A simple solution would be allowing `get_output_for()` to return a list, or a tuple, or a dictionary, or an OrderedDict of outputs alternatively to returning a single theano expression (we should pick one). A dictionary or OrderedDict would allow outputs to be named, such that the mask could be called `mask`, for example. This might improve readability of layers dealing with multiple inputs. As far as I can see, this could be added later without breaking existing code.
In addition, `get_output_shape_for()` would also have to be changed to return multiple shapes.

If we do this, the next question is what to stack on top. Only allow `MergeLayer` because it gets multiple inputs and input shapes? Or allow `Layer.get_output_for()` to accept a list/tuple/dictionary as its single `input` argument? How should it handle multiple input shapes then? Should we designate one of the outputs/output shapes as primary, so an ordinary `DenseLayer` or something could be stacked on top, oblivious of the extra inputs?

Do we have any other use cases that can help in design decisions?

f0k:
Thinking again, the second use case could theoretically be solved by designating a separate `InputLayer` for the mask and connecting it to each and every recurrent layer (which would then have to be a `MergeLayer` subclass). That's not terribly convenient, though.

ebenolson:
[...]

PR #522, opened 2015-12-03
Fill the initial hidden state of GRULayer with output from a previous layer

### LINKED ISSUE #462: Recurrent layers: Accept layer for hid_init
The recurrent layers currently have some custom behaviour when `hid_init` is a TensorVariable rather than a shared variable, callable or numpy array: They assume `hid_init` is a tensor of one order higher than otherwise, to include the batch dimension, and they assume `hid_init` is not to be learned.

With #404, parameters can be arbitrary Theano expressions anyway, and overriding that behaviour to assume a different dimensionality is a bit awkward. As discussed in https://github.com/Lasagne/Lasagne/issues/11#issuecomment-139221388 and following, a better solution would be allowing `hid_init` to be a `Layer`, and assuming it would include the batch dimension in this case. This would also be a step towards supporting the encoder-decoder architecture discussed in https://github.com/Lasagne/Lasagne/issues/391#issuecomment-144549176 and following.

almath123:
I have added an optional incoming layer `prefill_h` to the GRULayer so that it can be used as part of an encoder-decoder model. It should not be difficult to do this for the other recurrent layers. Also included is a simple example of its usage.

I have been using this approach in my own work and thought it worth feeding back into the project.

Closes #462.

### DIFF
--- a/lasagne/layers/recurrent.py
+++ b/lasagne/layers/recurrent.py
@@ -64,7 +64,7 @@ from .. import nonlinearities
 from .. import init
 from ..utils import unroll_scan
 
-from .base import MergeLayer
+from .base import MergeLayer, Layer
 from .input import InputLayer
 from .dense import DenseLayer
 from . import helper
@@ -121,7 +121,7 @@ class CustomRecurrentLayer(MergeLayer):
     nonlinearity : callable or None
         Nonlinearity to apply when computing new state (:math:`\sigma`). If
         None is provided, no nonlinearity will be applied.
-    hid_init : callable, np.ndarray, theano.shared or TensorVariable
+    hid_init : callable, np.ndarray, theano.shared, TensorVariable or Layer
         Initializer for initial hidden state (:math:`h_0`).  If a
         TensorVariable (Theano expression) is supplied, it will not be learned
         regardless of the value of `learn_init`.
@@ -210,12 +210,19 @@ class CustomRecurrentLayer(MergeLayer):
                  only_return_final=False,
                  **kwargs):
 
-        # This layer inherits from a MergeLayer, because it can have two
-        # inputs - the layer input, and the mask.  We will just provide the
-        # layer input as incomings, unless a mask input was provided.
+        # This layer inherits from a MergeLayer, because it can have three
+        # inputs - the layer input, the mask and the initial hidden state.  We
+        # will just provide the layer input as incomings, unless a mask input
+        # or initial hidden state was provided.
         incomings = [incoming]
+        self.mask_incoming_index = -1
+        self.hid_init_incoming_index = -1
         if mask_input is not None:
             incomings.append(mask_input)
+            self.mask_incoming_index = len(incomings)-1
+        if isinstance(hid_init, Layer):
+            incomings.append(hid_init)
+            self.hid_init_incoming_index = len(incomings)-1
 
         super(CustomRecurrentLayer, self).__init__(incomings, **kwargs)
 
@@ -314,6 +321,8 @@ class CustomRecurrentLayer(MergeLayer):
                     "When hid_init is provided as a TensorVariable, it should "
                     "have the same shape as hidden_to_hidden.output_shape")
             self.hid_init = hid_init
+        elif isinstance(hid_init, Layer):
+            self.hid_init = hid_init
         else:
             self.hid_init = self.add_param(
                 hid_init, (1,) + hidden_to_hidden.output_shape[1:],
@@ -356,7 +365,10 @@ class CustomRecurrentLayer(MergeLayer):
             part of the sequence or not.  `mask` should be a matrix of shape
             ``(n_batch, n_time_steps)`` where ``mask[i, j] = 1`` when ``j <=
             (length of sequence i)`` and ``mask[i, j] = 0`` when ``j > (length
-            of sequence i)``.
+            of sequence i)``. When the hidden state of this layer is to be
+            pre-filled (i.e. was set to a :class:`Layer` instance) `inputs`
+            should have length at least 2, and `inputs[-1]` is the hidden state
+            to prefill with.
 
         Returns
         -------
@@ -366,7 +378,12 @@ class CustomRecurrentLayer(MergeLayer):
         # Retrieve the layer input
         input = inputs[0]
         # Retrieve the mask when it is supplied
-        mask = inputs[1] if len(inputs) > 1 else None
+        mask = None
+        hid_init = None
+        if self.mask_incoming_index > 0:
+            mask = inputs[self.mask_incoming_index]
+        if self.hid_init_incoming_index > 0:
+            hid_init = inputs[self.hid_init_incoming_index]
 
         # Input should be provided as (n_batch, n_time_steps, n_features)
         # but scan requires the iterable dimension to be first
@@ -433,8 +450,10 @@ class CustomRecurrentLayer(MergeLayer):
             sequences = input
             step_fun = step
 
-        # When hid_init is provided as a TensorVariable, use it as-is
-        if isinstance(self.hid_init, T.TensorVariable):
+        if isinstance(self.hid_init, Layer):
+            pass
+        elif isinstance(self.hid_init, T.TensorVariable):
+            # When hid_init is provided as a TensorVariable, use it as-is
             hid_init = self.hid_init
         else:
             # The code below simply repeats self.hid_init num_batch times in
@@ -516,7 +535,7 @@ class RecurrentLayer(CustomRecurrentLayer):
     nonlinearity : callable or None
         Nonlinearity to apply when computing new state (:math:`\sigma`). If
         None is provided, no nonlinearity will be applied.
-    hid_init : callable, np.ndarray, theano.shared or TensorVariable
+    hid_init : callable, np.ndarray, theano.shared, TensorVariable or Layer
         Initializer for initial hidden state (:math:`h_0`).  If a
         TensorVariable (Theano expression) is supplied, it will not be learned
         regardless of the value of `learn_init`.
@@ -729,11 +748,11 @@ class LSTMLayer(MergeLayer):
     nonlinearity : callable or None
         The nonlinearity that is applied to the output (:math:`\sigma_h`). If
         None is provided, no nonlinearity will be applied.
-    cell_init : callable, np.ndarray, theano.shared or TensorVariable
+    cell_init : callable, np.ndarray, theano.shared, TensorVariable or Layer
         Initializer for initial cell state (:math:`c_0`).  If a
         TensorVariable (Theano expression) is supplied, it will not be learned
         regardless of the value of `learn_init`.
-    hid_init : callable, np.ndarray, theano.shared or TensorVariable
+    hid_init : callable, np.ndarray, theano.shared, TensorVariable or Layer
         Initializer for initial hidden state (:math:`h_0`).  If a
         TensorVariable (Theano expression) is supplied, it will not be learned
         regardless of the value of `learn_init`.
@@ -799,12 +818,24 @@ class LSTMLayer(MergeLayer):
                  only_return_final=False,
                  **kwargs):
 
-        # This layer inherits from a MergeLayer, because it can have two
-        # inputs - the layer input, and the mask.  We will just provide the
-        # layer input as incomings, unless a mask input was provided.
+        # This layer inherits from a MergeLayer, because it can have four
+        # inputs - the layer input, the mask, the initial hidden state and the
+        # inital cell state. We will just provide the layer input as incomings,
+        # unless a mask input, inital hidden state or initial cell state was
+        # provided.
         incomings = [incoming]
+        self.mask_incoming_index = -1
+        self.hid_init_incoming_index = -1
+        self.cell_init_incoming_index = -1
         if mask_input is not None:
             incomings.append(mask_input)
+            self.mask_incoming_index = len(incomings)-1
+        if isinstance(hid_init, Layer):
+            incomings.append(hid_init)
+            self.hid_init_incoming_index = len(incomings)-1
+        if isinstance(cell_init, Layer):
+            incomings.append(cell_init)
+            self.cell_init_incoming_index = len(incomings)-1
 
         # Initialize parent layer
         super(LSTMLayer, self).__init__(incomings, **kwargs)
@@ -884,6 +915,8 @@ class LSTMLayer(MergeLayer):
                     "When cell_init is provided as a TensorVariable, it should"
                     " have 2 dimensions and have shape (num_batch, num_units)")
             self.cell_init = cell_init
+        elif isinstance(cell_init, Layer):
+            self.cell_init = cell_init
         else:
             self.cell_init = self.add_param(
                 cell_init, (1, num_units), name="cell_init",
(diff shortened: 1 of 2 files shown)


skaae:
I think we have a Issue for this: https://github.com/Lasagne/Lasagne/issues/462

It is nearly supported through the `hid_init` argument which needs to be modified to accept input_layers. I think you are more than welcome to solve https://github.com/Lasagne/Lasagne/issues/462 ? @craffel ?

craffel:
Yes, better to solve this as part of #462.

f0k:
+1 for #462 -- however, you're reasonably close to this. We'd be happy if you could adapt this PR to use the proposed interface in #462 instead. That is, instead of a new argument `prefill_h`, allow `hid_init` to be a layer instance.
The encoder-decoder example is great as well and asked for a lot on the mailing list, but I think this might be suited better for [Lasagne/Recipes](https://github.com/Lasagne/Recipes) just because we're trying to keep the `examples/` directory to a bare minimum.

almath123:
Ok, sure. I have now had a look at #462 . I will use `hid_init` rather than `prefill_h` as the layer input and remove the encoder-decoder example.

lasagne/layers/recurrent.py
```diff
@@ -803,8 +819,20 @@ def __init__(self, incoming, num_units,
...
             incomings.append(mask_input)
+            self.mask_incoming_index = len(incomings)-1
+        if isinstance(hid_init, Layer):
+            incomings.append(hid_init)
+            self.hid_init_incoming_index = len(incomings)-1
+            learn_init = False
+        if isinstance(cell_init, Layer):
+            incomings.append(cell_init)
+            self.cell_init_incoming_index = len(incomings)-1
+            learn_init = False
```
f0k: You shouldn't set `learn_init = False` here -- otherwise providing just one of `hid_init` or `cell_init` as a Layer will disable learning of the other. As it's only used within the `self.add_param()` call further below, the value of `learn_init` doesn't matter when the initial state is provided as a Layer anyway. So for consistence, you can also remove the `learn_init = False` assignment for the other recurrent layer classes.

lasagne/layers/recurrent.py
```diff
@@ -1174,6 +1218,10 @@ class GRULayer(MergeLayer):
         Layer which allows for a sequence mask to be input, for when sequences
         are of variable length.  Default `None`, which means no mask will be
         supplied (i.e. all sequences are of the same length).
+    prefill_h : :class:`lasagne.layers.Layer
+        Layer which is used to fill the initial hidden state, used for building
+        encoder-decoder models. Default `None`, which means the initial hidden
+        state is not pre-filled from a previous layer.
```
f0k: Leftover from previous implementation.

lasagne/layers/recurrent.py
```diff
@@ -1304,15 +1361,20 @@ def get_output_for(self, inputs, **kwargs):
...
         inputs : list of theano.TensorType
             `inputs[0]` should always be the symbolic input variable.  When
-            this layer has a mask input (i.e. was instantiated with
-            `mask_input != None`, indicating that the lengths of sequences in
-            each batch vary), `inputs` should have length 2, where `inputs[1]`
-            is the `mask`.  The `mask` should be supplied as a Theano variable
+            this layer has a mask input (i.e. was instantiated with `mask_input
+            != None`, indicating that the lengths of sequences in each batch
+            vary), `inputs` should have length at least 2, where `inputs[1]` is
+            the `mask`.  The `mask` should be supplied as a Theano variable
```
f0k: Please leave it as before, to have a cleaner history (you only changed some line breaks here, no content).

lasagne/layers/recurrent.py
```diff
@@ -1304,15 +1361,20 @@ def get_output_for(self, inputs, **kwargs):
...
             part of the sequence or not.  `mask` should be a matrix of shape
             ``(n_batch, n_time_steps)`` where ``mask[i, j] = 1`` when ``j <=
             (length of sequence i)`` and ``mask[i, j] = 0`` when ``j > (length
-            of sequence i)``.
+            of sequence i)``. When the hidden state of this layer is to be
+            pre-filled (i.e. was instantiated with `isinstance(hid_init,
+            Layer)`) `inputs` should have length at least 2. If an input mask
+            was supplied then `inputs[2]` is the hidden state to prefill with.
+            If no input mask was supplied then `inputs[1] is the hidden state
+            to prefill with.
```
f0k: A bit shorter: 
```
When the hidden state [...] `inputs` should have length at least 2, and `inputs[-1]` is the hidden state to prefill with.
```

lasagne/tests/layers/test_recurrent.py
```diff
@@ -59,6 +59,50 @@ def test_recurrent_nparams_learn_init():
...
+
+def test_recurrent_nparams_hid_init_layer():
+    # test that you can see layers through hid_init
+    l_inp = InputLayer((2, 2, 3))
+    l_inp_h = InputLayer((2, 5))
+    l_inp_h_de = DenseLayer(l_inp_h, 7)
+    l_rec = RecurrentLayer(l_inp, 7, hid_init=l_inp_h_de)
+
+    x = T.tensor3()
+    h = T.matrix()
```
f0k: `x` and `h` are not used here.

lasagne/tests/layers/test_recurrent.py
```diff
@@ -59,6 +59,50 @@ def test_recurrent_nparams_learn_init():
...
+
+    # b, W_hid_to_hid and W_in_to_hid + W + b
+    assert len(lasagne.layers.get_all_params(l_rec, trainable=True)) == 5
```
f0k: I would first add a test directly checking whether you can see the layers through hid_init:
```
assert lasagne.layers.get_all_layers(l_rec) == [l_inp, l_inp_h, l_inp_h_de, l_rec]
```
Your test only checks whether the parameters of `hid_init` itself are included, not whether you can fully "see through" it to the incoming layers of `hid_init`.

lasagne/tests/layers/test_recurrent.py
```diff
@@ -438,6 +482,58 @@ def test_lstm_tensor_init():
...
+    # test that you can see layers through hid_init
+    l_inp = InputLayer((2, 2, 3))
+    l_inp_h = InputLayer((2, 5))
+    l_inp_h_de = DenseLayer(l_inp_h, 7)
+    l_inp_cell = InputLayer((2, 5))
+    l_inp_cell_de = DenseLayer(l_inp_cell, 7)
+    l_lstm = LSTMLayer(l_inp, 7, hid_init=l_inp_h_de, cell_init=l_inp_cell_de)
+
+    x = T.tensor3()
+    h = T.matrix()
```
f0k: Same as above: `x` and `h` are unused, and a `get_all_layers()` test would be good! Counting the parameters is not so important then, actually, it's a rather indirect test.

lasagne/tests/layers/test_recurrent.py
```diff
@@ -704,6 +800,52 @@ def test_gru_init_val_error():
...
+
+def test_gru_nparams_hid_init_layer():
+    # test that you can see layers through hid_init
+    l_inp = InputLayer((2, 2, 3))
+    l_inp_h = InputLayer((2, 5))
+    l_inp_h_de = DenseLayer(l_inp_h, 7)
+    l_gru = GRULayer(l_inp, 7, hid_init=l_inp_h_de)
+
+    x = T.tensor3()
+    h = T.matrix()
```
f0k: (Same as above.)

f0k:
Thank you, looks great except for a few minor comments. Kudos for tackling all three recurrent layers -- I didn't want to ask for it, but it's certainly appreciated! :+1:
In addition to my comments above, the docstrings need to be updated to mention that `hid_init` and `cell_init` can also be `Layer` instances now.

The `self.incoming_mask_index` etc. makes me think that it would be great if the `MergeLayer` also supported dictionaries -- i.e., take a dictionary like `{'input': layer1, `mask`: layer2, `hid_init`: layer3}` instead of a list `[layer1, layer2, layer3]` in the constructor and then provide that dictionary in `get_output_for()`. This would simplify the code a lot. That will need coordination with `lasagne.layers.get_output()` and #337, though.

@skaae: We could get rid of the `hid_init is TensorVariable` special case at the same time -- you don't need it any more, do you? That would remove some `if` clauses and make the docstrings easier. But we can also address this afterwards in a separate PR.

lasagne/layers/recurrent.py
almath123: What I changed here was to specify "length at least 2" rather than "length 2". I have changed it back for now. Though I think it is a necessary addition for clarity. Thoughts?

almath123:
Thanks for the comments! I have had a go at integrating them. If there are more fixes needed for this PR I'm happy to make them.

f0k:
Thanks, looks good to me! Can you squash your commits into one?
```
git reset --soft 7c0f477
git commit --amend -m 'Have recurrent layers accept `hid_init` and `cell_init` to be a Layer'
git push --force
```

skaae:
I missed your commet about tensorvariable. Im fine with removing it. Sorry for being slow

skaae:
@almath: it would be great if you could remove the tensorvariable option. The docstring and a few if statements should be modified and the tests should be removed. Do you want to do this? The pr looks good. Thanks!

f0k:
> @almath: it would be great if you could remove the tensorvariable option.

This could also be done in a follow-up PR. I wouldn't want to squash this in with the other commits anyway, so whatever you decide to do, please first squash what you have so far!

To not forget about it, when removing the explicit TensorVariable case and adapting the docstring (it's still indirectly supported via `add_param()`, but with changed semantics), it would be good to also document which shapes are expected for `hid_init` and `cell_init` (they differ when it's a Layer). But as I said, this can be done in a follow-up PR.

almath123:
I have squashed my commits. I will leave removing the TensorVariable case to another PR, which i'm happy to do though it could be a couple of days before I get round to it. (If anybody else would prefer to do it; go for it).

craffel:
I'll review this tomorrow. Thanks for it!

f0k:
I'm fine with merging, but another pair of eyes is welcome :)

lasagne/layers/recurrent.py
```diff
@@ -214,8 +214,14 @@ def __init__(self, incoming, input_to_hidden, hidden_to_hidden,
         # inputs - the layer input, and the mask.  We will just provide the
```
craffel: This comment needs updating now.

lasagne/layers/recurrent.py
```diff
@@ -356,7 +364,10 @@ def get_output_for(self, inputs, **kwargs):
             part of the sequence or not.  `mask` should be a matrix of shape
             ``(n_batch, n_time_steps)`` where ``mask[i, j] = 1`` when ``j <=
             (length of sequence i)`` and ``mask[i, j] = 0`` when ``j > (length
-            of sequence i)``.
+            of sequence i)``. When the hidden state of this layer is to be
+            pre-filled (i.e. was instantiated with `isinstance(hid_init,
+            Layer)`) `inputs` should have length at least 2, and `inputs[-1]`
```
craffel: This should probably say ` ``isinstance(hid_init, Layer) == True`` `.  Note the double backticks.

craffel:
> I'm fine with merging, but another pair of eyes is welcome :)

Well, this was something I was supposed to do, so I feel like I should at least look at it!

I made some pedantic comments.  It also feels like `incomings` should be a dict instead of a list, because having the semantics of different objects be defined by their absolute position in a list seems unpythonic to me, but that's just me.  It would also replace `self.mask_incoming_index = -1` and `self.mask_incoming_index > 0` with something much more readable, such as `'mask' in self.incomings`.  Also, I would love to see the tests testing actual expected behavior given some synthetic input values (i.e. compiling a Theano function and passing some dummy input).

f0k:
> It also feels like incomings should be a dict instead of a list, because having the semantics of different objects be defined by their absolute position in a list seems unpythonic to me, but that's just me.

No, it's not just you. There are use cases for which a list is appropriate (e.g., the ConcatLayer), and there are others for which a dict is appropriate (e.g., the RecurrentLayer). The easiest solution (in terms of minimal changes needed in Lasagne) would be a second MergeLayer base class that uses dictionaries instead of lists (for `incomings`, `inputs` and `input_shapes`). For the recurrent containers, we'd also like layers to output dictionaries rather than single tensors, though, and we might want to feed this into a MergeLayer as well (i.e., give it a single multiple-output layer rather than multiple single-output layers)... we should think about this. I'll add an issue.

f0k:
@almath123: There are two minor comments by @craffel -- would be cool if you could address them so we can merge! Otherwise I can try to do it as part of the merge. Cheers!

lasagne/layers/recurrent.py
f0k: I think `== True` can be omitted, it doesn't change that expression's value. Double backticks would be correct, yes. Or maybe reformulate as
```
When the hidden state of this layer is to be
pre-filled (i.e., `hid_init` was set to a
:class:`Layer` instance), ...
```

craffel: Well, it's awkward to just write `isinstance(hid_init, Layer)`, because there's no conditional there, it's just an expression.  It's sort of like saying "was instantiated with ``23*100``".  Huh?  I like your word-explanation better than either.

almath123:
@f0k sure I will do it later today

almath123:
I have made the suggested changes to the doc strings  and the comments. 

@craffel Thanks for the review and comments. I wasn't sure what needed to be done for further tests. I thought something like this might work (though I haven't pushed it):

```python
def test_recurrent_hid_init_layer_eval():
    # test hid_init as a Layer with some dummy input and init
    in_shp = (2, 2, 3)
    in_h_shp = (2, 5)
    n_units = 7

    # dummy init layers
    W = np.ones((in_h_shp[-1], n_units), dtype=theano.config.floatX)
    W_in_to_hid = np.ones((in_shp[-1], n_units), dtype=theano.config.floatX)
    W_hid_to_hid = np.ones((n_units, n_units), dtype=theano.config.floatX)

    l_inp = InputLayer(in_shp)
    l_inp_h = InputLayer(in_h_shp)
    l_inp_h_de = DenseLayer(l_inp_h, n_units, W=W, nonlinearity=None)
    l_rec = RecurrentLayer(l_inp, n_units, hid_init=l_inp_h_de,
                           W_in_to_hid=W_in_to_hid, W_hid_to_hid=W_hid_to_hid,
                           nonlinearity=None)

    X = T.tensor3()
    Xh = T.matrix()

    output = lasagne.layers.get_output(l_rec, {l_inp: X, l_inp_h: Xh})

    # test net with dummy input
    X_test = np.ones(in_shp, dtype=theano.config.floatX)
    Xh_test = np.ones(in_h_shp, dtype=theano.config.floatX)
    output_val = output.eval({X: X_test, Xh: Xh_test})

    # check output has the right type
    assert isinstance(output_val, np.ndarray)

    # check output is close to the expected in the first iteration
    H_de = np.dot(Xh_test, W)
    res = np.dot(X_test[:, 0, :], W_in_to_hid) + np.dot(H_de, W_hid_to_hid)
    assert np.allclose(output_val[:, 0, :], res)
```

craffel:
> @craffel Thanks for the review and comments. I wasn't sure what needed to be done for further tests. I thought something like this might work (though I haven't pushed it):

Looks reasonable to me!

almath123:
@f0k Thanks for picking that up. I have updated the comments I missed and squashed my commits. I haven't added that extra testcase. I was thinking of doing a similar test for each of the recurrent layers and adding it as another PR because I don't want to hold this up any longer.

f0k:
Great, let's merge this! Thanks again for your contribution and your patience in addressing our nitpicks.

> I was thinking of doing a similar test for each of the recurrent layers and adding it as another PR

Sounds good. It'd also be good to put the encoder/decoder example somewhere, e.g., on [Lasagne/Recipes](/Lasagne/Recipes). But take your time!

PR #533, opened 2015-12-11
Add RReLu layer

### LINKED ISSUE #530: Add RReLu nonlinearity
Some of you may have seen it, but the paper [Empirical Evaluation of Rectified Activations in Convolutional Network](http://arxiv.org/abs/1505.00853) compares several rectifier variants, and shows that Randomized ReLUs (uniform random a during training) can outperform PReLUs on CIFAR-10 & CIFAR-100 on the test set. RReLus have been [discussued](https://github.com/torch/nn/issues/380), [merged](https://github.com/torch/nn/pull/437), and [documented](https://github.com/torch/nn/blob/master/doc/transfer.md#rrelu) in Torch. That documentation link above also gives a fairly straightforward expression of what RReLus are.

I suspect it shouldn't be too hard to add support for RReLus in Lasagne, but I have little experience with the Lasagne internals. I believe it should simply be a matter of copying [`LeakyRectify`](https://github.com/Lasagne/Lasagne/blob/master/lasagne/nonlinearities.py#L160), replacing the leakiness with min+max, setting up a [Theano shared random stream](http://deeplearning.net/software/theano/tutorial/examples.html#brief-example), and using that to pick the leakiness. I'm not entirely sure how to make the leakiness be `(l+u)/2` during evaluation though...

jonhoo:
Fixes #530.

### DIFF
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -4,7 +4,9 @@ import theano.tensor as T
 from .. import init
 from .. import nonlinearities
 from ..utils import as_tuple
+from ..random import get_rng
 from .base import Layer, MergeLayer
+from theano.sandbox.rng_mrg import MRG_RandomStreams as RandomStreams
 
 
 __all__ = [
@@ -15,6 +17,8 @@ __all__ = [
     "TransformerLayer",
     "ParametricRectifierLayer",
     "prelu",
+    "RandomizedRectifierLayer",
+    "rrelu",
 ]
 
 
@@ -530,3 +534,125 @@ def prelu(layer, **kwargs):
     if nonlinearity is not None:
         layer.nonlinearity = nonlinearities.identity
     return ParametricRectifierLayer(layer, **kwargs)
+
+
+class RandomizedRectifierLayer(Layer):
+    """
+    A layer that applies a randomized leaky rectify nonlinearity to its input.
+
+    The randomized leaky rectifier was first proposed and used in the Kaggle
+    NDSB Competition, and later evaluated in [1]_. Compared to the standard
+    leaky rectifier :func:`leaky_rectify`, it has a randomly sampled slope
+    for negative input during training, and a fixed slope during evaluation.
+
+    Equation for the randomized rectifier linear unit during training:
+    :math:`\\varphi(x) = \\max((\\sim U(lower, upper)) \\cdot x, x)`
+
+    During evaluation, the factor is fixed to the arithmetic mean of `lower`
+    and `upper`.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
+
+    lower : Theano shared variable, expression, or constant
+        The lower bound for the randomly chosen slopes.
+
+    upper : Theano shared variable, expression, or constant
+        The upper bound for the randomly chosen slopes.
+
+    shared_axes : 'auto', 'all', int or tuple of int
+        The axes along which the random slopes of the rectifier units are
+        going to be shared. If ``'auto'`` (the default), share over all axes
+        except for the second - this will share the random slope over the
+        minibatch dimension for dense layers, and additionally over all
+        spatial dimensions for convolutional layers. If ``'all'``, share over
+        all axes, thus using a single random slope.
+
+    **kwargs
+        Any additional keyword arguments are passed to the `Layer` superclass.
+
+     References
+    ----------
+    .. [1] Bing Xu, Naiyan Wang et al. (2015):
+       Empirical Evaluation of Rectified Activations in Convolutional Network,
+       http://arxiv.org/abs/1505.00853
+    """
+    def __init__(self, incoming, lower=0.3, upper=0.8, shared_axes='auto',
+                 **kwargs):
+        super(RandomizedRectifierLayer, self).__init__(incoming, **kwargs)
+        self._srng = RandomStreams(get_rng().randint(1, 2147462579))
+        self.lower = lower
+        self.upper = upper
+
+        if not isinstance(lower > upper, theano.Variable) and lower > upper:
+            raise ValueError("Upper bound for RandomizedRectifierLayer needs "
+                             "to be higher than lower bound.")
+
+        if shared_axes == 'auto':
+            self.shared_axes = (0,) + tuple(range(2, len(self.input_shape)))
+        elif shared_axes == 'all':
+            self.shared_axes = tuple(range(len(self.input_shape)))
+        elif isinstance(shared_axes, int):
+            self.shared_axes = (shared_axes,)
+        else:
+            self.shared_axes = shared_axes
+
+    def get_output_for(self, input, deterministic=False, **kwargs):
+        """
+        Parameters
+        ----------
+        input : tensor
+            output from the previous layer
+        deterministic : bool
+            If true, the arithmetic mean of lower and upper are used for the
+            leaky slope.
+        """
+        if deterministic or self.upper == self.lower:
+            return theano.tensor.nnet.relu(input, (self.upper+self.lower)/2.0)
+        else:
+            shape = list(self.input_shape)
+            if any(s is None for s in shape):
+                shape = list(input.shape)
+            for ax in self.shared_axes:
+                shape[ax] = 1
+
+            rnd = self._srng.uniform(tuple(shape),
+                                     low=self.lower,
+                                     high=self.upper,
+                                     dtype=theano.config.floatX)
+            rnd = theano.tensor.addbroadcast(rnd, *self.shared_axes)
+            return theano.tensor.nnet.relu(input, rnd)
+
+
+def rrelu(layer, **kwargs):
+    """
+    Convenience function to apply randomized rectify to a given layer's output.
+    Will set the layer's nonlinearity to identity if there is one and will
+    apply the randomized rectifier instead.
+
+    Parameters
+    ----------
+    layer: a :class:`Layer` instance
+        The `Layer` instance to apply the randomized rectifier layer to;
+        note that it will be irreversibly modified as specified above
+
+    **kwargs
+        Any additional keyword arguments are passed to the
+        :class:`RandomizedRectifierLayer`
+
+    Examples
+    --------
+    Note that this function modifies an existing layer, like this:
+    >>> from lasagne.layers import InputLayer, DenseLayer, rrelu
+    >>> layer = InputLayer((32, 100))
+    >>> layer = DenseLayer(layer, num_units=200)
+    >>> layer = rrelu(layer)
+
+    In particular, :func:`rrelu` can *not* be passed as a nonlinearity.
+    """
+    nonlinearity = getattr(layer, 'nonlinearity', None)
+    if nonlinearity is not None:
+        layer.nonlinearity = nonlinearities.identity
+    return RandomizedRectifierLayer(layer, **kwargs)
--- a/lasagne/tests/layers/test_special.py
+++ b/lasagne/tests/layers/test_special.py
@@ -446,3 +446,107 @@ class TestParametricRectifierLayer:
         expected = np.maximum(expected, 0) + \
             np.minimum(expected, 0) * alpha_v
         assert np.allclose(output.eval(), expected)
+
+
+class TestRandomizedRectifierLayer:
+    @pytest.fixture
+    def RandomizedRectifierLayer(self):
+        from lasagne.layers.special import RandomizedRectifierLayer
+        return RandomizedRectifierLayer
+
+    def test_high_low(self, RandomizedRectifierLayer):
+        with pytest.raises(ValueError):
+            RandomizedRectifierLayer((None, 3, 28, 28), lower=0.9, upper=0.1)
+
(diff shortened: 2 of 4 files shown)


jonhoo:
Feedback welcome as I'm fairly new to the Lasagne and Theano internals. In particular, I'm not sure I'm handling dimensionality correctly. Also, do you think it is wasteful to generate a random number of the entire shape? We could use the same randomized leakiness across images in a batch and it should have mostly the same effect but with greater performance, right?

f0k:
> Also, do you think it is wasteful to generate a random number of the entire shape? We could use the same randomized leakiness across images in a batch and it should have mostly the same effect but with greater performance, right?

This should be configurable via the `shared_axes` argument, so people can configure it the way they want. You'd do something like:
```python
shape = list(input_shape)
for ax in self.shared_axes:
    shape[ax] = 1
rnd = self._srng.uniform(shape)
rnd = theano.tensor.addbroadcastable(rnd, *self.shared_axes)
return theano.tensor.nnet.relu(input, rnd)
```

You'll also need the `input_shape` distinction as in the `DropoutLayer`, otherwise it will fail if the `InputLayer` shape contains any `None`s.

lasagne/layers/special.py
```diff
@@ -530,3 +534,102 @@ def prelu(layer, **kwargs):
...
+        Any additional keyword arguments are passed to the `Layer` superclass.
+
+     References
+    ----------
+    .. [1] Bing Xu, Naiyan Wang et al. (2015):
+       Empirical Evaluation of Rectified Activations in Convolutional Network,
+       http://arxiv.org/abs/1505.00853
+    """
+    def __init__(self, incoming, lower=init.Constant(0.3),
+                 upper=init.Constant(0.8), shared_axes='auto',
```
f0k: `lower` and `upper` should just be ordinary Python scalars, just like the dropout rate in the `DropoutLayer`. If you change that, you don't need to include the function signature in the docstring any more (it's just needed if the default parameters contain some classes that don't have nice string representations).

lasagne/layers/special.py
```diff
@@ -530,3 +534,102 @@ def prelu(layer, **kwargs):
...
+                 **kwargs):
+        super(RandomizedRectifierLayer, self).__init__(incoming, **kwargs)
+        self._srng = RandomStreams(get_rng().randint(1, 2147462579))
+        self.min = lower
+        self.range = upper-lower
```
f0k: You should better directly store `self.lower` and `self.upper`, that's what users would expect from the constructor (they might want to access the attributes from outside later on). You can compute the range directly in `get_output_for()`.

lasagne/layers/special.py
```diff
@@ -530,3 +534,102 @@ def prelu(layer, **kwargs):
...
+        note that it will be irreversibly modified as specified above
+
+    **kwargs
+        Any additional keyword arguments are passed to the
+        :class:`RandomizedRectifierLayer`
+
+    Examples
+    --------
+    Note that this function modifies an existing layer, like this:
+    >>> from lasagne.layers import InputLayer, DenseLayer, prelu
```
f0k: rrelu, not prelu

lasagne/layers/special.py
```diff
@@ -530,3 +534,102 @@ def prelu(layer, **kwargs):
...
     return ParametricRectifierLayer(layer, **kwargs)
+
+
+class RandomizedRectifierLayer(Layer):
+    """
+    lasagne.layers.RandomizedRectifierLayer(incoming,
+    lower=init.Constant(0.3), upper=init.Constant(0.8),
+    shared_axes='auto', **kwargs)
+
+    The randomized leaky rectifier was first proposed and used in the Kaggle
```
f0k: The first sentence is used for the overview table in http://lasagne.readthedocs.org/en/latest/modules/layers.html, so it should be short and it must not include a reference (because that reference will not be defined where the table is rendered).

lasagne/layers/special.py
```diff
@@ -530,3 +534,102 @@ def prelu(layer, **kwargs):
...
+    NDSB Competition, and later evaluated in [1]_. Compared to the standard
+    leaky rectifier :func:`leaky_rectify`, it has a randomly sampled gradient
+    for negative input during training, and a fixed gradient during evaluation.
```
f0k: gradient -> slope

lasagne/layers/special.py
jonhoo: Why is this not a problem for `ParametricRectifierLayer`?

jonhoo:
Should we check and enforce that `lower < upper` and that they are both in `[0, 1)`?

jonhoo:
I'm also not entirely sure how to best write `test_rrelu`. Looking at `test_prelu`, I'm thinking one way to do it would be to ensure all the negative slopes have been limited to within `x * lower` and `x * upper`, but that doesn't seem very satisfactory. Any ideas?

lasagne/layers/special.py
f0k: Ah, correct... it's not necessarily the full first sentence, but everything up to the first dot or line break. The ParametricRectifierLayer was lucky,

f0k:
> Should we check and enforce that `lower < upper` and that they are both in `[0, 1)`?
And if so, should I raise an exception if this invariant is violated?

`lower < upper` -- yes, and raise a `ValueError` if not. Note that you can only check this if they are not symbolic variables. If `isinstance(lower < upper, theano.Variable)`, then silently skip the test in the constructor.
You could also allow `lower == upper` and fall back to not using any randomization in that case (can be merged with the `if deterministic` check in `get_output_for()`).
About the range: The formulation of the rectifier in Theano allows arbitrary slopes for the negative part, even negative slopes. So I think this shouldn't be constrained.

> I'm also not entirely sure how to best write `test_rrelu`. Looking at `test_prelu`, I'm thinking one way to do it would be to ensure all the negative slopes have been limited to within `x * lower` and `x * upper`, but that doesn't seem very satisfactory.

That's the best you can do, I guess. You could compute the slopes that were used for negative inputs and ensure they are in the correct range and not all the same (or, stricter, have a variance reasonably close to what's expected). In addition, you should have a test with `deterministic = True`. When adding a range check and exception to the constructor, you'll also need a short test for that, Coveralls is not forgiving us anything.

jonhoo:
This latest CI should pass. All the approximate math stuff was unnecessary on my machine, which is why I ended up pushing multiple non-`[skip ci]` commits. 

`TestRandomizedRectifierLayer::test_get_output_for` is now somewhat slow due to 4a1d989d530b3999c148b753719374b59b0ef3f8, but I'm not quite sure how to do the appropriate test as a matrix operation. Input welcome.

lasagne/tests/layers/test_special.py
```diff
@@ -446,3 +446,107 @@ def test_prelu(self, init_alpha):
...
+            else:
+                assert y[i] <= layer.lower * x[i] and y[i] > layer.upper * x[i]
+
+    def test_get_output_for(self, RandomizedRectifierLayer):
+        input_shape = (3, 3, 28, 28)
+
+        # ensure slope never exceeds [lower,upper)
+        input = np.random.randn(*input_shape).astype(theano.config.floatX)
+        layer = RandomizedRectifierLayer(input_shape, shared_axes=0)
+        self.assert_between(layer, input, layer.get_output_for(input).eval())
```
f0k: I'd directly compute and check the slopes:
```python
slopes = layer.get_output_for(input).eval() / input
slopes = slopes[input < 0]
assert slopes.min() >= layer.lower
assert slopes.max() <= layer.upper
```

lasagne/tests/layers/test_special.py
```diff
@@ -446,3 +446,107 @@ def test_prelu(self, init_alpha):
...
+        # from here on, we want to check parameter sharing
+        # this is easier to check if the input is all ones
+        input = np.ones(input_shape).astype(theano.config.floatX) * -1
+
+        # default: parameters shared along all but 2nd axis
+        layer = RandomizedRectifierLayer(input_shape)
+        out = layer.get_output_for(input).eval()
+        assert np.allclose(out[0].flatten(), out[-1].flatten())
+        assert not np.allclose(out[0][0].flatten(), out[0][-1].flatten())
+        assert np.allclose(out[0][0][0], out[0][0][-1])
```
f0k: I think it's easier to read if you just check the variances:
```python
assert [out.var(axis=a) == 0 for a in range(4)] == [False, True, False, False]
```
You can use this scheme for all the tests. The variance should be zero across all axes the randomization is shared over.

lasagne/tests/layers/test_special.py
```diff
@@ -446,3 +446,107 @@ def test_prelu(self, init_alpha):
...
+        # share across 1st and 4th axes
+        layer = RandomizedRectifierLayer(input_shape, shared_axes=(0, 3))
+        out = layer.get_output_for(input).eval()
+        assert np.allclose(out[0].flatten(), out[-1].flatten())
+        assert not np.allclose(out[0][0].flatten(), out[0][-1].flatten())
+        assert not np.allclose(out[0][0][0].flatten(), out[0][0][-1].flatten())
+        assert np.allclose(out[0][0][0], out[-1][0][0])
+
+    def test_rrelu(self):
+        import lasagne
```
f0k: Not sure it matters, but we usually just import what's needed in the tests:
```python
from lasagne.layers import InputLayer, DenseLayer, rrelu
from lasagne.nonlinearities import identity
```
(Haven't checked whether the ParametricRectifyLayer does that. It's probably not super important, but this way it clearly states which components are involved in a particular test.)

lasagne/layers/special.py
```diff
@@ -14,7 +16,9 @@
     "InverseLayer",
     "TransformerLayer",
     "ParametricRectifierLayer",
+    "RandomizedRectifierLayer",
     "prelu",
+    "rrelu",
```
f0k: Minor thing, but please move `RandomizedRectifierLayer` one line down so it matches the order in the source code. Thanks!

lasagne/layers/special.py
```diff
@@ -530,3 +534,125 @@ def prelu(layer, **kwargs):
...
+        The layer feeding into this layer, or the expected input shape
+
+    lower : Theano shared variable, expression, or constant
+        The lower bound for the randomly chosen leaky slope.
+
+    upper : Theano shared variable, expression, or constant
+        The upper bound for the randomly chosen leaky slope.
+
+    shared_axes : 'auto', 'all', int or tuple of int
+        The axes along which the parameters of the rectifier units are
```
f0k: For clarity, I'd suggest `random slopes` instead of `parameters`.

lasagne/tests/layers/test_special.py
jonhoo: Ah, yes, that's clever. Will do that.

lasagne/layers/special.py
```diff
@@ -530,3 +534,125 @@ def prelu(layer, **kwargs):
...
+    and `upper`.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
+
+    lower : Theano shared variable, expression, or constant
+        The lower bound for the randomly chosen leaky slope.
+
```
f0k: (Ooops, sorry, seems I forgot to send this one.) I'd prefer just `slope`, I don't think `leaky slope` is a well-defined term. Actually, I'd even prefer `slopes`, because generally there will be multiple ones.

f0k:
Great so far, almost there! What's still missing in the end is adding it to the docs: You'll need to add it to `docs/modules/layers.rst` and `docs/modules/layers/special.rst`, using the PReLU entries as a template (it's pretty simple).

jonhoo:
Would you want me to do some commit squashing before merging, or are you happy to merge with all the minor commits in there too?

lasagne/layers/special.py
```diff
@@ -530,3 +534,125 @@ def prelu(layer, **kwargs):
...
+    upper : Theano shared variable, expression, or constant
+        The upper bound for the randomly chosen slopes.
+
+    shared_axes : 'auto', 'all', int or tuple of int
+        The axes along which the random slopes of the rectifier units are
+        going to be shared. If ``'auto'`` (the default), share over all axes
+        except for the second - this will share the random slope over the
+        minibatch dimension for dense layers, and additionally over all
+        spatial dimensions for convolutional layers. If ``'all'``, share over
+        all axes, which corresponds to a single scalar parameter.
```
f0k: "a single random slope.", or at least something without "parameter". Sorry, didn't check this one again before.

f0k:
Just one more nitpick!

> Would you want me to do some commit squashing before merging

Yes, I'd prefer to have everything in a single commit. There's no intermediate step worth preserving for the future, the commits were mostly part of the reviewing process. You can fix the nitpick and then squash before (force-)pushing. Thanks a lot!

About the "skip ci": If you push multiple commits at once, Travis will run for the last commit only, so there's no need to mark minor commits if you don't push them separately (but it doesn't hurt either).

jonhoo:
Or just `git rebase -i origin/master`, `s/pick/fixup/`. Anyway, done and pushed.

f0k:
I should try interactive rebase some time as well.

Thanks a lot for your contribution, merging!

benanne:
This looks great, thanks a lot!

PR #540, opened 2015-12-15
Adamax updates

benanne:
This is an implementation of Adamax, a variant of Adam described in the latest version of the paper ( http://arxiv.org/abs/1412.6980 ) which uses the inifinity norm instead of the L2 norm. This is argued to be more stable in the case of sparse updates (e.g. for embeddings).

I have not included tests yet since I noticed that the existing updates implementations are all tested against a Torch implementation. @ebenolson, I believe you set this up, so it would be great if you could chime in :) We may want to include a separate set of tests that do not depend on Torch, so that it's easier to add new update rules.

### DIFF
--- a/lasagne/updates.py
+++ b/lasagne/updates.py
@@ -77,6 +77,7 @@ __all__ = [
     "rmsprop",
     "adadelta",
     "adam",
+    "adamax",
     "norm_constraint",
     "total_norm_constraint"
 ]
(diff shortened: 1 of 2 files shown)


ebenolson:
Luckily Torch `optim` has implemented adamax already, so it's an easy addition. 

I've updated the [gist](https://gist.github.com/ebenolson/931e879ed38f257253d2), the calculated values (for `learning_rate=0.01`) are `[0.90211749000754,0.90211748762402,0.90211748682951]`

benanne:
That's great, thanks! It'll do for this PR, but in the long term we might want to add more update rules that are not necessarily in Torch (or you might not be available to run the script). So I still think we should figure out a way to not make this a hard dependency.

ebenolson:
Well, the current test doesn't actually require Torch or the script, just calculating 3 values for each optimizer - and the problem is simple enough that it could probably even be done by hand if needed. 

Alternatively someone could write numpy versions of each, but that seems like wasted effort, and probably still more prone to errors than comparing to a fully independent implementation.

lasagne/updates.py
```diff
@@ -598,6 +598,65 @@ def adam(loss_or_grads, params, learning_rate=0.001, beta1=0.9,
     return updates
 
 
+def adamax(loss_or_grads, params, learning_rate=0.002, beta1=0.9,
+         beta2=0.999, epsilon=1e-8):
```
f0k: pep8: two more blanks needed

nouiz: Maybe the algo is already in downhill that use Theano. This could be used
in some cases:

http://downhill.readthedocs.org/en/stable/

benanne:
> Alternatively someone could write numpy versions of each, but that seems like wasted effort, and probably still more prone to errors than comparing to a fully independent implementation.

I think it would be a useful addition. The current setup might put people off contributing, because they cannot be bothered to install Torch and run the script (and then that only works when the algorithm in question is already implemented in Torch to begin with). It's great to have these tests, it would just be nice to have something a little more self-contained in addition to that.

benanne:
I added a test and fixed the PEP8 issue. I'll make a separate issue about more self-contained tests for the updates functions :)

f0k:
Looks good, can you please squash the two commits? Cheers!

benanne:
done!

f0k:
Merging, thanks again!

Issue #27, opened 2014-09-13
Conv3DLayer implementation


benanne:
I'd like to have `Conv1DLayer`, `Conv2DLayer` and `Conv3DLayer` available. Unfortunately Theano's support for 3D convolutions is a bit of a mess imho, there are some different implementations (see [documentation](http://deeplearning.net/software/theano/library/tensor/nnet/conv.html) for details):
- `theano.tensor.nnet.Conv3D.conv3D`, which I guess is the default. Its expected input shape configuration is `b01tc` (b = batch, 0 = width, 1 = height, t = time, c = channels, note that 0, 1 and t are basically interchangeable), which is sort of unusual because the channel dimension is last. In practice this is apparently CPU only because the GPU implementation is slow (the docs even encourage using `conv3d2d` on GPU instead).
- `theano.tensor.nnet.conv3d2d.conv3d` has `btc01` as its input shape configuration. Also kind of weird to have the channel dimension in the middle there. This works by turning the 3D convolution into a 2D convolution, and given the recent improvements for the 2D convolution, this is probably the most useful implementation.
- `theano.sandbox.cuda.fftconv.conv3d_fft` has `bct01` as its input shape configuration, which is the most natural way to do things imho (since `conv2d` uses `bc01`). This is a 3D version of `conv2d_fft`. I don't know how practical this is because FFT convolutions use a lot of memory. There is no gradient implemented
- There is an optimization that swaps `theano.tensor.nnet.Conv3D.conv3D` for `theano.sandbox.cuda.fftconv.conv3d_fft`. It must be enabled manually.

Since `conv3d_fft` has no gradient implemented, there is no point in supporting it directly. Instead we should support `Conv3D.conv3D`.

I think we should also support `conv3d2d.conv3d`. Afaik there is no optimization to replace  `Conv3D.conv3D` by `conv3d2d.conv3d`. So that means the `Conv3DLayer` class needs a switch to choose the implementation. This can just be an `implementation` keyword argument on the constructor (which is how I've tackled this issue in the past in layers.py).

In general I think having an `implementation` keyword for all of the `Conv*DLayers` is a good idea, there are a lot of different ways to do things and although there are sometimes optimizations that can be enabled manually, those aren't always available. And for the 1D convolution there is currently no implementation at all (but I have a bunch of code for that already).

[...]

PR #541, opened 2015-12-16
Revamp Convolutional Layers

### LINKED ISSUE #524: Revamp Convolutional Layers
We should refactor/clean up the different convolutional layers in Lasagne:

* [x] Refactor the existing code to have a common base class for all convolution dimensionalities and all implementations (as discussed in #498)
* [x] Reconsider the interface(s) of the convolutional layers in light of the new `conv2d()` interface in Theano -- can `Conv2DLayer` support more than it does now, and should we drop some of the implementation-specific layers?
* [x] Add a `Deconv2DLayer`. Theano provides a placeholder Op for the gradient wrt. input now, so we can provide a more direct route than via the `InverseLayer`.
* [x] Document that the implementation-specific layers not only live outside the `lasagne.layers` namespace (I think this is probably clear from the way it's documented), but also that their modules are not imported by default. (#535)

(Side issue: At least the first step would be interesting for pooling as well.)

f0k:
Starting to address #524. First two commits:
* Add a `BaseConvLayer` class that has all the common code and derive all other convolutional layer implementations from that. Was pretty easy except for all the special-casing in `Conv2DCCLayer`: It has a `dimshuffle` argument that optionally makes it operate directly on data in c01b layout, and it stored `self.filter_size`, `self.pad` etc. as a single integer despite being a 2D layer. I had to change the latter because I couldn't call the super constructor otherwise, also I think it's better to be consistent with other layers.
* Added a check to `BaseConvLayer` to test whether the input dimensionality fits the convolution dimensionality. There were some reports on the mailing list about users getting this wrong and getting difficult-to-interpret error messages.

Next step: Make use of the new conv2d interface in Theano -- specifically, get rid of the custom padding code in Conv2DLayer and Conv1DLayer, rename `image_shape` to `input_shape`, and move the `filter_flip` argument to the base class and add it to Conv2DLayer (adding it to Conv1DLayer is too much work for now, this would require changing all the underlying provided convolution implementations).

### DIFF
--- a/lasagne/__init__.py
+++ b/lasagne/__init__.py
@@ -13,7 +13,7 @@ http://lasagne.readthedocs.org/en/latest/user/installation.html#theano"""
 except ImportError:  # pragma: no cover
     raise ImportError("Could not import Theano." + install_instr)
 else:
-    if not hasattr(theano.tensor.nnet, 'h_softmax'):  # pragma: no cover
+    if not hasattr(theano.tensor.nnet, 'abstract_conv'):  # pragma: no cover
         raise ImportError("Your Theano version is too old." + install_instr)
     del install_instr
     del theano
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -80,7 +80,202 @@ def conv_output_length(input_length, filter_size, stride, pad=0):
     return output_length
 
 
-class Conv1DLayer(Layer):
+class BaseConvLayer(Layer):
+    """
+    lasagne.layers.BaseConvLayer(incoming, num_filters, filter_size,
+    stride=1, pad=0, untie_biases=False,
+    W=lasagne.init.GlorotUniform(), b=lasagne.init.Constant(0.),
+    nonlinearity=lasagne.nonlinearities.rectify,
+    n=None, **kwargs)
+
+    Convolutional layer base class
+
+    Base class for performing an `n`-dimensional convolution on its input,
+    optionally adding a bias and applying an elementwise nonlinearity. Note
+    that this class cannot be used in a Lasagne network, only its subclasses
+    can (e.g., :class:`Conv1DLayer`, :class:`Conv2DLayer`).
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape. Must
+        be a tensor of 2+`n` dimensions:
+        ``(batch_size, num_input_channels, <n spatial dimensions>)``.
+
+    num_filters : int
+        The number of learnable convolutional filters this layer has.
+
+    filter_size : int or iterable of int
+        An integer or an `n`-element tuple specifying the size of the filters.
+
+    stride : int or iterable of int
+        An integer or an `n`-element tuple specifying the stride of the
+        convolution operation.
+
+    pad : int, iterable of int, 'full', 'same' or 'valid' (default: 0)
+        By default, the convolution is only computed where the input and the
+        filter fully overlap (a valid convolution). When ``stride=1``, this
+        yields an output that is smaller than the input by ``filter_size - 1``.
+        The `pad` argument allows you to implicitly pad the input with zeros,
+        extending the output size.
+
+        A single integer results in symmetric zero-padding of the given size on
+        all borders, a tuple of `n` integers allows different symmetric padding
+        per dimension.
+
+        ``'full'`` pads with one less than the filter size on both sides. This
+        is equivalent to computing the convolution wherever the input and the
+        filter overlap by at least one position.
+
+        ``'same'`` pads with half the filter size (rounded down) on both sides.
+        When ``stride=1`` this results in an output size equal to the input
+        size. Even filter size is not supported.
+
+        ``'valid'`` is an alias for ``0`` (no padding / a valid convolution).
+
+        Note that ``'full'`` and ``'same'`` can be faster than equivalent
+        integer values due to optimizations by Theano.
+
+    untie_biases : bool (default: False)
+        If ``False``, the layer will have a bias parameter for each channel,
+        which is shared across all positions in this channel. As a result, the
+        `b` attribute will be a vector (1D).
+
+        If ``True``, the layer will have separate bias parameters for each
+        position in each channel. As a result, the `b` attribute will be an
+        `n`-dimensional tensor.
+
+    W : Theano shared variable, expression, numpy array or callable
+        Initial value, expression or initializer for the weights.
+        These should be a tensor of 2+`n` dimensions with shape
+        ``(num_filters, num_input_channels, <n spatial dimensions>)``.
+        See :func:`lasagne.utils.create_param` for more information.
+
+    b : Theano shared variable, expression, numpy array, callable or ``None``
+        Initial value, expression or initializer for the biases. If set to
+        ``None``, the layer will have no biases. Otherwise, biases should be
+        a 1D array with shape ``(num_filters,)`` if `untied_biases` is set to
+        ``False``. If it is set to ``True``, its shape should be
+        ``(num_filters, <n spatial dimensions>)`` instead.
+        See :func:`lasagne.utils.create_param` for more information.
+
+    nonlinearity : callable or None
+        The nonlinearity that is applied to the layer activations. If None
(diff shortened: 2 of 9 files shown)


f0k:
Added a third commit that addresses what I mentioned as "next step" above.

This required some changes to the conv1d implementations in `lasagne.theano_extensions` (and accordingly more tests), so they're compatible with the new `conv2d` interface (specifically, support custom padding and filter flipping).
I also changed the tests to include test cases with custom padding for all conv2d tests (before this was only done for conv1d, and there was a single simplified padding test case per conv2d layer that just checked whether the output shape was correct -- probably because custom padding was unsupported in Conv2DLayer when the tests were written).
To reduce the number of test cases again, `untie_biases` is now a single test case rather than a parameter that's used with both settings for every test case.
Finally, now that all conv2d layers support`flip_filters`, I got rid of the `extra_kwargs` technique that modified all conv2d layers defaulting to `flip_filters=False` to use `flip_filters=True` in the tests. Instead, *all* conv2d test cases are now run with `flip_filters=True` except for a single extra test case trying `flip_filters=False`. This should get us covered without too much overhead.
We could probably drastically reduce the number of conv2d test cases, by the way, there are a lot of them right now:
```python
    for pad in (0, 1, 2, 'full', 'same'):
        for stride in (1, 2, 3):
            for filter_size in (1, 3):
```
We probably don't need a grid search like this and could just use a hand-picked selection of parameter combinations instead.

f0k:
Hmm, Travis fails because the MNIST example doesn't finish within 10 minutes. Strange, will try to run it locally.

f0k:
It runs fine locally. I deleted the Travis caches and ran again (to reinstall Theano from scratch), but it still doesn't get through the MNIST example. It will need to recompile some parts, but it shouldn't take *that* long. No idea what's going on here.

benanne:
Nice work so far! I'll try to go through the code this weekend.

f0k:
> I'll try to go through the code this weekend.

Cool! It will help to go through it commit by commit -- each commit groups a set of related changes passing the tests. The third one is the toughest due to the conv1d changes.

f0k:
It's a pain. There are two issues:

1. The MNIST example suddenly is terribly slow on Travis, between 7 and 15 seconds per batch (of which there are 100, so it's up to 25 minutes in total just for a single MNIST epoch). I've investigated this by disabling stdout capturing of pytest and adding some more print statements, then mitigated this by enabling the examples to run for 0.1 epochs only.
2. `CorrMM` (on CPU) doesn't work any longer, it fails with: `ImportError: ('The following error happened while compiling the node', CorrMM{valid, (1, 1)}(InplaceDimShuffle{0,x,1,2}.0, Subtensor{::, ::, ::int64, ::int64}.0), '\n', '/home/travis/.theano/compiledir_Linux-3.13--generic-x86_64-with-debian-wheezy-sid-x86_64-2.7.9-64/tmp6qWZLL/630e56d53a13c31e8be0eafd6612c44d.so: undefined symbol: dgemm_', '[CorrMM{valid, (1, 1)}(<TensorType(float64, (False, True, False, False))>, <TensorType(float64, (False, True, False, False))>)]')`. I've tried fixing this by removing Theano's compile dir before running the tests (after all, this PR also includes an update of Theano so maybe the cached compile dir doesn't play nicely with it), but it doesn't help.

These two issues could even be related. Maybe something's wrong with BLAS. It's cumbersome to investigate on a remote machine without shell access.

f0k:
> These two issues could even be related.

Seems they are indeed, judging from [the documentation of THEANO_FLAGS=blas.ldflags](http://deeplearning.net/software/theano/library/config.html#config.config.blas.ldflags):
>    Default: ‘-lblas’
>
>    Link arguments to link against a (Fortran) level-3 blas implementation. The default will test if ‘-lblas’ work. If not, we will disable our c code for BLAS.

It seems this is exactly what's happening. The C Code for BLAS is disabled because `-lblas` doesn't work, so the MNIST example becomes extremely slow, and then CorrMM fails because in contrast to gemm, it doesn't check whether BLAS is actually available. This must be caused by a change in Theano between Theano/Theano@a85a44fc and Theano/Theano@24ace594. Unfortunately I don't see a way to do a git-bisect on Travis.

f0k:
@nouiz: It seems https://github.com/Theano/Theano/pull/3654 broke BLAS detection for our Travis test suite. Can you shed some light on what changed, or how we can further debug this? It seems that the former `GCC_compiler.try_flags(["-lblas"])` works fine in the setup we have on Travis, but the new way doesn't.
The system running at Travis is debian-based and we install `libblas-dev`, `liblapack-dev` and `gfortran` before running the tests. It worked fine with a Theano version from November 12, but not with a version from December 15, and I'm quite positive https://github.com/Theano/Theano/pull/3654 is the culprit.
Thanks!

nouiz:
Can you print the value of res there:

https://github.com/Theano/Theano/pull/3654/files#diff-e0a2d5b52d2b6d9bdd9f87220b300c00R346

It would help identify what the problem is, compilation, linker or
execution.

f0k:
> Can you print the value of res there:

That's a bit difficult, I'd need to set up the modified file somewhere Travis can pull it from. But I've found something crazy in that commit and sent a PR: https://github.com/Theano/Theano/issues/3802

f0k:
Yes, finally! @nouiz, my Theano PR fixes the problem at Travis -- please merge as soon as you've got a chance. That gave me a lot of headaches. Thank you!

f0k:
My fix has been merged upstream, I've updated the requirements accordingly (did this in an extra commit because it was independent of the requirement needed for abstract_conv2d, could also squash this with the previous commit if anybody prefers that). Good to merge from my side, ready to be reviewed!

benanne:
Thanks for the tedious detective work :) Didn't get around to looking at this over the weekend, but I might be able to do it now!

benanne:
Went through this commit by commit, so some of my comments on the older commits may already be irrelevant :) Sorry about that!

Since my comments were on individual commits, it looks like they don't show up here either... but I guess you know where to find them!

We should probably make an issue for `Conv3DLayer`. This refactor should make it really easy to implement one or more 3D convolution layers.

f0k:
> Went through this commit by commit, so some of my comments on the older commits may already be irrelevant :)

Great, thank you! They shouldn't be irrelevant, the commits are supposed to be non-destructive, so to say (they don't reflect a history of patches to fix things done wrongly in another commit, they just group together the changes in easily digestible chunks).

> We should probably make an issue for Conv3DLayer.

What about #27? :)

> This refactor should make it really easy to implement one or more 3D convolution layers in addition to the cuDNN-based one.

For a general `Conv3DLayer` we'll first have to add an `abstract_conv3d()` to Theano that's as complete as `abstract_conv2d()`. This could mean refactoring the Ops to support 2D and 3D at the same time and then just providing separate interfaces. It would be possible to add layers for the other specific implementations (there's a gemm-based version, for example), but I don't know if it's worth it. Until there are users asking for a 3D convolution that do not have access to cuDNN I assume they don't exist ;)

f0k:
Just figured that we will lose your comments when I edit the first commit now. So let's decide on the [marker classes](https://github.com/f0k/Lasagne/commit/f37d25936167dbaef3b41145fdea09be1515664b#commitcomment-15108328) first: Shall we drop `CCLayer`, `DNNLayer` and the like or keep them for backwards compatibility? And shall we rename the `convolution` constructor argument to `Conv1DLayer` and `Conv2DLayer` or keep them for backwards compatibility? I'm leaning towards keeping everything compatible in this PR and using another PR for breaking compatibility if we want that.

benanne:
> What about #27? :)

Right, let's give that a bump then.

Re: backward compatibility: let's get rid of the marker classes. They were never officially part of the API anyway, so I think we can do what we want with them. It seems exceedingly unlikely that someone is using them for type checking, but if there is, it should be an easy fix (and an easy error message to figure out).

Re: `convolution` argument, that *is* part of the API, so we should probably keep that as is. If we don't we should add deprecation warnings, and maybe that just isn't worth it.

f0k:
So, finally pushed what I did in the train home for christmas! Good to merge, I think. The marker classes are gone and your other comments have been addressed as well (extra test for `NotImplementedError` in the convolutional base class).

benanne:
Excellent work! Merging :)

PR #549, opened 2015-12-23
Remove TensorVariable Init

skaae:
Removes TensorVariable special case as discussed here: https://github.com/Lasagne/Lasagne/issues/462

* Removes TensorVariable Init.
* Removes related Tests

### DIFF
--- a/lasagne/layers/recurrent.py
+++ b/lasagne/layers/recurrent.py
@@ -121,18 +121,14 @@ class CustomRecurrentLayer(MergeLayer):
     nonlinearity : callable or None
         Nonlinearity to apply when computing new state (:math:`\sigma`). If
         None is provided, no nonlinearity will be applied.
-    hid_init : callable, np.ndarray, theano.shared, TensorVariable or Layer
-        Initializer for initial hidden state (:math:`h_0`).  If a
-        TensorVariable (Theano expression) is supplied, it will not be learned
-        regardless of the value of `learn_init`.
+    hid_init : callable, np.ndarray, theano.shared or :class:`Layer`
+        Initializer for initial hidden state (:math:`h_0`).
     backwards : bool
         If True, process the sequence backwards and then reverse the
         output again such that the output from the layer is always
         from :math:`x_1` to :math:`x_n`.
     learn_init : bool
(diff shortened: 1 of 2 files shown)


skaae:
I noticed that `hid_init` does not have a reference to `lasagne.layers.Layer` .
`:class:lasagne.layers.Layer` does not fit the linewidth but :class:Layer` does.

Should I change add `:class:Layer` and maybe also change the rest of the recurrent docs to use
`:class:Layer` because its more consistent with the rest of the docs?

f0k:
> Removes TensorVariable special case

Looks good!

> Should I change add :class:Layer and maybe also change the rest of the recurrent docs to use
:class:Layer because its more consistent with the rest of the docs?

Yes, would be good to change that! :class:Layer is enough because we tell Sphinx that we're in the lasagne.layers namespace. The docstring change should be in a separate commit this time, i.e., not squashed into the first one. Let me know when you're done so we can merge! (github doesn't send out a notification for additional commits to a PR, only for additional comments.)

skaae:
I added references to the layer class in the doc string and squashed the commits.

lasagne/layers/recurrent.py
```diff
@@ -748,22 +730,16 @@ class LSTMLayer(MergeLayer):
...
         None is provided, no nonlinearity will be applied.
-    cell_init : callable, np.ndarray, theano.shared, TensorVariable or Layer
-        Initializer for initial cell state (:math:`c_0`).  If a
-        TensorVariable (Theano expression) is supplied, it will not be learned
-        regardless of the value of `learn_init`.
-    hid_init : callable, np.ndarray, theano.shared, TensorVariable or Layer
-        Initializer for initial hidden state (:math:`h_0`).  If a
-        TensorVariable (Theano expression) is supplied, it will not be learned
-        regardless of the value of `learn_init`.
+    cell_init : callable, np.ndarray, theano.shared or Layer
```
f0k: you missed that one!

f0k:
One `Layer` reference missed, other than that, it's good to merge. Can you please fix that (and squash)?

skaae:
Sorry about that. I fixed it.

f0k:
Thanks, merging!

Issue #551, opened 2015-12-23
The line 119 in Lasagne/lasagne/layers/merge.py has a bug


ciaua:
Line 119:
`slices_by_input = [[] for i in range(ndim)]`

Should be:
`slices_by_input = [[] for i in range(len(inputs))]`

ndim is the dimension of the tensors.
The ndim should be replaced with len(inputs) because each element in the list is corresponding to an input.

benanne:
Have you run into problems because of this? I don't fully understand the code, but to me it seems equally likely that the comment above it is wrong :) Either way we need to fix it. Thanks for the report!

ciaua:
Yes, I got a DisconnectedInputError because some inputs are discarded due to this issue. 

For example, it happens when I concatenate a list of 5 inputs of tensor dimension 4. At the line 159 after doing the zip(inputs, slices_by_input), the length of the inputs is reduced to 4 while it should be 5.

After I correct line 119 in the source code, it is all good now.

benanne:
In that case it's weird that this hasn't come up before! I guess people usually have fewer inputs than dimensions. Anyway, if you want to do a PR that would be very welcome :) If not then I'm sure someone else will take care of it soon. We'll definitely want to add a test for this case (more inputs than dimensions) as well so the bug is never reintroduced.

skaae:
I couldn't figure out how to checkout this PR. I created a new pr that adds a test https://github.com/Lasagne/Lasagne/pull/555

PR #552, opened 2015-12-23
Correct slices_by_input

ciaua:
As discussed in issue #551, the ndim should be replaced with len(inputs) in line 119 in lasagne/layers/merge.py

This branch did:
1. Replace ndim with len(inputs) in line 119
2. Add an assert to check the arguments in zip() have the same length at line 157

### DIFF
--- a/lasagne/layers/merge.py
+++ b/lasagne/layers/merge.py
@@ -116,7 +116,7 @@ def autocrop(inputs, cropping):
 
         # Nested list of slices; each list in `slices` corresponds to
         # an input and contains a slice for each dimension
-        slices_by_input = [[] for i in range(ndim)]
+        slices_by_input = [[] for i in range(len(inputs))]
 
         # If there are more dimensions than cropping entries, pad
         # the cropping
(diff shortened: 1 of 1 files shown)


benanne:
Thanks! I'm a bit uncomfortable with having assert statements in the code though, I think it would be more appropriate to add one or more test cases here: https://github.com/Lasagne/Lasagne/blob/fa4b16852adb39c2879cb9f3dbaf48864a8c6934/lasagne/tests/layers/test_merge.py

It'd be great if you could move it there, but if you don't have time to figure this out just let us know and I'm sure someone else will take care of it!

ciaua:
Sorry I don't have time to figure it out right know.

benanne:
I just realised we already have an outstanding PR for this, also with a missing test: #517
Anyone reckon they could add it anytime soon? Looks like a few people are running into this issue so we should fix it sooner rather than later.

skaae:
I can fix it tomorrow if no one else volunteers :)

PR #554, opened 2015-12-24
Replace masked code in all recurrent layers by T.switch command

### LINKED ISSUE #550: Use theano.tensor.switch in LSTM Layer in masked step
I'm curious why theano.tensor.switch isn't used in the following place: https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/recurrent.py#L1077

not_mask = 1 - mask_n
cell = cell*mask_n + cell_previous*not_mask
hid = hid*mask_n + hid_previous*not_mask

we can rewrite this code as following:

cell = T.switch(mask_n, cell, cell_previous)
hid = T.switch(mask_n, hid, hid_previous)

How I see it runs with the same speed but anyway why you don't use 'switch'?

avostryakov:
Replace masked code in all recurrent layers by T.switch command

Closes #550.

### DIFF
--- a/lasagne/layers/recurrent.py
+++ b/lasagne/layers/recurrent.py
@@ -439,7 +439,7 @@ class CustomRecurrentLayer(MergeLayer):
             # Skip over any input with mask 0 by copying the previous
             # hidden state; proceed normally for any input with mask 1.
             hid = step(input_n, hid_previous, *args)
-            hid_out = hid*mask_n + hid_previous*(1 - mask_n)
+            hid_out = T.switch(mask_n, hid, hid_previous)
             return [hid_out]
 
         if mask is not None:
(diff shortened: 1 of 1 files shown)


skaae:
It looks good to me. Thanks.

benanne:
Looks good to me as well. I'll let you do the honours :)

f0k:
Looks good, it's cleaner and shouldn't hurt (the previous implementation allows to do some tricks with non-binary masks, but probably nobody did that).
@craffel: Was there a specific reason for not using `T.switch()`? Did you observe that multiplication to be faster (contrary to the cursory test in #550)?

skaae:
I think i did the masking implementation. I don't think there was a particular reason for not using `T.switch`

craffel:
I think there was some reference implementation somewhere from the Theano folks who used the multiplication trick.  Great that switch is faster, thanks for this.

f0k:
All right, merging then! Thanks again!

PR #555, opened 2015-12-25
replace ndim with len(inputs) in line 119 + test

### LINKED ISSUE #551: The line 119 in Lasagne/lasagne/layers/merge.py has a bug

skaae:
Adds a test to PR https://github.com/Lasagne/Lasagne/pull/552. PR https://github.com/Lasagne/Lasagne/pull/552 fixes issue https://github.com/Lasagne/Lasagne/issues/551.

Fixes #551, closes #552, closes #517.

1) replace ndim with len(inputs) in line 119
2) test that the number of outputs is correct when the number of inputs
is larger than ndim of the inputs.

### DIFF
--- a/lasagne/tests/layers/test_merge.py
+++ b/lasagne/tests/layers/test_merge.py
@@ -93,6 +93,12 @@ class TestAutocrop:
         crop_test(crop_x, [x0, x1, x2],
                   [x0[:1, :2, :, :], x1[:1, :2, :, :], x2[:1, :2, :, :]])
 
+        # test that num outputs is correct when the number of inputs is
+        # larger than ndim of the inputs.
+        crop_test(crop_x, [x0, x1, x2, x0, x1, x2],
+                  [x0[:1, :2, :, :], x1[:1, :2, :, :], x2[:1, :2, :, :],
+                   x0[:1, :2, :, :], x1[:1, :2, :, :], x2[:1, :2, :, :]])
+
         with pytest.raises(ValueError):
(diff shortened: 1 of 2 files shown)


benanne:
LGTM! Anyone else wanna take a look?

lasagne/tests/layers/test_merge.py
```diff
@@ -102,6 +110,8 @@ def crop_test(cropping, inputs, expected):
             crop_test(crop_bad, [x0[:, :, :, 0], x1, x2[:, :, :, :, None]],
                       [x0[:1, :2, :, :], x1[:1, :2, :, :], x2[:1, :2, :, :]])
 
+        num_inputs_larger_than_ndim_input()
```
f0k: Wouldn't it be easier to just add another `crop_test()` call above such as:
```python
crop_test(crop_x, [x0, x1, x2, x0, x1, x2],
          [x0[:1, :2, :, :], x1[:1, :2, :, :], x2[:1, :2, :, :],
           x0[:1, :2, :, :], x1[:1, :2, :, :], x2[:1, :2, :, :]])
```
This avoids adding another function.
```

skaae:
I udated the test with @f0k suggestion and squashed commits.

f0k:
Perfect, merging!

PR #556, opened 2015-12-25
clipped activation

alrojo:
Implemented the clipped rectify as in: http://arxiv.org/abs/1412.5567


f0k:
Did you close this intentionally? The implementation looks good except for the PEP8 errors.

alrojo:
Yeah sorry, I am new with Travis so I just defaulted when I couldn't figure how to debug from the error trace. I think I have got it now and will PR when it is fixed :)

f0k:
Note that you can also run the tests locally, as described in the documentation: http://lasagne.readthedocs.org/en/latest/user/development.html#how-to-contribute
Via `py.test -k nonlin` you can run the nonlinearity tests only. This way you don't have to wait for Travis.

I'll have a closer look at the PR later, thank you for your work!

alrojo:
That would've been more convenient. Thanks

alrojo:
Hi @f0k, any news?

f0k:
Sorry, this completely slipped off my mind! Will review now.

lasagne/nonlinearities.py
```diff
@@ -286,6 +286,71 @@ def softplus(x):
     return theano.tensor.nnet.softplus(x)
 
 
+# Clipped Activation
+class ClippedActivation(object):
+    """Clipped Activation activation function
+    :math:`\\varphi(x,c,a) = \\min(a(x),c)`
+
+    This function clips the activation function. Usefull for non-saturating
```
f0k: *Useful

lasagne/nonlinearities.py
```diff
@@ -286,6 +286,71 @@ def softplus(x):
...
+    functions such as the rectify to avoid exploding activations.
+
+    Parameters
+    ----------
+    x : float32
+        The activation (the summed, weighted input of a neuron).
+    clip : float32
+        The upper bound for the activation function.
+    activation : function
+        An activation function that maps x onto a float32
```
f0k: `x` is not a constructor parameter and must not be listed here. For activation, I'd suggest:
```
activation : callable
    An activation function to clip
```

lasagne/nonlinearities.py
```diff
@@ -286,6 +286,71 @@ def softplus(x):
...
+    >>> l1 = DenseLayer(l_in, num_units=200, nonlinearity=custom_activation)
+
+    Alternatively, you can use the provided instance for clip=20,
+    activation=rectify
+
+    >>> from lasagne.nonlinearities import clipped_rectify
+    >>> l2 = DenseLayer(l_in, num_units=200, nonlinearity=clipped_rectify)
+    """
+
+    def __init__(self, clip=20, activation=rectify):
```
f0k: If we call it `ClippedActivation`, maybe it should support all of the functionality of `T.clip()`. We could have two arguments `maximum` and `minimum`, with `minimum` defaulting to `None`. Depending on which bounds are given, `__call__` would use `T.minimum`, `T.maximum` or `T.clip`.

lasagne/nonlinearities.py
```diff
@@ -286,6 +286,71 @@ def softplus(x):
...
+    def __call__(self, x):
+        return theano.tensor.minimum(self.activation(x), self.clip)
+
+
+# shortcut with default activation and clip
+clipped_rectify = ClippedActivation()
+clipped_rectify.__doc__ = """clipped_rectify(x)
+
+    Instance of :class:`ClippedActivation' such as used in [1]_.
+    :math:`\\clip=20.0 \\activation=rectify`
```
f0k: To fit into the overview table (http://lasagne.readthedocs.org/en/latest/modules/nonlinearities.html), it should be a single line and must not have references. See the `LeakyRectifier` for how I solved it (giving the reference in the `See also` section of the `LeakyRectifier` class). The single sentence should directly give the activation function formula, `:math:\\clip=20 \\activation=rectify` will not even render correctly. It's helpful to render the documentation locally to check: http://lasagne.readthedocs.org/en/latest/user/development.html#documentation

lasagne/nonlinearities.py
```diff
@@ -286,6 +286,71 @@ def softplus(x):
...
+
+    References
+    ----------
+    .. [1] Awni Hannun, Carl Case, Jared Casper, Bryan Catanzaro, Greg Diamos,
+            Eric Elsen, Ryan Prenger, Sanjeev Satheesh, Shubho Sengupta,
+            Adam Coates, Andrew Y. Ng (2014):
+       Deep Speech: Scaling up end-to-end speech recognition,
+       http://arxiv.org/abs/1412.5567
```
f0k: Sphinx can be very picky about indentations -- if it doesn't render correctly, you may have to indent everything the same.

lasagne/nonlinearities.py
f0k: And maybe it should state more directly that it modifies an existing activation function:
```
Wraps an existing activation function to clip its output range. Useful for non-saturating
functions such as the linear rectifier to avoid exploding activations [1]_.
```

lasagne/nonlinearities.py
```diff
@@ -286,6 +286,71 @@ def softplus(x):
     return theano.tensor.nnet.softplus(x)
 
 
+# Clipped Activation
+class ClippedActivation(object):
+    """Clipped Activation activation function
```
f0k: What about `Clip an activation function`? The duplicate `Activation activation` reads funny.

PR #560, opened 2016-01-03
Add Adamax to docs

kmike:
Adamax was not here: http://lasagne.readthedocs.org/en/latest/modules/updates.html


f0k:
Oh, didn't see that while reviewing. Thanks!

PR #561, opened 2016-01-03
Python 3 compatible code in installation docs

kmike:


docs/user/installation.rst
```diff
@@ -222,7 +222,7 @@ To check whether it is found by Theano, run the following command:
 
 .. code-block:: bash
 
-  python -c "import theano; print theano.sandbox.cuda.dnn.dnn_available() or theano.sandbox.cuda.dnn.dnn_available.msg"
+  python -c "import theano; print(theano.sandbox.cuda.dnn.dnn_available() or theano.sandbox.cuda.dnn.dnn_available.msg)"
```
f0k: If we change this line, I'd like to shorten it while we're at it:
```bash
python -c "from theano.sandbox.cuda.dnn import dnn_available as d; print(d() or d.msg)"
```
Can you amend your commit to do that (i.e., still keep this PR to a single commit)? Thanks!

kmike: :+1:

f0k:
Perfect, thanks!

PR #566, opened 2016-01-08
don't use O(N^2) algorithm in lasagne.utils.unique

kmike:
Probably it doesn't matter much since this function is only used in get_all_params.

### DIFF
--- a/lasagne/utils.py
+++ b/lasagne/utils.py
@@ -153,9 +153,11 @@ def unique(l):
         A list of elements of `l` without duplicates and in the same order.
     """
     new_list = []
+    seen = set()
     for el in l:
-        if el not in new_list:
+        if el not in seen:
             new_list.append(el)
+            seen.add(el)
 
     return new_list
 


f0k:
The downside is that this only works for hashable items now, i.e., not for unification of a list of lists or dictionaries.

If we're truly optimizing for speed, we can use one of the [`itertools` recipes](https://docs.python.org/3/library/itertools.html#itertools-recipes) instead:
```python
def unique(iterable):
    seen = set()
    seen_add = seen.add
    for element in itertools.ifilterfalse(seen.__contains__, iterable):
        seen_add(element)
        yield element
```
It's a bit faster because it iterates over duplicate elements in C (within `ifilterfalse`) and avoids looking up `seen.add` repeatedly. It changes semantics to be a generator rather than returning a list, but it could be adapted accordingly, or `get_all_params()` could wrap it in a `list()` (if we assume that nobody else uses it, we can change the semantics).

Thanks for reviewing some of our code, by the way, it's appreciated!

kmike:
That's a good point about hashable types. Do you think there are cases in Lasagne when handling those unhashable types is needed? If so, what about adding a comment about O(N^2) complexity to the docsting instead of changing the algorithm? 

If there are no such cases it may be better to fix it (not sure it worths to micro-optimize though): users who call unique for non-hashable types will get an exception, and that's good because they'll be notified about a problem, while with the existing implementation it is possible to use O(N^2) algorithm without knowing that.

f0k:
> Do you think there are cases in Lasagne when handling those unhashable types is needed?

No, can't think of any right now. And we can still change it back or extend it if needed.

> not sure it worths to micro-optimize though

It doesn't hurt either, does it? I find the more optimized version easy enough to read (especially as you know what it's supposed to do). To have it still return a list, it could be:
```python
def unique(iterable):
    result = list()
    seen = set()
    result_append = result.append
    seen_add = seen.add
    for element in itertools.ifilterfalse(seen.__contains__, iterable):
        seen_add(element)
        result_append(element)
    return result
```
Shorter (this is a joke):
```python
def unique(iterable):
    seen = set()
    seen_add = seen.add
    return list(itertools.ifilterfalse(lambda x: x in seen or seen_add(x), iterable))
```

---

With some experimentation, I find that the `itertools` variant actually slows things down (Python 2.7). Let's either merge your PR as it is now, or optimize by pulling out all required lookups:
```python
def unique(iterable):
    result = list()
    seen = set()
    have_seen = seen.__contains__
    seen_add = seen.add
    result_append = result.append
    for element in iterable:
        if not have_seen(element):
            seen_add(element)
            result_append(element)
    return result
```

Nah, let's just merge it.

PR #570, opened 2016-01-11
Add ScaleLayer and standardize function

craffel:
The motivation behind this layer is this: Usually, part of your ML pipeline is to compute some statistics on the training set (mean and standard deviation, or min and max) and then use them to normalize/standardize all inputs to the network.  Currently, this needs to be done "outside of the model", i.e. when you save your model you must also save the training set statistics to use later when passing data through the model.  There's no convenient way to store this information in the model itself.  This layer simply stores these statistics, and then applies them in `get_output` in a smart way which hopefully covers most use-cases.  See the example in the docstring for an example.

As another (personally relevant) example, say you are passing sequences of feature tensors into a convolutional network.  Your input data is of shape `(n_batch, n_filters, n_time_steps, n_feature)`.  You want to normalize over the `n_filters` and `n_feature` (feature) dimensions, not the `n_batch` or `n_time_steps` (sample) dimensions.  After reshaping/dimshuffling your training data (call it `X`) to shape `(n_batch*n_time_steps, n_filters, n_feature)`, you could compute the mean and standard deviation over these dimensions as `offset = X.mean(axis=0)`, `scale = X.std(axis=0)` and then apply them using the `StandardizationLayer` as follows:

```Python
l_in = lasagne.layers.InputLayer((None, n_filters, None, n_features))
l_std = lasagne.layers.normalization.StandardizationLayer(
    l_in, scale, offset, axis=(1, 3))
```

If this makes sense and seems useful, I will create some unit tests for it and/or add/remove/change features.

### DIFF
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -12,6 +12,7 @@ from theano.sandbox.rng_mrg import MRG_RandomStreams as RandomStreams
 __all__ = [
     "NonlinearityLayer",
     "BiasLayer",
+    "ScaleLayer",
     "ExpressionLayer",
     "InverseLayer",
     "TransformerLayer",
@@ -114,6 +115,67 @@ class BiasLayer(Layer):
             return input + self.b.dimshuffle(*pattern)
         else:
             return input
+
+
+class ScaleLayer(Layer):
+    """
+    lasagne.layers.ScaleLayer(incoming, scales=lasagne.init.Constant(1),
+    shared_axes='auto', **kwargs)
+
+    A layer that scales its inputs by learned coefficients.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
+
+    scales : Theano shared variable, expression, numpy array, or callable
+        Initial value, expression or initializer for the scale.  The scale
+        shape must match the incoming shape, skipping those axes the scales are
+        shared over (see the example below).  See
+        :func:`lasagne.utils.create_param` for more information.
+
+    shared_axes : 'auto', int or tuple of int
+        The axis or axes to share scales over. If ``'auto'`` (the default),
+        share over all axes except for the second: this will share scales over
+        the minibatch dimension for dense layers, and additionally over all
+        spatial dimensions for convolutional layers.
+
+    Notes
+    -----
+    The scales parameter dimensionality is the input dimensionality minus the
+    number of axes the scales are shared over, which matches the bias parameter
+    conventions of :class:`DenseLayer` or :class:`Conv2DLayer`. For example:
+
+    >>> layer = ScaleLayer((20, 30, 40, 50), shared_axes=(0, 2))
+    >>> layer.scales.get_value().shape
+    (30, 50)
+    """
+    def __init__(self, incoming, scales=init.Constant(1), shared_axes='auto',
+                 **kwargs):
+        super(ScaleLayer, self).__init__(incoming, **kwargs)
+
+        if shared_axes == 'auto':
+            # default: share scales over all but the second axis
+            shared_axes = (0,) + tuple(range(2, len(self.input_shape)))
+        elif isinstance(shared_axes, int):
+            shared_axes = (shared_axes,)
+        self.shared_axes = shared_axes
+
+        # create scales parameter, ignoring all dimensions in shared_axes
+        shape = [size for axis, size in enumerate(self.input_shape)
+                 if axis not in self.shared_axes]
+        if any(size is None for size in shape):
+            raise ValueError("ScaleLayer needs specified input sizes for "
+                             "all axes that scales are not shared over.")
+        self.scales = self.add_param(
+            scales, shape, 'scales', regularizable=False)
+
+    def get_output_for(self, input, **kwargs):
+        axes = iter(range(self.scales.ndim))
+        pattern = ['x' if input_axis in self.shared_axes
+                   else next(axes) for input_axis in range(input.ndim)]
+        return input * self.scales.dimshuffle(*pattern)
 
 
 class ExpressionLayer(Layer):
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -13,6 +13,7 @@ __all__ = [
     "NonlinearityLayer",
     "BiasLayer",
     "ScaleLayer",
+    "standardize",
     "ExpressionLayer",
     "InverseLayer",
     "TransformerLayer",
@@ -176,6 +177,62 @@ class ScaleLayer(Layer):
         pattern = ['x' if input_axis in self.shared_axes
                    else next(axes) for input_axis in range(input.ndim)]
         return input * self.scales.dimshuffle(*pattern)
+
+
+def standardize(layer, offset, scale, shared_axes):
+    """
+    Convenience function for standardizing inputs by applying a fixed offset
+    and scale.  This is usually useful when you want the input to your network
+    to, say, have zero mean and unit standard deviation over the feature
+    dimensions.  This layer allows you to include the appropriate statistics to
+    achieve this normalization as part of your network, and applies them to its
+    input.  The statistics are supplied as the `offset` and `scale` parameters,
+    which are applied to the input by subtracting `offset` and dividing by
+    `scale`, sharing dimensions as specified by the `shared_axes` argument.
+
+    Parameters
+    ----------
+    layer : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape.
+    offset : Theano shared variable, expression, or numpy array
+        The offset to apply (via subtraction) to the axis/axes being
+        standardized.
+    scale : Theano shared variable, expression or numpy array
+        The scale to apply (via division) to the axis/axes being standardized.
+    shared_axes : 'auto', int or tuple of int
+        The axis or axes to share the offset and scale over. If ``'auto'`` (the
+        default), share over all axes except for the second: this will share
+        scales over the minibatch dimension for dense layers, and additionally
+        over all spatial dimensions for convolutional layers.
+
+    Examples
+    --------
+    Assuming your training data exists in a 2D numpy ndarray called
+    ``training_data``, you can use this function to scale input features to the
+    [0, 1] range based on the training set statistics like so:
+
+    >>> import lasagne
+    >>> import numpy as np
+    >>> training_data = np.random.standard_normal((100, 20))
+    >>> input_shape = (None, training_data.shape[1])
+    >>> l_in = lasagne.layers.InputLayer(input_shape)
+    >>> offset = training_data.min(axis=0)
+    >>> scale = training_data.max(axis=0) - training_data.min(axis=0)
+    >>> l_std = standardize(l_in, offset, scale, shared_axes=0)
+
+    Alternatively, to z-score your inputs based on training set statistics, you
+    could set ``offset = training_data.mean(axis=0)`` and
+    ``scale = training_data.std(axis=0)`` instead.
+    """
+    # Subtract the offset
+    layer = BiasLayer(layer, -offset, shared_axes)
+    # Do not optimize the offset parameter
+    layer.params[layer.b].remove('trainable')
+    # Divide by the scale
+    layer = ScaleLayer(layer, T.inv(scale), shared_axes)
+    # Do not optimize the scales parameter
+    layer.params[layer.scales].remove('trainable')
+    return layer
 
 
 class ExpressionLayer(Layer):
--- a/lasagne/tests/layers/test_special.py
+++ b/lasagne/tests/layers/test_special.py
@@ -2,6 +2,7 @@ from mock import Mock
 import numpy as np
 import pytest
 import theano
+from lasagne.layers import InputLayer, standardize, get_output
 
 
 class TestExpressionLayer:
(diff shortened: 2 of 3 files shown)


f0k:
+1 for adding a layer that just scales and shifts, this is currently not easily possible. The way it's presented might evoke too high expectations, though (like it would actually take care of computing the standardization). What about just neutrally calling it a `ScaleShiftLayer`, renaming `axes` to `shared_axes`, giving it a `trainable` keyword argument and pointing out your use case in the documentation? Would that work as well or would that be too generic / too difficult to find?

Regarding the implementation, I'm not sure about the `reshape` in the end -- wouldn't it be better to follow the `dimshuffle` pattern as in `BiasLayer`, for example?

craffel:
>  The way it's presented might evoke too high expectations, though (like it would actually take care of computing the standardization).

Yes, that's true; sort of a naming issue.

> What about just neutrally calling it a ScaleShiftLayer,

Or `OffsetScaleLayer` based on my current naming.

> renaming axes to shared_axes,

Sure.

> giving it a trainable keyword argument and pointing out your use case in the documentation?   Would that work as well or would that be too generic / too difficult to find?

If there's a use-case for a generic (potentially trainable) offset-and-scale layer, I'd personally think it would be convenient/helpful to also have a named layer which suggests it is used for standardization and sets trainable to `False` by default.

> Regarding the implementation, I'm not sure about the reshape in the end -- wouldn't it be better to follow the dimshuffle pattern as in BiasLayer, for example?

Hm, I see that.  Is one better than the other in terms of efficiency?  The reshape pattern is more natural to me personally.  The way I have it implemented here works for the use-cases I tested.

f0k:
> Hm, I see that. Is one better than the other in terms of efficiency?

Every symbolic `.shape` introduces a Theano Op, so it might a little less inefficient / more difficult to optimize for Theano (but I don't know).

> The reshape pattern is more natural to me personally.

I think the most natural way would be making the parameters correctly broadcastable right in the constructor and then not care about the shared axes any more. It's just that existing layers already went the route of reducing the dimensionality of their biases (e.g., have a vector for the convolutional layers) and dimshuffling them for applying them, so I followed suit in `BiasLayer`.

> Or `OffsetScaleLayer` based on my current naming.

Fine as well. `ScaleShiftLayer` would actually imply scaling first, so disregard this.

> If there's a use-case for a generic (potentially trainable) offset-and-scale layer

Not sure about actual use cases, it would just feel inconsistent to me to add a layer that is not even trainable by default (because it would be the only one).

> I'd personally think it would be convenient/helpful to also have a named layer which suggests it is used for standardization and sets trainable to `False` by default.

What about a macro function? If we go for a macro function, we could also just add a `ScaleLayer` to `lasagne.layers.special` and add a `trainable` keyword argument to the `BiasLayer` -- or let the macro function remove the `trainable` flag from the layers:
```python
def standardize(layer, offset, scale, axes=..., trainable=False):
    layer = BiasLayer(layer, offset, axes)
    if not trainable:
        layer.params[layer.b].remove('trainable')
    layer = ScaleLayer(layer, scale, axes)
    if not trainable:
        layer.params[layer.scale].remove('trainable')
    return layer
```
Might lead to less code duplication in the long run, and it would be about as convenient to use as your `StandardizationLayer`. What do you think?

craffel:
> I think the most natural way would be making the parameters correctly broadcastable right in the constructor and then not care about the shared axes any more. It's just that existing layers already went the route of reducing the dimensionality of their biases (e.g., have a vector for the convolutional layers) and dimshuffling them for applying them, so I followed suit in BiasLayer.

I don't personally like the idea of the layer changing the shape of the parameters that the user passes to it, but if `BiasLayer` already does it one way and we're sticking to that it's a moot discussion anyways.

> Not sure about actual use cases, it would just feel inconsistent to me to add a layer that is not even trainable by default (because it would be the only one).

Ok.

> What about a macro function? If we go for a macro function, we could also just add a ScaleLayer to lasagne.layers.special and add a trainable keyword argument to the BiasLayer -- or let the macro function remove the trainable flag from the layers:

I am ok with this, with the minor caveat that I don't think there should be a `trainable` argument to the macro - it doesn't make any sense in my mind for the offset/scale to be optimizable when they are used for standardization.

f0k:
> I don't personally like the idea of the layer changing the shape of the parameters that the user passes to it

No no, it doesn't. It was just a question of what dimensionality the parameters are defined as in the first place. My idea was that a `BiasLayer` on 4D input should always have a 4D bias tensor with broadcastable singleton dimensions for all shared axes, instead of creating, say, a 2D tensor and adding the correct broadcastable dimensions in `get_output_for()`. It does the latter now to be compatible with existing biases, but this also means it's got to have the somewhat complex `dimshuffle` pattern generator.

> I am ok with this, with the minor caveat that I don't think there should be a `trainable` argument to the macro - it doesn't make any sense in my mind for the offset/scale to be optimizable when they are used for standardization.

Okay for me as well. One might want to try continuing training from the standardization, but that only makes sense if the next layer doesn't start with a linear transformation anyway. That's probably not going to happen very often.

Do you want to add the `ScaleLayer` and `standardize()` macro (the latter could reuse a lot of your current docstring)? The `ScaleLayer` should probably go into `layers/special.py` after the `BiasLayer`, about `standardize()` I'm not sure.

craffel:
> No no, it doesn't. It was just a question of what dimensionality the parameters are defined as in the first place. My idea was that a BiasLayer on 4D input should always have a 4D bias tensor with broadcastable singleton dimensions for all shared axes, instead of creating, say, a 2D tensor and adding the correct broadcastable dimensions in get_output_for(). It does the latter now to be compatible with existing biases, but this also means it's got to have the somewhat complex dimshuffle pattern generator.

I'll follow the existing convention, if it's something which we decide to change later on I'll leave it to then.

> Do you want to add the ScaleLayer and standardize() macro (the latter could reuse a lot of your current docstring)? The ScaleLayer should probably go into layers/special.py after the BiasLayer, about standardize() I'm not sure.

Sure, I will do this now.

craffel:
Ok, added `ScaleLayer` and `standardize` macro, ready for review.

benanne:
Looks nice :) I had a quick look, didn't do a detailed review. It seems that coveralls is unhappy because the case of `scales=None` is untested.

craffel:
Oops, I actually meant to take that out (I took that case out of `get_output_for`) - I can't really think of a reason anyone would want to use a `ScaleLayer` with no scaling.  If you want the no-scaling, `scales is None` case, I can re-add that back in.

benanne:
No, I think it's fine to get rid of it altogether probably :)

craffel:
Not sure where to put the docs for `standardize`, any ideas are appreciated.

lasagne/layers/normalization.py
```diff
@@ -101,3 +103,58 @@ def get_output_for(self, input, **kwargs):
...
         return input / scale
+
+
+def standardize(layer, offset, scale, shared_axes):
+    """
+    Convenience function for standardizing inputs by applying a fixed offset
+    and scale.  This is usually useful when you want the input to your network
+    to, say, have zero mean and unit standard deviation over the feature
+    dimensions.  This layer allows you to compute those statistics over your
+    training set and include them as part of your network.  The statistics are
```
f0k: "This layer allows you to compute those statistics" -- sounds as if the layer would do this. Can you rephrase such that it's clear that the layer only allows you to store the normalization offsets/factors within the network, but does not help you with computing them?

lasagne/layers/normalization.py
```diff
@@ -101,3 +103,58 @@ def get_output_for(self, input, **kwargs):
...
+    layer : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape.
+    offset : Theano shared variable, expression, numpy array or callable
+        The offset to apply (via subtraction) to the axis/axes being
+        standardized.
+    scale : Theano shared variable, expression, numpy array or callable
+        The scale to apply (via division) to the axis/axes being standardized.
+    shared_axes : 'auto', int or tuple of int
+        The axis or axes which will be shared (i.e. are not being standardized
+        over).
```
f0k: Mhm, that's not really what it does, is it? `shared_axes` names the axes over which the offset and scale will be shared/broadcasted. It also needs an explanation for `'auto'` (which can be copied over from the `BiasLayer`).

f0k: Comparing to the `BatchNormLayer`, those axes are actually the axes standardized over, not the axes *not* standardized over. I.e., to compute the offset for `shared_axes=(1,3)`, you would compute `np.mean(X, axis=(1,3))`. It's confusing. I think it may be safer to copy and adapt the explanation from `BiasLayer` or `ScaleLayer`.

lasagne/layers/special.py
```diff
@@ -112,6 +113,67 @@ def get_output_for(self, input, **kwargs):
             return input
 
 
+class ScaleLayer(Layer):
+    """
+    lasagne.layers.ScaleLayer(incoming, scales=lasagne.init.Constant(1),
+    shared_axes='auto', **kwargs)
+
+    A layer that scales its inputs along certain axes.
```
f0k: "scale along axes" confuses me -- what about just saying `A layer that scales its inputs by learned coefficients.`? Sharing coefficients over axes is explained fine in the `shared_axes` description.

f0k:
> Not sure where to put the docs for `standardize`, any ideas are appreciated.

If you rebase onto the latest master (sorry about the merge conflict), `layers/normalize.rst` will magically appear for you to put the docs into. I'm still not sure if this function should live in `normalize` or `special`, though (hint: if you move it to `special`, you avoid the merge conflict, if that's a good reason ;) ).

lasagne/layers/normalization.py
```diff
@@ -101,3 +103,58 @@ def get_output_for(self, input, **kwargs):
...
+    Alternatively, to z-score your inputs based on training set statistics, you
+    could set ``offset = training_data.mean(axis=0)`` and
+    ``scale = training_data.std(axis=0)`` instead.
+    """
+    # Subtract the offset
+    layer = BiasLayer(layer, -offset, shared_axes)
+    # Do not optimize the offset parameter
+    layer.params[layer.b].remove('trainable')
+    # Divide by the scale
+    layer = ScaleLayer(layer, 1./scale, shared_axes)
```
f0k: This is dangerous again -- if `scale` is a numpy array, `1./scale` will upcast it to float64. You could use `T.inv()`, with the fun side effect that it immediately turns a numpy array into a Theano constant that cannot be learned. Alternatively, you could use `lasagne.utils.floatX(1) / scale`.

Aiming for constants might enable some more optimizations for Theano, though (if they can be folded with other operations). We could supply the bias as a constant as well, instead of a shared variable.

lasagne/layers/special.py
craffel: Ok, changed.

lasagne/layers/normalization.py
craffel: Ok, changed wording.

lasagne/layers/normalization.py
craffel: I understand it in the opposite way (e.g. "standardizing over" means "make these dimensions standardize").  I copied the explanation.

lasagne/layers/normalization.py
craffel: Ok, I switched to using `T.inv`.  I also removed the documentation saying that a callable could be used, which is not true.

craffel:
> If you rebase onto the latest master (sorry about the merge conflict), layers/normalize.rst will magically appear for you to put the docs into. I'm still not sure if this function should live in normalize or special, though (hint: if you move it to special, you avoid the merge conflict, if that's a good reason ;) ).

Ok, I moved it.  Also rebased.  Should be ready for re-review.

lasagne/layers/special.py
```diff
@@ -116,6 +118,124 @@ def get_output_for(self, input, **kwargs):
...
+    achieve this normalization as part of your network, and applies them to its
+    input.  The statistics are supplied as the `offset` and `scale` parameters,
+    which are applied to the input by subtracting `offset` and dividing by
+    `scale`, sharing dimensions as specified by the `shared_axes` argument.
+
+    Parameters
+    ----------
+    layer : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape.
+    offset : Theano shared variable, expression, numpy array or callable
```
f0k: Cannot be a callable the way it's implemented.

lasagne/layers/special.py
```diff
@@ -116,6 +118,124 @@ def get_output_for(self, input, **kwargs):
...
+        The scale to apply (via division) to the axis/axes being standardized.
+    shared_axes : 'auto', int or tuple of int
+        The axis or axes to share the offset and scale over. If ``'auto'`` (the
+        default), share over all axes except for the second: this will share
+        scales over the minibatch dimension for dense layers, and additionally
+        over all spatial dimensions for convolutional layers.
+
+    Examples
+    --------
+
```
f0k: Just for consistency, could you drop this newline, please? Cheers!

lasagne/tests/layers/test_special.py
```diff
@@ -2,6 +2,7 @@
 import numpy as np
 import pytest
 import theano
+import lasagne
```
f0k: I'd prefer `from lasagne.layers import InputLayer, standardize, get_output` down in `test_standardize()` instead of a global import. (But also just for consistency, because most test suites were written to avoid a global `lasagne` import -- I don't know who started it. A possible advantage is that it's easier to track which Lasagne components specific test suites depend on.)

lasagne/tests/layers/test_special.py
```diff
@@ -219,6 +220,79 @@ def test_undefined_shape(self, BiasLayer):
...
+    l_std = lasagne.layers.standardize(
+        l_in, X.min(axis=0), (X.max(axis=0) - X.min(axis=0)), shared_axes=0)
+    out = lasagne.layers.get_output(l_std).eval({l_in.input_var: X})
+    assert np.allclose(out.max(axis=0), 1.)
+    assert np.allclose(out.min(axis=0), 0.)
+    # More complicated example
+    X = np.random.standard_normal(
+        (50, 3, 100, 10)).astype(theano.config.floatX)
+    mean = X.transpose(0, 2, 1, 3).mean(axis=0).mean(axis=0)
+    std = X.transpose(0, 2, 1, 3).std(axis=0).std(axis=0)
```
f0k: It would be easier to follow as:
```python
mean = X.mean(axis=(0,2))
std = X.std(axis=(0, 2))
```
Unless you had something specific in mind when computing the standard deviation of standard deviations?

f0k:
> Ok, I moved it. Also rebased. Should be ready for re-review.

Thanks, looks very good except for a few comments!

lasagne/layers/special.py
craffel: Oops, missed that one, fixed.

lasagne/layers/special.py
craffel: Ok.

lasagne/tests/layers/test_special.py
craffel: Ok.

lasagne/tests/layers/test_special.py
craffel: Yes, this is better, changed.

craffel:
Ok, addressed your comments.

f0k:
Nice! Will merge tomorrow if there are no further comments.

craffel:
Bump!

benanne:
I'm not @f0k, but I'll merge it as there are no further comments :) Thanks!

PR #569, opened 2016-01-11
Raise error if pooling layers get unexpected input dimension

danstowell:
Fixes #568. First time looking at the codebase, feedback welcome.

PR #572, opened 2016-01-12
Added Pool3DDNNLayer and MaxPool3DDNNLayer

gyglim:
Hi

This adds the cuDnn implementation of the Pool3DDNNLayer. Required by https://github.com/Lasagne/Recipes/pull/41.

Cheers,
Michael

### DIFF
--- a/lasagne/layers/dnn.py
+++ b/lasagne/layers/dnn.py
@@ -16,6 +16,8 @@ if not theano.config.device.startswith("gpu") or not dnn.dnn_available():
 __all__ = [
     "Pool2DDNNLayer",
     "MaxPool2DDNNLayer",
+    "Pool3DDNNLayer",
+    "MaxPool3DDNNLayer",
     "Conv2DDNNLayer",
     "Conv3DDNNLayer",
 ]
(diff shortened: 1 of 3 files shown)


lasagne/layers/dnn.py
```diff
@@ -16,6 +16,8 @@
 __all__ = [
     "Pool2DDNNLayer",
     "MaxPool2DDNNLayer",
+    "Pool3DDNNLayer",
+    "MaxPool3DDNNLayer",
```
f0k: Please also add this to `docs/modules/layers.rst`!

lasagne/layers/dnn.py
```diff
@@ -125,6 +127,114 @@ def __init__(self, incoming, pool_size, stride=None,
...
+    -----
+    The value used to pad the input is chosen to be less than
+    the minimum of the input, so that the output of each pooling region
+    always corresponds to some element in the unpadded input region.
+
+    """
+    def __init__(self, incoming, pool_size, stride=None, pad=(0, 0, 0),
+                 ignore_border=True, mode='max', **kwargs):
+        super(Pool3DDNNLayer, self).__init__(incoming, **kwargs)
+        self.pool_size = as_tuple(pool_size, 3)
```
f0k: It was merged into `master` just minutes ago, but could you please also copy/paste the dimensionality check of #569 here?

lasagne/layers/dnn.py
```diff
@@ -125,6 +127,114 @@ def __init__(self, incoming, pool_size, stride=None,
...
+                                             pad=self.pad[0],
+                                             ignore_border=True,
+                                             )
+
+        output_shape[3] = pool_output_length(input_shape[3],
+                                             pool_size=self.pool_size[1],
+                                             stride=self.stride[1],
+                                             pad=self.pad[1],
+                                             ignore_border=True,
+                                             )
```
f0k: Nitpick: Just for consistency, please also add a newline here.

lasagne/tests/layers/test_pool.py
```diff
@@ -502,6 +523,88 @@ def test_not_implemented(self):
...
+            raise
+        #    pytest.skip()
+
+    def test_not_implemented(self):
+        try:
+            from lasagne.layers.dnn import MaxPool3DDNNLayer
+        except ImportError:
+            pytest.skip("cuDNN not available")
+        with pytest.raises(NotImplementedError) as exc:
+            layer = MaxPool3DDNNLayer((1, 2, 3, 4), pool_size=2,
```
f0k: Should be `(1, 2, 3, 4, 5)` (but you'd have noticed after adding the dimensionality check).

lasagne/layers/dnn.py
f0k: (It will probably be easier if you rebase your PR onto the latest master, so the existing 2d pooling tests will include the dimensionality check. Let me know if you need help.)

f0k:
Thank you, looks good except for a few minor things! If you can fix them and squash your PR into a single commit, this is good to merge. (If you need any help regarding rebasing/squashing, just let me know.) Cheers!

gyglim:
Great, thank. I made the changes. 
On a related note: why is the indexing done with [0],[1] and so on for the Pooling? Could be done from the back, no ([-1], [-2] and so on)? This would remove the constraint that the tensor needs to have exactly 5 (or 4 in the 2D pooling case) dimensions, as long as there are as many as there are pooled over?

As for the squashing: Yes, I've never done it and would appreciate some help

f0k:
> This would remove the constraint that the tensor needs to have exactly 5 (or 4 in the 2D pooling case) dimensions, as long as there are as many as there are pooled over?

It would be easy for us to change the implementation in `get_output_shape_for()`, but the cuDNN implementation requires 4D or 5D tensors. We'd need to include some reshaping in `get_output_for()` then to squash the first dimensions before `dnn_pool()` and get them back out afterwards (if the input dimensionality is too large), or to introduce additional first dimensions and remove them later (if the input dimensionality is too low). As long as there are not obvious use cases for this, I'd leave this burden up to the user.

> Great, thank. I made the changes. 

Great! You'll also need to add a test for the dimensionality check, like this one: https://github.com/Lasagne/Lasagne/pull/569/files#diff-05ecd45be63aafa8322cac30a5c0c66eR522
Just adapt and add it to the end of `TestMaxPool3DNNLayer`. You're lucky that Travis doesn't have a GPU, otherwise coveralls would have complained.

> As for the squashing: Yes, I've never done it and would appreciate some help

The easiest way to remember is to run `git rebase --interactive master`. You'll be greeted with a text editor listing all your commits and giving some guidance below. You'll have to leave the first commit as it is (unless you want to reword it, using `r`). All the other commits can be changed to `f` (for fixup). When you save and exit, git will squash all `fixup` commits into the first one.

f0k:
It seems you pulled in a commit of @danstowell into your PR now? This creates a merge conflict... you can go back with `git reset --hard 1ab40a4; git push --force`. To get the upstream changes into your PR, it's better to rebase it. Use `git fetch upstream; git rebase upstream/master` to rebase your PR onto the latest master (assuming you've got the official Lasagne repository named `upstream` in your clone).

gyglim:
Ok, I added the ndim test. 
In the test code from https://github.com/Lasagne/Lasagne/pull/569/files there was a small error, it used  MaxPool2DNNLayer instead of  MaxPool2DDNNLayer. Fixed this aswell. Now I just need to get the squashing right!

lasagne/tests/layers/test_pool.py
```diff
@@ -525,13 +546,107 @@ def test_fail_on_mismatching_dimensionality(self):
         except ImportError:
             pytest.skip("cuDNN not available")
         with pytest.raises(ValueError) as exc:
-            MaxPool2DNNLayer((10, 20, 30), 3, 2)
+            MaxPool2DDNNLayer((10, 20, 30), 3, 2)
         assert "Expected 4 input dimensions" in exc.value.args[0]
         with pytest.raises(ValueError) as exc:
-            MaxPool2DNNLayer((10, 20, 30, 40, 50), 3, 2)
+            MaxPool2DDNNLayer((10, 20, 30, 40, 50), 3, 2)
```
f0k: Oops, good catch. Forgot that we can't rely on Travis for any tests involving the GPU. I'll remember to test everything locally next time.

gyglim:
Hi. Ok, now I believe it is all correct, right?

f0k:
> Ok, now I believe it is all correct, right?

Looks good, except that the newline (https://github.com/Lasagne/Lasagne/pull/572#discussion_r49480283) seems to have disappeared again? Didn't you fix this already?

After fixing this, do `git add lasagne/layers/dnn.py; git commit --amend --no-edit; git push --force` to integrate the change into the current commit (so you don't have to squash again).

gyglim:
I added one, but forgot another. Fixed it now

f0k:
Looks good to me! Tests pass on my machine. Will merge tomorrow if there are no further comments.

gyglim:
Great, thank you for guiding me through the process

benanne:
Awesome, thanks a lot for contributing!

f0k:
Merging, thanks again!

Issue #563, opened 2016-01-04
dropout layers leak float64 into graph


n291hhJJA:
When using a dropout layer, the output upcasts to float64, causing every layer after it to be upcast to float64. The problem seems to be caused by the way numpy treats operations that upcast integers. For example:

x = 1 - numpy.float32(0.5)

causes x to have type numpy.float64

This happens to me on numpy version 1.9.0, but I think it happens on other versions as well.

So, here is a code fix:

Line 73 of noise.py - which reads:
retain_prob = 1 - self.p

should be changed to something like this:
if theano.config.floatX == 'float64' :
   retain_prob = np.float64(1 - self.p)
else :
   retain_prob = np.float32(1 - self.p)

benanne:
it's odd that this only comes up now, but maybe it is something that happens only with newer numpy versions? Anyone care doing some tests? At any rate we should avoid it. I think a more future-proof fix is probably `retain_prob = np.array(1 - self.p, dtype=theano.config.floatX)`.

f0k:
> it's odd that this only comes up now

No, if you set `p` to a Python float instead of a numpy scalar, then everything's fine. We should allow both, though.

> I think a more future-proof fix is probably retain_prob = np.array(1 - self.p, dtype=theano.config.floatX).

Nope, `p` could be symbolic. It should be `retain_prob = lasagne.utils.floatX(1) - self.p`.

@n291hhJJA, would you like to send a PR or shall we?

n291hhJJA:
I'd like to do the pull request if you don't mind waiting a day or two.

I agree on the symbolic issue- didn't think of that case originally.

f0k:
> I'd like to do the pull request if you don't mind waiting a day or two.

That's okay! Check the [contribution guide](http://lasagne.readthedocs.org/en/latest/user/development.html#how-to-contribute), and let us know if you need any help.

PR #578, opened 2016-01-18
fix float64 leak in Dropout layer

n291hhJJA:
This resolves issue #563

### DIFF
--- a/lasagne/layers/noise.py
+++ b/lasagne/layers/noise.py
@@ -2,7 +2,7 @@ import theano
 
 from .base import Layer
 from ..random import get_rng
-
+from .. import utils
 from theano.sandbox.rng_mrg import MRG_RandomStreams as RandomStreams
 
 
@@ -70,7 +70,7 @@ class DropoutLayer(Layer):
         if deterministic or self.p == 0:
             return input
         else:
-            retain_prob = 1 - self.p
+            retain_prob = utils.floatX(1) - self.p
             if self.rescale:
                 input /= retain_prob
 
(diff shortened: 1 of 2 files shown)


lasagne/layers/noise.py
```diff
@@ -2,7 +2,7 @@
 
 from .base import Layer
 from ..random import get_rng
-
+from .. import utils
```
f0k: Can you please re-add the blank line (after `from .. import utils`)?

lasagne/tests/layers/test_noise.py
```diff
@@ -28,6 +28,11 @@ def layer_p_02(self, input_layer):
         from lasagne.layers.noise import DropoutLayer
         return DropoutLayer(input_layer, p=0.2)
 
+    def test_float64_leaks(self, input_layer) :
+        from lasagne.layers.noise import DropoutLayer
+	theano.config.floatX = 'float32'
+	assert DropoutLayer(input_layer,p=numpy.float32(0.5)).get_output_for(theano.shared(numpy.ones((100,100),dtype=numpy.float32))).eval().dtype is numpy.dtype('float32')
```
f0k: The indentation is wrong, you're missing a space for the second and third line. In addition, the third line is much too long (the limit is 79 characters).
The test is not good either, but you couldn't know: When you set `theano.config.floatX = 'float32'`, this will affect all future tests. You could save the previous value and restore it (after obtaining the dtype of the result, but before asserting it is float32).

n291hhJJA: OK, I'll fix it soon. Didn't realize that globals stuck around for all tests.

Issue #557, opened 2015-12-26
Raise a warning when a keyword passed to `get_output()` is unused


benanne:
`lasagne.layers.get_output()` can be called with keyword arguments, which are passed to all traversed layers. This makes it possible to control layers with multiple behaviours (e.g. dropout, which can be turned on or off, or batch normalization, which can use minibatch statistics or aggregated statistics).

The most common use case for this is the `deterministic` keyword, which is used to turn off any regularizing noise in the net at test time (i.e. dropout).

The handling of these keywords is implemented in the `get_output_for()` method of each `Layer` class, which is required to take a catchall `**kwargs` argument to absorb any keywords that the layer doesn't recognize.

One issue with this is that a `Layer` instance might unintentionally ignore a certain keyword, and this is basically a silent failure. If you misspell `deterministic` as `determinstic` for example, all layers will ignore it and no warning or exception will be raised.

This is not very user friendly. Recently I had a brief discussion about this issue on this reddit thread: https://www.reddit.com/r/MachineLearning/comments/3xom8a/discussion_what_would_you_like_to_see_in_a/cy6hqg8

One of the developers of climin ( https://github.com/BRML/climin ) chimed in and pointed me to their solution for this issue: https://github.com/BRML/climin/blob/master/climin/util.py#L140-L151

Basically, check the called function's argument specification through introspection, and see if any arguments were provided that aren't listed there. If so, raise a warning.

I think such a solution may be possible for Lasagne as well. We could have `get_output()` inspect the argument specification for all `get_output_for()` calls it makes, and keep track of the set of all accepted keyword arguments across all layers. Once the graph is traversed, we can check whether any of the provided keyword arguments are not in this set.

As far as I can tell this would only require modifications in `get_output()` itself. It would complicate the code somewhat, but not excessively so, and I think the usability benefit should not be underestimated.

There is one big caveat: not all `get_output_for()` implementations specify all keywords arguments they accept in the method signature. Some implementations work directly with the `kwargs` dictionary, most importantly the new batch normalization implementation (#467).

So we'd have to figure out a workaround for that. The simplest approach would be to always have all accepted keyword arguments in the method signature, but I don't know if that is desirable. Another solution would be to provide an alternative way for such methods to 'declare' which keyword arguments they understand, but I think that might get a little too complicated.

I might try whipping up a PR for this, but before I do that we should probably figure out what to do with accepted keyword arguments that are not explicitly declared in the method signature.

[...]

PR #585, opened 2016-01-20
Raise a warning when a keyword passed to `get_output()` is unused

Joshua-Chin:
This resolves issue #557. `get_output()` uses introspection to retrieve the key word arguments from `get_output_for()`. All layers now have the attribute `get_output_kwargs`, which allows a user to explicitly specify the kwargs `get_output_for` will use. Also, I changed the signature of `BatchNormalizationLayer.get_output_for` to explicitly state the kwargs that it uses.

### DIFF
--- a/lasagne/layers/helper.py
+++ b/lasagne/layers/helper.py
@@ -1,5 +1,8 @@
 from collections import deque
+from difflib import get_close_matches
+from inspect import getargspec
 from itertools import chain
+from warnings import warn
 
 import theano
 import numpy as np
@@ -147,6 +150,8 @@ def get_output(layer_or_layers, inputs=None, **kwargs):
     """
     from .input import InputLayer
     from .base import MergeLayer
+    # track accepted kwargs used by get_output_for
+    accepted_kwargs = {'deterministic'}
     # obtain topological ordering of all layers the output layer(s) depend on
     treat_as_input = inputs.keys() if isinstance(inputs, dict) else []
     all_layers = get_all_layers(layer_or_layers, treat_as_input)
@@ -184,6 +189,27 @@ def get_output(layer_or_layers, inputs=None, **kwargs):
                                  "mapping this layer to an input expression."
                                  % layer)
             all_outputs[layer] = layer.get_output_for(layer_inputs, **kwargs)
+            try:
+                names, _, _, defaults = getargspec(layer.get_output_for)
+            except TypeError:
+                # If introspection is not possible, skip it
+                pass
+            else:
+                if defaults is not None:
+                    accepted_kwargs |= set(names[-len(defaults):])
+            accepted_kwargs |= set(layer.get_output_kwargs)
+    unused_kwargs = set(kwargs.keys()) - accepted_kwargs
+    if unused_kwargs:
+        suggestions = []
+        for kwarg in unused_kwargs:
+            suggestion = get_close_matches(kwarg, accepted_kwargs)
+            if suggestion:
+                suggestions.append('%s (perhaps you meant %s)'
+                                   % (kwarg, suggestion[0]))
+            else:
+                suggestions.append(kwarg)
+        warn("get_output() was called with unused kwargs:\n\t%s"
+             % "\n\t".join(suggestions))
     # return the output(s) of the requested layer(s) only
     try:
         return [all_outputs[layer] for layer in layer_or_layers]
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -37,6 +37,7 @@ class Layer(object):
 
         self.name = name
         self.params = OrderedDict()
+        self.get_output_kwargs = []
 
         if any(d is not None and d <= 0 for d in self.input_shape):
             raise ValueError((
@@ -241,6 +242,7 @@ class MergeLayer(Layer):
                              for incoming in incomings]
         self.name = name
         self.params = OrderedDict()
+        self.get_output_kwargs = []
 
     @Layer.output_shape.getter
     def output_shape(self):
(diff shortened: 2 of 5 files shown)


benanne:
Awesome, thanks for contributing! It seems that Travis fails -- I wanted to check out why, but for some reason it's not showing me the logs right now. Can anybody else see them? I'm wondering if it's just me or if something is wrong with Travis at the moment.

Joshua-Chin:
According to https://www.traviscistatus.com/, Travis is performing emergency database upgrades for the build log system.

Joshua-Chin:
`mock` and `inspect` don't work together very well in python2. I am trying to find a workaround.

lasagne/layers/helper.py
```diff
@@ -183,6 +189,13 @@ def get_output(layer_or_layers, inputs=None, **kwargs):
                                  "mapping this layer to an input expression."
                                  % layer)
             all_outputs[layer] = layer.get_output_for(layer_inputs, **kwargs)
+            if not isinstance(layer, Mock) and \
+                    not isinstance(layer.get_output_for, Mock):
```
f0k: I'd prefer not to test for a `Mock` instance. What about `if isinstance(layer, Layer)` or `hasattr(layer, 'get_output_for')`?

lasagne/layers/helper.py
```diff
@@ -183,6 +189,13 @@ def get_output(layer_or_layers, inputs=None, **kwargs):
...
+                unused_kwargs -= set(getargspec(layer.get_output_for)[0])
+            unused_kwargs -= set(layer.get_output_kwargs)
+    if unused_kwargs:
+        template = ", ".join(unused_kwargs)
+        warn("get_output() was called with unused kwargs: %s" % template)
```
f0k: Could we go the other way instead -- collect a list of all accepted keywords and compare it to `set(kwargs.keys())` afterwards? This would allow us to not only print the keyword arguments that were not understood, but also the list of keyword arguments that would have been accepted (or even make suggestions using `difflib.get_close_matches`).

lasagne/layers/helper.py
```diff
@@ -183,6 +189,13 @@ def get_output(layer_or_layers, inputs=None, **kwargs):
                                  "mapping this layer to an input expression."
                                  % layer)
             all_outputs[layer] = layer.get_output_for(layer_inputs, **kwargs)
+            if not isinstance(layer, Mock) and \
+                    not isinstance(layer.get_output_for, Mock):
+                unused_kwargs -= set(getargspec(layer.get_output_for)[0])
```
f0k: It might be good to just collect the actual keyword arguments, i.e., the ones that have default values set:
```
names, _, _, defaults = getargspec(layer.get_output_for)
unused_kwargs -= set(names[-len(defaults):])
```

Joshua-Chin:
I just squashed the commits into a single commit. Is there anything else I should do/fix?

lasagne/layers/helper.py
```diff
@@ -184,6 +190,24 @@ def get_output(layer_or_layers, inputs=None, **kwargs):
...
+                    accepted_kwargs |= set(names[-len(defaults):])
+            accepted_kwargs |= set(layer.get_output_kwargs)
+    unused_kwargs = set(kwargs.keys()) - accepted_kwargs
+    if unused_kwargs:
+        suggestions = []
+        for kwarg in unused_kwargs:
+            suggestion = get_close_matches(kwarg, accepted_kwargs)
+            if suggestion:
+                suggestions.append('%s (perhaps you meant %s)'
+                                   % (kwarg, suggestion[0]))
```
f0k: Yes, that's what I had in mind :) Very cool! :+1:

lasagne/layers/helper.py
```diff
@@ -184,6 +190,24 @@ def get_output(layer_or_layers, inputs=None, **kwargs):
                                  "mapping this layer to an input expression."
                                  % layer)
             all_outputs[layer] = layer.get_output_for(layer_inputs, **kwargs)
+            if isinstance(layer.get_output_for, MethodType):
+                # Mock in python2 does not support introspection
```
f0k: `layer.get_output_for` could also be a function if somebody monkey-patches an instance. This won't be a common use case, but I'm still a bit worried about making the implementation less generic just to cater for deficits in the test framework.
To make it a little more robust, what about either:
```python
if isinstance(layer.get_output_for, (MethodType, FunctionType)):
```
or:
```python
try:
    names, _, _, defaults = getargspec(layer.get_output_for)
except TypeError:
    pass  # if introspection is not possible, skip it
else:
    if defaults is not None:
        accepted_kwargs |= set(names[-len(defaults):])
```

Joshua-Chin:
I changed the introspection code to catch `TypeError`s.

f0k:
Great! I'm really happy with this, thanks a lot for your work! I've also ran the full test suite locally, everything works (Travis skips the GPU tests). And the `difflib` feature is excellent :)

The only open question is (@benanne, @ebenolson): Do we want to warn about the "deterministic" keyword for a network without dropouts/noise or do we want to allow "deterministic" no matter what? This could be implemented by changing [a single line](https://github.com/Lasagne/Lasagne/pull/585/files#diff-84ce884e5f993565a6a1db7b062daa20R154), so it's just a question of what we'd prefer.

An argument for always allowing "deterministic" is that it would encourage people to always pass `deterministic=True` when creating the validation/test graph, making it less likely to be forgotten. If there's a warning as long as you don't use dropout layers, people may remove it (possibly without understanding) and not add it back later. What do you think?

ebenolson:
I'm in favor of letting `deterministic` through - I don't think you should need to modify training loop code due to changes in the network definition. But as it's only a warning I'm ok with either case.

benanne:
I think @ebenolson's argument for letting it through is compelling: I like to keep my training loop code and actual model definitions in separate files, and I wouldn't want to have to touch the training loop code to get rid of a pesky warning, just because I happen to have a model without dropout layers :)

Joshua-Chin:
I added `deterministic` as an accepted kwarg and squashed the commits again.

f0k:
Perfect, thanks again for your work and for your patience! Let's finally merge this!

PR #586, opened 2016-01-23
Implemented thin plate spline transformation layer #579

### LINKED ISSUE #579: Thin Spline Transform Transformer Layer
Are there any plans to implement the thin spline transform used in the Jaderberg paper?

arimorcos:
This code implements the thin plate spline spatial transformation layer as described in the Jaderberg NIPS 2015 paper. 

Closes #579

### DIFF
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -1,5 +1,6 @@
 import theano
 import theano.tensor as T
+import numpy as np
 
 from .. import init
 from .. import nonlinearities
@@ -17,6 +18,7 @@ __all__ = [
     "ExpressionLayer",
     "InverseLayer",
     "TransformerLayer",
+    "TPSTransformerLayer",
     "ParametricRectifierLayer",
     "prelu",
     "RandomizedRectifierLayer",
@@ -376,9 +378,10 @@ class TransformerLayer(MergeLayer):
 
     References
     ----------
-    .. [1]  Spatial Transformer Networks
-            Max Jaderberg, Karen Simonyan, Andrew Zisserman, Koray Kavukcuoglu
-            Submitted on 5 Jun 2015
+    .. [1]  Max Jaderberg, Karen Simonyan, Andrew Zisserman,
+            Koray Kavukcuoglu (2015):
+            Spatial Transformer Networks. NIPS 2015,
+            http://papers.nips.cc/paper/5854-spatial-transformer-networks.pdf
 
     Examples
     --------
@@ -425,10 +428,10 @@ class TransformerLayer(MergeLayer):
     def get_output_for(self, inputs, **kwargs):
         # see eq. (1) and sec 3.1 in [1]
         input, theta = inputs
-        return _transform(theta, input, self.downsample_factor)
+        return _transform_affine(theta, input, self.downsample_factor)
 
 
-def _transform(theta, input, downsample_factor):
+def _transform_affine(theta, input, downsample_factor):
     num_batch, num_channels, height, width = input.shape
     theta = T.reshape(theta, (-1, 2, 3))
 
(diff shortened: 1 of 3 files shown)


arimorcos:
Ok! It took a few tries to get all the tests to run properly, but should be ready now. 

@skaae, looking forward to hearing what you think.

benanne:
You can run the tests locally: http://lasagne.readthedocs.org/en/latest/user/development.html#testing
It's a lot quicker than waiting for Travis, usually :)

arimorcos:
Haha, thanks. I had been running the individual class using the -k argument, and I guess that caused it to skip the pep8 checks.

benanne:
Oh, I see! That makes sense.

skaae:
> @skaae, looking forward to hearing what you think.

I think it looks awesome :) I have to admit that it'll take me a while to understand the code. 

Could we try to reproduce some of the MNIST results from the SPN paper or have you already tried that?

benanne:
I haven't read the code in detail, but are we sure we want all of this in a single class? It might make things quite hard to read (especially because one of the "branches", i.e. the regular affine ST layer is much easier to understand than the other). If there is a lot of shared code perhaps we should be using inheritance instead?

arimorcos:
@skaae Thanks! Unfortunately, the math behind thin plate splines is not super easy to understand from the [main paper](http://user.engineering.uiowa.edu/~aip/papers/bookstein-89.pdf). But the meat of it is all in section 1D.

I haven't tried reproducing the MNIST results. I just performed checks to ensure that it's efficiently and accurately applying the warp. It should be trivial to just plug it in to [the notebook in recipes](https://github.com/Lasagne/Recipes/blob/master/examples/spatial_transformer_network.ipynb) and then compare the performance of the affine layer to the thin plate spline layer on the elastic warped MNIST images. The elastic warp from the STN paper is just scaling followed by a thin plate spline warp. I'm going to be pretty busy over the next week or two, but I'll see if I can find some time late at night. ;) 

@benanne All the shared code is currently implemented as helper functions in special.py, so it wouldn't be challenging to break them up.

f0k:
Looks like a good start, and a nice addition to Lasagne!

Haven't read the code in detail either, but my first thought was also to split this into two classes, since there's not a lot of shared code and some of the constructor arguments only apply to thin-plate splines. Sorry I didn't participate in #579!
Can you split it into a separate class? It probably doesn't even make sense to inherit from `TransformerLayer` since you can't call the super constructor (due to its input shape checks). On a side note, your PR contains a lot of unrelated white space changes -- please configure your editor to not auto-format the full file! Thanks :)

> It should be trivial to just plug it in to the notebook in recipes and then compare the performance of the affine layer to the thin plate spline layer on the elastic warped MNIST images.

Would surely be nice to have a working example in Recipes as well! Either as part of the existing Recipe, or as a separate one, whichever is easier.

In any case, thanks for your work so far!

skaae:
I think the `_Interpolate` function is shared? But we can seperate that out as a helper function?

arimorcos:
`_interpolate`, `_linsapce`, and `_meshgrid` are all required for both the affine and thin plate spline layers, but they are already factored out, so no challenge there. 

@f0k I think I was getting pep-8 errors from some of the unrelated classes, but I can change those and try the checks again. 

I'll go ahead and split the transformer into two classes tonight. What do you guys think about naming? We probably don't want to change the name of the current layer. So `TransformerLayer` (affine) and `ThinPlateSplineTransformerLayer`? It's a little long, so we could use `TPSTransformerLayer`.

f0k:
> What do you guys think about naming?

Ah yes, forgot that again! `ThinPlateSplineTransformerLayer` came to my mind first as well, but I like `TPSTransformerLayer` better. FWIW, they also use "TPS" throughout [the paper](http://papers.nips.cc/paper/5854-spatial-transformer-networks.pdf) you mentioned in the first post (well, it's the obvious abbreviation).

arimorcos:
> ...I like TPSTransformerLayer better.

I'm inclined to agree. I'll go ahead and change it to that. I'll update the PR later tonight.

arimorcos:
Ok, separated them into separate classes. Sorry for the barrage of commits. I made a few silly errors.

lasagne/layers/special.py
```diff
@@ -447,11 +449,11 @@ def _transform(theta, input, downsample_factor):
     # dimshuffle input to  (bs, height, width, channels)
     input_dim = input.dimshuffle(0, 2, 3, 1)
     input_transformed = _interpolate(
-        input_dim, x_s_flat, y_s_flat,
-        out_height, out_width)
+            input_dim, x_s_flat, y_s_flat,
+            out_height, out_width)
```
f0k: Can you undo this whitespace change please?

lasagne/layers/special.py
```diff
@@ -447,11 +449,11 @@ def _transform(theta, input, downsample_factor):
...
 
-    output = T.reshape(
-        input_transformed, (num_batch, out_height, out_width, num_channels))
+    output = T.reshape(input_transformed,
+                       (num_batch, out_height, out_width, num_channels))
```
f0k: As well as this whitespace change?

lasagne/layers/special.py
```diff
@@ -544,6 +546,296 @@ def _meshgrid(height, width):
...
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape. The
+        output of this layer should be a 4D tensor, with shape
+        ``(batch_size, num_input_channels, input_rows, input_columns)``.
+
+    localization_network : a :class:`Layer` instance
+        The network that calculates the parameters of the thin plate spline
+        transformation as the x and y coordinates of the destinations of each
+        control point. The output of the localization network must be twice
+        the number of control points.
```
f0k: Maybe a little more explicit: The output of the localization network must be a 2D tensor, with shape ``(batch_size, 2 * num_control_points)``.

lasagne/layers/special.py
```diff
@@ -544,6 +546,296 @@ def _meshgrid(height, width):
...
+    num_control_points : integer
+        The number of control points to be used for the thin plate spline
+        transformation. These points will be arranged as a grid along the
+        image, so the value must be a perfect square. Default is 16.
+
+    References
+    ----------
+    .. [1]  Spatial Transformer Networks
+            Max Jaderberg, Karen Simonyan, Andrew Zisserman, Koray Kavukcuoglu
+            Submitted on 5 Jun 2015
```
f0k: Could you fix this reference while we're at it? I know you copied it from `TransformerLayer`, but "Submitted on 5 Jun 2015" is not actually a valid source :)
We're using the format "Authors (Year): Title. Source" for most references, e.g.: https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/noise.py#L45-L48
So here it could be:
```
Max Jaderberg, Karen Simonyan, Andrew Zisserman, Koray Kavukcuoglu (2015):
Spatial Transformer Networks. NIPS 2015,
http://papers.nips.cc/paper/5854-spatial-transformer-networks.pdf
```
Thanks!

lasagne/layers/special.py
```diff
@@ -544,6 +546,296 @@ def _meshgrid(height, width):
...
+    ...                                  np.linspace(-1, 1, grid_size))
+    >>> # offset for the control points
+    >>> x_offset = 0
+    >>> y_offset = 0
+    >>> x_control_dest = x_source.flatten() + x_offset
+    >>> y_control_dest = y_source.flatten() + y_offset
+    >>> dest_points = np.vstack((x_control_dest, y_control_dest)).flatten()
+    >>>
+    >>> # We'll set the bias to be the source points
+    >>> b = np.vstack((x_source.flatten(), y_source.flatten())).flatten()
```
f0k: This seems like a pretty complicated setup. Wouldn't it be easier to have the localization network predict the *offsets* instead of the positions? This way, a network outputting zeros would correspond to the identity transform.

arimorcos: This is a good point. The only worry I had about that strategy was that it delinks the localization network output from the source points and might lead people to think that the offsets would generalize to different grid configurations. For ease, I implemented it such that the control points are on a grid, but there's no reason for that to be true, and one could actually use the control point configuration as a way to express a prior about the type of transformation. (For example, if one's prior was that horizontal warps were more likely than vertical warps you could position the grid as a vertically-oriented rectangle, providing for finer horizontal warps and coarser vertical warps.)

That said, now that I think about it a little more, I don't actually think that keeping them as destination points does much to combat that misunderstanding and offsets are definitely easier to initialize.

arimorcos: Also, looking at this example again, I realize I actually don't even use `dest_points`. `dest_points` should be used to initialize the bias so that the network starts with a given transform. I'll change to offsets and fix that.

f0k: > (For example, if one's prior was that horizontal warps were more likely than vertical warps you could position the grid as a vertically-oriented rectangle, providing for finer horizontal warps and coarser vertical warps.)

That was the first thing I thought of when reading the docstring for `num_control_points`. It would be nice to eventually allow specifying a tuple of two ints. I'm not sure whether we should change the interface to cater for that possible future addition, I think `num_control_points=(10,5)` for a 10x5 grid would read well enough. That's why I didn't say anything :)

For curiosity, is there anything in the code that makes use of the fact that it's a square grid, or a grid at all? Or would it be easy to allow the control points to be initialized to any sequence of coordinates in (-1, 1)?

arimorcos: I like the `control_points` idea you listed. Looking through the code actually, I think it shouldn't be so hard to implement that. I have to give a presentation tomorrow, so I won't have time before then, but let me see if I can get that working tomorrow afternoon.

arimorcos:
As an aside, @benanne, congrats on mastering Go!

lasagne/layers/special.py
f0k: > Looking through the code actually, I think it shouldn't be so hard to implement that. I have to give a presentation tomorrow, so I won't have time before then, but let me see if I can get that working tomorrow afternoon.

No need to hurry! If you think this can be done, let's just rename the argument to cater for this. It doesn't have to be part of the initial PR! Good luck with your presentation if it's not done already, otherwise I hope it went well :)

arimorcos:
Ok, I've updated the parameter `num_cluster_points` to just be `cluster_points`. In a future PR, I'll change it to be more flexible. Are there any other suggested changes I've missed? 

And thanks @f0k! The presentation went well. I'm a neuroscientist with a passion for ML, so getting experimental neuroscientists to internalize ML concepts is always a bit hit or miss.

lasagne/layers/special.py
```diff
@@ -179,7 +181,7 @@ def get_output_for(self, input, **kwargs):
         return input * self.scales.dimshuffle(*pattern)
 
 
-def standardize(layer, offset, scale, shared_axes):
+def standardize(layer, offset, scale, shared_axes='auto'):
```
f0k: This was actually part of another PR that you may have accidentally copied in... seems it doesn't impede mergeability, but maybe it's better to remove. Don't know exactly if it will change the history in any way. Please wait until you read my upcoming comment on squashing commits, though :)

lasagne/layers/special.py
```diff
@@ -544,6 +547,291 @@ def _meshgrid(height, width):
...
+    ----------
+    .. [1]  Max Jaderberg, Karen Simonyan, Andrew Zisserman,
+            Koray Kavukcuoglu (2015):
+            Spatial Transformer Networks. NIPS 2015,
+            http://papers.nips.cc/paper/5854-spatial-transformer-networks.pdf
+    .. [2]  Fred L. Bookstein (1989):
+            Principal warps: thin-plate splines and the decomposition of
+            deformations. IEEE Transactions on
+            Pattern Analysis and Machine Intelligence.
+            http://doi.org/10.1109/34.24792
```
f0k: You're missing a newline here (so there's a blank line before `Examples`). I think it may not render correctly otherwise.

lasagne/layers/special.py
```diff
@@ -544,6 +547,291 @@ def _meshgrid(height, width):
...
+    make everything invariant to the shape of the image, the x and y range
+    of the image is normalized to [-1, 1] as in ref [1]_. To replicate an
+    identity transform, we'll simply set the offsets to 0. More complicated
+    transformations can easily be implemented using different x and y offsets
+    (importantly, each control point can have it's own pair of offsets).
+
+    >>> import numpy as np
+    >>> import lasagne
+    >>>
+    >>> # We'll set the bias to be the all offsets of 0
```
f0k: This reads funny. Maybe "We'll set the bias to have all offsets be zero"

lasagne/layers/special.py
```diff
@@ -544,6 +547,291 @@ def _meshgrid(height, width):
...
+    >>> num_points = 16
+    >>> b = np.zeros(2*num_points)
+    >>>
+    >>> # Create the network
+    >>> W = lasagne.init.Constant(0.0)
+    >>> l_in = lasagne.layers.InputLayer((None, 3, 28, 28))
+    >>> l_loc = lasagne.layers.DenseLayer(l_in, num_units=2*num_points,
+    ...                                   W=W, b=b, nonlinearity=None)
```
f0k: Thinking again, there's no need to set `b` to a numpy array, you can just set `b = lasagne.init.Constant(0)`, and that's the default for dense layers. Maybe leave out the bias initialization above, and say:
```
# Create the network
# we'll initialize the weights and biases to zero, so it defaults
# to the identity transform (all control point offsets are zero)
>>> W = b = lasagne.init.Constant(0.0)
```

lasagne/layers/special.py
```diff
@@ -544,6 +547,291 @@ def _meshgrid(height, width):
...
+        The layer feeding into this layer, or the expected input shape. The
+        output of this layer should be a 4D tensor, with shape
+        ``(batch_size, num_input_channels, input_rows, input_columns)``.
+
+    localization_network : a :class:`Layer` instance
+        The network that calculates the parameters of the thin plate spline
+        transformation as the x and y coordinates of the destination offsets of
+        each control point. The output of the localization network must
+        should be a 2D tensor, with shape
+        ``(batch_size, 2 * num_control_points)``
```
f0k: remove `must` (and then `should be` should fit into the previous line, and maybe you can avoid the line break before `shape`)

lasagne/tests/layers/test_special.py
```diff
@@ -400,19 +498,48 @@ def test_transform_downsample():
...
+        )
+        image = np.zeros(shape=(28, 28))
+        image[[0, -1], :] = 1
+        image[:, [0, -1]] = 1
+        inputs = np.tile(image, (batchsize, 3, 1, 1))
+        shifted_input = np.ones(shape=(28, 28))
+        shifted_input[:13, :13] = 0
+        shifted_input[13, :13] = 0.50000271
+        shifted_input[:13, 13] = 0.50000271
+        shifted_input[13, 13] = 0.75000271
```
f0k: Just checking: These *are* the values to be expected, not `0.5` and `0.75`?

f0k:
> Are there any other suggested changes I've missed?

No! But I've got a last round of (very minor) comments. When fixing them, after committing and before doing `git push`, could you please squash your commits? The easiest way (in terms of user-friendliness) seems to be running `git rebase --interactive upstream/master`. This will give you a list with all your commits in a text editor. Leave the first one as it is, and change the other ones from `pick` to `fixup`, as explained in the help text you'll see as a comment in the editor. Save and close, and git will do its magic. In the end, run `git push --force` to upload your state of affairs, replacing the commits of the current PR (as they're no longer part of the history).
If anything goes wrong, let me know (everything that was committed can be recovered).

lasagne/tests/layers/test_special.py
arimorcos: Yep, I'm pretty sure (though not 100% positive). I think it's due to the some minor rounding in the interpolation. I know for certain that the transformed points for the new grid are correct though because I've checked them against another [library](https://github.com/olt/thinplatespline).

arimorcos:
It failed two tests for the standardize layer because I removed the 'shared_axes' argument. Other than that, it's ready to go. What should I do?

benanne:
Thanks :) Very nice work by the way, thanks a lot for contributing! Not sure what to do about the tests... there seems to be a weird merge-related issue here that I can't wrap my head around. Maybe @f0k can!

arimorcos:
Thanks! 

I think that just adding back in the default `shared_axes` argument should fix the tests. If the change is already present in lasagne's master branch, I don't think it should cause a merge conflict.

benanne:
Sounds good to me!

arimorcos:
Ok! Looks like that fixed it. Should be ready to go I think.

f0k:
> I think that just adding back in the default `shared_axes` argument should fix the tests. If the change is already present in lasagne's master branch, I don't think it should cause a merge conflict.

To solve the riddle: When you did `git rebase --interactive upstream/master`, your commits were replayed on top of the upstream master, so they included Colin's addition of `shared_axes` and the accompanying new tests. Keeping `shared_axes` was the correct thing to do. In the diff for this PR you can see that `shared_axes` is not mentioned as part of your changeset any longer.

Thanks a lot for your work, merging in 3.. 2... 1...!

PR #588, opened 2016-01-25
Don't use T.inv in standardize

### LINKED ISSUE #583: When using standardize, scales doesn't get returned as a parameter
```Python
In [1]: net = lasagne.layers.standardize(
   ...:     lasagne.layers.InputLayer((10, 20)), np.zeros(20), np.ones(20), 0)

In [2]: lasagne.layers.get_all_params(net)
Out[2]: [b]

In [3]: net.params
Out[3]: OrderedDict([(Elemwise{inv,no_inplace}.0, set())])

In [4]: net.get_params()
Out[4]: []

In [5]: theano.printing.debugprint(net.scales)
Elemwise{inv,no_inplace} [id A] ''
 |TensorConstant{(20,) of 1.0} [id B]
```

Maybe because it's using [`T.inv` on a `np.ndarray`](https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/special.py#L232) (which gets converted to a `TensorConstant`, not a `SharedVariable`), so `collect_shared_vars` misses it?  I can't think of why someone would not want `scales` to be considered a parameter or to actually be a non-shared Theano variable, so I think requiring the user to pass in a shared variable or a `np.ndarray` is reasonable, and then just use `1./scale` instead of `T.inv(scale)`.

craffel:
Using `T.inv` with `type(scale) = np.ndarray` resulted in `scales` not being returned by `get_all_params` because it would be converted to a Theano constant, not a shared variable, so `collect_all_shared` missed it.  Here, I've changed to `utils.floatX(1.)/scale` instead.  This means `standardize` can't accept Theano expressions, which is fine, that is sort of a crazy use-case for the intended purpose.  I also added a simple additional test to make sure the problem is fixed.  Should resolve #583

### DIFF
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -3,7 +3,7 @@ import theano.tensor as T
 
 from .. import init
 from .. import nonlinearities
-from ..utils import as_tuple
+from ..utils import as_tuple, floatX
 from ..random import get_rng
 from .base import Layer, MergeLayer
 from theano.sandbox.rng_mrg import MRG_RandomStreams as RandomStreams
@@ -194,10 +194,10 @@ def standardize(layer, offset, scale, shared_axes='auto'):
     ----------
     layer : a :class:`Layer` instance or a tuple
         The layer feeding into this layer, or the expected input shape.
-    offset : Theano shared variable, expression, or numpy array
+    offset : Theano shared variable or numpy array
         The offset to apply (via subtraction) to the axis/axes being
         standardized.
-    scale : Theano shared variable, expression or numpy array
+    scale : Theano shared variable or numpy array
         The scale to apply (via division) to the axis/axes being standardized.
     shared_axes : 'auto', int or tuple of int
         The axis or axes to share the offset and scale over. If ``'auto'`` (the
@@ -229,7 +229,7 @@ def standardize(layer, offset, scale, shared_axes='auto'):
     # Do not optimize the offset parameter
     layer.params[layer.b].remove('trainable')
     # Divide by the scale
-    layer = ScaleLayer(layer, T.inv(scale), shared_axes)
+    layer = ScaleLayer(layer, floatX(1.)/scale, shared_axes)
     # Do not optimize the scales parameter
     layer.params[layer.scales].remove('trainable')
     return layer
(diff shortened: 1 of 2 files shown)


benanne:
the MNIST example fails because yann.lecun.com was down... how do you retrigger Travis again? :)

f0k:
> how do you retrigger Travis again? :)

There's a "restart build" button when you're logged in and view a failed build (it's a round button with a "reload"-like arrow). I've clicked it. Otherwise, closing and reopening the PR also works (that's the way to go for a submitter who doesn't have full rights on Travis).

f0k:
> This means `standardize` can't accept Theano expressions

Really? Why? `np.array(1) / some_expression` should be a valid Theano expression. Same goes for the bias. Is there anything I'm missing?

craffel:
> Really? Why? np.array(1) / some_expression should be a valid Theano expression. Same goes for the bias. Is there anything I'm missing?

Because then the params won't get returned by `get_all_params`.

f0k:
> Because then the params won't get returned by `get_all_params`.

Depends on the expression, of course -- only shared variables included in the expression will be returned. But I'm fine with only advertising numpy arrays and shared variables if you prefer.

craffel:
Oh, this is the problem again where I forget what the term is for "a Theano expression with no shared variables".  That's the thing we don't want to encourage.  I was considering advertising compatibility with shared variables to include "Theano expressions with only shared variables".

f0k:
> Oh, this is the problem again where I forget what the term is for "a Theano expression with no shared variables".

Don't think there is one :)

> That's the thing we don't want to encourage.

... under the assumption that everybody using `standardize()` needs the offset and scale to be delivered by `get_all_param_values()`. I agree that using `T.inv()` was a bad idea because it unnecessarily hides things from `get_all_param_values()`, but I don't think it's vital that everything you supply becomes part of the network params. Sometimes the sole use of `standardize()` could be to have the standardization be part of the forward pass and performed on GPU.

Anyway, I'm fine with the current PR! Will merge tomorrow if there are no further comments (and nothing unforeseeable happens).

craffel:
> Sometimes the sole use of standardize() could be to have the standardization be part of the forward pass and performed on GPU.

Yes, I wasn't thinking of it that way, I was thinking of it as a way to avoid having to separately save training set statistics alongside your network param files.

craffel:
Bump.

f0k:
Oops! Sorry and thanks for the reminder!

PR #589, opened 2016-01-26
Try alternative URLs for MNIST dataset download

### LINKED ISSUE #587: yann.lecun.com is down --> examples/mnist.py does not work
When running the mnist.py example, it can't download the required data files since yann.lecun.com seems to be down. I just spent 30 minutes searching for an alternate location, but there don't seem to be any direct mirrors (all other mnist data sets I could find seemed to use a different file format). This is probably just a temporary glitch, but unfortunate nevertheless. Would it make sense to package these files with the repo instead (according to the cached version of Yann LeCun's page they're only a few megabytes in size)? Or are there other known mirrors? If so, would be great to list them in the source code somewhere, to make it easy to swap them in when needed.

f0k:
Fixes #587. But it makes things a bit ugly... I tend towards holding back until download availability really becomes a problem.
Note: This will work on Python 2 and 3 on an English Ubuntu, but I don't know if it's portable to other languages or Windows, Mac. Would be interesting to know what Exception exactly `urlretrieve('http://foo.bar', 'foo')` throws on other systems.

### DIFF
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -33,12 +33,28 @@ def load_dataset():
     # We first define a download function, supporting both Python 2 and 3.
     if sys.version_info[0] == 2:
         from urllib import urlretrieve
+        URLError = IOError
     else:
-        from urllib.request import urlretrieve
+        from urllib.request import urlretrieve, URLError
 
-    def download(filename, source='http://yann.lecun.com/exdb/mnist/'):
+    def download(filename,
+                 sources=['http://yann.lecun.com/exdb/mnist/',
+                          'https://web.archive.org/web/20150906081542/'
+                          'http://yann.lecun.com/exdb/mnist/',
+                          'https://s3.amazonaws.com/lasagne/recipes/'
+                          'datasets/mnist/']):
         print("Downloading %s" % filename)
-        urlretrieve(source + filename, filename)
+        for source in sources:
+            try:
+                urlretrieve(source + filename, filename)
+            except URLError as exc:
+                if exc.args[-1].args == (-2, 'Name or service not known'):
+                    sources.remove(source)  # do not try this source again
+            else:
+                break  # we were successful!
+        else:
+            print("Could not download from any location. Are you offline?")
+            raise RuntimeError()
 
     # We then define functions for loading MNIST images and labels.
     # For convenience, they also download the requested files if needed.


liebald:
Instead of building your own redundancy scheme in code, why not simply point to a cloud-based storage location, such as the S3 bucket in your list (or alternatively any Dropbox/Gdrive/CDN)? These should have high enough availability to basically make this a non-issue (as opposed to yann.lecun.com, which seems to be hosted somewhere at NYU, perhaps just a single machine sitting under someone's desk)

f0k:
> why not simply point to a cloud-based storage location, such as the S3 bucket in your list

Well, firstly because we (@ebenolson, actually) pay for S3 and secondly because I'd prefer to include the official and probably most permanent location. Not sure if the costs for S3 would noticeable, but people use the example code as a basis for other experiments and so the link could spread quite a bit.

It's the first time I've seen LeCun's website being down, by the way -- from my (limited) experience, it's highly available. That's why I propose to just leave it the way it is for now.

As a compromise, we could just list the alternative download locations in a comment so they're easy to swap in (this wouldn't help with continuous integration, but at least make it easier for end users).

examples/mnist.py
```diff
@@ -33,12 +33,28 @@ def load_dataset():
...
+                          'https://s3.amazonaws.com/lasagne/recipes/'
+                          'datasets/mnist/']):
         print("Downloading %s" % filename)
-        urlretrieve(source + filename, filename)
+        for source in sources:
+            try:
+                urlretrieve(source + filename, filename)
+            except URLError as exc:
+                if exc.args[-1].args == (-2, 'Name or service not known'):
+                    sources.remove(source)  # do not try this source again
```
craffel: Keep in mind `sources` is mutable.  I think you might be doing this intentionally, but it's setting off all my Python gotcha alarms.

examples/mnist.py
```diff
@@ -33,12 +33,28 @@ def load_dataset():
...
+            else:
+                break  # we were successful!
+        else:
+            print("Could not download from any location. Are you offline?")
```
craffel: Why not `raise RuntimeError("Could not download from any location. Are you offline?")`

examples/mnist.py
f0k: Thought the comment would explain. But yes, it's on purpose. If the primary location is down when downloading the first file, I don't want it to be tried again for the next three downloads. I didn't call it ugly for no reason ;) We could move the sources list into the outer context to make it clearer.

examples/mnist.py
f0k: I figured it would be easier to spot if it comes right after the "Downloading %s" line, instead of being mixed in somewhere with the exception backtrace.

PR #592, opened 2016-01-29
don't use O(N^2) algorithm in get_all_params

Joshua-Chin:
`sum` allocates space for all the intermediary lists. For timings, see [this post](http://stackoverflow.com/questions/952914/making-a-flat-list-out-of-list-of-lists-in-python/952952#952952) on StackOverflow.

### DIFF
--- a/lasagne/layers/helper.py
+++ b/lasagne/layers/helper.py
@@ -1,4 +1,5 @@
 from collections import deque
+from itertools import chain
 
 import theano
 import numpy as np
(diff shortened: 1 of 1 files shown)


f0k:
Yes, good idea! Not likely to be a performance issue, but as `utils.unique` just requires an iterable, it's not necessary to construct a list first. Merging right away as it's a trivial change that won't spark a discussion. Thank you!

benanne:
Thanks :)

PR #594, opened 2016-01-31
Additional test cases for hid_init as a Layer in recurrent layers.

almath123:
I have written test cases for Recurrent, LSTM and GRU layers which pass some dummy inputs through `hid_init` (and cell_init where applicable) as a Layer and check the output. Essentially I check that you get the same result from passing a dummy input as an InputLayer and a np.array.

The need for this sort of test case was mentioned in PR #522

### DIFF
--- a/lasagne/tests/layers/test_recurrent.py
+++ b/lasagne/tests/layers/test_recurrent.py
@@ -104,6 +104,53 @@ def test_recurrent_hid_init_mask():
     output = lasagne.layers.get_output(l_rec, inputs)
 
 
+def test_recurrent_hid_init_layer_eval():
+    # Test `hid_init` as a `Layer` with some dummy input. Compare the output of
+    # a network with a `Layer` as input to `hid_init` to a network with a
+    # `np.array` as input to `hid_init`
+    n_units = 7
+    n_test_cases = 2
+    in_shp = (n_test_cases, 2, 3)
+    in_h_shp = (1, n_units)
+
+    # dummy inputs
+    X_test = np.ones(in_shp, dtype=theano.config.floatX)
+    Xh_test = np.ones(in_h_shp, dtype=theano.config.floatX)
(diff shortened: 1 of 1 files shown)


lasagne/tests/layers/test_recurrent.py
```diff
@@ -104,6 +104,57 @@ def test_recurrent_hid_init_mask():
...
+    output_nparray = lasagne.layers.get_output(l_rec_nparray, {l_inp: X})
+
+    # test both nets with dummy input
+    output_val_inp_layer = output_inp_layer.eval({X: X_test,
+                                                  Xh: Xh_test_batch})
+    output_val_nparray = output_nparray.eval({X: X_test})
+
+    # check output has the right type
+    assert isinstance(output_val_inp_layer, np.ndarray)
+    assert isinstance(output_val_nparray, np.ndarray)
```
f0k: Not sure about this -- won't the result be a `CudaNdarray` when things run on GPU? `np.allclose()` won't complain because calling `np.asarray()` on a `CudaNdarray` will copy the matrix to CPU, but the `isinstance()` check would fail. Can you remove it, or is there a special reason for it?

almath123:
I have removed the `isinstance()` checks on the output of `eval()` as it is not really necessary. I did, however, test it on the GPU and it passed. I'm assuming that when `eval()` is called it copies the output values back to local memory, though it wouldn't surprise me if some combination of `borrow` arguments would lead to a `CudaNdarray` being returned.

f0k:
Thanks, can you please squash the two commits into one?

almath123:
No problem. Commits squashed.

f0k:
I'm fine with merging! @craffel or @skaae, masters of `lasagne.layers.recurrent`, are you okay as well?

craffel:
Looks reasonable to me.

f0k:
All right, merging! Thanks again!

Issue #317, opened 2015-07-02
Sparse input to Lasagne


Ivanopolo:
I'm currently working on a problem that I want to solve with multilayered perceptron, yet the data is very sparse and it would be much simpler if Lasagne had an InputLayer that would accept sparse input. Theano does have the needed theano.sparse module that works with csr / csc matrices from scipy. What is the easiest way to use sparse matrices as an input in Lasagne? 

The only idea I came up with is to reimplement InputLayer + Dense layers so that they would accept sparse matrices and employ sparse operations introduced by Theano, but it seems like an overkill. What would you recommend? Is there any PR that I missed with needed functionality? Would that be interesting for a PR if I implemented the mentioned Layers?

f0k:
> The only idea I came up with is to reimplement InputLayer + Dense layers so that they would accept sparse matrices and employ sparse operations introduced by Theano, but it seems like an overkill.

I guess there are two possible solutions:

1. Create a `lasagne.layers.sparse` submodule with layers that accept sparse input and possibly produce sparse output, using `theano.sparse`
2. Adapt some of the existing layers (`InputLayer`, `DenseLayer`, but probably not `Conv2DLayer`) such that they accept both dense and sparse input and document that

The first solution is about what you proposed, and I don't think it's overkill. It nicely separates the sparse layers from regular ones, and allows to freely choose which layers to (re-)implement with sparsity support, possibly derived from their dense counterparts. The second solution has the potential to turn into a mess.

> Would that be interesting for a PR if I implemented the mentioned Layers?

Hmm, I'm not sure if there's enough need for that. If we put it into a `lasagne.layers.sparse` submodule that is not auto-imported, it wouldn't affect mainland Lasagne that much, though (we'd only be stuck with maintaining it in case we change the API in any way). For a start, I'd suggest you to implement your layers locally, outside of the Lasagne source tree.

Side question you should investigate: What happens if you run with `device=gpu` and have a `theano.sparse.dot(sparse_input, W)` with a shared float32 variable `W` living on the GPU? Will it be copied to CPU for each dot product or is there a way to force it to CPU once? I'm thinking that you might want to combine CPU (for sparse computations) and GPU (for dense computations).

elanmart:
For anyone who finds this thread in the future:
I've asked a question about keeping some shared vars on the CPU and got the following [answer](https://groups.google.com/forum/#!topic/theano-users/z-SmcsubLGo):
[...]

PR #595, opened 2016-02-01
Sparse input to Lasagne #317

eduardo4jesus:
I recently started to work with Natural Language Processing, which contains lots of sparse data. After noticing that Lasagne doesn't accept sparse inputs for its layers. I quickly realized that I would need to make some adjusts to make it work for my application. Thankfully, Theano does support sparse representation. I took me just a few tweaks.

Later, I found that there is an open issue regarding [Sparse input to Lasagne #317](https://github.com/Lasagne/Lasagne/issues/317). In this issue, it was suggested to create a `lasagne.layers.sparse` submodule. However, since it is just about using sparse variable as inputs, I thought more convenient to just adapt the `DenseLayer`. After all the Layer will be still produce a dense model `W`.

So, for a quick start, I adapted the Dense Layer and the Dropout to work with sparse inputs, this way I was able to use a network exactly* similar to the MLP in the MNIST example below. (* I just changed the input `shape` ). 

```
# ##################### Build the neural network model #######################
# This script supports three types of models. For each one, we define a
# function that takes a Theano variable representing the input and returns
# the output layer of a neural network model built in Lasagne.

def build_mlp(num_features, num_targets, input_var=None):
    # This creates an MLP of two hidden layers of 800 units each, followed by
    # a soft_max output layer of 10 units. It applies 20% dropout to the input
    # data and 50% dropout to the hidden layers.

    # Input layer, specifying the expected input shape of the network
    # (unspecified batch size, number of features) and
    # linking it to the given Theano variable `input_var`, if any:
    l_in = lasagne.layers.InputLayer(shape=(None, num_features),
                                     input_var=input_var)

    # Apply 20% dropout to the input data:
    l_in_drop = lasagne.layers.DropoutLayer(l_in, p=0.2)

    # Add a fully-connected layer of 800 units, using the linear rectifier, and
    # initializing weights with Glorot's scheme (which is the default anyway):
    l_hid1 = lasagne.layers.DenseLayer(
            l_in_drop, num_units=800,
            nonlinearity=lasagne.nonlinearities.rectify,
            W=lasagne.init.GlorotUniform())

    # We'll now add dropout of 50%:
    l_hid1_drop = lasagne.layers.DropoutLayer(l_hid1, p=0.5)

    # Another 800-unit layer:
    l_hid2 = lasagne.layers.DenseLayer(
            l_hid1_drop, num_units=800,
            nonlinearity=lasagne.nonlinearities.rectify)

    # 50% dropout again:
    l_hid2_drop = lasagne.layers.DropoutLayer(l_hid2, p=0.5)

    # Finally, we'll add the fully-connected output layer, of 10 softmax units:
    l_out = lasagne.layers.DenseLayer(
            l_hid2_drop, num_units=num_targets,
            nonlinearity=lasagne.nonlinearities.softmax)

    # Each layer is linked to its incoming layer(s), so we only need to pass
    # the output layer to give access to a network in Lasagne:
    return l_out
```
 
I also have to tweak the Theano variable for the inputs. In my case I my dataset uses `scipy.sparse.csr_matrix`.

```
# Prepare Theano variables for inputs and targets. Here we use sparse matrices instead of tensors
    input_var = theano.sparse.csr_matrix(name='inputs', dtype='float32')
    target_var = T.ivector('targets')
```

With this step up I was able to use sparse inputs.

### DIFF
--- a/lasagne/layers/dense.py
+++ b/lasagne/layers/dense.py
@@ -87,7 +87,7 @@ class DenseLayer(Layer):
 
         # It needs a proper call in case the input is a SparseVariable.
         # Though it might be the case, the layer activation will remain
-        # dense since W represents a dense matrix. 
+        # dense since W represents a dense matrix.
         if type(input) == sparse.basic.SparseVariable:
             activation = sparse.basic.dot(input, self.W)
         else:
--- a/lasagne/layers/dense.py
+++ b/lasagne/layers/dense.py
@@ -1,5 +1,6 @@
 import numpy as np
 import theano.tensor as T
+import theano.sparse as sparse
 
 from .. import init
 from .. import nonlinearities
@@ -84,7 +85,14 @@ class DenseLayer(Layer):
             # batch of feature vectors.
             input = input.flatten(2)
 
-        activation = T.dot(input, self.W)
+        # It needs a proper call in case the input is a SparseVariable.
+        # Though it might be the case, the layer activation will remain
+        # dense since W represents a dense matrix. 
+        if type(input) == sparse.basic.SparseVariable:
+            activation = sparse.basic.dot(input, self.W)
+        else:
+            activation = T.dot(input, self.W)
+
         if self.b is not None:
             activation = activation + self.b.dimshuffle('x', 0)
         return self.nonlinearity(activation)
--- a/lasagne/layers/noise.py
+++ b/lasagne/layers/noise.py
@@ -73,6 +73,7 @@ class DropoutLayer(Layer):
         else:
             retain_prob = 1 - self.p
             if self.rescale:
+                # It needs a proper call in case the input is a SparseVariable
                 if type(input) == sparse.basic.SparseVariable:
                     input = sparse.basic.mul(input, 1/retain_prob)
                 else:
--- a/lasagne/layers/noise.py
+++ b/lasagne/layers/noise.py
@@ -1,4 +1,5 @@
 import theano
+import theano.sparse as sparse
 
 from .base import Layer
 from ..random import get_rng
@@ -72,7 +73,10 @@ class DropoutLayer(Layer):
         else:
             retain_prob = 1 - self.p
             if self.rescale:
-                input /= retain_prob
+                if type(input) == sparse.basic.SparseVariable:
+                    input = sparse.basic.mul(input, 1/retain_prob)
+                else:
+                    input /= retain_prob
 
             # use nonsymbolic shape for dropout mask if possible
             input_shape = self.input_shape


f0k:
Well, the case distinction is rather minimal, but what if we accumulate more case distinctions like these? I'm still leaning towards a separate module. The classes can still be derived from the existing ones and just override `get_output_for()`. What do others think? Shall we treat sparse matrices as first-class citizens in some existing layers or move that to a separate module?

hmendozap:
I'm new in Lasagne development, but I would argue for a ```` lasagne.layers.sparse```` submodule, as Theano does it and also scipy

I noticed that if  ````get_output(net, INPUT) ```` is called, it raises a Theano Type error, because it only checks against normal tensor types.

benanne:
I would also be strongly in favour of using inheritance for this and keeping it separate from the main ("dense") implementations. This aligns better with our design philosophy, in particular:

> Restraint: Do not obstruct users with features they decide not to use. Both in using and in extending components, it should be possible for users to be fully oblivious to features they do not need.

The code for dealing with sparse input makes it harder to read and understand, and since a lot of people read this code, I think we should avoid complicating it. Especially for a fairly uncommon use case such as this one (but still a very valid one, of course!).

eduardo4jesus:
@f0k, @hmendozap and @benanne:

> I'm new in Lasagne development, but I would argue for a lasagne.layers.sparse submodule, as Theano does it and also scipy

I really thought about it. Actually, at first, when I decided to try to adapt a sparse input to a dense layer, I did copy the `DenseLayer`, made the tweaks and called it `SparseDenseLayer`. I didn't like this name because it is so confusing, sparse is the opposite of dense. In the other hand, I was not comfortable to call it `SparseLayer`. The major reason is that this layer would not present any sparse effect to the network. It would be exactly our old `DenseLayer` that doesn't crash with a sparse matrix as input. 

I completely agree with a  `lasagne.layers.sparse submodule`. I miss understood it at first. I thought that you would expect such module to contain only things related to actual sparse layers and networks. But just now it came to my mind that what you guys are constantly saying is that I could move my changes from `lasagne.layers.DenseLayer` to `lasagne.layers.sparse.DenseLayer` and do the same with the `DropoutLayer`, right? Thus, `lasagne.layers.sparse` would house these small tweaks and also actual sparse layers that might come up one day.

@hmendozap , I am using `get_output` without any problem.

```
network = build_mlp(num_features=num_features, num_targets=num_targets, input_var=input_var)
.
.
.
prediction = lasagne.layers.get_output(network)
```

However, I have omit the `INPUT` argument and let `get_output` uses its default value `INPUT=None`, which is supported:

```
inputs : None, Theano expression, numpy array, or dict
        If None, uses the input variables associated with the
        :class:`InputLayer` instances.
```

I am really sorry for the confusion, I am still new with collaborative development. I will make the necessary changes soon.

f0k:
> But just now it came to my mind that what you guys are constantly saying is that I could move my changes from lasagne.layers.DenseLayer to lasagne.layers.sparse.DenseLayer and do the same with the DropoutLayer, right?

Yes, `lasagne.layers.sparse` would house all layers (or adaptions of layers) that support sparse input and/or output via `theano.tensor.sparse`.
Note that you don't need to (and shouldn't) copy the full `DenseLayer` implementation over, but derive a new class from `lasagne.layers.dense.DenseLayer` and just override what you need (i.e., the `get_output_for()` method, since both the constructor and the output shape computation stay the same).
If we call it `lasagne.layers.sparse.DenseLayer`, then `lasagne.layers.sparse.*` will not be auto-imported into the `lasagne.layers` namespace (similarly to `lasagne.layers.dnn`), but needs to be imported separately. That's quite okay for a special-purpose module, though.

f0k:
Superseded by #596.

PR #596, opened 2016-02-03
Sparse input to Lasagne

eduardo4jesus:
Recently, I started to work with sparse data. Then I found lasagne, a very easy-to-use and handful module. However, it didn't support sparse input for networks. I started to make some small tweaks to `lasagne.layers.DenseLayer` and `lasagne.layers.DropoutLayer` to get it working with the `MLP` from MNIST example. Later, I found that other people might find it useful as well, [Sparse input to Lasagne #317](https://github.com/Lasagne/Lasagne/issues/317).

A couple days ago I did a pull request of these changes. See [Sparse input to Lasagne #595] (https://github.com/Lasagne/Lasagne/pull/595). After getting nice feedbacks, I am submitting it again for the community analysis.

Changes:
- A sparse submodule was introduced: `lasagne.layers.sparse`
- Two new classes were added, `lasagne.layers.sparse.DenseLayer` and `lasagne.layers.sparse.DropoutLayer`, that extend from `lasagne.layers.DenseLayer` and `lasagne.layers.DropoutLayer`, respectively. Both classes just overrides the `get_output_for(  )` in order to use sparse matrices instead of `theano.tensor`.

### DIFF
--- a/lasagne/layers/sparse/__init__.py
+++ b/lasagne/layers/sparse/__init__.py
@@ -1,0 +1,2 @@
+from .dense import *
+from .noise import *
--- a/lasagne/layers/sparse/dense.py
+++ b/lasagne/layers/sparse/dense.py
@@ -1,11 +1,6 @@
-import numpy as np
-import theano.tensor as T
 import theano.sparse as sparse
 
-from ... import init
-from ... import nonlinearities
-
-from ..base import Layer
+from ..dense import DenseLayer as Dense
 
 
 __all__ = [
@@ -13,71 +8,14 @@ __all__ = [
 ]
 
 
-class DenseLayer(Layer):
-    """
-    lasagne.layers.sparse.DenseLayer(incoming, num_units,
-    W=lasagne.init.GlorotUniform(), b=lasagne.init.Constant(0.),
-    nonlinearity=lasagne.nonlinearities.rectify, **kwargs)
-
-    A fully connected layer.
-
-    Parameters
-    ----------
-    incoming : a :class:`Layer` instance or a tuple
-        The layer feeding into this layer, or the expected input shape
-
-    num_units : int
-        The number of units of the layer
-
-    W : Theano shared variable, expression, numpy array or callable
-        Initial value, expression or initializer for the weights.
-        These should be a matrix with shape ``(num_inputs, num_units)``.
-        See :func:`lasagne.utils.create_param` for more information.
-
-    b : Theano shared variable, expression, numpy array, callable or ``None``
-        Initial value, expression or initializer for the biases. If set to
-        ``None``, the layer will have no biases. Otherwise, biases should be
-        a 1D array with shape ``(num_units,)``.
-        See :func:`lasagne.utils.create_param` for more information.
-
-    nonlinearity : callable or None
-        The nonlinearity that is applied to the layer activations. If None
-        is provided, the layer will be linear.
-
-    Examples
-    --------
-    >>> from lasagne.layers import InputLayer
-    >>> from lasagne.layers.sparse import DenseLayer
-    >>> l_in = InputLayer((100, 20))
-    >>> l1 = DenseLayer(l_in, num_units=50)
-
-    Notes
-    -----
-
-    This layer expects a sparse matrix as input.
-    """
-    def __init__(self, incoming, num_units, W=init.GlorotUniform(),
-                 b=init.Constant(0.), nonlinearity=nonlinearities.rectify,
-                 **kwargs):
-        super(DenseLayer, self).__init__(incoming, **kwargs)
-        self.nonlinearity = (nonlinearities.identity if nonlinearity is None
-                             else nonlinearity)
-
-        self.num_units = num_units
-
-        num_inputs = int(np.prod(self.input_shape[1:]))
-
-        self.W = self.add_param(W, (num_inputs, num_units), name="W")
-        if b is None:
-            self.b = None
-        else:
-            self.b = self.add_param(b, (num_units,), name="b",
-                                    regularizable=False)
-
-    def get_output_shape_for(self, input_shape):
-        return (input_shape[0], self.num_units)
-
+class DenseLayer(Dense):
     def get_output_for(self, input, **kwargs):
+        """
+        Parameters
+        ----------
+        input : theano sparse matrix
+            output from the previous layer
+        """
         assert type(input) == sparse.basic.SparseVariable
 
         activation = sparse.basic.dot(input, self.W)
--- a/lasagne/layers/sparse/dense.py
+++ b/lasagne/layers/sparse/dense.py
@@ -1,0 +1,89 @@
+import numpy as np
+import theano.tensor as T
+
+from .. import init
+from .. import nonlinearities
+
+from .base import Layer
+
+
+__all__ = [
+    "DenseLayer",
+]
+
+
+class DenseLayer(Layer):
+    """
+    lasagne.layers.DenseLayer(incoming, num_units,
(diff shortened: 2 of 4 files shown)


lasagne/layers/sparse/dense.py
```diff
@@ -0,0 +1,24 @@
...
+    def get_output_for(self, input, **kwargs):
+        """
+        Parameters
+        ----------
+        input : theano sparse matrix
+            output from the previous layer
+        """
+        assert type(input) == sparse.basic.SparseVariable
+
+        activation = sparse.basic.dot(input, self.W)
```
nouiz: space dot have many implementation. By default, theano.tensor.dot(a,b) will call theano.sparse.basic.structured_dot().
The main difference is in the grad. Just double check you have the gradient behave like you want.

f0k: So the default `DenseLayer` will work with sparse input already?

If there are different implementations, then the user should probably be able to choose one in the constructor -- but this is something that could be added later.

lasagne/layers/sparse/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .dense import *
+from .noise import *
```
f0k: You should specifically import what you need, not `*`.

lasagne/layers/sparse/noise.py
```diff
@@ -0,0 +1,40 @@
...
+            If true dropout and scaling is disabled, see notes
+        """
+        if deterministic or self.p == 0:
+            return input
+        else:
+            assert type(input) == sparse.basic.SparseVariable
+
+            retain_prob = 1 - self.p
+            if self.rescale:
+                input = sparse.basic.mul(input, 1/retain_prob)
```
f0k: In both occasions, instead of using `1`, you should use `lasagne.utils.floatX(1)` -- otherwise the result will be upcast to float64 in case `self.p` is a numpy matrix.

lasagne/layers/sparse/dense.py
eduardo4jesus: > By default, theano.tensor.dot(a,b) will call theano.sparse.basic.structured_dot().

@nouiz , I am not sure about this.

```
>>> import scipy.sparse as sp
>>> import theano
>>> import theano.tensor as T
>>> import theano.sparse as sparse
>>> x = sparse.csr_matrix('x')
>>> y = sparse.csr_matrix('y')
>>> z = T.dot(x, y)
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/theano/tensor/basic.py", line 4872, in dot
    a, b = as_tensor_variable(a), as_tensor_variable(b)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/theano/tensor/basic.py", line 161, in as_tensor_variable
    "Variable type field must be a TensorType.", x, x.type)
theano.tensor.var.AsTensorError: ('Variable type field must be a TensorType.', SparseVariable{csr,float64}, Sparse[float64, csr])
>>> 
```

> The main difference is in the grad. Just double check you have the gradient behave like you want.

As far as I understood, the `sparse.dot` and `sparse.structured_dot` would only differ, in our case, when computing the grad wrt to the sparse input. As the main propose of these class adaptation are just to allow an sparse input to the network, the grad wrt to the input would not be relevant, since there is no previous layers for learning its parameters.

eduardo4jesus: > So the default DenseLayer will work with sparse input already?

It's original constructor does support a sparse input type. However, it crashes when calling `get_output_for` since `T.dot(input, self.W)` expects a `TensorType`

lasagne/layers/sparse/noise.py
eduardo4jesus: @f0k 
I don't know the reason, but after adding `lasagne.utils.floatX`, my CPU computation increased dramatically. In my experiment, I have a `csr_matrix` dataset with `shape == (728497, 53097)`.

Before including `lasagne.utils.floatX`, my process was using only one core, and each epoch was taking around 146 seconds. After adding it, all the 16 cores were getting considerably busy and each epoch was taking around 463 seconds. Maybe this is the reason why `layers.DropoutLayer.get_output_for()` does NOT use `lasagne.utils.floatX`.

So, I am going to revert this change. I hope that it's ok for you.

f0k: Weird! Seems it makes a difference whether `retain_prob` is a numpy scalar or a Python float, but in theory it shouldn't. Maybe Theano passes is on to scipy as-is, and scipy uses a different algorithm that needlessly uses multi-threading.
In this case, we need to check whether `self.p` is symbolic or not, and use `float(1 - self.p)` if it is not symbolic:
```python
if isinstance(self.p, theano.Variable):
    retain_prob = 1 - self.p
    inv_retain_prob = T.inv(self.p)
else:
    retrain_prob = float(1 - self.p)
    inv_retain_prob = 1 / retrain_prob
if self.rescale:
    input = sparse.basic.mul(input, inv_retain_prob)
```

lasagne/layers/sparse/dense.py
nouiz: Right, it don't make a difference in this case. So if you wish, with theano.dot(), you could make the dense layer work with both Type.

lasagne/layers/sparse/noise.py
nouiz: If you want to track the speed difference, you can use Theano profiler and
compare both case. You can run it for just one epoch.

Just use the Theano flag profile=True

eduardo4jesus:
I was away for 2 months, than I tried one of my experiments again, it is much slower than before. O remember one epoch taking about 140 seconds. Now it is taking around 550 seconds. What would be the best way to identify why it is happening?

f0k:
> What would be the best way to identify why it is happening?

Run with THEANO_FLAGS=profile=1. Switch back to the Theano version you used when it was taking 140 seconds, and also run with THEANO_FLAGS=profile=1. Compare what's different. If you can't find anything obvious, use "git bisect" to find which commit in Theano introduced the performance regression.

eduardo4jesus:
I already tried an older Theano and Lasagne version. The slowness still remained. I also checked for other processes that could be running. However, I didn't find anything that was using the same GPU I was. Oddly enough, after finally get the admin to reboot the machine, the epoch computation time on the training came down to 140 seconds with both old and bleeding edge versions. Thank you.

nouiz:
I already saw this. Which GPU are in that box? I saw some GPU get
underclocked and that stay like that. I never understood what was the
problem.

hmendozap:
Hi all,

Sorry to ask, but is this PR available? If it's not, I could help finish it. I'm using the small tweaks on #317 and #595. I have time now and if there is anything to be done I could help.

Thanks,
HM

f0k:
I'd be fine with you taking over! We can wait a little bit to hear from @eduardo4jesus.

The code is good in general, the dropout layer should be updated to match the changes in #661 and the tests should be extended to get 100% coverage (haven't checked the coverage report in detail).

hmendozap:
Hi,

I have some questions about the implementation. Is ok if I open a PR and we discuss it there or here is ok?

f0k:
A new PR is fine!

ironfist2:
Can you also implement it for Conv1DLayer and Conv2DLayer too.

f0k:
> Can you also implement it for Conv1DLayer and Conv2DLayer too.

From a quick research, the following Caffe fork supports this, but it requires Intel CPUs: https://github.com/IntelLabs/SkimCaffe
Porting this to Theano and then integrating it into Lasagne is nontrivial and will require the Intel compiler afterwards. If you want to do research on sparse convolutions, you may want to try their project first!

Issue #12, opened 2014-09-12
Training loops


benanne:
So far the only thing that's been implemented is a bunch of tools to generate Theano expressions for neural nets. There is no actual training code in the library yet.

We should provide some premade 'training loops' for common use cases, that take care of things like compiling the necessary Theano functions and updating the parameters given a dataset.

It would be great if we could rely on Python generators for this - although at this point I'm not sure if they offer enough flexibility. But if they do, it would be great to be able to avoid adding another class / abstraction.

We could provide a few different types of training loops, for different dataset sizes and approaches. For example, some datasets fit into GPU memory, so we should provide a loop that loads up the data into a shared variable and then iterates over that in batches. But a lot of datasets actually don't, so then we'd have to load a new 'chunk' of data into the shared variable at regular intervals.

Until now I've always reimplemented this type of thing specifically for the problems I was working on (layers.py only provided tools to generate Theano expressions, nothing else). But I've definitely felt like I was reinventing the wheel a couple of times :)

I don't have a concrete idea yet of how we should implement this, input is very welcome.

f0k:
Was inadvertently closed on merging in recurrent layers, due to a commit message that actually referenced an issue in @craffel's fork.

benanne:
I'd like to revive this discussion again. I think it would be nice if the next release of Lasagne came with some training loop code that covers the most common use cases.

I know there is still some discussion about whether we need this in Lasagne at all, but I think we do. It would be nice to provide a more complete package. Of course the focus remains firmly on neural networks, but training is a pretty essential aspect of NNs so I think this still fits within our philosophy of "do one thing and do it well". And as with everything in Lasagne it would of course be completely optional to use.

I haven't thought too much about the API yet. The only thing I'm certain of is that we should rely on Python generators where we can for maximal flexibility. I think it would be nice to have something that lets you say "okay, here is my network architecture definition, now please train it on this dataset and use sensible defaults for all the things I didn't specify."

Preferably it should be possible to do this in 2-3 lines of code. Currently a lot of code uses the same boilerplate to construct shared variables for the dataset, iterate through it in batches, evaluate periodically, ... I'm pretty sure we could provide something that would be sufficient for 90% of use cases. This would save people a lot of thinking and typing (or copy-pasting).

[...]

PR #599, opened 2016-02-03
[WIP] Add training loops

Joshua-Chin:
This adds a convenience class the helps train and evaluate networks, using a scikit-learn style interface.


benanne:
This is great, but I'm not sure if we would want something like this in the main library at this point. I'm open to discussing it though! Note that there is also nolearn which provides the same functionality.

I still plan to come up with some training loop utilities that don't involve encapsulating the model into a class, which tends to fit my own workflow better -- but I guess that's mostly a matter of preference. Hopefully I can draft a proposal for that sometime soon (I've been postponing it for a while now...)

Hopefully I'll have some time to take a closer look at this this weekend. At a glance, it seems like it might actually be fairly limited in scope (e.g. only supporting (X, y)-style problems, i.e. supervised learning) and we might want something more broadly applicable. There is a fairly lengthy discussion at #12 which is probably worth (re-)reading as it touches some of these concerns.

f0k:
Same here... we'd love to have some generic training loop code in Lasagne to make it easier to get started, but we need to get it right! This means the following restrictions of your proposal would need to be lifted:
* restriction to "single input, single output" network
* restriction to "X -> y" problems
* restriction to datasets that fit into host memory at once

The toughest part probably is that the training loop should cooperate with any kind of data iterator. We can provide one for the simple case of iterating over a numpy array, but other than that, it should support whatever people have cooked up already. This will still need some discussion.

Thank you for bringing this back to our attention, though, and feel free to add your two cents to #12! Meanwhile, [nolearn](https://github.com/dnouri/nolearn) provides what you proposed and then some more.

Joshua-Chin:
Generic training loop code would be great for Lasagne, but common use cases should be easy. Many of the applications for neural networks can be phrased as "X -> y" problems.

I have also been working on a more generic batch iterator, that handles both numpy arrays and generators over individual examples. I plan to create a pull request for it sometime this coming week.

benanne:
I agree that (X, y) should be the default, but if we add this to the main library it should definitely be more flexible than that.

PR #602, opened 2016-02-05
Increased speed of thin plate spline transformer layer dramatically

### LINKED ISSUE #601: Speed up thin plate spline transformer layer
Currently, the thin plate spline transformer layer I just implemented is fairly slow for larger images. The main culprit is '_get_transformed_points_tps'. The way it works now is that `_get_transformed_points_tps` takes in all of the points in the grid (equivalent to the number of pixels if no downsampling is applied), and shifts them according the to calculated thin plate spline coefficients. The shifted grid is then run through the bilinear interpolation along with the input image to create the output. 

Unfortunately, while the computation for an affine transform is simply a dot product of the parameters and the source grid, the computation for the thin plate spline transform is more complicated. The main culprit for the slow down is that the computation requires taking the squared difference between each input point and each source point and then multiplying this squared difference by it's log. [This computation](https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/special.py#L734) is taking up over 80% of the run time. 

Thankfully, we don't actually need to compute it as part of the theano graph! Because this portion only relies on the source points and the original image size and downsample factor, we can compute it in numpy when the layer is constructed. I see that in the note for [`_meshgrid`](https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/special.py#L527) that grid construction was implemented in theano to allow for symbolic grid sizes because there wasn't a noticable effect on performance but there is clearly one here. 

It seems to me that for this layer it is clearly worth increasing the speed at the cost of having to prespecify the image size when the layer is constructed (likely defaulting to the shape of the incoming layer). What do you guys think?

arimorcos:
Moved the computation of the distances between the grid and the source points to the initialization of the layer. This prevents this computation from having to be done on every pass. I detailed this concern in #601. 

The one downside of this is that it requires the grid and thus, the image size, to be specified at layer construction time. 

Closes #601.

### DIFF
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -581,6 +581,15 @@ class TPSTransformerLayer(MergeLayer):
         transformation. These points will be arranged as a grid along the
         image, so the value must be a perfect square. Default is 16.
 
+    precompute_grid : 'auto' or boolean
+        Flag to precompute the U function [2]_ for the grid and source
+        points. If 'auto', will be set to true as long as the input height
+        and width are specified. If true, the U function is computed when the
+        layer is constructed for a fixed input shape. If false, grid will be
+        computed as part of the Theano computational graph, which is
+        substantially slower as this computation scales with
+        num_pixels*num_control_points. Default is 'auto'.
+
     References
     ----------
     .. [1]  Max Jaderberg, Karen Simonyan, Andrew Zisserman,
@@ -623,7 +632,7 @@ class TPSTransformerLayer(MergeLayer):
     """
 
     def __init__(self, incoming, localization_network, downsample_factor=1,
-                 control_points=16, **kwargs):
+                 control_points=16, precompute_grid='auto', **kwargs):
         super(TPSTransformerLayer, self).__init__(
                 [incoming, localization_network], **kwargs)
 
@@ -648,9 +657,20 @@ class TPSTransformerLayer(MergeLayer):
                              "output shape: (batch_size, num_input_channels, "
                              "input_rows, input_columns)")
 
+        # Process precompute grid
+        can_precompute_grid = all(s is not None for s in input_shp[2:])
+        if precompute_grid == 'auto':
+            precompute_grid = can_precompute_grid
+        elif precompute_grid and not can_precompute_grid:
+            raise ValueError("Grid can only be precomputed if the input "
+                             "height and width are pre-specified.")
+        self.precompute_grid = precompute_grid
+
         # Create source points and L matrix
-        self.source_points, self.L_inv = _initialize_tps(
-                control_points)
+        self.right_mat, self.L_inv, self.source_points, self.out_height, \
+            self.out_width = _initialize_tps(
+                control_points, input_shp, self.downsample_factor,
+                precompute_grid)
 
     def get_output_shape_for(self, input_shapes):
         shape = input_shapes[0]
@@ -662,15 +682,16 @@ class TPSTransformerLayer(MergeLayer):
         # see eq. (1) and sec 3.1 in [1]
         # Get input and destination control points
         input, dest_offsets = inputs
-        return _transform_thin_thin_plate_spline(dest_offsets,
-                                                 input,
-                                                 self.source_points,
-                                                 self.L_inv,
-                                                 self.downsample_factor)
+        return _transform_thin_plate_spline(
+                dest_offsets, input, self.right_mat, self.L_inv,
+                self.source_points, self.out_height, self.out_width,
+                self.precompute_grid, self.downsample_factor)
 
 
-def _transform_thin_thin_plate_spline(dest_offsets, input, source_points,
-                                      L_inv, downsample_factor):
+def _transform_thin_plate_spline(
+        dest_offsets, input, right_mat, L_inv, source_points, out_height,
+        out_width, precompute_grid, downsample_factor):
+
     num_batch, num_channels, height, width = input.shape
     num_control_points = source_points.shape[1]
 
@@ -682,18 +703,25 @@ def _transform_thin_thin_plate_spline(dest_offsets, input, source_points,
     # Solve as in ref [2]
     coefficients = T.dot(dest_points, L_inv[:, 3:].T)
 
-    # Transformed grid
-    out_height = T.cast(height / downsample_factor[0], 'int64')
-    out_width = T.cast(width / downsample_factor[1], 'int64')
-    orig_grid = _meshgrid(out_height, out_width)
-    orig_grid = orig_grid[0:2, :]
-    orig_grid = T.tile(orig_grid, (num_batch, 1, 1))
+    if precompute_grid:
 
-    # Transform each point on the source grid (image_size x image_size)
-    transformed_points = _get_transformed_points_tps(orig_grid, source_points,
-                                                     coefficients,
-                                                     num_control_points,
-                                                     num_batch)
+        # Transform each point on the source grid (image_size x image_size)
+        right_mat = T.tile(right_mat.dimshuffle('x', 0, 1), (num_batch, 1, 1))
+        transformed_points = T.batched_dot(coefficients, right_mat)
+
+    else:
+
+        # Transformed grid
+        out_height = T.cast(height / downsample_factor[0], 'int64')
+        out_width = T.cast(width / downsample_factor[1], 'int64')
+        orig_grid = _meshgrid(out_height, out_width)
+        orig_grid = orig_grid[0:2, :]
+        orig_grid = T.tile(orig_grid, (num_batch, 1, 1))
+
+        # Transform each point on the source grid (image_size x image_size)
+        transformed_points = _get_transformed_points_tps(
+                orig_grid, source_points, coefficients, num_control_points,
+                num_batch)
 
     # Get out new points
     x_transformed = transformed_points[:, 0].flatten()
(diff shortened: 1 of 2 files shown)


arimorcos:
As @skaae mentioned in #601, we could add a flag to either compute grid at layer construction time or as part of the theano graph. This implementation doesn't have a flag. I could add one in, but I'm having trouble thinking of a use case where the image size would change dynamically. 

Can anyone else think of any?

benanne:
Are we talking about the input image size, or the output of the transformer? If it's the former, I think it wouldn't be a bad idea to support varying sizes where needed. I can imagine a situation where you'd want to apply the network to arbitrarily sized images (maybe downloaded from the web) and want the transformer to focus on a particular object in it, or something to that extent. Of course training would probably still happen on fixed size inputs to get the speed advantage. How much messier would it really get?

I seem to remember a similar case where we gave the option to either precompute or not, but I can't remember what it was...

arimorcos:
It would be the input image size, so the input to the transformer, though that obviously affects the output of the transformer. I can go ahead and add a flag. It'll be a little messy, but still readable certainly.

arimorcos:
Ok! I added a flag to precompute the grid (`precompute_grid`) and set it to default to `True`.

f0k:
> I seem to remember a similar case where we gave the option to either precompute or not, but I can't remember what it was...

It was the original transformer, I experimented with precomputing something offline in numpy, but it didn't really help performance in the end so I dropped it again.

> Ok! I added a flag to precompute the grid (precompute_grid) and set it to default to True.

Instead of a separate flag, can you precompute whenever the input image size is fixed (or whatever part of it needs to be fixed) and compute dynamically otherwise? You'd just need to check whether there are `None`s in `self.input_shape` after calling the super constructor. Having `None` in the first dimension (the batch size) or the second dimension (the channel count) won't be a problem, if I remember correctly.

arimorcos:
> Instead of a separate flag, can you precompute whenever the input image size is fixed (or whatever part of it needs to be fixed) and compute dynamically otherwise? You'd just need to check whether there are Nones in self.input_shape after calling the super constructor. Having None in the first dimension (the batch size) or the second dimension (the channel count) won't be a problem, if I remember correctly.

Ya, no problem, that should be an easy swap. The only worry I'd have about that is that the choice wouldn't be transparent. The speed difference is substantial, so it seems like we should do what we can to make sure users are aware of that. I could just add a note in the documentation. Does anyone have a better idea of how to deal with that?

f0k:
> The speed difference is substantial, so it seems like we should do what we can to make sure users are aware of that.

What about defaulting `precompute='auto'`, which precomputes when possible? This would document the behaviour (and allow people to use the slower version even if they don't need to, hooray ;) ).

arimorcos:
I think that's a great solution! So I implemented it. ;)

lasagne/layers/special.py
```diff
@@ -648,9 +658,25 @@ def __init__(self, incoming, localization_network, downsample_factor=1,
                              "output shape: (batch_size, num_input_channels, "
                              "input_rows, input_columns)")
 
+        if type(precompute_grid) is not bool and precompute_grid != 'auto':
+            raise ValueError("precompute_grid must be a boolean or set to "
+                             "'auto'.")
```
f0k: I think I wouldn't do an explicit type check. Sometimes you'll want `0` to be an accepted false-ish value. What about a structure as follows:
```python
can_precompute = all(s is not None for s in input_shp[2:])
if precompute_grid == 'auto':
    precompute_grid = can_precompute
elif precompute_grid and not can_precompute:
    raise ValueError(...)
self.precompute_grid = precompute_grid
```

arimorcos: I agree, I think that's clearer. Any other suggestions before I commit? I can rebase this one into a single commit if you're otherwise pleased with the PR.

lasagne/layers/special.py
```diff
@@ -662,38 +688,46 @@ def get_output_for(self, inputs, **kwargs):
...
         input, dest_offsets = inputs
-        return _transform_thin_thin_plate_spline(dest_offsets,
-                                                 input,
-                                                 self.source_points,
-                                                 self.L_inv,
-                                                 self.downsample_factor)
+        return _transform_thin_thin_plate_spline(
+                dest_offsets, input, self.right_mat, self.L_inv,
+                self.source_points, self.out_height, self.out_width,
+                self.precompute_grid, self.downsample_factor)
```
f0k: This should have failed -- with your current implementation, if `precompute_grid='auto'`, you set `self.precompute_grid = 'auto'` (not True/False), so `_transform_thin_thin_plate_spline()` (wait, why is `thin` duplicated, by the way?) would treat it as `precompute_grid=True` no matter if the input shape was fully-defined or not.
If it didn't fail, you're probably missing a test for `precompute_grid='auto'` and variable input shape?

lasagne/layers/special.py
```diff
@@ -662,38 +688,46 @@ def get_output_for(self, inputs, **kwargs):
...
-def _transform_thin_thin_plate_spline(dest_offsets, input, source_points,
-                                      L_inv, downsample_factor):
     num_batch, num_channels, height, width = input.shape
     num_control_points = source_points.shape[1]
 
     # reshape destination offsets to be (num_batch, 2, num_control_points)
     # and add to source_points
     dest_points = source_points + T.reshape(
-            dest_offsets, (num_batch, 2, num_control_points))
+        dest_offsets, (num_batch, 2, num_control_points))
```
f0k: I'd prefer double indent as before (make sure to use the same pep8 version as in `requirements-dev.txt`).

lasagne/layers/special.py
```diff
@@ -702,8 +736,8 @@ def _transform_thin_thin_plate_spline(dest_offsets, input, source_points,
     # dimshuffle input to  (bs, height, width, channels)
     input_dim = input.dimshuffle(0, 2, 3, 1)
     input_transformed = _interpolate(
-            input_dim, x_transformed, y_transformed,
-            out_height, out_width)
+        input_dim, x_transformed, y_transformed,
+        out_height, out_width)
```
f0k: I'd prefer double indent as before (make sure to use the same pep8 version as in `requirements-dev.txt`).

lasagne/layers/special.py
```diff
@@ -815,8 +864,8 @@ def _initialize_tps(num_control_points):
         for point_2 in range(point_1, num_control_points):
 
             L[point_1 + 3, point_2 + 3] = _U_func_numpy(
-                    source_points[0, point_1], source_points[1, point_1],
-                    source_points[0, point_2], source_points[1, point_2])
+                source_points[0, point_1], source_points[1, point_1],
+                source_points[0, point_2], source_points[1, point_2])
```
f0k: I'd prefer double indent as before (make sure to use the same pep8 version as in `requirements-dev.txt`).

lasagne/layers/special.py
```diff
@@ -796,13 +842,16 @@ def _initialize_tps(num_control_points):
 
     # Create 2 x num_points array of source points
     source_points = np.vstack(
-            (x_control_source.flatten(), y_control_source.flatten()))
+        (x_control_source.flatten(), y_control_source.flatten()))
```
f0k: I'd prefer double indent as before (make sure to use the same pep8 version as in `requirements-dev.txt`).

lasagne/tests/layers/test_special.py
```diff
@@ -514,7 +533,7 @@ def test_transform_thin_plate_spline_identity(self):
         inputs = floatX(np.arange(np.prod(l_in.shape)).reshape(l_in.shape))
         outputs = layer.get_output_for([constant(inputs),
                                         constant(dest_offset)]).eval()
-        np.testing.assert_allclose(inputs, outputs, atol=1e-5)
+        np.testing.assert_allclose(inputs, outputs, atol=5e-4)
```
f0k: Is it less accurate than before?

lasagne/layers/special.py
arimorcos: Good catch on both! I removed the extra `thin` and added in a test for `precompute_grid='auto'` with a variable input shape.

f0k:
Finished going through the PR in its current form! So apart from the comments, I'm fine with it (although the code might be reorganized a bit -- but that's not really crucial now).

arimorcos:
Whoops! Didn't refresh and see some of the new comments. Ignore the update I just sent. I still have to fix some of your comments.

lasagne/tests/layers/test_special.py
arimorcos: I tried chasing this down, and had trouble figuring out exactly what was up. My best guess is that it's a precision difference between the way that numpy is computing it and theano was, but I'm not sure exactly.

arimorcos:
Ok, ready to go! The only change I didn't address is the two-fold change in tolerance needed for the identity test. As mentioned in the line comment, I tried chasing it down, but had trouble. My best guess is that it had something to with a precision difference between numpy and theano, but I'm not sure.

f0k:
> Ok, ready to go!

Cool! Looks good to me. Can anybody else have a look?

> The only change I didn't address is the two-fold change in tolerance needed for the identity test.

It's a 50-fold increase, actually. Probably still small enough to not cause any trouble, but still funny that it makes such a difference. Let's leave it at that, though.

benanne:
I skimmed the changes and nothing jumped out at me, but I haven't looked at every line in detail.

arimorcos:
> It's a 50-fold increase, actually. Probably still small enough to not cause any trouble, but still funny that it makes such a difference. Let's leave it at that, though.

Of course. I must have been thinking about it with positive exponents for some reason. Now I feel silly. If you guys feel it's important to track down the precise cause of it, I can spend some more time to do that. I probably won't be able to do for a few days though.

f0k:
All right, since there are no further comments, I'll merge! Thanks again!

PR #603, opened 2016-02-07
Dropout layer output type should match input type when rescale is False

### LINKED ISSUE #546: Word dropout
Using dropout on the inputlayer for dropping out words in a sentence, say an int32 matrix doesn't work since the self._srng.binomial is called with dtype floatX. Maybe it should just use the input dtype?

Joshua-Chin:
Closes #546.

### DIFF
--- a/lasagne/tests/layers/test_noise.py
+++ b/lasagne/tests/layers/test_noise.py
@@ -48,6 +48,11 @@ class TestDropoutLayer:
         assert 0.4 < result_eval.mean() < 0.6
         assert (numpy.unique(result_eval) == [0., 1.]).all()
 
+    def test_get_output_for_no_rescale_dtype(self, layer_no_rescale):
+        input = theano.shared(numpy.ones((100, 100), dtype=numpy.int32))
+        result = layer_no_rescale.get_output_for(input)
+        assert result.dtype == input.dtype
+
     def test_get_output_for_p_02(self, layer_p_02):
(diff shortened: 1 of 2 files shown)


lasagne/tests/layers/test_noise.py
```diff
@@ -48,6 +48,12 @@ def test_get_output_for_no_rescale(self, layer_no_rescale):
         assert 0.4 < result_eval.mean() < 0.6
         assert (numpy.unique(result_eval) == [0., 1.]).all()
 
+    def test_get_output_for_no_rescale_dtype(self, layer_no_rescale):
+        input = theano.shared(numpy.ones((100, 100), dtype=numpy.int32))
+        result = layer_no_rescale.get_output_for(input)
+        result_eval = result.eval()
+        assert result_eval.dtype == input.dtype
```
f0k: It would be enough to do:
```
assert result.dtype == input.dtype
```
This avoids the `eval()` step which requires a compilation. Theano already keeps track of dtypes in the graph fine.

f0k:
Thank you, looks good except that the test can be sped up!

Joshua-Chin:
Okay, test sped up and commit squashed.

f0k:
Thank you, merging! (And I removed "issue" from your PR description so it follows the syntax in https://github.com/blog/1506-closing-issues-via-pull-requests and closes the issue automatically on merging.)

benanne:
Nice, thanks!

PR #606, opened 2016-02-10
Change GPU check to allow theano.sandbox.cuda.use()

### LINKED ISSUE #379: Change theano.config.device.startswith("gpu") check
As discussed in https://groups.google.com/forum/#!topic/lasagne-users/AWbQ0vvJjrQ, the existing `theano.config.device.startswith("gpu")` checks are too strict -- Theano provides a second way to use a GPU, via `theano.sandbox.cuda.use`, and Lasagne wrongly prevents `lasagne.layers.dnn` to be imported in that case.

From https://groups.google.com/d/msg/lasagne-users/AWbQ0vvJjrQ/neQZc-vLBgAJ, it seems the most universal solution for now would be replacing it with `theano.sandbox.cuda.use.device_number >= 0`. We might want to check if there's a solution that also works in the new GPU backend, though.

In any case, this would require changes in the mocks for readthedocs as well.

f0k:
Closes #379 by replacing `theano.config.device.startswith("gpu")` with `theano.sandbox.cuda.cuda_enabled`. This allows CUDA to be enabled in a script via `theano.sandbox.cuda.use("gpu0")` instead of only via the Theano configuration.

While we're at it, this changes the import error messages to point at the corresponding documentation on readthedocs.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -322,6 +322,7 @@ import theano
 import theano.sandbox.cuda
 
 theano.config = Mock(device='gpu')
+theano.sandbox.cuda.cuda_enabled = True
 theano.sandbox.cuda.dnn = Mock(dnn_available=lambda: True)
 
 import sys
--- a/lasagne/layers/corrmm.py
+++ b/lasagne/layers/corrmm.py
@@ -17,8 +17,10 @@ __all__ = [
 ]
 
 
-if not theano.config.device.startswith("gpu"):
-    raise ImportError("requires a GPU to work")  # pragma: no cover
+if not theano.sandbox.cuda.cuda_enabled:
+    raise ImportError(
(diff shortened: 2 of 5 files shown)


benanne:
Very nice :) is the `cuda_enabled` field a new thing, or has it always been there and we just didn't know about it? In the former case, should we also bump the minimum required version of Theano?

nouiz:
This is old stuff in Theano.

f0k:
We just didn't know about it. @nouiz suggested it in https://github.com/Lasagne/Lasagne/issues/379#issuecomment-135796199.

> Very nice :)

I take this as "looks good to me" and merge! :)

benanne:
I think that was a good interpretation :) Thanks!

Issue #447, opened 2015-09-30
InverseLayer with untied weights


f0k:
This was asked for on the mailing list: https://groups.google.com/forum/#!topic/lasagne-users/vYIzqT1Wn7g

For the implementation, the constructor would need to create new shared variables for all params of the layer to be inverted, and `get_output_for()` would need to clone the expression it would usually return, replacing all params of the layer to be inverted with its own copies. Neat trickery!

benanne:
cool :)

f0k:
Moved to v0.3, would be nice to have, but is not critical for the release.

PR #610, opened 2016-02-15
Add TransposedConv2DLayer

f0k:
Addresses the third TODO item in issue #524. Untested work in progress, just to mark that I've started it.

### DIFF
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -23,7 +23,7 @@ def conv_output_length(input_length, filter_size, stride, pad=0):
 
     Parameters
     ----------
-    input_length : int
+    input_length : int or None
         The size of the input.
 
     filter_size : int
@@ -52,13 +52,14 @@ def conv_output_length(input_length, filter_size, stride, pad=0):
 
     Returns
     -------
-    int
-        The output size corresponding to the given convolution parameters.
+    int or None
+        The output size corresponding to the given convolution parameters, or
+        ``None`` if `input_size` is ``None``.
 
     Raises
     ------
-    RuntimeError
-        When an invalid padding is specified, a `RuntimeError` is raised.
+    ValueError
+        When an invalid padding is specified, a `ValueError` is raised.
     """
     if input_length is None:
         return None
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -11,6 +11,8 @@ from .base import Layer
 __all__ = [
     "Conv1DLayer",
     "Conv2DLayer",
+    "TransposedConv2DLayer",
+    "Deconv2DLayer",
 ]
 
 
@@ -79,6 +81,75 @@ def conv_output_length(input_length, filter_size, stride, pad=0):
     output_length = (output_length + stride - 1) // stride
 
     return output_length
+
+
+def conv_input_length(output_length, filter_size, stride, pad=0):
+    """Helper function to compute the input size of a convolution operation
+
+    This function computes the length along a single axis, which corresponds
+    to a 1D convolution. It can also be used for convolutions with higher
+    dimensionalities by using it individually for each axis.
+
+    Parameters
+    ----------
+    output_length : int or None
+        The size of the output.
+
+    filter_size : int
+        The size of the filter.
+
+    stride : int
+        The stride of the convolution operation.
+
+    pad : int, 'full' or 'same' (default: 0)
+        By default, the convolution is only computed where the input and the
+        filter fully overlap (a valid convolution). When ``stride=1``, this
+        yields an output that is smaller than the input by ``filter_size - 1``.
+        The `pad` argument allows you to implicitly pad the input with zeros,
+        extending the output size.
+
+        A single integer results in symmetric zero-padding of the given size on
+        both borders.
+
+        ``'full'`` pads with one less than the filter size on both sides. This
+        is equivalent to computing the convolution wherever the input and the
+        filter overlap by at least one position.
+
+        ``'same'`` pads with half the filter size on both sides (one less on
+        the second side for an even filter size). When ``stride=1``, this
+        results in an output size equal to the input size.
+
+    Returns
+    -------
+    int or None
+        The smallest input size corresponding to the given convolution
+        parameters for the given output size, or ``None`` if `output_size` is
+        ``None``. For a strided convolution, any input size of up to
+        ``stride - 1`` elements larger than returned will still give the same
+        output size.
+
+    Raises
+    ------
+    ValueError
+        When an invalid padding is specified, a `ValueError` is raised.
+
+    Notes
+    -----
+    This can be used to compute the output size of a convolution backward pass,
+    also called transposed convolution, fractionally-strided convolution or
+    (wrongly) deconvolution in the literature.
+    """
+    if output_length is None:
+        return None
+    if pad == 'valid':
+        pad = 0
+    elif pad == 'full':
+        pad = filter_size - 1
+    elif pad == 'same':
+        pad = filter_size // 2
+    if not isinstance(pad, int):
+        raise ValueError('Invalid pad: {0}'.format(pad))
+    return (output_length - 1) * stride - 2 * pad + filter_size
 
 
 class BaseConvLayer(Layer):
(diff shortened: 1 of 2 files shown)


kli-nlpr:
Hi, @f0k 
```
def convolve(self, input, **kwargs):
        border_mode = 'half' if self.pad == 'same' else self.pad
        op = T.nnet.abstract_conv.AbstractConv2d_gradInputs(
            imshp=self.output_shape,
            kshp=self.get_W_shape(),
            subsample=self.stride, border_mode=self.pad)
        conved = op(self.W, input, self.output_shape[2:])
        return conved
```
I can't understand this code, especially the op variable.
I find little information about ```AbstractConv2d_gradInputs```, I don't know what's going on here.
I know that ```input``` is the input feature map, and ```W``` is the kernel.
What's the purpose this convolve function? What does it return?
Thank you very much.

kli-nlpr:
Oh, I find it out!!!
I get inspiration from caffe’s code.
```
template <typename Dtype>
void DeconvolutionLayer<Dtype>::Forward_cpu(const vector<Blob<Dtype>*>& bottom,
      const vector<Blob<Dtype>*>& top) {
  const Dtype* weight = this->blobs_[0]->cpu_data();
  for (int i = 0; i < bottom.size(); ++i) {
    const Dtype* bottom_data = bottom[i]->cpu_data();
    Dtype* top_data = top[i]->mutable_cpu_data();
    for (int n = 0; n < this->num_; ++n) {
      this->backward_cpu_gemm(bottom_data + n * this->bottom_dim_, weight,
          top_data + n * this->top_dim_);
      if (this->bias_term_) {
        const Dtype* bias = this->blobs_[1]->cpu_data();
        this->forward_cpu_bias(top_data + n * this->top_dim_, bias);
      }
    }
  }
}
```
I think that ```op``` is the same as ```backward_cpu_gemm```.

f0k:
> I can't understand this code, especially the op variable.

That's the way Theano graphs are constructed. For example, if you call the `T.nnet.abstract_conv.conv2d()` function, all it does is instantiate the `ConvOp` op with the parameters that are fixed at compile time (i.e., the shapes, stride and padding), and then call it with the symbolic variables to create an `Apply` node: https://github.com/Theano/Theano/blob/e160bc3/theano/tensor/nnet/abstract_conv.py#L117-L138
Here, we instantiate the Op manually, because there was no convenience function at the time of writing (in between, `T.nnet.abstract_conv.conv2d_grad_wrt_inputs` has been added).
Have a look at http://deeplearning.net/software/theano/extending/graphstructures.html to understand how Ops, Apply nodes and Variables make up the Theano graph.

> I think that `op` is the same as `backward_cpu_gemm`.

Not exactly the same, but yes, the Op is a placeholder that under circumstances can result in the gemm-based convolution backward pass being used in the compiled Theano function.

trungnt13:
This layer cannot get back the exact shape before doing the convolution if we use stride and the shape is floored, and it is also not possible to infer the accurate original shape because of this:
>    output_length = (output_length + stride - 1) **//** stride

I think an you should add optional argument **output_shape** to **__init__** that help user define desire _"Deconvolve"_ shape

kli-nlpr:
I hope this one can merge soon, since deconv operation is frequently used in modern cnn architecture for image segmentation tasks.

f0k:
Should have time for this next week.

> This layer cannot get back the exact shape before doing the convolution

It was not necessarily meant to invert a convolution, so any output shape would be fine. There may be use cases that warrant customizing the output shape, I'll figure out a solution when I resume this PR. Thanks for the input!

madratman:
Awaiting this to get merged eagerly! Also, how can one do this with current Lasagne? InverseLayer?

f0k:
> Awaiting this to get merged eagerly!

Sorry for the long wait!

> Also, how can one do this with current Lasagne? InverseLayer?

Sort of, it will have tied weights with the layer you invert, though. The best solution for now will be to copy/paste the implementation from this PR into your script.

madratman:
By copy paste you mean it works(barring custom output case) ? From the comments, I don't get what the typo is referring to.

f0k:
> From the comments, I don't get what the typo is referring to.

Yes, the comment about the typo had a typo. I've fixed it now.

> By copy paste you mean it works(barring custom output case) ?

I didn't try, but it should!

liuhantang:
Hi, this is really useful. I copied and pasted the code so that I can use it. I find some typos in the current code. I fixed the typos to make it work. When I initialize the network with an `InputLayer` of shape (1, 3, None, None), the building of computation graph goes well.  However, when I compute a forward pass with a real input of size (1, 3, 500, 500), the `Conv2DTransposeLayer` signals an error `AsTensorError: ('Cannot convert (None, None) to TensorType', <type 'tuple'>)`. Doesn't it support arbitrary input shape now? I'm still studying how to use theano and lasagne, so there may be something that I'm missing? Thank you.

f0k:
> Doesn't it support arbitrary input shape now?

No, I still need to add this. Wait a minute.

f0k:
Should work with variable input shape now, but note that I made several other modifications (see commit message). In particular, I changed it to return the minimum possible shape for strided convolution instead of the maximum possible shape -- this is what you get when dilating the input by only placing zeros between input elements, not after the last input elements as well. See https://github.com/vdumoulin/conv_arithmetic.

We still need a way to allow custom output shapes. For example, blocks defines two additional parameters `unused_edge` and `original_image_size` for this: https://github.com/mila-udem/blocks/blob/b853b570f14e7494c5f95179f1cd977ffa69e91b/blocks/bricks/conv.py#L189-L204
The first one gives the number of additional zeros to pad the right and bottom border of the input with, allowing values between `0` and `stride - 1`. The second one directly gives the output size (a bit more difficult to check for correctness, but probably easier for the user). Which one(s) should we support?

Both won't give a convenient way of inverting a convolution with variable input size, as only a `MergeLayer` can query the symbolic input shape of a convolution to be inverted. Probably we should leave that job for the `InverseLayer`, possibly along with #447.

liuhantang:
Thank you for your quick respond and nice code. I've tested the new `TransposedConv2DLayer`, it works smooth on my test cases with different input sizes for now. However, in the new version, there's a line `del self.pad` while `self.pad` is used later in `convole`, I commented `del self.pad` to make it work. I guess the `self.pad` used in `convolve` should be `self.crop`?

f0k:
> I guess the `self.pad` used in `convolve` should be `self.crop`?

Yes, sorry, thanks for trying the code!

fvisin:
> We still need a way to allow custom output shapes. For example, blocks defines two additional parameters unused_edge and original_image_size for this: https://github.com/mila-udem/blocks/blob/b853b570f14e7494c5f95179f1cd977ffa69e91b/blocks/bricks/conv.py#L189-L204
The first one gives the number of additional zeros to pad the right and bottom border of the input with, allowing values between 0 and stride - 1. The second one directly gives the output size (a bit more difficult to check for correctness, but probably easier for the user). Which one(s) should we support?

I agree that the second one is probably easier for the user. The typical use case I would expect is an shrinking encoding pathway followed by one or more `TransposeConv2DLayer` to restore the original resolution. To specify the `unused_edge` the user would have to do some (possibly nasty) trickery to detect the amount of *implicit cropping* done by the encoding layers (e.g., a valid convolution 2x2 will implicitly crop one column/row if the input is odd, but will not crop anything on an even input). Specifying the desired shape should instead be easy.

> Both won't give a convenient way of inverting a convolution with variable input size, as only a MergeLayer can query the symbolic input shape of a convolution to be inverted. Probably we should leave that job for the InverseLayer, possibly along with #447.

I didn't look into your code yet, but I guess both arguments could be allowed to be symbolical values (meaning accept either symbolical or numerical values). If that's possible, then the user can just find his own way to get a tensor with the desired (variable) padding or shape. That would be ideal.

f0k:
Added some basic tests and squashed into a single commit. Tests pass and coverage is complete: https://coveralls.io/builds/5756402
For some reason, coverage is not reported in our PRs any longer. What's up with that?

Things to be discussed:
* `pad` is renamed to `crop` here since it actually defines how much is cropped from the zero-padding border. We could also have a `pad` keyword argument that is internally "inverted" (valid <-> full, etc.), but this would probably make it more confusing for people who want to "undo" a particular convolution in their network, because then they'd have to invert the argument as well.
* There's no way to specify a custom output shape yet. I'm thinking of adding an `output_size` argument akin to the `filter_size` and `stride` argument -- i.e., it can be a single integer or a tuple. This will require some more changes in documentation and tests -- shall we keep this for a future PR or integrate it here? The trouble is that I'm busy with a ton of other things and don't want to hold this PR back for much longer.

Against @fvsin's recommendation, I'm still using `AbstractConv2d_gradInputs` directly. This allows the code to cleanly distinguish between compile-time output shape and run-time output shape.

benanne:
I agree, no need to hold this up for much longer, better to add the `output_size` argument in a separate PR since it wouldn't be backwards incompatible anyway.

f0k:
Added remaining tests and fixed a bug (untied biases wouldn't work because of the self.pad/self.crop renaming). Still has [full test coverage](https://coveralls.io/builds/5802409). Ready to merge!

benanne:
LGTM, haven't gone through all the code/tests in detail though. Maybe some other people in this thread (particularly the ones who plan to use the code) should also take a look :)

f0k:
@fvisin, would you like to briefly check if everything looks okay to you (except that I still use `AbstractConv2d_gradInputs` as explained above, and that `output_size` will only be added later)?

fvisin:
@f0k I already read the code out of curiosity some days ago and LGTM. 
I planned to look into it more in detail probably tomorrow night, but I am behind my schedule with two other projects so I am not sure I will make it. I'll try my best though!

lasagne/layers/conv.py
```diff
@@ -80,6 +82,74 @@ def conv_output_length(input_length, filter_size, stride, pad=0):
...
+def conv_input_length(output_length, filter_size, stride, pad=0):
+    """Helper function to compute the input size of a convolution operation
+
+    This function computes the length along a single axis, which corresponds
+    to a 1D convolution. It can also be used for convolutions with higher
+    dimensionalities by using it individually for each axis.
+
+    Parameters
+    ----------
+    output_length : int
```
fvisin: Should be `int` or `None`

lasagne/layers/conv.py
```diff
@@ -80,6 +82,74 @@ def conv_output_length(input_length, filter_size, stride, pad=0):
     return output_length
 
 
+def conv_input_length(output_length, filter_size, stride, pad=0):
+    """Helper function to compute the input size of a convolution operation
```
fvisin: To me 'output size of a transposed convolution operation' would be more clear, but it's very subjective.

lasagne/layers/conv.py
```diff
@@ -536,3 +606,176 @@ def convolve(self, input, **kwargs):
...
+    Performs the backward pass of a 2D convolution (also called transposed
+    convolution, fractionally-strided convolution or deconvolution in the
+    literature) on its input and optionally adds a bias and applies an
+    elementwise nonlinearity.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape. The
+        output of this layer should be a 4D tensor, with shape
```
fvisin: Again, very subjective but "this layer" here might refer to both the TransposedConv one and the incoming one. Something along the lines of "The output of this incoming layer" would not be ambiguous.

lasagne/layers/conv.py
```diff
@@ -536,3 +606,176 @@ def convolve(self, input, **kwargs):
...
+
+    References
+    ----------
+    .. [1]: Vincent Dumoulin, Francesco Visin (2016):
+            A guide to convolution arithmetic for deep learning. arXiv.
+            http://arxiv.org/abs/1603.07285,
+            https://github.com/vdumoulin/conv_arithmetic
+    """
+    def __init__(self, incoming, num_filters, filter_size, stride=(1, 1),
+                 crop=0, untie_biases=False,
```
fvisin: Minor comment, but line break here is not necessary.

lasagne/layers/conv.py
```diff
@@ -536,3 +606,176 @@ def convolve(self, input, **kwargs):
...
+                      in zip(input_shape[2:], self.filter_size,
+                             self.stride, crop)))
+
+    def convolve(self, input, **kwargs):
+        border_mode = 'half' if self.crop == 'same' else self.crop
+        op = T.nnet.abstract_conv.AbstractConv2d_gradInputs(
+            imshp=self.output_shape,
+            kshp=self.get_W_shape(),
+            subsample=self.stride, border_mode=border_mode,
+            filter_flip=not self.flip_filters)
```
fvisin: Why is the flag inverted? I am not sure if this can be confusing for the final user.

fvisin:
Ok, done! Everything LGTM, I had to refresh my memory on `AbstractConv2d_gradInputs` but the types of the arguments that you pass to the constructor and to the node are correct. I can't wait for `output_size` to be added, but I am happy this will be merged in the meanwhile. Nice job!

lasagne/layers/conv.py
f0k: Yes, I find that confusing as well. We need to change *all* existing docstrings, that should become a separate PR. I'd leave it here for consistency.

lasagne/layers/conv.py
f0k: If you see the operation as: "convolution of dilated input", then `self.flip_filters` should tell whether we do a convolution (`self.flip_filters=True`) or correlation (`self.flip_filters=False`). As gradInputs does exactly the opposite (a correlation if the forward pass is a convolution, and a convolution if the forward pass is a correlation), we need to invert it here. The reasoning is about the same as for renaming `pad` to `crop` -- I would like to present the operation as something straightforward, not as the backward pass of a strided convolution (it just happens to use that for its implementation).

lasagne/layers/conv.py
f0k: Good point! Will fix both this and `conv_output_length`.

lasagne/layers/conv.py
f0k: But then we'd need all the explanations for the stride/dilation and padding/cropping we have in `TransposedConv2DLayer`. I find it much easier to document this as computing the input size of a convolution for which you know the output size.

lasagne/layers/conv.py
f0k: Thanks, but that's for consistency with the existing layers. I guess the idea was to have `W` and `b` on the same line.

lasagne/layers/conv.py
fvisin: Fair enough. As I wrote it depends on the terminology you are used to. :)

lasagne/layers/conv.py
fvisin: Agreed.

lasagne/layers/conv.py
fvisin: Wow, I would have never thought about consistency on break lines on the arguments! That's very accurate. Works for me anyway!

lasagne/layers/conv.py
fvisin: I see your point. Works for me.

lasagne/layers/conv.py
f0k: To be fair, the consistency isn't because I've paid extra close attention when writing the code, but because I started with copy/pasting the Conv2DLayer :)

PR #613, opened 2016-02-17
Remove forgotten `mode` argument from BatchNormLayer constructor

f0k:
My bad. At some point during development, `BatchNormLayer` had a `mode` argument. It was inadvertently left in the constructor when removing it from the implementation.


benanne:
LGTM, merging

PR #614, opened 2016-02-17
Clarifications for multi-output networks

jonathanstrong:
A few additions to clarify how multi-output networks should be handled, per the discussion here: https://groups.google.com/forum/#!topic/lasagne-users/61GbgRqKjAo


f0k:
Thanks! Some of the lines are too long, though, we have a hard limit of 79 characters to comply with PEP8. (Not your fault, you couldn't tell when using the web interface.)

And I'm not sure where to put the clarification, actually. When re-reading the existing docstrings, they already say thing like "Its main use is to collect all layers of a network just given the output layer(s)." -- so they advertise supplying all output layers. The first sentence of `get_all_params` should probably say `network` instead of `layer`, though. What do you think? Is there a less verbose way of clarifying things such that you'd have known what to do?

PR #618, opened 2016-02-22
update squared_error docstring

kli-nlpr:
Change
```
Returns
    -------
    Theano tensor
        An expression for the item-wise squared difference.
```

To
```
Returns
    -------
    Theano tensor
        An expression for the element-wise squared difference.
```


f0k:
:+1:, thanks!

PR #619, opened 2016-02-22
Update multiclass_hinge_loss docstring

kli-nlpr:
According to the code,
I think the original description

```
L_i = \\max_{j \\not = p_i} (0, t_j - t_{p_i} + \\delta)
```

Maybe should change to this

```
L_i = \\max_{j \\not = t_i} (0, p_j - p_{t_i} + \\delta)
```

where `L_i` is the loss for the `i` th sample, `t_i`  is the ground truth label of the `i` th sample.
`p` is the predicted probability vector of the `i` th sample.

Becasue in the original description, `t_{p_i}`  is meaningless since `p_i`  is the predicted probabilty which is a decimal.

### DIFF
--- a/lasagne/objectives.py
+++ b/lasagne/objectives.py
@@ -250,7 +250,7 @@ def binary_hinge_loss(predictions, targets, binary=True, delta=1):
 
 
 def multiclass_hinge_loss(predictions, targets, delta=1):
-    """Computes the item-wise multi-class hinge loss between predictions and targets.
+    """Computes the multi-class hinge loss between predictions and targets.
 
     .. math:: L_i = \\max_{j \\not = t_i} (0, p_j - p_{t_i} + \\delta)
 


lasagne/objectives.py
```diff
@@ -250,9 +250,9 @@ def binary_hinge_loss(predictions, targets, binary=True, delta=1):
 
 
 def multiclass_hinge_loss(predictions, targets, delta=1):
-    """Computes the multi-class hinge loss between predictions and targets.
+    """Computes the item-wise multi-class hinge loss between predictions and targets.
```
f0k: This makes the line too long. Please undo.

lasagne/objectives.py
```diff
@@ -250,9 +250,9 @@ def binary_hinge_loss(predictions, targets, binary=True, delta=1):
...
 
-    .. math:: L_i = \\max_{j \\not = p_i} (0, t_j - t_{p_i} + \\delta)
+    .. math:: L_i = \\max_{j \\not = t_i} (0, p_j - p_{t_i} + \\delta)
```
f0k: Good catch!

f0k:
Can you squash this into a single commit? You'll need to clone the repository to your machine with `git`, then do `git checkout patch-2`, `git reset --soft 9f10cad`, `git commit --amend --no-edit`, and `git push --force`.

f0k:
Will merge from the command line (to squash the two commits) after #846.

Sentient07:
I'm not able to add @kli-nlpr 's fork of Lasagne to my remote. It shows unknown repository here (on the top). So I will open a new PR

Sentient07:
I think this PR can be closed? (Master seems to be same as the diff that this proposes)

f0k:
No, master still has the wrong version: https://github.com/Lasagne/Lasagne/blob/master/lasagne/objectives.py#L305
Otherwise this PR would have been marked as conflicting.
But I'll merge this myself, no need for a new PR, thanks!

Sentient07:
> No, master still has the wrong version

I had made this change previously and forgot to commit. It hence was moving around my other branches as well. My bad, sorry

f0k:
Merged manually. Closing.

PR #620, opened 2016-02-22
Update  LeakyRectify docstring

kli-nlpr:


lasagne/nonlinearities.py
```diff
@@ -159,8 +159,8 @@ def rectify(x):
 
 # leaky rectify
 class LeakyRectify(object):
-    """Leaky rectifier :math:`\\varphi(x) = \\max(\\alpha \\cdot x, x)`
-
+    """Leaky rectifier :math:`\\varphi(x) = (x > 0)? x : \\alpha \\cdot x`
+
```
f0k: Good point, the previous formulation is only correct for alpha < 1. The `(x > 0) ? ` formulation is not actually valid mathematics, though. I know I've used it somewhere else because I didn't have a better idea. The tough part is that it needs to fit into a single line in order to be displayed correctly in the overview table, and a real case distinction will need too much markup.

In any case, you introduced trailing whitespace in line 163, PEP8 doesn't like this! Can you remove this and squash your commits into one (or amend your commit)?

f0k:
Can you squash this into a single commit? You'll need to clone the repository to your machine with `git`, then do `git checkout patch-2`, `git reset --soft aaff8cd`, `git commit --amend --no-edit`, and `git push --force`.

f0k:
Mhm, somehow the commit message from #619 got here... can you do
```bash
git checkout patch-2
git commit --amend -m "Update LeakyRectify docstring"
git push --force`?
```

f0k:
Good to merge after #846.

PR #622, opened 2016-02-22
Update conf.py

ebenolson:
Currently equation rendering is broken on RTD when using https.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -39,6 +39,10 @@ extensions = [
 #    'sphinx.ext.napoleon',  # alternative to numpydoc -- looks a bit worse.
 ]
 
+# See https://github.com/rtfd/readthedocs.org/issues/283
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -39,6 +39,10 @@
 #    'sphinx.ext.napoleon',  # alternative to numpydoc -- looks a bit worse.
 ]
 
+# See https://github.com/rtfd/readthedocs.org/issues/283
+mathjax_path = ('https://cdn.mathjax.org/mathjax/latest/MathJax.js?'
+                'config=TeX-AMS-MML_HTMLorMML')
```
f0k: What about `mathjax_path = ('//:cdn.mathjax.org/...'` so it uses the same protocol RTD is viewed with? Hmm no, this would probably break viewing files locally over the `file://` protocol. Defaulting to https seems reasonable then.

f0k:
Merging, thanks!

PR #627, opened 2016-03-01
Issue #625 Make FlattenLayer support `None` in input shape

### LINKED ISSUE #625: Make FlattenLayer support `None` in input shape
Currently, `FlattenLayer` does not work if any of the axes that are flattened have a shape of `None`. So this does not work:

```
l = lasagne.layers.InputLayer((None, None, None))
l = lasagne.layers.FlattenLayer(l)
```

Instead, users get an unhelpful error message because Lasagne attempts to compute the output shape nevertheless: 

```
TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'
```

(see https://groups.google.com/forum/#!topic/lasagne-users/0fa-GRH8wwo )

The shape computation is here: https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/shape.py#L46-L49

I think we can easily adapt this to simply return `None` for the flattened dimension of any of the constituent dimensions from the input shape have size `None`.

Any takers? :) If nobody wants to I can probably do a PR this weekend. Shouldn't be too much work, but I don't have much time to spare at the moment.

chuzui:
Fixes #625

### DIFF
--- a/lasagne/layers/shape.py
+++ b/lasagne/layers/shape.py
@@ -44,9 +44,14 @@ class FlattenLayer(Layer):
             raise ValueError('Dim must be >0, was %i', outdim)
 
     def get_output_shape_for(self, input_shape):
-        shp = [input_shape[i] for i in range(self.outdim-1)]
-        shp += [int(np.prod(input_shape[(self.outdim-1):]))]
-        return tuple(shp)
+        to_flatten = input_shape[self.outdim - 1:]
+
+        if any(s is None for s in to_flatten):
+            flattened = None
+        else:
+            flattened = int(np.prod(to_flatten))
+
+        return input_shape[:self.outdim - 1] + (flattened,)
 
     def get_output_for(self, input, **kwargs):
         return input.flatten(self.outdim)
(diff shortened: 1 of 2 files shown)


f0k:
That's wrong, unfortunately, `get_output_shape_for()` must not return `None`. We don't have a set of generic `Layer` tests that all layers have to go through, maybe we should? It's difficult to accomplish, though.

The old implementation:
```python
shp = [input_shape[i] for i in range(self.outdim-1)]
shp += [int(np.prod(input_shape[(self.outdim-1):]))]
return tuple(shp)
```
should change to
```python
to_flatten = input_shape[self.outdim:]
if any(s is None for s in to_flatten):
    flattened = None
else:
    flattened = int(np.prod(to_flatten))
return input_shape[:self.outdim-1] + (flattened,)
```

chuzui:
Tnank you, f0k. I misunderstood this function.

f0k:
Perfect, thanks! Will merge tomorrow if there are no further comments.

benanne:
LGTM, thanks for contributing!

PR #628, opened 2016-03-01
Update to reflect changes in Theano Tutorial

f0k:
The [Theano Tutorial](http://deeplearning.net/software/theano/tutorial/) has changed: The "Graph Structures" part has been moved from right after "More Examples" to "Further Readings" near the very end. As we asked people to read up to "Graph Structures", this means they'd now have to read basically everything. I'd change this to suggest to read up to "More Examples".


benanne:
Is the graph structures part important? I suppose not since they moved it all the way to the back? Otherwise it's fine by me :)

f0k:
It's not terribly important, I just found it nice to know: http://deeplearning.net/software/theano/extending/graphstructures.html
I've added it back in with some note.

benanne:
LGTM :)

Issue #537, opened 2015-12-15
MergeLayer accepting dictionaries


f0k:
Right now, our MergeLayer base class, the only layer class that can handle multiple inputs, requires its input layers to be given as a list. This is a good fit for the ConcatLayer, for example, which concatenates multiple input tensors in a specific order. In #522, we found that this is a bad fit for the recurrent layers, though, which accept multiple input layers meaning very different things: the actual input, a mask, and an initial hidden state. For this use case, it would be much more suitable to access the different input layers (and inputs, and input shapes) by name: that is, `incomings` in the constructor, `input_shapes` in `get_output_shape_for()` and `inputs` in `get_output_for()` should be dictionaries rather than lists.

What can we do to allow this?
a) Add another `MergeLayer` base class (NamedMergeLayer` or `DictMergeLayer` or something) that has a different interface and adapt `lasagne.layers.get_output()` accordingly?
b) Expand `MergeLayer` to accept either a list or a dictionary in its constructor and change its behavior accordingly? -> This would mean that existing merge layers need to cope with the possibility of being handed a dictionary.
c) ~~Change the `Layer` base class to always allow multiple input layers?~~ -> This wouldn't make sense for most layers such as the `DenseLayer`.

There's also the side issue of #337: We may want to allow a single layer to output a list or dictionary of tensors. How would that interplay with merge layers? Is is to be treated completely orthogonal (i.e., a merge layer is for merging multiple layers, not for processing multiple inputs) or do we need to account for this?

At first glance, a) seems like the cleanest solution, with the disadvantage that having two different merge layer base classes will be confusing, especially if we cannot rename the current one.

craffel:
> a) Add another MergeLayer base class (NamedMergeLayer or DictMergeLayer or something) that has a different interface and adapt lasagne.layers.get_output() accordingly?

I think the current `MergeLayer` has semantics which are different enough (expects a list of inputs which can be concatenated, where order matters) from a "named" merge layer (accepts certain inputs) that having them separate is ok with me.  If we change `MergeLayer` to accept dicts, that would also maybe break user-created `MergeLayer` subclasses (if a dict was passed).  So I'm in support of a).

skaae:
> There's also the side issue of #337: We may want to allow a single layer to output a list or dictionary of tensors. How would that interplay with merge layers? Is is to be treated completely orthogonal (i.e., a merge layer is for merging multiple layers, not for processing multiple inputs) or do we need to account for this?

[...]

Issue #356, opened 2015-08-06
Code duplication in recurrent layers


craffel:
As noted by @f0k here https://github.com/Lasagne/Lasagne/pull/294#issuecomment-128004387, there is a lot of code duplication across the recurrent layers.  It would be nice to merge this either into a common base class, or into common functions.

PR #629, opened 2016-03-02
Added recurrent containers so that any graph of layers can be iterated over

stevenxxiu:
This PR addresses #425. Some main points in the implementation are:
- The ability to iterate over a sequence using any graph of layers which maps inputs to outputs.
- The ability to iterate where a sequence is absent, by specifying the # of steps to iterate.
- Cells (layers which have a state per time step) and normal layers (layers which do not) can be stacked together to form the graph.
- Information about whether an initial state is updated or not is passed on to `scan`, so that forms of RNN dropout can be implemented more efficiently.


stevenxxiu:
@f0k Tests are passing now.

setup.cfg
```diff
@@ -4,7 +4,5 @@ dev = develop easy_install lasagne[testing]
 [pytest]
 addopts =
     -v --doctest-modules
-    --cov=lasagne --cov-report=term-missing
```
benanne: Did you mean to commit this? We definitely don't want to get rid of coverage testing :)

benanne:
This looks pretty beefy, I guess it would be good to get some input from @skaae and @craffel as well :)

craffel:
tbh, I'm a bit overwhelmed by the changes here -- @stevenxxiu can you post a few usage examples and/or an overview of the usage of the new classes so I can get an idea of what your new API is?  I also would not object to more docstrings/comments.

setup.cfg
stevenxxiu: @benanne Sorry it's a bit unclear but I moved coverage over to `.travis.yml` instead. The thing is pycharm is a bit grumpy and the debugger breaks when `--cov` is in there. I thought this was a good compromise.

stevenxxiu:
Sure, the api looks like this (there are some more test cases [here](https://github.com/stevenxxiu/Lasagne/blob/recurrent/lasagne/tests/layers/test_recurrent.py#L1107)):

    # layered implementation
    l_rec_1 = lasagne.layers.ReshapeLayer(l_inp, (num_batch * seq_len, -1))
    l_rec_1 = lasagne.layers.DenseLayer(l_rec_1, num_units_1)
    l_rec_1 = lasagne.layers.ReshapeLayer(l_rec_1, (num_batch, seq_len, -1))
    l_rec_1 = LSTMLayer(l_rec_1, num_units=num_units_2)
    l_rec_1 = LSTMLayer(l_rec_1, num_units=num_units_3)

    # equivalent cell implementation
    cell_inp = lasagne.layers.InputLayer((num_batch, n_features))
    cell = DenseLayer(cell_inp, num_units_1)
    cell = LSTMCell(cell, num_units_2)['output']
    cell = LSTMCell(cell, num_units_3)['output']
    l_rec_2 = lasagne.layers.RecurrentContainerLayer({cell_inp: l_inp}, cell)

The idea is that each cell can accept multiple inputs and outputs, and similarly for the container layer. Multiple inputs is already implemented by specifying which cell's input binds to which layer's input, which provides the sequence.

An example of the ability to iterate over where a sequence is absent:

    cell_inp = InputLayer((num_batch, n_features))
    cell = CustomRecurrentCell(
        None, None, DenseLayer(cell_inp, n_features), hid_init=l_inp)['output']
    l_rec_1 = lasagne.layers.RecurrentContainerLayer(
        {}, cell, n_steps=seq_len, only_return_final=True)

There's some more discussion over at #425 of my implementation if you're interested.

setup.cfg
f0k: It's good with respect to Travis, but it also means that users cannot simply enter `py.test`, but have to run the full command as in `.travis.yml`, though. I'm not sure we want that. It makes it more intuitive to just test a single (you don't need to know about `-k`), but harder to run the full test suite. Isn't this more of a bug in PyCharm?

stevenxxiu: Fair enough, I'll just get rid of that commit and use `skip-worktree` on my end so I can still debug.

f0k:
I think it's a very good start! Sure, there are a lot of changes and it would need more comments/documentation, but I'm fine with deferring that a bit until we know we're okay with the API. The examples I've seen look very reasonable, we may want to double-check that it covers all use cases.

The single thing that bugs me most is the change to `MergeLayer` -- we should finish up the discussion in #537 and separate this out into its own pull request, then rebase this one on top of it. We'll almost surely want a separate class for the dictionary case. It's good that this long-standing issue is receiving some pressure now!

Some other design decisions also deserve closer scrutinization. I'll try to go through the diff and ask you some questions soon. In any case, I'm excited we're finally addressing this!

skaae:
I'm not sure i understand the api.

```Python
# equivalent cell implementation
cell_inp = lasagne.layers.InputLayer((num_batch, n_features))
cell = DenseLayer(cell_inp, num_units_1)
cell = LSTMCell(cell, num_units_2)['output']
cell = LSTMCell(cell, num_units_3)['output']
l_rec_2 = lasagne.layers.RecurrentContainerLayer({cell_inp: l_inp}, cell)
```
`cell` is the top of the network inside scan? What does the dictionary mean? I don't see l_inp being defined anywhere?

Its not clear to me what the new possibilities are with the container layer. I'll just name a few usecases that are not supported in the the current recurrent layers and maybe you can tell if they would be supported with containers? 

1. Combining an RNN with the spatial transformer layer and feed the image glimpse at timestep t-1 into the RNN at timestep t? 
2. Encoder/Decoder model with attention? (If we implemented an attention layer)

> An example of the ability to iterate over where a sequence is absent:

I guess this covers RNN autoencoders where the last hidden state of the encoder is used as input at every timestep for the decoder?

Maybe we should split recurrent.py into two files? It was already pretty long?

skaae:
btw does our code handle randomness inside scan?

stevenxxiu:
> cell is the top of the network inside scan? What does the dictionary mean? I don't see l_inp being defined anywhere?

Yes cell is the top. The dictionary just maps a layer to be iterated over to a cell input. I've edited the above post, `l_inp` is just any layer to be iterated over.

> I guess this covers RNN autoencoders where the last hidden state of the encoder is used as input at every timestep for the decoder?

I think so, but I'm not 100% sure as I haven't implemented a text autoencoder yet. Same goes with attention. I think the soft attention in "Show, Attend and Tell: Neural Image Caption Generation with Visual Attention, Kelvin Xu et al." can be made into a layer, as it's just re-weighting the same vectors {a}_i in each time step using a function of the previous lstm hidden state and {a}_i.

But that does suggest a problem with the current api, I need a way to feed the output at t-1 into a different cell's input at time t. Take a dummy example, say I want to feed the output of a lstm's into the next time step's input. Here's one solution I thought of:

    cell_inp_1 = lasagne.layers.InputLayer((num_batch, n_features))
    cell_inp_2 = lasagne.layers.InputLayer((num_batch, n_features))
    cell = ConcatLayer(cell_inp_1, cell_inp_2)
    cell = LSTMCell(cell_inp_1, num_units_2)['output']
    l_rec_2 = lasagne.layers.RecurrentContainerLayer({cell_inp_1: l_inp}, {cell_inp_2: cell}, cell)

Here `cell_inp_1` is the layer input, and `cell_inp_2` is the lstm's output. `{cell_inp_1: l_inp}` specifies which sequence we are feeding into which cell. `{cell_inp_2: cell}` specifies which cell's previous state we are feeding into which cell.

> Combining an RNN with the spatial transformer layer and feed the image glimpse at timestep t-1 into the RNN at timestep t? 

Which paper is this? I haven't read about spatial transformer networks yet.

> btw does our code handle randomness inside scan?

There can be a dropout mask that is unchanged throughout inside scan. Is this what you mean?

skaae:
> Which paper is this? I haven't read about spatial transformer networks yet.

Something similar to RAM: http://arxiv.org/abs/1406.6247

skaae:
> I haven't implemented a text autoencoder or attentional model yet, but I think the soft attention in Show, Attend and Tell: Neural Image Caption Generation with Visual Attention, Kelvin Xu et al. can be made into a layer, as it's just re-weighting the same vectors {a}_i in each time step using a function of the previous lstm hidden state and {a}_i.

Yes. You need a layer that takes the output from the decoder and transform it into a weighted sum.Something like:

```
rnn_input = attentionlayer(decoder_output, h_[t-1])
h_t = lstmcell(rnn_input)
```
Where `decoder_ouput` is a `num_batch, decoder_len, num_unts` matrix and the attention layer sums over the decoder_len dimension.

f0k:
> There can be a dropout mask that is unchanged throughout inside scan. Is this what you mean?

No, he's referring to #429. This is orthogonal to this PR, though (the fix is always the same, no matter if we do it before or after revamping the recurrent layers).

> But that does suggest a problem with the current api, I need a way to feed the output at t-1 into a different cell's input at time t.

Yes, that's one of the use cases we definitely want to cover: https://github.com/Lasagne/Lasagne/issues/425#issuecomment-157064360
Other use cases were listed here: https://github.com/Lasagne/Lasagne/issues/425#issuecomment-157744032

> Here's one solution I thought of:
```python
cell_inp_1 = lasagne.layers.InputLayer((num_batch, n_features))
cell_inp_2 = lasagne.layers.InputLayer((num_batch, n_features))
cell = ConcatLayer(cell_inp_1, cell_inp_2)
cell = LSTMCell(cell, num_units_2)['output']
l_rec = lasagne.layers.RecurrentContainerLayer({cell_inp_1: l_inp}, cell, {cell_inp_2: cell})
```
Here `cell_inp_1` is the layer input at each time step, and `cell_inp_2` is the lstm's output. `{cell_inp_1: l_inp}` specifies which sequence we are feeding into which cell. `{cell_inp_2: cell}` specifies which cell's previous state we are feeding into which cell.

Shouldn't `cell_inp_2` have shape `(num_batch, num_units_2)` to match the `LSTMCell`? Otherwise `{cell_inp_2: cell}` won't match up. If that was just a mistake, then I think I understand what you're suggesting and it looks good from the outside. It would allow you to have an arbitrarily complex processing chain within the container and still feed back the output (or the LSTM/GRU cell state) of some intermediate cell to any input cell.

craffel:
Sorry, I'm confused about something fundamental - none of your `InputLayer`s in any of your examples have a "number of time steps" dimension.  This makes sense for cells, but not for recurrent layers.  For example, I don't see how sequences are passed in to this example from above:

```Python
l_inp = lasagne.layers.InputLayer((num_batch, n_inputs))
cell_inp = lasagne.layers.InputLayer((num_batch, n_features))
cell = DenseLayer(cell_inp, num_units_1)
cell = LSTMCell(cell, num_units_2)['output']
cell = LSTMCell(cell, num_units_3)['output']
l_rec_2 = lasagne.layers.RecurrentContainerLayer({cell_inp: l_inp}, cell)
```

Sorry for being slow here, but I want to understand the API before I dig into the code.

f0k:
> Sorry, I'm confused about something fundamental - none of your InputLayers in any of your examples have a "number of time steps" dimension.

Right, seems he forgot when posting here, the tests have them: https://github.com/stevenxxiu/Lasagne/blob/b223654/lasagne/tests/layers/test_recurrent.py#L1107

> Sorry for being slow here

No need to, we still shouldn't rush things!

craffel:
> Right, seems he forgot when posting here, the tests have them:

Ah, ok, I see that.

After spending a bit looking over the code, I think I have a decent idea of what is going on at a high level - a `CellLayer` is essentially a `MergeLayer` with potentially multiple inputs/outputs and the ability to accept initial states for each input.  Recurrent layers are built by passing their input sequence one step at a time to their `CellLayer`.  This seems reasonable, but I don't think I'm going to be able to do a full parse-from-scratch of the code without a lot more comments/documentation/examples with the amount of bandwidth I have right now.

f0k:
> Recurrent layers are built by passing their input sequence one step at a time to their `CellLayer`

Or, more general, to their graph of `CellLayer` and ordinary feed-forward `Layer` instances. The container can have multiple inputs/outputs as well.

> This seems reasonable, but I don't think I'm going to be able to do a full parse-from-scratch of the code without a lot more comments/documentation/examples with the amount of bandwidth I have right now.

Same here. For now, the code is a bit difficult to take in. But I also see that fully documenting everything may be wasted effort if we want further changes on the outside. So I'd suggest to have a look at the end-user API first: Does it cover all the use cases, in readable end-user code? If we're fine with this, we can scrutinize the internals, and this will need correct docstrings especially for the new Cell methods, and some more comments (we can ask for them in the diff then).

What I like from a first look is that all the redundant code seems to have been centralized, which is exactly what we wanted (#356).

craffel:
> So I'd suggest to have a look at the end-user API first: Does it cover all the use cases, in readable end-user code?

I think the best way to test this is to write some examples, e.g. an attention network, text prediction ala Karpathy, etc.

Sandy4321:
Pls help find code example for sentiment analysis for short texts

f0k:
> I think the best way to test this is to write some examples, e.g. an attention network, text prediction ala Karpathy, etc.

I think short code snippets would do for now. Let's just collect the primitives we'd want to see.

* map N >= 0 input sequences to M > 0 output sequences
  * single sequence to single sequence
  * single sequence to two sequences
  * two sequences to single sequence
  * zero sequences to single sequence
* initial hidden state variants:
  * fixed initial hidden states
  * learned initial hidden states
  * initial hidden states fed in from (an)other layer(s)
* vanilla RNN, LSTM, GRU (guess we can skip that, it's the same as before)
* convolutional RNN
* convolutional LSTM or GRU (a sketch is enough -- this would probably entail writing a custom cell?)
* stack multiple LSTMs/GRUs in a single container
* feed back output as input
* vanilla RNN with dropout mask constant over time

Is there anything missing? This covers the sequence-to-one, one-to-sequence, and sequence-to-sequence cases, so it also covers encode/decode architectures, and it covers text generation.

craffel:
I think the attention network also requires some additional machinery that might not be covered there, because it would need to use the output of two recurrent processes (`h` and `s`) in order to compute the next step in `s`.  E.g.
 1. `h_t = f_1(x_t, h_{t - 1})` (normal recurrent computation)
 1. `c_t = f_2(s_{t - 1}, h)` (depends on `s_{t - 1}` and the entire sequence `h_{1, ... T}`)
 1. `s_t = f_3(c_t, s_{t - 1})` (normal recurrent computation)

I can try to write that once I see the other examples.

benanne:
This looks excellent so far :)

>  The dictionary just maps a layer l_inp to be iterated over to a cell input.

One concern I have about this is that currently, dictionaries are used in Lasagne to:
- map names to layers, e.g. `{ 'foo': layer_foo }`
- map layers to Theano expressions, e.g. `{ layer_foo: x }`

These two cases are easily distinguished visually because the former will typically have string keys, and the latter will have layer references as keys.

If I understand correctly, the `RecurrentContainerLayer` would take a dictionary mapping `Layer` instances to other `Layer` instances, which is not easily visually distinguishable from the 2nd case. As a seasoned Lasagne user it would definitely confuse me if I saw it for the first time. Maybe there is some way we can make this usage more 'distinct'? This is not a showstopper of course, I'm just thinking about how we can make the API as intuitive as possible.

stevenxxiu:
Many thanks for all the positivity, feedback and suggestions guys!

I've been working hard the past week on some more precompute issues, refactorings to make the code better, and two of the above use cases mentioned:

- feeding a step into at time t from a cell at time t-1 (I think this is enough for attention, but I still haven't read the attentional paper mentioned by @skaae yet, will get to that!)
- feeding a layer outside the cell graph as a fixed input into the recurrence

For now I'll write some code examples to @f0k's primitives. Don't have a solution to @benanne's questions yet but hope it will become clearer once I write the code examples.

skaae:
> read the attentional paper mentioned by @skaae yet, will get to that!)

I dont don't feed the glimpses into the RNN in the spatial RNN paper. Don't be confused if you cant find it :)

stevenxxiu:
Ok I'll put examples in https://github.com/stevenxxiu/lasagne_recurrence_examples. seq to seq examples are there now. This time I've run the code so there should be no bugs, in the examples at least :)

> I dont don't feed the glimpses into the RNN in the spatial RNN paper. Don't be confused if you cant find it :)

Even the image attentional one, I'm not that familiar with computer vision as you can see (I'm currently doing sentiment analysis), but would definitely like to know more!

craffel:
Thanks for making that!  Some questions (some of these could be cleared up by documentation/comments - if you'd rather go that way that'd be great!)

1. https://github.com/stevenxxiu/lasagne_recurrence_examples/blob/master/primitives.py#L22 So outputs from the cells are retrieved with a dict-like __getitem__ format?  Why this and not `lasagne.layers.get_output(cell, output='output')` or something?
1. https://github.com/stevenxxiu/lasagne_recurrence_examples/blob/master/primitives.py#L23 So the `RecurrentContainerLayer` accepts a dict which maps `InputLayer`s for cells to an input layer, which should be iterated over to pass data to the cell?
1. https://github.com/stevenxxiu/lasagne_recurrence_examples/blob/master/primitives.py#L37 What is the identity layer doing?  Some sort of combining of the cells' outputs which is not concatenation?  Unless I am misunderstanding, `IdentityLayer` is not what I would call that.
1. https://github.com/stevenxxiu/lasagne_recurrence_examples/blob/master/primitives.py#L59 So how exactly are the output of these two cells being combined?  Concatenation, summing, etc?
1. https://github.com/stevenxxiu/lasagne_recurrence_examples/blob/master/primitives.py#L77 I think what Jan meant by "fixed hidden state" was that it would not be returned by `get_all_params`.  Is that the behavior when you pass a numpy constant as you're doing here?

Looking forward to those attention examples!  Thank you again for this!

stevenxxiu:
> Thanks for making that! Some questions (some of these could be cleared up by documentation/comments - if you'd rather go that way that'd be great!)

Do you mean in the code or examples? Putting all the docs in the code might take some time and it might be better for the api to be fixed first. But good questions, to respond:

1. The reason is that the item can be made into a layer. This has many advantages, including using that layer with other layers down the track. I took the syntax from @fok's suggestion [here](https://github.com/Lasagne/Lasagne/issues/337#issuecomment-165122642).

2. Indeed, that's exactly what it does. The current code also allows mapping the previuos output of a cell to the input of a cell, or a cell outside the graph to the input of a cell, in the third argument.

3. Not quite, the identity layer literally returns what's fed into it. This is a way to achieve a dictionary output layer using individual layers as input.

4. It's basically the following:

    for each time step:
        concatenate cell_1[t] and cell_2[t] (the concatenation is done through ConcatLayer)
        feed this into a dense recurrent layer to generate the new output

5. Ok this isn't quite what my example does. Whether init is fixed or not is set by `learn_init=False` for the entire recurrent container, which isn't really ideal. I think it's better for this to be set at the cell level and each init being optionally learnable. A possible api is:

    DenseRecurrentCell(..., W_in_to_hid=init.Uniform(), W_in_to_hid_learnable=True, ...

But that seems kind of convoluted, another approach may be to move the learnable into the init:

    DenseRecurrentCell(..., W_in_to_hid=init.Uniform(learnable=True), ...

> Looking forward to those attention examples! Thank you again for this!

My pleasure! I could possibly do some simple soft attention example, a working one if there's a suitable toy problem, but if there isn't any I could still provide the concept. Hard attention with the mask being a random variable is better but it's easier to implement soft attention first.

stevenxxiu:
@f0k What exactly is a convolutional lstm, is there a paper associated with it?

Sandy4321:
Steven
May you help find working example rnn rprop for short texts sentiment
analysis or classification
Or
Lstm or gru for texts classification
With word2vec embeddings
Thanks
Sander
Any other optimizations like quickprop are welcome

fok:
Guys, please stop mentioning or CCing me all the time. I'm @fok and have nothing to do with this project. You probably mean @f0k. Thx!

craffel:
> Do you mean in the code or examples? Putting all the docs in the code might take some time and it might be better for the api to be fixed first.

In the code.  I'm personally of the opinion that it's better to document first, and code later, because with docs I probably wouldn't have any of these questions!  But, do whichever you want.

> 1. The reason is that the item can be made into a layer. This has many advantages, including using that layer with other layers down the track. I took the syntax from @fok's suggestion here.

Ah, I see, thanks for the ref.  My input is late, but I don't think this is really natural/intuitive unless we also add a `__index__` method to the `Layer` baseclass which returns `get_output_for` when `'output'` is requested, so the syntax is universal.

> 3. Not quite, the identity layer literally returns what's fed into it. This is a way to achieve a dictionary output layer using individual layers as input.

Hmm, isn't this what @f0k's `DictMergeLayer` was supposed to do?  That's a much more intuitive name to me than `IdentityLayer`.

> 4. It's basically the following:
> ```
> for each time step:
>     concatenate cell_1[t] and cell_2[t] (the concatenation is done through ConcatLayer)
>     feed this into a dense recurrent layer to generate the new output
> ```

For me, it would be more intuitive to require the user to do the concatenation themselves, rather than implicit concatenating.

> Ok this isn't quite what my example does. Whether init is fixed or not is set by learn_init=False for the entire recurrent container, which isn't really ideal. I think it's better for this to be set at the cell level and each init being optionally learnable.

Agreed.

> any comments on which is preferable or any other suggestions?

Hmm, I forget if there is a way for the user to supply tags for a given parameter when constructing a layer, or if that can only happen in the `add_param` call.  This would be the most natural way, I think.

> Guys, please stop mentioning or CCing me all the time. I'm @fok and have nothing to do with this project. You probably mean @f0k. Thx!

Lol. @f0k I blame you for the ambiguous circle-shaped character in your name!  I don't think this will stop as long as new people are contributing to @f0k-associated projects.

stevenxxiu:
> In the code. I'm personally of the opinion that it's better to document first, and code later, because with docs I probably wouldn't have any of these questions! But, do whichever you want.

I see your point :)

> Ah, I see, thanks for the ref. My input is late, but I don't think this is really natural/intuitive unless we also add a __index__ method to the Layer baseclass which returns get_output_for when 'output' is requested, so the syntax is universal.

Actually yes this is what I implemented, with a small difference. Given that `l` is a `MergeLayer`, `l['output']` returns another layer, not the result of `get_output_for`. `CellLayer` just happens to be a `MergeLayer` that requires this quite often from the user-perspective (we don't usually need the hidden states except for the output!).

> Hmm, isn't this what @f0k's DictMergeLayer was supposed to do? That's a much more intuitive name to me than IdentityLayer.

I think `DictMergeLayer` is intended to be a base class, but I could indeed name it `DictMergeLayer` instead of `IdentityLayer` if the following are implemented instead of raising `NotImplementedError` as they do now:

            return input_shapes

            return inputs

The code explains why I've named it `IdentityLayer` for now, I guess.

> For me, it would be more intuitive to require the user to do the concatenation themselves, rather than implicit concatenating.

Yes this is what's happening, see https://github.com/stevenxxiu/lasagne_recurrence_examples/blob/master/primitives.py#L53.

> Hmm, I forget if there is a way for the user to supply tags for a given parameter when constructing a layer, or if that can only happen in the add_param call. This would be the most natural way, I think.

There is the option of `trainable=True` or `trainable=False` in `add_param`, but we need to expose that to the user somehow.

craffel:
> Given that l is a MergeLayer, l['output'] returns another layer, not the result of get_output_for.

Oops, yes, this is what I meant (I think).

> I think DictMergeLayer is intended to be a base class, but I could indeed use DictMergeLayer directly if the following are implemented instead of raising NotImplementedError as they do now, but I wonder if this is a good idea for base classes:

Yeah, I'm not sure what the best way to do that is.

> Yes this is what's happening, see https://github.com/stevenxxiu/lasagne_recurrence_examples/blob/master/primitives.py#L53.

Oops, you're right, sorry I missed that!

> There is the option of trainable=True or trainable=False in add_param, but we need to expose that to the user somehow .

Exactly.

elanmart:
@f0k I'm not a Lasagne dev, but I think a more general version of

> feed back output as input

would be
- feed any collections of `h_{i, tm1} ... h_{j, tm1}` to `h_{k, t}`

stevenxxiu:
@craffel Glad I've cleared things up a bit!

> > There is the option of trainable=True or trainable=False in add_param, but we need to expose that to the user somehow .

> Exactly.

I've added an `inits_fixed` argument to each cell, which can be specified like: `LSTMCell(..., inits_fixed={'cell', 'output'})`. I used fixed instead of trainable since params default to being trainable. I'll update the examples shortly.

Also got rid of the previous way of specifying dropout masks as a fixed hidden state. After actually implementing a value dropout LSTM I found that this didn't make much sense as the LSTM shouldn't need to output it's dropout mask. The previous approach enforced this since `get_output_shape()` was needed to determine the init param's shape, and whether an output changed or not was inferred by comparing `get_output_shape()` and `inits`. 

So it makes more sense to add the dropout mask a non-trainable param directly in the cell instead of this being managed by `RecurrentContainerLayer`.

@elanmart what is tm_1? The current code allows feeding any $h_{i,t-1}$ to $h_{j,t}$. Collections of hidden states cannot be fed in, but I don't think this is a huge problem as each cell already supports a dictionary of hidden states, or are you thinking of something else.

stevenxxiu:
Another thing which can be improved on is `CustomRecurrentCell`. It currently only supports a single sequence as input and adds the init output and hidden output layer. This can be made much more flexible, so that it supports many sequences as input and uses a bunch of layers to transform the hidden states at time t-1 to hidden states at time t.

This may appear to be like what `RecurrentContainerLayer` is doing already, but I think it still makes sense to have a `CustomRecurrentCell` since to me it makes more sense to have inits specified for cells instead of in the container.

But implementing precompute will take some work, so it might be easier to mark `CustomRecurrentCell` as internal for now and leave it for the future.

stevenxxiu:
@f0k I've added all the primitives you've mentioned except for the convolutional lstm/gru as I'm not sure as to what they are.

Would you guys mind having a look at the value dropout example? I'm not sure if the api is quite ideal there, as `seq_incoming` needs to be passed into the `BernoulliDropout` layer. This is since `incoming` can change per time step and doesn't make sense for inits.

trungnt13:
Just some comments concern both the old and new implementation:

- The logic of enabling peep_holes for LSTM is ambiguous, logically you can turn on (off) the peepholes by setting W_cell of Gate to None, simple check of None should be possible to do this. This code is so bad design
```
        if self.peepholes:
            self.W_cell_to_ingate = self.add_param(
                ingate.W_cell, (num_units, ), name="W_cell_to_ingate")

            self.W_cell_to_forgetgate = self.add_param(
                forgetgate.W_cell, (num_units, ), name="W_cell_to_forgetgate")

            self.W_cell_to_outgate = self.add_param(
                outgate.W_cell, (num_units, ), name="W_cell_to_outgate")
```
What happen if we enable peephole and set W_cell to None, vice versa, setting the W_cell and have the peephole = False (it just implicitly ignore W_cell). 
=> My point is remove the argument peepholes, and check W_cell of Gate, if it is not None, we enable it (like what we did for bias in dense layer), and during the step function, simple check like _if hasattr(self, 'W_cell_to_forgetgate')_ would do a good job. User have full control to enable the peephole for only some gates they want.
```
        if self.peepholes:
            # Compute peephole connections
            ingate += cell_previous*self.W_cell_to_ingate
            forgetgate += cell_previous*self.W_cell_to_forgetgate
```

- It will be nice if the Cell have methods get_constants() which return the constants for non_sequences, one possible use case for this is same dropout mask for all time step during recurrent computation.

---

For more general design, I think we should implement the RecurrentContainer to have multiple Cell:

- A Cell can have memory (LSTM) or a non-memory cell (GRU)
- I think it is better to constraint RecurrentContainer always has **1** hidden state, and it pass this states to all Cell during computation.
- The  hidden state is passed to all cell during recurrent. Each cell will return: _[hid_new, cell_new, outputs_new]_ . _cell_new_ will be None if cell don't have memory, and the cell can return multiple outputs or no output, then outputs_new is None also. Checking this would be easy based on the init states return by CellLayer.
- The the RecurrentContainer has to handle merging multiple new hidden states, since only 1 hidden states allowed to passed.

The biggest issue is allowing inter-cells communication:
- 2 or more Cells can share their memory.
- One Cells can receives previous cell states from all other cells.
- The order of cell computation (i.e. one Cell can take the new hidden_states, cell_states, outputs_state from previous computed Cell in the step). If we handle this precisely, it is extremely complex since you have to define a mapping between cells. Simplest approach would be **parallels** or **sequential** computation option for RecurrentContainer, then mind the users about the order they add Cell to RecurrentContainer.

---

@f0k : Can you explain more about the: Convolutional LSTM or GRU, the Gate output is always 2D, if we use Convolution within the Gate, it turns out to have at least 8 Convolution running in the Cell. I think the possible solution is reshape the Gate back to 4D and do the same for normal LSTM ?

stevenxxiu:
> For more general design, I think we should implement the RecurrentContainer to have multiple Cell:
> ...
> The hidden state is passed to all cell during recurrent. Each cell will return: [hid_new, cell_new, outputs_new] . cell_new will be None if cell don't have memory, and the cell can return multiple outputs or no output, then outputs_new is None also. Checking this would be easy based on the init states return by CellLayer.
> ...

I think this is a little restrictive as it only corresponds to the LSTM. The current implementation actually allows you to map use any hidden and output states - you take what you need and the implementation is not very complicated.

> 2 or more Cells can have the same precompute_input, I am not sure theano will optimize duplicated computation in this case, but we shouldn't have duplicated precompute_input as well.

Quite sure theano handles removing duplicate expressions so this should work, as we just precompute the expressions before passing them on to scan.

> 2 or more Cells can share their memory.
> One Cells can receives previous cell states from all other cells.

I think both are covered by the existing implemetation. You just need to feed the new state as step input.

> Simplest approach would be parallels or sequential computation option for RecurrentContainer, then mind the users about the order they add Cell to RecurrentContainer.

I'm not quite sure what you mean. Some things can't be done in parallel such as feeding the previous state to the new cell. Do you mind looking at my [examples repo](https://github.com/stevenxxiu/lasagne_recurrence_examples) and implement an example where the current api is too complicated?

> One more thing: is it important to enable strict mode and pass all parameters to the non_sequences of scan ? It would make the implementation simpler if we turn of strict mode ?

Just so you know the implementation already works :) and yes it does enforce strict mode.

trungnt13:
I read the example, really powerful implementation.
For the sequential and parallel RecurrentContainer, this is the [Design](https://drive.google.com/file/d/0B9wVmnaJ18ngdFI5cE5jTFpLUzA/view?usp=sharing)

> I think this is a little restrictive as it only corresponds to the LSTM
Simple Recurrent, GRU, LSTM ... they all have hidden states, I haven't seen Recurrent Neural Network without hidden states, at least 1 hidden state.

Cell can have memory or just be memoryless (in your implementation it just means that get_out_for return only one _{output:..}_ or also include other states _{cell:...,output:..}_. In general, it just a CellLayer with initial state or no initial state.

> I think both are covered by my existing implemetation. You just need to feed the new state as step input. The second one however requires at least the number of cell states compared to the states you want to feed into it, as cell states can only be tensors, not dictionaries (a cell already supports multiple states using a dictionary, e.g. return {'hidden': state_1, 'output': state_2}, I don't think we need a dictionary inside a dictionary).
Yes, I check the example, both are well-implemented. Just some comments:

- A mapping between InputLayer-InputLayer with different shape make me so confusing, but your explanation above is clear, I am not sure how new users will catch up with this changes.
- We have to re-specifed many duplicated information, finding cleaner way might constraint the API.
- I am confusing about your _ lstm_dropout_gal_ example, how do you keep the dropout_mask constant for all time_step:

```
        non_seqs, non_seqs_index = self._get_cell_params(
            step=True, precompute_input=self.precompute_input), {}
        for input_m, cell_m in self.step_incomings.items():
            if cell_m not in self.cells:
                non_seqs.append(inputs[input_m])
                non_seqs_index[input_m] = len(non_seqs) - 1
```
The step_incoming is mapping: cell_input to cell; it means (n_batch, n_features) will be divided into (1, n_features) then feed into each step for given cell or the whole (n_batch, n_features) for each time step (in this case step_incoming is constants added to non_sequences right?).

And, what else will be added to non_sequences in this part ? All the inits that not the outputs will be constants ?
```
        for cell_m in self.cells:
            if isinstance(cell_m, CellLayer):
                for name in set(cell_m.inits) - set(cell_m.output_shape):
                    input_layer = cell_m.input_layers[name]
                    non_seqs.append(inputs[input_layer])
                    non_seqs_index[input_layer] = len(non_seqs) - 1
```

stevenxxiu:
> For the sequential and parallel RecurrentContainer, this is the Design

Ah thanks then I think both are covered.

> Cell can have memory or just be memoryless (in your implementation it just means that get_out_for return only one {output:..} or also include other states {cell:...,output:..}. In general, it just a CellLayer with initial state or no initial state.

Yes you got it! But if a cell is really memoryless (it doesn't even read the previous output) then you can just use a normal layer. A memoryless cell with an initial state just allows a sequence of length 0 to have a custom output otherwise it defaults to 0.

> A mapping between InputLayer-InputLayer with different shape make me so confusing, but your explanation above is clear, I am not sure how new users will catch up with this changes.

Thanks. I guess I should add some docs now since everybody seems to be generally happy with the api.

> We have to re-specifed many duplicated information, finding cleaner way might constraint the API.

I couldn't think of a way to simplify or remove the input mapping as you need to know where the for loop begins. Anyway if there's some simplification you can think of which works you can comment here.

> I am confusing about your _ lstm_dropout_gal_ example, how do you keep the dropout_mask constant for all time_step:

Easy. Putting it in as the third argument as `step_incoming` makes it constant (it goes to output_info in scan), there's no partitioning or anything else going on. Putting it in as the first argument as `incoming` makes it different each time (since it becomes part of the cell graph).

> And, what else will be added to non_sequences in this part ? All the inits that not the outputs will be constants ?

Yes precisely.

trungnt13:
> I couldn't think of a way to simplify or remove the input mapping as you need to know where the for loop begins. Anyway if there's some simplification you can think of which works you can comment here.

For this example:
```
    l_inp_1 = InputLayer((n_batch, seq_len, n_features))
    l_inp_2 = InputLayer((n_batch, seq_len, n_features))
    cell_inp_1 = InputLayer((n_batch, n_features))
    cell_inp_2 = InputLayer((n_batch, n_features))
    cell = ConcatLayer([cell_inp_1, cell_inp_2])
    cell = DenseRecurrentCell(cell, n_units)['output']
    l_rec = RecurrentContainerLayer({cell_inp_1: l_inp_1, cell_inp_2: l_inp_2}, cell)
```
4 InputLayer involve in this example, the fact we only need to specified *l_inp_1* and *l_inp_2*, it can be simplified as:
```
    l_inp_1 = InputLayer((n_batch, seq_len, n_features))
    l_inp_2 = InputLayer((n_batch, seq_len, n_features))

    cell = ConcatLayer([l_inp_1, l_inp_2])
    cell = DenseRecurrentCell(cell, n_units)['output']
    l_rec = RecurrentContainerLayer(cell)
```
When adding the Cell, you travel back the tree of DenseRecurrentCell and find all its InputLayer, then you infer the shape of its shape of l_inp_1 and l_inp_2.

btw, as you stated: ['output'] return other layer, but I cannot find place where you define __getitem__ for CellLayer, what is the purpose of calling it before you add cell to RecurrentContainer?

stevenxxiu:
> 4 InputLayer involve in this example, the fact we only need to specified l_inp_1 and l_inp_2, it can be simplified as:

That is the most direct way but it runs into the problem mentioned above, because `ConcatLayer` can accept 3-d tensors too, it's hard to tell when the recurrence starts. Also the input shapes are wrong to `ConcatLayer`.

> btw, as you stated: ['output'] return other layer, but I cannot find place where you define getitem for CellLayer, what is the purpose of calling it before you add cell to RecurrentContainer?

You don't have to, but then you get both the cell and output states. The `__getitem__` makes sure you only get the output state. It is not specific to cells, and is [implemented here](https://github.com/Lasagne/Lasagne/pull/629/files#diff-1aca1664756f89ab20b6716803219df1R260).

trungnt13:
```
That is the most direct way but it runs into the problem mentioned above, because ConcatLayer can accept 3-d tensors too, it's hard to tell when the recurrence starts. Also the input shapes are wrong to ConcatLayer.
```
Yes, but when I am using InputLayer I expected Input, the meaning of InputLayer in this case is a PlaceHolder for the Cell. This is a little re-design, idea is anything does recurrent it is Cell:
```
    l_inp_1 = InputLayer((n_batch, seq_len, n_features))
    cell1 = DenseRecurrentCell(l_inp_1, n_units)

    l_inp_2 = InputLayer((n_batch, seq_len, n_features))
    cell2 = DenseRecurrentCell(l_inp_2, n_units)

    cell3 = MergeCell((cell1, cell2), mode='sum')
    cell3 = SplitLayer(cell3, key='output')
    
    l_rec = RecurrentContainerLayer(cell)
```

Another Example for seq_1_to_2:
```
    l_inp = InputLayer((n_batch, seq_len, n_features))

    cell_1 = DenseRecurrentCell(l_inp, n_units)
    cell_2 = DenseRecurrentCell(l_inp, n_units)

    cell = MergeCell((cell_1, cell_2), mode='identity')
    cell = SplitLayer(cell, key='output')
    
    l_rec = RecurrentContainerLayer(cell)
```

I don't agree with the expression **['output']**. I know it is shortcut for convenient, I used to love shortcut, but it is very difficult for new users to interpret the code.

stevenxxiu:
> I don't agree with the expression ['output']. I know it is shortcut for convenient, I used to love shortcut, but it is very difficult for new users to interpret the code. 

You can still do it the long way by `SplitLayer(l, 'output')`. Btw removing it and having predefined outputs isn't good, as it would unnecessarily restrict the interface to the outputs the cells expose, and this creates problems say when you want to share the cell for whatever reason.

But note that the syntax is not specific to cells and has other uses too. I took it from @fok's suggestion [here](https://github.com/Lasagne/Lasagne/issues/337#issuecomment-165122642).

MInner:
Hi @stevenxxiu ! :) I wounder if I can already clone your `recurrent` branch and try to implement a custom `CellLayer`, or you would rather say that it is still not ready real-world usage? If I try to use it, I only need to override `get_output_shape_for` to return a `{'cell':.., 'output':..}` and `get_output_for` to return `{'cell': cell, 'output': hid}` - correct? Or have I missed something? :)

I actually need an RNN with a custom step function now, and it seems like only two options available now are: 
* tensorflow rnn module (with imho wiered matrix api) and
*  @stevenxxiu 's Lasagne fork - and I like waaay more :)

Thank you.

stevenxxiu:
@MInner Hi I'm really glad someone is interested in this :) I am using it myself for some projects of mine and there haven't been any bugs, so I think it's more or less ready for use. I don't think the final interface is going to change a lot judging by discussions above since people seem to be mostly happy with it.

And yes, you can override those 2 methods, `get_output_shape_for` and `get_output_for`, but also `__init__` so scan knows what sequences, parameters and initial states you're using. I think this example is helpful: https://github.com/stevenxxiu/lasagne_recurrence_examples/blob/master/primitives.py#L339.

MInner:
@stevenxxiu I guess there are some pieces of code where short notes\docstrings are somewhat vital to make it usable by other people :)

Could you please write a short speudocode explaining how does `RecurrentContainerLayer` deals with cells passed to it and what happens to `incomings`, `cell` and `step_incomings` passed to `CellLayer` and what is the life-cycle of `incomings`, `inits`, `inits_fixed` and when are they available\updatable?

From what I could understand by now, the pipleline looks somewhat like this:

When we pass a dict to `RecurrentContainerLayer`, it passes `incomings` (dict value) sliced among time-direction (axis=1)  to (dict key)-cell (usually one dimension less then value) one by one to finally compute `cell` argument. `cell_input_layer` (the dict-key) is usually just `InputLayer` that passes this "time-sliced X" further to `CellLayer` instance and passes it as an `incommings` argument that is used to initialize internal state of cell. Cell input and output on each timestep (call of `get_output_for`) are both dicts that have same keys - those which are passed to the `CellLayer.__init__` on cell creation:
* `incomings` dict represents which inputs are coming from lower layers (just passed further from them); each key from this dict will be available inside `inputs` dict during each step computation and represents a single time slice; (or it shows the dict we should initialize `key`s from?)
* `inits` - initial values additionally (to the incoming input) passes inside `inputs` dict; when we return something in `get_output_for` we can override\update them this way (or we can override only 'output'? why some values have output in `get_output_shape_for` and some are not - e.g. `dropout` is in `inits`, but we don't update it and do not specify it's shape?) 

Why didn't we pass anything here? I (hopefully) understand that `step_incomings` kind of 'redispatches' `output_t` to `input_{t+1}`, but where does the first input comes from if `incomings` is empty?

    cell_input = InputLayer((n_batch, n_features))
    cell_zero_state_input = InputLayer((n_batch, n_units_1))
    cell_h1 = LSTMCell(cell_input, n_units_1, hid_init=cell_zero_state_input)['output']
    cell_output = GRUCell(cell_h1, n_units_2)['output']
    l_rec = RecurrentContainerLayer(incomings={}, cell=cell_output, 
                                 step_incomings={cell_input: cell_output}, n_steps=seq_len) ##??

But `LSTMDropoutMCLSCell` example totally blows up my mind in terms of what happens on initialization :)) 

`incomings={'x': incoming, 'xi': incoming, 'xf': incoming, 'xc': incoming, 'xo': incoming,}`

- where does those `x, xi, xf, ..` come from? are they all equal each other in the beginning = `incoming`? or they are initialized like `x = incoming['x']` but how does those fields appear in incoming? but then we redefine them in `precompute_for(..)` - now there're all now equal anymore? why did we make them equal in the first place?

Sorry for such a mess and a huge "wtf I just read?"-message :), I might really have too little experience reading scientific code, yet :) Thank you!

lasagne/utils.py
```diff
@@ -136,7 +136,7 @@ def one_hot(x, m=None):
     return T.eye(m)[T.cast(x, 'int32')]
 
 
-def unique(l):
+def unique(l, return_index=False):
```
kmike: Is this change used?

fdabek1:
I have not gotten a chance to read through the entirety of the code and the discussion here, so I'm not sure of the exact implementation, but as a new user of Lasagne and a machine learning researcher I think it would be very useful to make sure that the recurrent implementation supports modifying the step function separately from the rest of the class.

For example, to implement a Clockwork RNN (http://arxiv.org/abs/1402.3511) with Lasagne I have had to take the CustomRecurrentLayer class and copy the entirety of it and then modify it to have "modules".  However, if it would be possible to subclass the CustomRecurrentLayer class and only only modify the step function then this would allow for easier implementations and much less code reuse.

This is just my two cents.

mjwillson:
Seems like something along these lines could be very useful to help implement an efficient beam search for sequence prediction models, where you need to compile a function that applies a single step of the RNN at a time.

Perhaps implementing a beam search on top of it (or a sampling procedure) could be a useful test of the new functionality?

justheuristic:
It might be a bit late, but we just happened to implement this "Custom recurrent graph" library for out tinkering in deep reinforcement learning, attention and long term memory architectures - but we only managed to make it public 4 days ago.

It might be not as polished as core Lasagne, but i still hope it will be of use someone since it was focused on ease of prototyping.

Take a look:
* Repo: https://github.com/yandexdataschool/AgentNet
* [Simplistic Deep Reinforcement Learning example](https://github.com/yandexdataschool/AgentNet/blob/master/examples/Basic%20tutorial%20on%20Boolearn%20Reasoning%20problem.ipynb)
* [Stack-augmented GRU language model](https://github.com/yandexdataschool/AgentNet/blob/master/examples/Stack%20RNN%20for%20formal%20sequence%20modelling.ipynb)
* [A bit more serious about reinforcement learning](https://github.com/yandexdataschool/AgentNet/blob/master/examples/Advanced%20MDP%20tools%20and%20wikicat.ipynb)

f0k:
> Repo: https://github.com/yandexdataschool/AgentNet

Cool, the Readme looks promising!

---

Sorry for being so unresponsive on this thread the last weeks, it's still a burning issue but I was busy with other tasks. I'll try to go through the discussion soon. I guess the most important decisions to be made are in #337 and #537, this will probably also tell us if we need any (minor) changes of the API for the recurrent containers. Again I'm sorry for holding this up so long, your work is certainly appreciated, @stevenxxiu!

stevenxxiu:
@MInner  sorry for the very late reply

> where does those x, xi, xf, .. come from? are they all equal each other in the beginning = incoming? or they are initialized like x = incoming['x'] 

They are just precompute dictionaries. I've changed the api very slightly (see new recurrent examples) so precompute only adds new sequences to the input dictionary instead of modifying existing sequences. So no need for the initializations from before which were actually useless.

I'll try add some docs.

@f0k Sure I'll try and get those two resolved in my code soon. Have you looked at it in the meantime?

f0k:
> Sure I'll try and get those two resolved in my code soon. Have you looked at it in the meantime?

Yes, I figured #678 should be the solution, and be implemented before your PR. I'm just not sure what'd be the best choice with respect to backwards compatibility vs. cleanness (and didn't have much time to think about it). In any case, there's no need to change your code in this respect for now. Documentation is always good, though :)

stevenxxiu:
@f0k Any updates on this? I've just rebased the PR so it's up to date with master.

f0k:
> Any updates on this? I've just rebased the PR so it's up to date with master.

Hey, great! I'm sorry for letting this sit for so long, and relieved to hear you're still on board :) We should release 0.2 before tackling #678, which in turn is required for your PR. However, I'm working on my thesis and can't (or shouldn't...) invest too much time into Lasagne right now. There are a few easy and a few slightly more complex issues in https://github.com/Lasagne/Lasagne/milestone/2, any help is appreciated!

stevenxxiu:
Cool, no worries and good luck on your thesis, Ill take a look when I have time.

PR #633, opened 2016-03-06
add broadcasting to elementwise merge layers

### LINKED ISSUE #584: [ElementwiseMergeLayer] "Mismatch: not all input shapes are the same"
Is it necessary to force all Input shapes must be the same in ElementwiseMergeLayer ?
This code can be run and provide output normally, however, as soon as I add one more layer over it and the function _get_output_shape_for_ is called, it crashed:
```
    l_in = lasagne.layers.InputLayer(name='input', shape=(None, maxlen, dim))
    l_mask = lasagne.layers.InputLayer(name='mask', shape=(None, maxlen))
    l_mask = lasagne.layers.DimshuffleLayer(l_mask, pattern=(0, 1, 'x'))
    l_in = lasagne.layers.ElemwiseMergeLayer((l_in, l_mask), tensor.mul)
```
The naive approach can be choose the shape of first input as default output_shape with a note in documentation ?

In short, is there any better way to multiply mask with input in Lasagne?

skaae:
Fix https://github.com/Lasagne/Lasagne/issues/584 
Adds a check to ElemwiseMergeLayer that checks if a dimension is broadcastable. 

I think the layer will fail if the tensor is not broadcastable but the shapes are. Should we take care of that or is it the users reponsiblity to not use non-broadcastable tensors?

### DIFF
--- a/lasagne/layers/merge.py
+++ b/lasagne/layers/merge.py
@@ -302,9 +302,8 @@ class ElemwiseMergeLayer(MergeLayer):
 
     Parameters
     ----------
-    incomings : a list of :class:`Layer` instances or tuples
-        the layers feeding into this layer, or expected input shapes,
-        with all incoming shapes being equal
+    incomings : Unless `cropping` is given, all shapes must be equal, except
+        for dimensions that are undefined (``None``) or broadcastable (``1``).
 
     merge_function : callable
         the merge function to use. Should take two arguments and return the
@@ -324,9 +323,36 @@ class ElemwiseMergeLayer(MergeLayer):
         super(ElemwiseMergeLayer, self).__init__(incomings, **kwargs)
         self.merge_function = merge_function
         self.cropping = cropping
+        self.broadcastable = None
 
     def get_output_shape_for(self, input_shapes):
         input_shapes = autocrop_array_shapes(input_shapes, self.cropping)
+
+        input_dims = [len(shp) for shp in input_shapes]
+        if not all(input_dim == input_dims[0] for input_dim in input_dims):
+            raise ValueError('Input dimensions must be the same but were %s' %
+                             ", ".join(map(str, input_shapes)))
+
+        def broadcasting(input_dim):
+            # Identify dimensions that will be broadcasted.
+            sorted_dim = sorted(input_dim,
+                                key=lambda x: x if x is not None else -1)
+            if isinstance(sorted_dim[-1], int) and sorted_dim[-1] != 1 \
+                    and all([d == 1 for d in sorted_dim[:-1]]):
+                size_after_broadcast = sorted_dim[-1]
+                broadcast = [True if d == 1 else None for d in input_dim]
+                return ((size_after_broadcast,)*len(input_dim), broadcast)
+            else:
+                return (input_dim, [None]*len(input_dim))
+
+        # if the dimension is broadcastable we replace 1's with the size
+        # after broadcasting.
+        input_dims, broadcastable = list(zip(
+            *[broadcasting(input_dim)for input_dim in zip(*input_shapes)]))
+
+        self.broadcastable = list(zip(*broadcastable))
+        input_shapes = list(zip(*input_dims))
+
         # Infer the output shape by grabbing, for each axis, the first
         # input size that is not `None` (if there is any)
         output_shape = tuple(next((s for s in sizes if s is not None), None)
@@ -344,6 +370,13 @@ class ElemwiseMergeLayer(MergeLayer):
 
     def get_output_for(self, inputs, **kwargs):
         inputs = autocrop(inputs, self.cropping)
+        # modify broadcasting pattern.
+        if self.broadcastable is not None:
+            for n, broadcasting_dim in enumerate(self.broadcastable):
+                for dim, broadcasting in enumerate(broadcasting_dim):
+                    if broadcasting:
+                        inputs[n] = T.addbroadcast(inputs[n], dim)
+
         output = None
         for input in inputs:
             if output is not None:
@@ -360,9 +393,8 @@ class ElemwiseSumLayer(ElemwiseMergeLayer):
 
     Parameters
     ----------
-    incomings : a list of :class:`Layer` instances or tuples
-        the layers feeding into this layer, or expected input shapes,
-        with all incoming shapes being equal
+    incomings : Unless `cropping` is given, all shapes must be equal, except
+        for dimensions that are undefined (``None``) or broadcastable (``1``).
 
     coeffs: list or scalar
         A same-sized list of coefficients, or a single coefficient that
--- a/lasagne/tests/layers/test_merge.py
+++ b/lasagne/tests/layers/test_merge.py
@@ -184,12 +184,18 @@ class TestElemwiseSumLayer:
         assert layer.get_output_shape_for([(3, 2), (3, None)]) == (3, 2)
         assert layer.get_output_shape_for([(None, 2), (3, 2)]) == (3, 2)
         assert layer.get_output_shape_for([(None, 2), (None, 2)]) == (None, 2)
+        assert layer.get_output_shape_for([(None, 5), (None, 1)]) == (None, 5)
+        assert layer.get_output_shape_for([(None, 1), (None, 5)]) == (None, 5)
+        assert layer.get_output_shape_for(
+            [(None, None), (None, 5)]) == (None, 5)
         with pytest.raises(ValueError):
             layer.get_output_shape_for([(3, None), (4, 2)])
         with pytest.raises(ValueError):
             layer.get_output_shape_for([(3, 2), (4, None)])
         with pytest.raises(ValueError):
             layer.get_output_shape_for([(None, 2), (3, 2), (4, 2)])
+        with pytest.raises(ValueError):
+            layer.get_output_shape_for([(None, 2), (3, 2, 10), (4, 2)])
 
     def test_get_output_for(self, layer):
         a = numpy.array([[0, 1], [2, 3]])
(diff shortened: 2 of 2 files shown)


lasagne/layers/merge.py
```diff
@@ -303,8 +303,9 @@ class ElemwiseMergeLayer(MergeLayer):
     Parameters
     ----------
     incomings : a list of :class:`Layer` instances or tuples
-        the layers feeding into this layer, or expected input shapes,
-        with all incoming shapes being equal
+        the layers feeding into this layer, or expected input shapes.
+        All input shapes must be equal or broadcastable. A dimension is
+        broadcastable if all non None dimensions are the same size or 1.
```
f0k: Hmm, I don't think this makes it really clear, and it's also wrong since we've got auto-cropping. Maybe write:
```
Unless `cropping` is given, all shapes must be equal,
except for dimensions that are undefined (``None``)
or broadcastable (``1``).
```

 "with all incoming shapes being equal or broadcastable to the same size".

lasagne/layers/merge.py
```diff
@@ -327,14 +328,35 @@ def __init__(self, incomings, merge_function, cropping=None, **kwargs):
...
+                return [broadcast_size]*len(input_dim)
+            else:
+                return input_dim
+
+        # if dimension is broadcastable and we can infer the size then then
+        # replace all sizes with the broadcasting dimension.
+        input_dims = [broadcasting(input_dim)
+                      for input_dim in zip(*input_shapes)]
+        input_shapes = zip(*input_dims)
+
```
f0k: This is a lot of code. Can't you just change the `output_shape` computation to grab the largest input size per axis instead of the first, and then change the `match` code to treat `1` like `None`? As far as I see, you won't need to add any more code then.

f0k:
> I think the layer will fail if the tensor is not broadcastable but the shapes are. Should we take care of that or is it the users reponsiblity to not use non-broadcastable tensors?

I think this is not always in the user's hands. In `get_output_for()`, we should make dimensions broadcastable that have a size of `1` and need broadcasting.

lasagne/layers/merge.py
skaae: I like the top one best. 

I'm not sure how we shoud handle 
input1 [32, 1]
input2 [32, 5]
input3 [32, None]

Should that give an error or just return None in the last dimensions? I think it's easier to debug if it raises an error?

skaae:
>This is a lot of code. Can't you just change the output_shape computation to grab the largest input size per axis instead of the first, and then change the match code to treat 1 like None? As far as I see, you won't need to add any more code then.

Yes. That is much cleaner. 

I modified the code to:

```Python
        # Infer the output shape by grabbing, for each axis, the maximum
        # input size that is not `None` (if there is any)
        output_shape = tuple(max(sizes,
                                 key=lambda x: x if x is not None else -1)
                             for sizes in zip(*input_shapes))

        def match(shape1, shape2):
            return (len(shape1) == len(shape2) and
                    all(s1 in [None, 1] or s2 in [None, 1] or s1 == s2
                        for s1, s2 in zip(shape1, shape2)))

        # Check for compatibility with inferred output shape
        if not all(match(shape, output_shape) for shape in input_shapes):
            raise ValueError("Mismatch: not all input shapes are the same")
        return output_shape

        inputs = autocrop(inputs, self.cropping)
        for axis, (out_shp, in_shps) in enumerate(zip(
                self.output_shape, zip(*self.input_shapes))):
            for input_num, in_shp in enumerate(in_shps):
                if out_shp != in_shp and in_shp == 1:
                    inputs[input_num] = T.addbroadcast(inputs[input_num], axis)
        output = None
        for input in inputs:
            if output is not None:
                output = self.merge_function(output, input)
            else:
                output = input
        return output

``` 

This works fine when i test it, but when i run the test suite I get errors with Mock() having no len? 
The error is related to the broadcasting code in `get_output_for` ?

lasagne/layers/merge.py
f0k: > Should that give an error or just return None in the last dimensions?

No, it should return (32, 5) for the output shape. That's the only possible outcome. We can't raise an error because the run-time shape of input3 could be correct.

f0k:
> `key=lambda x: x if x is not None else -1`

This won't work, an axis of all-None is replaced by -1 now and returned as such in the end. This should fail some of the tests. You will need to replace -1 by None afterwards. (If it wasn't for Python 3, you could omit the `key=` -- in Python 2, `None` is smaller than all ints anyway. Unfortunately, in Python 3, `None` cannot be compared to an int.)

> `s1 in [None, 1]`

Considered bad practice. Comparison against `None` should always use `is`, even if it makes the code a bit longer.

> This works fine when i test it, but when i run the test suite I get errors with Mock() having no len?

You will have to investigate closer. Where exactly does it try to access the length of a Mock? Sometimes the Mocks used for the tests require some extra attributes (like a valid output shape) to be added when a new feature is added to the code.

Sentient07:
Is someone working on this PR? Else shall I complete it? I had a situation to merge two 4D tensors, of shape `(N, 1, 32, 32) and (N, 3, 32, 32)` for which I tried this diff and it did not work for me. I did a rapid fix for this in my project where I had the user specify which dimension has to be broadcasted. (which I think is the subset of what this PR intends to do, that or automatically apply broadcasting like numpy).

f0k:
Nobody is working on this, you're welcome to have a go!

Lasagne has changed a bit in the meantime such that the tensors should already be broadcastable if their size is known to Lasagne (i.e., if you have a 4D tensor with a single channel, the channel dimension should be broadcastable). The MergeLayer will currently still complain when checking for matching shapes, this needs to be changed.

Sentient07:
> Lasagne has changed a bit in the meantime such that the tensors should already be broadcastable

Could you please point to the code that you say has changed ? I am bit confused about this..

Sentient07:
I tested with the following [code ](http://dpaste.com/1VK4K8B) after rebasing branch with the current master to mimic my scenario. It looks like I misunderstood your comment. Here, both the tensors have all the dimensions non broadcastable and Theano complains. I will look more deeper, what is happening

f0k:
> Could you please point to the code that you say has changed ?

#715, #794, and #792.

> I tested with the following code after rebasing branch with the current master to mimic my scenario.

Ah, yes, if you create the input variable yourself instead of letting `InputLayer` create it, then it will not be made broadcastable. The best solution might be to change `InputLayer.get_output()` to add broadcasts if necessary. Feel free to take this!

Sentient07:
> Ah, yes, if you create the input variable yourself instead of letting InputLayer create it, then it will not be made broadcastable.

So, I should be getting the output of the `InputLayer` to be the input for theano graph? 

> Sometimes you may not even want a variable to become broadcastable just because you change the number of input channels from 3 to 1, for example (this will change the variable type and possibly trigger recompilations). 

I'm not sure I understand this. What is the recompilation? 

> Adding it to the MergeLayer is ugly as well, because it shouldn't be the MergeLayer's responsibility to fix other layers' mistakes.

I think it should be a functionality of the MergeLayer because there is actually no need for other layers to keep a dimension broadcastable, whether it is desired or not. I am not sure if there are other Layers that require broadcasting of dimensions in Lasagne. Are there any ? If so what you said perfectly makes sense to me. Else,  I think the broadcasting should be done from the MergeLayer itself as the MergeLayer would do arithmetic over a particular dimension (of adding/multiplying etc..)

f0k:
> So, I should be getting the output of the `InputLayer` to be the input for theano graph?

If you do:
```python
network = InputLayer(...)
input_var = network.input_var
```
then you'll have a correctly broadcastable input variable. But we don't want to enforce this usage pattern on the user. Just using a `T.tensor4()` should work as well.

> I'm not sure I understand this. What is the recompilation?

If you create a network:
```python
net = Conv2DLayer(InputLayer((None, 1, 32, 32)), 16, 3)
```
and compile a function using it, and then change the network to
```python
net = Conv2DLayer(InputLayer((None, 3, 32, 32)), 16, 3)
```
the input variable type for the convolution will change from a tensor4 with pattern (False, True, False, False) to a tensor4 with pattern (False, False, False, False), so Theano will have to compile a new convolution Op when you compile a function for the modified network. If you change the batchsize from `None` to `1` it will even have to recompile all Ops of the network. If you don't actually need anything to be broadcastable, that's a nuisance. So that'd be an argument for leaving it up to the layers that potentially require broadcasts at all.

> I am not sure if there are other Layers that require broadcasting of dimensions in Lasagne. Are there any ?

Some custom layer, or a `ExpressionLayer`, or maybe a custom loss function or whatever users may do with the Theano expression returned by `get_output()`. For this purpose it'd be nice if broadcasts were always inferred correctly. But I've also seen that it leads to a lot of problems if broadcasts are only inferred 90% of the time, because then you end up with mismatching variable types here and there.

So let's leave this up to the merge layer for now. If Theano gets proper shape inference, we'll have to add this to the InputLayer or `get_output()`, though.

f0k:
Another use case where it would be helpful if all layers correctly maintained broadcastable patterns: https://groups.google.com/forum/#!topic/lasagne-users/P0VovqFyPBA
In that case, the resulting expression of a SliceLayer was not made broadcastable, but a following BatchNormLayer inferred from the shape that it should be broadcastable and prepared the `mean` and `inv_std` parameters accordingly. This lead to a mismatch for the default updates.

We could have a `lasagne.utils.ensure_broadcasts(var, shape)` helper function that is called for all expressions in `lasagne.layers.get_output()` (that'd be a brute-force solution), or called in `get_output_for()` for all layers that need it.

Sentient07:
Hi, 

> Just using a T.tensor4() should work as well

This is what I have in my code and this doesn't seem to work. I'm not sure if it is desirable to get the input variable to the `InputLayer` using `input_var = network.input_var`. The latter seems like a little hack to me..

> and compile a function using it, and then change the network to 

I'm not sure I understand well, but can't we rather use a different variable ? 
We can actually add a check [here](https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/input.py#L66-L69), depending on the shape that Lasagne is aware of ? 

> or maybe a custom loss function or whatever users may do

But isn't it advisable to use(inherit) `MergeLayer` when someone is dealing with any Elemwise operation over two `Layers` ?

PR #634, opened 2016-03-08
Fix comments (input shape description) in LSTM and GRU classes

avostryakov:


skaae:
I think the original comment is correct?

avostryakov:
A shape of input variable should be (n_time_steps, n_batch, n_features). It is input to LSTM layer. And it is multiplied on W_in_stacked which has a shape (n_features, 4*num_units):

(n_time_steps, n_batch, n_features) x (n_features, 4*num_units) = (n_time_steps, n_batch, 4*num_units)

I can't multiply (n_time_steps, n_batch, 4*num_units) x (n_features, 4*num_units)

skaae:
True. I missed the input above. I'm fine with merging.

f0k:
> I can't multiply (n_time_steps, n_batch, 4num_units) x (n_features, 4num_units)

That's not what the comment wants to say. It's meant to say what `input` looks like *after* that multiplication. Granted, it's not formulated very clearly!

Closing a) because it's not the correct fix, and b) because it conflicts with #629. Comment fixes should better be made there to avoid rebasing, please feel free to comment on the diff of that PR!

PR #643, opened 2016-03-19
Extending get_params to optionally return Theano expressions

### LINKED ISSUE #641: Make get_params optionally return expressions?
While implementing a form of [BinaryConnect](https://github.com/MatthieuCourbariaux/BinaryConnect) I noticed that get_params returns only the shared variables, but not expressions. That makes perfect sense e.g. for low-rank dense layers (where W = UV is a Theano expression with U and V as shared variables). I now have the opposite use case: I'd like to register an expression as a parameter of which I later want to compute the gradient, so I want get_param to return that expression instead of the shared variables contained. In principle that is easy to implement: one could add a keyword "only_shared=True" in get_params and modify the return statement. 

If you like the idea I can open a pull request.

[Further info on use case]: My use case is as follows: I have a parameter that is binary (e.g. 0 or 1), but I'd still like to use SGD to optimize that parameter. To this end I follow BinaryConnect and define a new real-valued hidden variable *p* with support *[0,1]* (through clipping). The binary parameter is then implemented as a simple Bernoulli process with probability *p*. The parameter *p* is then updated in the direction of the gradient with respect to the binary parameter (so e.g. if the binary parameter b was 1, and the is positive with respect to this parameter, then p is increased).

wielandbrendel:
This PR fixes #641. It adds the ability to make get_params and get_all_params return Theano expressions if any of the parameters was registered as a Theano expression. To this end this PR adds the keyword *unwrap_shared* (default: True) to both methods. The default behaviour of both methods is unchanged, yet setting unwrap_shared=False will yield the Theano expressions instead of the contained shared variables.

### DIFF
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -49,9 +49,10 @@ class Layer(object):
     def output_shape(self):
         return self.get_output_shape_for(self.input_shape)
 
-    def get_params(self, **tags):
+    def get_params(self, unwrap_shared=True, **tags):
         """
-        Returns a list of Theano shared variables that parameterize the layer.
+        Returns a list of Theano shared variables or expressions that
+        parameterize the layer.
 
         By default, all shared variables that participate in the forward pass
         will be returned (in the order they were registered in the Layer's
@@ -62,12 +63,19 @@ class Layer(object):
         regularized (e.g., by L2 decay).
 
         If any of the layer's parameters was set to a Theano expression instead
-        of a shared variable, the shared variables involved in that expression
-        will be returned rather than the expression itself. Tag filtering
-        considers all variables within an expression to be tagged the same.
+        of a shared variable, `unwrap_shared` controls whether to return the
+        shared variables involved in that expression (``unwrap_shared=True``,
+        the default), or the expression itself (``unwrap_shared=False``). In
+        either case, tag filtering applies to the expressions, considering all
+        variables within an expression to be tagged the same.
 
         Parameters
         ----------
+        unwrap_shared : bool (default: True)
+            Affects only parameters that were set to a Theano expression. If
+            ``True`` the function returns the shared variables contained in
+            the expression, otherwise the Theano expression itself.
+
         **tags (optional)
             tags can be specified to filter the list. Specifying ``tag1=True``
             will limit the list to parameters that are tagged with ``tag1``.
@@ -77,7 +85,7 @@ class Layer(object):
 
         Returns
         -------
-        list of Theano shared variables
+        list of Theano shared variables or expressions
             A list of variables that parameterize the layer
 
         Notes
@@ -98,7 +106,10 @@ class Layer(object):
             result = [param for param in result
                       if not (self.params[param] & exclude)]
 
-        return utils.collect_shared_vars(result)
+        if unwrap_shared:
+            return utils.collect_shared_vars(result)
+        else:
+            return result
 
     def get_output_shape_for(self, input_shape):
         """
--- a/lasagne/layers/helper.py
+++ b/lasagne/layers/helper.py
@@ -290,9 +290,10 @@ def get_output_shape(layer_or_layers, input_shapes=None):
         return all_shapes[layer_or_layers]
 
 
-def get_all_params(layer, **tags):
+def get_all_params(layer, unwrap_shared=True, **tags):
     """
-    Returns a list of Theano shared variables that parameterize the layer.
+    Returns a list of Theano shared variables or expressions that
+    parameterize the layer.
 
     This function gathers all parameter variables of all layers below one or
     more given :class:`Layer` instances, including the layer(s) itself. Its
@@ -311,6 +312,11 @@ def get_all_params(layer, **tags):
         The :class:`Layer` instance for which to gather all parameters, or a
         list of :class:`Layer` instances.
 
+    unwrap_shared : bool (default: True)
+        Affects only parameters that were set to a Theano expression. If
+        ``True`` the function returns the shared variables contained in
+        the expression, otherwise the Theano expression itself.
+
     **tags (optional)
         tags can be specified to filter the list. Specifying ``tag1=True``
         will limit the list to parameters that are tagged with ``tag1``.
@@ -321,7 +327,8 @@ def get_all_params(layer, **tags):
     Returns
     -------
     params : list
-        A list of Theano shared variables representing the parameters.
+        A list of Theano shared variables or expressions representing
+        the parameters.
 
     Examples
     --------
@@ -334,10 +341,12 @@ def get_all_params(layer, **tags):
 
     Notes
     -----
-    If any layer's parameter was set to a Theano expression instead of a shared
-    variable, the shared variables involved in that expression will be returned
-    rather than the expression itself. Tag filtering considers all variables
-    within an expression to be tagged the same.
+    If any of the layers' parameters was set to a Theano expression instead
+    of a shared variable, `unwrap_shared` controls whether to return the
+    shared variables involved in that expression (``unwrap_shared=True``,
+    the default), or the expression itself (``unwrap_shared=False``). In
+    either case, tag filtering applies to the expressions, considering all
+    variables within an expression to be tagged the same.
     >>> import theano
     >>> import numpy as np
     >>> from lasagne.utils import floatX
@@ -349,7 +358,8 @@ def get_all_params(layer, **tags):
     True
     """
     layers = get_all_layers(layer)
-    params = chain.from_iterable(l.get_params(**tags) for l in layers)
+    params = chain.from_iterable(l.get_params(
+            unwrap_shared=unwrap_shared, **tags) for l in layers)
     return utils.unique(params)
 
 
--- a/lasagne/tests/layers/test_helper.py
+++ b/lasagne/tests/layers/test_helper.py
@@ -718,6 +718,23 @@ class TestGetAllParams:
                 (l2.get_params(regularizable=True) +
                  l3.get_params(regularizable=True)))
 
+    def test_get_all_params_with_unwrap_shared(self):
+        from lasagne.layers import (InputLayer, DenseLayer, get_all_params)
+        import theano.tensor as T
+        from lasagne.utils import floatX
+
+        l1 = InputLayer((10, 20))
+        l2 = DenseLayer(l1, 30)
+
+        W1 = theano.shared(floatX(numpy.zeros((30, 2))))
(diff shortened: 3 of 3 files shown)


f0k:
Thanks! There's a merge conflict because you branched off an old `master` (from January 13th). Can you please do (assuming you added the original Lasagne/Lasagne repository as `upstream`):
```bash
git checkout get_params_extension
git fetch upstream
git rebase upstream/master
# edit files to fix conflicts
git add lasagne/layers/base.py lasagne/layers/helper.py
git rebase --continue
git push --force
```
There are a few things I'd document differently, but let's rebase first and then I'll comment on the diff. Also you'll need to add a test case to `lasagne/tests/layers/test_helper.py` that assures the new keyword works correctly!

wielandbrendel:
Thanks for detailed instruction! I was a bit puzzled about the fact that my Lasagne branch was so old, but it turns out that I was forking Lasagne a second time and didn't notice that this won't update the fork I created the first time in January.

Please let me know what changes in the documentation you'd like to see. I also added a test to test_helper.py.

lasagne/layers/base.py
```diff
@@ -63,11 +63,18 @@ def get_params(self, **tags):
 
         If any of the layer's parameters was set to a Theano expression instead
         of a shared variable, the shared variables involved in that expression
-        will be returned rather than the expression itself. Tag filtering
-        considers all variables within an expression to be tagged the same.
+        will be returned by default rather than the expression itself. Tag
+        filtering considers all variables within an expression to be tagged
+        the same. To retrieve the original Theano expression set
+        ``unwrap_shared=False``.
```
f0k: I'd prefer:
```
If any of the layer's parameters was set to a Theano expression instead
of a shared variable, `unwrap_shared` controls whether to return the
shared variables involved in that expression (``unwrap_shared=True``,
the default), or the expression itself (``unwrap_shared=False``). In
either case, tag filtering applies to the expressions, considering all
variables within an expression to be tagged the same.
```

lasagne/layers/base.py
```diff
@@ -77,7 +84,7 @@ def get_params(self, **tags):
 
         Returns
         -------
-        list of Theano shared variables
+        list of Theano shared variables (default) or expressions
```
f0k: I'd omit `(default)`, it's not really clear what default it refers to.

lasagne/layers/helper.py
```diff
@@ -290,9 +290,10 @@ def get_output_shape(layer_or_layers, input_shapes=None):
         return all_shapes[layer_or_layers]
 
 
-def get_all_params(layer, **tags):
+def get_all_params(layer, unwrap_shared=True, **tags):
     """
-    Returns a list of Theano shared variables that parameterize the layer.
+    Returns a list of Theano shared variables (default) or expressions that
+    parameterize the layer.
```
f0k: Again, I'd omit `(default)`. Also check how this renders in the overview table at http://lasagne.readthedocs.org/en/latest/modules/layers.html. Maybe it has to be kept on one line. Note that the initial sentence of `Layer.get_params()` has to be adapted as well.

lasagne/layers/helper.py
```diff
@@ -311,6 +312,11 @@ def get_all_params(layer, **tags):
         The :class:`Layer` instance for which to gather all parameters, or a
         list of :class:`Layer` instances.
 
+    unwrap_shared : bool (default: True)
+        affects only parameters that were set to a Theano expression. If
```
f0k: *Affects

lasagne/layers/helper.py
```diff
@@ -321,7 +327,8 @@ def get_all_params(layer, **tags):
     Returns
     -------
     params : list
-        A list of Theano shared variables representing the parameters.
+        A list of Theano shared variables (default) or expressions representing
+        the parameters.
```
f0k: Again, I'd omit `(default)`.

lasagne/layers/helper.py
```diff
@@ -336,8 +343,9 @@ def get_all_params(layer, **tags):
     -----
     If any layer's parameter was set to a Theano expression instead of a shared
     variable, the shared variables involved in that expression will be returned
-    rather than the expression itself. Tag filtering considers all variables
-    within an expression to be tagged the same.
+    by default rather than the expression itself. Tag filtering considers all
+    variables within an expression to be tagged the same. To retrieve the
+    original Theano expression set ``unwrap_shared=False``.
```
f0k: Same suggestion for the text as above.

lasagne/layers/helper.py
```diff
@@ -349,7 +357,8 @@ def get_all_params(layer, **tags):
     True
     """
     layers = get_all_layers(layer)
-    params = chain.from_iterable(l.get_params(**tags) for l in layers)
+    params = chain.from_iterable(l.get_params(
+                          unwrap_shared=unwrap_shared, **tags) for l in layers)
```
f0k: We usually use 8 characters for indentation in this case. Would be good to keep this consistent.

lasagne/tests/layers/test_helper.py
```diff
@@ -718,6 +718,29 @@ def test_get_all_params(self):
...
+    def test_get_all_params_with_unwrap_shared(self):
+        from lasagne.layers import (InputLayer, DenseLayer, get_all_params)
+        from theano.tensor.var import TensorVariable
+        l1 = InputLayer((10, 20))
+        l2 = DenseLayer(l1, 30)
+        l3 = DenseLayer(l2, 40)
+
+        assert get_all_params(l3) == get_all_params(l3, unwrap_shared=False)
+
+        l2.add_param((l2.W).dot(l2.W.T), (20, 20), name="W.WT")
```
f0k: I'd prefer to create an expression for `W` of `l3` before creating `l3`. Maybe:
```python
W1 = theano.shared(floatX(np.zeros((30,2))))
W2 = theano.shared(floatX(np.zeros((2,40))))
W_expr = T.dot(W1, W2)
l3 = DenseLayer(l2, W=W_expr)
```
Then directly check that the correct instances are returned (in correct order!) with different settings for `unwrap_shared`, just using two assertions.

wielandbrendel:
I added the changes you requested, the only thing I was unsure about is the intendation. Are you counting 8 spaces from the beginning of the line (after the standard intendation of the paragraph) or after the equality sign?

lasagne/layers/helper.py
```diff
@@ -334,10 +341,12 @@ def get_all_params(layer, **tags):
 
     Notes
     -----
-    If any layer's parameter was set to a Theano expression instead of a shared
-    variable, the shared variables involved in that expression will be returned
-    rather than the expression itself. Tag filtering considers all variables
-    within an expression to be tagged the same.
+    If any of the layer's parameters was set to a Theano expression instead
```
f0k: Minor nitpick: `If any layer's parameter` allows multiple layers, `If any of the layer's parameters` assumes a single layer (as in `Layer.get_params`). Please either restore the first variant or use yet another option: `If any of the layers' parameters`.

lasagne/layers/helper.py
```diff
@@ -349,7 +358,8 @@ def get_all_params(layer, **tags):
     True
     """
     layers = get_all_layers(layer)
-    params = chain.from_iterable(l.get_params(**tags) for l in layers)
+    params = chain.from_iterable(l.get_params(
+                     unwrap_shared=unwrap_shared, **tags) for l in layers)
```
f0k: I meant 8 characters from the beginning of `params` here. For a line break directly after an opening brace, in this codebase we usually do a double indent (= 8 characters instead of 4) in the next line.

f0k:
Thank you, just two minor comments left! After fixing and committing locally, please squash your commits into one. Do `git rebase -i upstream/master`, and in the text editor coming up, change all but your first commit from `pick` to `fixup`. Then do `git push --force` to update this PR. Thanks!

wielandbrendel:
Ok - done!

lasagne/layers/helper.py
```diff
@@ -334,10 +341,12 @@ def get_all_params(layer, **tags):
...
-    variable, the shared variables involved in that expression will be returned
-    rather than the expression itself. Tag filtering considers all variables
-    within an expression to be tagged the same.
+    If any of the layers' parameters was set to a Theano expression instead
+    of a shared variable, `unwrap_shared` controls whether to return the
+    shared variables involved in that expression (``unwrap_shared=True``,
+    the default), or the expression itself (``unwrap_shared=False``). In
+    either case, tag filtering applies to the expressions, considering all
+    variables within an expression to be tagged the same.
     >>> import theano
```
f0k: Just noticed that this example never rendered properly, but I'll fix it afterwards.

f0k:
Thank you! I'm fine with merging. @benanne, do you want to review as well?

benanne:
LGTM. one minor comment is that the docstring for `get_all_params()` is perhaps a bit confusing now: the first paragraph talks about shared variables or expressions, but in the second and third paragraphs only "parameter variables" are mentioned. The note at the end clears it up, but maybe reading from top to bottom could be confusing. I don't know.

f0k:
Valid point... the third paragraph is fine because it starts with "By default" (which originally meant "if no tags were specified", but now also means "if unwrap_shared is left at its default setting"). The second paragraph is only fine if "parameter variables" is interpreted as "isinstance(theano.Variable)", which holds for both shared variables and expressions. Shall we write "parameters" instead, in the second paragraph?

I'd advocate for leaving the `unwrap_shared` explanation in the Notes, because most users will not need to bother.

benanne:
Yeah, no need to move that up. Your proposed change is a better solution.

lasagne/layers/helper.py
```diff
@@ -290,9 +290,10 @@ def get_output_shape(layer_or_layers, input_shapes=None):
...
 
 
-def get_all_params(layer, **tags):
+def get_all_params(layer, unwrap_shared=True, **tags):
     """
-    Returns a list of Theano shared variables that parameterize the layer.
+    Returns a list of Theano shared variables or expressions that
+    parameterize the layer.
 
     This function gathers all parameter variables of all layers below one or
```
f0k: @wielandbrendel, as we just discussed, could you change "parameter variables" to "parameters" here, amending your commit or squashing afterwards? Thanks! (I don't feel fully comfortable doing this as part of the merge, I think github will not render the PR as "merged" but as "closed" then. I should try that sometime.)

f0k:
Merging and addressing the remaining minor documentation change in a separate PR along with fixing the example (https://github.com/Lasagne/Lasagne/pull/643#discussion_r59223160), to not longer hold this back.

PR #644, opened 2016-03-22
Update Theano dependency to 0.8.0

f0k:
Probably a good idea to update to the new Theano release.

The dependency check in `__init__.py` just checks for the new "feature" we use, which is the new name and location for the `max_pool_2d` function. We don't actually require a more recent Theano version than the one that introduced this. Is that okay? Any arguments against that?

### DIFF
--- a/lasagne/__init__.py
+++ b/lasagne/__init__.py
@@ -13,7 +13,9 @@ http://lasagne.readthedocs.org/en/latest/user/installation.html#theano"""
 except ImportError:  # pragma: no cover
     raise ImportError("Could not import Theano." + install_instr)
 else:
-    if not hasattr(theano.tensor.nnet, 'abstract_conv'):  # pragma: no cover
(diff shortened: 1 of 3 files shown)


benanne:
LGTM :) It's good to be able to depend on a release version again. At least for a while!

f0k:
> It's good to be able to depend on a release version again.

Maybe we should also change the dependency to use PyPI instead of git?

benanne:
Sure, maybe that's a bit cleaner. We should probably think about another Lasagne release as well, a lot of people assume that "pip install lasagne" is enough, and to be honest it bothers me a bit that it isn't :)

f0k:
> Sure, maybe that's a bit cleaner.

Did so. We will have to change back to git if we need any new features though.

> We should probably think about another Lasagne release as well

Yes, let's finally tackle that in April!

PR #650, opened 2016-03-30
Spatial Pyramid Pooling

PGrothaus:
Hi everyone,

This PR would include spatial pyramid pooling following reference http://arxiv.org/abs/1406.4729.
The SpatialPyramidPoolingLayer will reduce the output of the convolutional/pooling part of a DNN to a fixed size independent of the original image dimension. Hence, a dense network structure with a fixed number of nodes may be added even though the input dimension is undetermined.

This implementation allows for any level of pooling (the paper uses 3-level and 4-level structures, but in principle it is arbitrary). All this is possible because dnn.dnn_pool can now take Theano expressions as input (https://github.com/Theano/Theano/pull/3965).

I hope this is of interest. 

best,

Philipp

### DIFF
--- a/lasagne/layers/dnn.py
+++ b/lasagne/layers/dnn.py
@@ -27,6 +27,7 @@ __all__ = [
     "MaxPool3DDNNLayer",
     "Conv2DDNNLayer",
     "Conv3DDNNLayer",
+    "SpatialPyramidPoolingDNNLayer",
 ]
 
 
@@ -139,7 +140,6 @@ class MaxPool2DDNNLayer(Pool2DDNNLayer):
                                                 **kwargs)
 
 
-
 class Pool3DDNNLayer(Layer):
     """
     3D pooling layer
@@ -208,7 +208,6 @@ class Pool3DDNNLayer(Layer):
             raise NotImplementedError("Pool3DDNNLayer does not support "
                                       "ignore_border=False.")
 
-
     def get_output_shape_for(self, input_shape):
         output_shape = list(input_shape)  # copy / convert to mutable list
 
@@ -225,7 +224,7 @@ class Pool3DDNNLayer(Layer):
                                              pad=self.pad[1],
                                              ignore_border=True,
                                              )
-        
+
         output_shape[4] = pool_output_length(input_shape[4],
                                              pool_size=self.pool_size[2],
                                              stride=self.stride[2],
@@ -503,3 +502,92 @@ class Conv3DDNNLayer(BaseConvLayer):
                                 conv_mode=conv_mode
                                 )
         return conved
+
+
+class SpatialPyramidPoolingDNNLayer(Layer):
+    """
+    Spatial Pyramid Pooling Layer
+
+    Performs spatial pyramid pooling (SPP) over the input.
+    It will turn a 2D input of arbitrary size into an output of fixed
+    dimension.
+    Hence, the convolutional part of a DNN can be connected to a dense part
+    with a fixed number of nodes even if the dimensions of the
+    input image are unknown.
+
+    The pooling is performed over :math:`l` pooling levels.
+    Each pooling level :math:`i` will create :math:`M_i` output features.
+    :math:`M_i` is given by :math:`n_i * n_i`,
+    with :math:`n_i` as the number of pooling operation per dimension in
+    level :math:`i`, and we use a list of the :math:`n_i`'s as a
+    parameter for SPP-Layer.
+    The length of this list is the level of the spatial pyramid.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or tuple
+        The layer feeding into this layer, or the expected input shape.
+
+    pool_dims : list of integers
+        The list of :math:`n_i`'s that define the output dimension of each
+        pooling level :math:`i`. The length of pool_dims is the level of
+        the spatial pyramid.
+
+    mode : string
+        Pooling mode, one of 'max', 'average_inc_pad' or 'average_exc_pad'.
+        Defaults to 'max'.
+
+    **kwargs
+        Any additional keyword arguments are passed to the :class:`Layer`
+        superclass.
+
+    Notes
+    -----
+    This layer should be inserted between the convolutional part of a
+    DNN and its dense part. Convolutions can be used for
+    arbitrary input dimensions, but the size of their output will
+    depend on their input dimensions. Connecting the output of the
+    convolutional to the dense part then usually demands us to fix
+    the dimensions of the network's InputLayer.
+    The spatial pyramid pooling layer, however, allows us to leave the
+    network input dimensions arbitrary. The advantage over a global
+    pooling layer is the added robustness against object deformations
+    due to the pooling on different scales.
+
+    References
+    ----------
+    .. [1] He, Kaiming et al (2015):
+           Spatial Pyramid Pooling in Deep Convolutional Networks
+           for Visual Recognition.
+           http://arxiv.org/pdf/1406.4729.pdf.
+    """
+    def __init__(self, incoming, pool_dims=[4, 2, 1], mode='max', **kwargs):
+            super(SpatialPyramidPoolingDNNLayer, self).__init__(incoming,
+                                                                **kwargs)
+            if len(self.input_shape) != 4:
+                raise ValueError("Tried to create a SPP layer with "
+                                 "input shape %r. Expected 4 input dimensions "
+                                 "(batchsize, channels, 2 spatial dimensions)."
+                                 % (self.input_shape,))
+            self.mode = mode
+            self.pool_dims = pool_dims
+
+    def get_output_for(self, input, **kwargs):
+        input_size = tuple(symb if fixed is None else fixed
+                           for fixed, symb
+                           in zip(self.input_shape[2:], input.shape[2:]))
+        pool_list = []
+        for pool_dim in self.pool_dims:
+            win_size = tuple((i + pool_dim - 1) // pool_dim
+                             for i in input_size)
+            str_size = tuple(i // pool_dim for i in input_size)
+
+            pool = dnn.dnn_pool(input, win_size, str_size, self.mode, (0, 0))
+            pool = pool.flatten(2)
+            pool_list.append(pool)
+
+        return theano.tensor.concatenate(pool_list, axis=1)
+
+    def get_output_shape_for(self, input_shape):
+        num_features = sum(p*p for p in self.pool_dims)
+        return (input_shape[0], input_shape[1], num_features)
--- a/lasagne/tests/layers/test_pool.py
+++ b/lasagne/tests/layers/test_pool.py
@@ -116,6 +116,29 @@ def upscale_2d(data, scale_factor):
     return upscaled
 
 
+def spatial_pool(data, pool_dims):
+
+    def ceildiv(a, b):
+        return (a + b - 1) // b
+
+    def floordiv(a, b):
+        return a // b
+
+    input_size = data.shape[2:]
+    pooled_data_list = []
+    for pool_dim in pool_dims:
+        pool_size = tuple(ceildiv(i, pool_dim) for i in input_size)
+        stride_size = tuple(floordiv(i, pool_dim) for i in input_size)
+
+        pooled_part = max_pool_2d_ignoreborder(
+                data, pool_size, stride_size, (0, 0))
+        pooled_part = pooled_part.reshape(
+                np.shape(data)[0], np.shape(data)[1] * pool_dim ** 2)
+        pooled_data_list.append(pooled_part)
+
+    return np.concatenate(pooled_data_list, axis=1)
+
+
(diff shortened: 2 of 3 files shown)


f0k:
Nice! I think this would be worth including, to have a plug-and-play solution. It's not possible to implement without writing a custom layer anyway.

There are a couple of changes we'd need, though:
* it should live in `layers/dnn.py`, as it requires cuDNN, and probably it should have DNN in its name then in case we ever add a non-cuDNN equivalent
* casting to float32 and back may result in the division taking place on GPU (haven't tested, but I think Theano is quite greedily moving computation to GPU). It'd probably be more efficient not to cast and use integer division (`//` instead of `/`). This will give you `floor`, for `ceil` you need to add the denominator minus one to the numerator.
* it needs a docstring
* it needs tests
* there are a couple of PEP8 violations that needs fixing

Would you be willing to complete this PR accordingly? Have a look at http://lasagne.readthedocs.org/en/latest/user/development.html#development-setup and following to see how to setup Lasagne for building the documentation and running the tests (including PEP8 checks) locally.

PGrothaus:
Great! Thanks for the positive feedback. I am definitely up to finish everything according to the development setup and will make the changes. I'll let you know when I have moved forward with all this.

Philipp

PGrothaus:
OK, I went through the list and included all the changes. Let me know if there is anything else missing or if more changes are needed.

Philipp

lasagne/layers/dnn.py
```diff
@@ -503,3 +502,95 @@ def convolve(self, input, **kwargs):
...
+
+    Notes
+    -----
+    This layer should be inserted between the convolutional part of a
+    DNN and its dense part. Convolutions can be used for
+    arbitrary input dimensions, but the size of their output will
+    depend on their input dimensions. Connecting the output of the
+    convolutional to the dense part then usually demands us to fix
+    the dimensions of the network's InputLayer.
+    The SPPLayer, however, allows us to leave the network input
```
f0k: That's a bit confusing because it's not called `SPPLayer` -- probably just write `spatial pooling layer` instead.

lasagne/layers/dnn.py
```diff
@@ -503,3 +502,95 @@ def convolve(self, input, **kwargs):
...
+    dimensions arbitrary.
```
f0k: Maybe add a final sentence `The advantage over a global pooling layer is that ...`? Because global average pooling also solves the problem you describe here.

lasagne/layers/dnn.py
```diff
@@ -503,3 +502,95 @@ def convolve(self, input, **kwargs):
...
+                             "(batchsize, channels, 2 spatial dimensions)."
+                             % (self.input_shape,))
+            self.num_features = 0
+            for pool_dim in pool_dims:
+                self.num_features += pool_dim * pool_dim
+            self.pool_dims = pool_dims
+
+    def get_output_for(self, input, **kwargs):
+        input_dim_2 = theano.tensor.shape(input)[2]
+        input_dim_3 = theano.tensor.shape(input)[3]
```
f0k: I guess you could still use the nonsymbolic shape in cases where it's fixed. Something like:
```python
input_size = tuple(symb if fixed is None else fixed
                   for fixed, symb in zip(self.input_shape[2:], input.shape[2:]))
```

lasagne/layers/dnn.py
```diff
@@ -503,3 +502,95 @@ def convolve(self, input, **kwargs):
...
+            super(SpatialPyramidPoolingDNNLayer, self).__init__(incoming,
+                                                             **kwargs)
+            if len(self.input_shape) != 4:
+                raise ValueError("Tried to create a SPP layer with "
+                             "input shape %r. Expected 4 input dimensions "
+                             "(batchsize, channels, 2 spatial dimensions)."
+                             % (self.input_shape,))
+            self.num_features = 0
+            for pool_dim in pool_dims:
+                self.num_features += pool_dim * pool_dim
```
f0k: Just for conciseness, I think I'd prefer `self.num_features = sum(p*p for p in pool_dims)`. And I wouldn't store it as an attribute, but compute it in `get_output_for()` directly. You don't need `num_features` anywhere else!

lasagne/layers/dnn.py
```diff
@@ -503,3 +502,95 @@ def convolve(self, input, **kwargs):
...
+            win_2 = (input_dim_2 + pool_dim - 1) // pool_dim
+            win_3 = (input_dim_3 + pool_dim - 1) // pool_dim
+
+            str_2 = input_dim_2 // pool_dim
+            str_3 = input_dim_3 // pool_dim
+
+            pool = dnn.dnn_pool(input,
+                                (win_2, win_3),
+                                (str_2, str_3),
+                                'max', (0, 0))
```
f0k: Can you make the `mode` configurable just as for the `Pool2DLayer`?

lasagne/tests/layers/test_pool.py
```diff
@@ -116,6 +116,38 @@ def upscale_2d(data, scale_factor):
...
+    num_features = 0
+    for pool_dim in pool_dims:
+        pool_size_2 = input_dim_2 + pool_dim - 1
+        pool_size_2 = pool_size_2 // pool_dim
+
+        pool_size_3 = input_dim_3 + pool_dim - 1
+        pool_size_3 = pool_size_3 // pool_dim
+
+        stride_2 = input_dim_2 // pool_dim
+        stride_3 = input_dim_3 // pool_dim
```
f0k: For increased legibility, you could also do:
```python
input_size = data.shape[2:]
ceildiv = lambda a, b: (a + b - 1) // b
floordiv = lambda a, b: a // b
for pool_dim in pool_dims:
    pool_size = tuple(ceil_div(i, pool_dim) for i in input_size)
    stride = tuple(floor_div(i, pool_dim) for i in input_size)
    ...
```

lasagne/tests/layers/test_pool.py
```diff
@@ -116,6 +116,38 @@ def upscale_2d(data, scale_factor):
...
+            data,
+            (pool_size_2, pool_size_3),
+            (stride_2, stride_3),
+            (0, 0)
+        )
+        pooled_part = pooled_part.reshape(
+            np.shape(data)[0],
+            np.shape(data)[1] * pool_dim ** 2
+        )
+        num_features += pool_dim**2
```
f0k: `num_features` is unused.

lasagne/tests/layers/test_pool.py
```diff
@@ -116,6 +116,38 @@ def upscale_2d(data, scale_factor):
...
+
+        stride_2 = input_dim_2 // pool_dim
+        stride_3 = input_dim_3 // pool_dim
+
+        pooled_part = max_pool_2d_ignoreborder(
+            data,
+            (pool_size_2, pool_size_3),
+            (stride_2, stride_3),
+            (0, 0)
+        )
```
f0k: We don't have this indentation style anywhere else. We'd probably do (along with my suggested change above):
```python
pooled_part = max_pool_2d_ignoreborder(
        data, pool_size, stride, (0, 0))
```
I don't have strong feelings about this, though.

lasagne/tests/layers/test_pool.py
```diff
@@ -815,3 +847,75 @@ def test_get_output_for(self, layer):
...
+        try:
+            input = floatX(np.random.randn(8, 16, 17, 13))
+            input_layer = self.input_layer(input.shape)
+            input_theano = theano.shared(input)
+
+            result = self.layer(
+                input_layer,
+                pool_dims,
+            ).get_output_for(input_theano)
+
```
f0k: Again, we'd probably indent this as
```python
result = self.layer(input_layer, pool_dims).get_output_for(
        input_theano)
```
Not so important for me, though.

f0k:
Thanks, the documentation and tests look very good! I've only got a few minor comments. In the end, could you please squash everything into a single commit (and leave a comment again, because github doesn't send out notifications for added or modified commits)? Thanks!

PGrothaus:
Thanks, I included the suggestions and squashed into one commit. I've made the mode flexible now as well.
While writing this contribution I have merged changes from the master branch at some point, so you will see changes in 'Files changed' that I didn't change myself. Don't know if that's an issue.
Anyway, I'm happy to contribute :smile:

f0k:
> Thanks, I included the suggestions and squashed into one commit. I've made the mode flexible now as well.

Great, that was fast!

> While writing this contribution I have merged changes from the master branch at some point, so you will see changes in 'Files changed' that I didn't change myself. Don't know if that's an issue.

Oh yes, that's an issue, but luckily it can be fixed! You should always rebase instead of merging. `git pull upstream master --rebase` will do the trick whenever you need that. To fix it afterwards, please run `git rebase upstream/master`. Afterwards you will need to force-push with `git push --force origin`. Let me know when you did so we can do a final review!

PGrothaus:
OK, this looks better.

lasagne/layers/dnn.py
```diff
@@ -503,3 +502,95 @@ def convolve(self, input, **kwargs):
                                 conv_mode=conv_mode
                                 )
         return conved
+
+
+class SpatialPyramidPoolingDNNLayer(Layer):
+    """
+    lasagne.layers.SpatialPyramidPoolingDNNLayer(self, incoming,
+    pool_dims=[4,2,1], mode='max'**kwargs)
```
erfannoury: Shouldn't there be a comma and space before `**kwargs`?

PGrothaus: Absolutely!

f0k: Good catch!

lasagne/layers/dnn.py
```diff
@@ -503,3 +502,95 @@ def convolve(self, input, **kwargs):
                                 conv_mode=conv_mode
                                 )
         return conved
+
+
+class SpatialPyramidPoolingDNNLayer(Layer):
+    """
+    lasagne.layers.SpatialPyramidPoolingDNNLayer(self, incoming,
+    pool_dims=[4,2,1], mode='max', **kwargs)
```
f0k: Actually you can just omit this. We only need it when the default arguments include object instances like `lasagne.init.Uniform()`, because they would be rendered as `<lasagne.init.Uniform instance at 0x12345678>` by default.

lasagne/layers/dnn.py
```diff
@@ -503,3 +502,95 @@ def convolve(self, input, **kwargs):
...
+        input_size = tuple(symb if fixed is None else fixed
+                           for fixed, symb
+                           in zip(self.input_shape[2:], input.shape[2:]))
+        pool_list = []
+        for pool_dim in self.pool_dims:
+            win_size = tuple((i + pool_dim - 1) // pool_dim
+                             for i in input_size)
+            str_size = tuple(i // pool_dim for i in input_size)
+
+            pool = dnn.dnn_pool(input, win_size, str_size, self.mode, (0, 0))
```
f0k: Sorry for noticing so late, but does that actually need to use cuDNN? Can't we just use `pool = T.signal.pool.pool_2d(input, win_size, st=str_size, mode=self.mode, ignore_border=False)`? This would allow to move the class to `lasagne/layers/pool` and make it available to everyone.

PGrothaus: I'm afraid that pool.pool_2d doesn't allow for theano expressions as input.

f0k: Ah yes, I forgot about that on the way! Thanks for the reminder!

PGrothaus:
OK, removed that part of the docstring.

f0k:
Perfect! I'll merge tonight unless there are any other comments.

f0k:
Merging, thanks again for your fine work!

PGrothaus:
Happy to contribute! I hope people use it ;)

lasagne/layers/dnn.py
```diff
@@ -503,3 +502,92 @@ def convolve(self, input, **kwargs):
...
+            pool = pool.flatten(2)
+            pool_list.append(pool)
+
+        return theano.tensor.concatenate(pool_list, axis=1)
+
+    def get_output_shape_for(self, input_shape):
+        num_features = sum(p*p for p in self.pool_dims)
+        return (input_shape[0], input_shape[1], num_features)
```
kli-nlpr: Maybe the get_output_shape_for should return ```(input_shape[0],input_shape[1]*num_features)```

kli-nlpr: Maybe ```get_output_shape_for``` should return ```(input_shape[0],num_features*input_shape[1])```?

f0k: You're right, this doesn't match what's returned by the layers. Apparently I haven't read the tests closely enough, the `get_output_for` test should have checked `get_output_shape_for` as well. I think we should change `get_output_for` to return a tensor of (batchsize, channels, num_features) rather than changing `get_output_shape_for`. I'll have a stab at it.

PGrothaus: Yes, good catch. Thanks for taking care of it..

PR #651, opened 2016-03-30
Fix example in InverseLayer docstring

### LINKED ISSUE #639: [Docs] Error example for InverseLayer
I ran the following example, the layer **l_u** inverse jump from l2 back to l1 which course the input shape mismatch error:
```
import lasagne
from lasagne.layers import InputLayer, Conv2DLayer, DenseLayer
from lasagne.layers import InverseLayer
l_in = InputLayer((100, 3, 28, 28))
l1 = Conv2DLayer(l_in, num_filters=16, filter_size=5)
l2 = DenseLayer(l1, num_units=20)
l_u = InverseLayer(l2, l1)  # As Deconv2DLayer
```
I think it should be (inverse the dense layer first to get the right shape for Deconv)
```
l_u2 = InverseLayer(l2, l2)  # As TransposeDense
l_u1 = InverseLayer(l_u, l1)  # As Deconv2DLayer
```

p/s: just a stupid question, as I understand this layer calculate the partial derivative with respect to its own input, but I cannot imagine the math behind it.
For example: the dense layer
_y = f(x) = a*X + b_ => its inverse: _X = f(y) = (y - b)/a_
Calculate the dervative of f(x) with respect to X: give **a** ... I lost here, as I know the slope of a function and its inverse is multiplicative inverse, but not sure how to drive the inverse function from its derivative.

f0k:
Fixes #639. Thanks for reporting!

### DIFF
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -335,7 +335,8 @@ class InverseLayer(MergeLayer):
     >>> l_in = InputLayer((100, 3, 28, 28))
     >>> l1 = Conv2DLayer(l_in, num_filters=16, filter_size=5)
     >>> l2 = DenseLayer(l1, num_units=20)
-    >>> l_u = InverseLayer(l2, l1)  # As Deconv2DLayer
+    >>> l_u2 = InverseLayer(l2, l2)  # backprop through l2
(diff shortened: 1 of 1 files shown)


benanne:
LGTM, merging

PR #652, opened 2016-04-01
Add link to "From Zero to Lasagne" guides

f0k:
This adds a link to the new platform-specific installation guides.

### DIFF
--- a/docs/user/installation.rst
+++ b/docs/user/installation.rst
@@ -10,9 +10,10 @@ tight coupling to Theano, you will have to install a recent version of Theano
 (usually more recent than the latest official release!) fitting the version of
 Lasagne you choose to install.
 
-Most of the instructions below assume you are running a Linux or Mac system;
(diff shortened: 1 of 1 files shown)


benanne:
Wow, excellent work! LGTM, but maybe we would want to move some of this to the docs eventually? I guess the wiki format is more inviting to additions and updates though.

Any Mac users around that would be willing to complete the set? :)

f0k:
> LGTM, but maybe we would want to move some of this to the docs eventually?

Not sure... almost everything of the additional information in there is really platform-specific and not useful to put into the general instructions. Also the general instructions look much easier and less intimidating the way they are :)
The only thing we could do would be creating extra `From Zero to Lasagne` pages in the main documentation. But as you said, the wiki is much easier to edit, that's why I chose it. I'd like people who encounter problems or explore alternative solutions to easily share their insight.

benanne:
Fair enough :)

PR #653, opened 2016-04-01
Update dnn layers to support new theano backend 

jacobzweig:
I noticed that the new theano backend (sandbox.gpuarray) wasn't currently supported by dnn layers. I put together some quick adjustments to allow compatibility. It seems to be working for me based on a preliminary test for 2D convolution. I didn't notice a function in the new theano backend for 3D convolution, so as it stands that's not currently working. Hopefully this is helpful or a decent start!

### DIFF
--- a/lasagne/layers/dnn.py
+++ b/lasagne/layers/dnn.py
@@ -8,20 +8,16 @@ from .conv import conv_output_length, BaseConvLayer
 from .pool import pool_output_length
 from ..utils import as_tuple
 
-if theano.sandbox.cuda.cuda_enabled:
+if theano.sandbox.cuda.dnn.dnn_available():
     from theano.sandbox.cuda import dnn
     print("Using Old Theano Backend")     
-    
-    if not theano.sandbox.cuda.dnn.dnn_available():
-        raise ImportError(
-                "cuDNN not available: %s\nSee http://lasagne.readthedocs.org/en/"
-                "latest/user/installation.html#cudnn" %
-                dnn.dnn_available.msg)  # pragma: no cover
-
 elif theano.sandbox.gpuarray.dnn.dnn_present():
     from theano.sandbox.gpuarray import dnn 
     print("Using New Theano Backend")
-
+else:
+    raise ImportError(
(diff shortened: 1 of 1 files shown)


lasagne/layers/dnn.py
```diff
@@ -9,15 +8,16 @@
 from .pool import pool_output_length
 from ..utils import as_tuple
 
-if not theano.sandbox.cuda.cuda_enabled:
+if theano.sandbox.cuda.dnn.dnn_available():
```
f0k: When device=cpu, you cannot access `theano.sandbox.cuda` if it's not explicitly imported. So this would need to be done differently.

f0k:
:+1: for looking into supporting the new backend. Rather than fixing a single module, we should probably fix all references to `theano.sandbox.cuda`, though. Plus we should amend the configuration for Travis-CI to run tests with OpenCL under the new backend (this doesn't require a GPU): http://stackoverflow.com/questions/29986942/continuous-integration-service-for-gpu-package

lasagne/layers/dnn.py
```diff
@@ -9,15 +8,16 @@
...
+    from theano.sandbox.cuda import dnn
+    print("Using Old Theano Backend")     
+elif theano.sandbox.gpuarray.dnn.dnn_present():
```
HapeMask: The latest version of Theano has moved gpuarray out of the sandbox, so this should just be`theano.gpuarray.dnn.dnn_present()` (and likewise for the rest) for those who are trying to use this PR with the dev version of Theano.

f0k:
More discussion in #698.

gitclem:
Theano has changed:

https://github.com/Theano/Theano/wiki/Converting-to-the-new-gpu-back-end%28gpuarray%29

theano.sandbox.cuda 

needs to be replaced by:
theano.gpuarray

for backwards compatibility, Lasagne should check for both.

f0k:
Closing as this is superseded by #836.

> for backwards compatibility, Lasagne should check for both.

Yes, that's what we're planning!

PR #661, opened 2016-04-14
Fix lasagne updates

### LINKED ISSUE #402: lasagne.updates.adam yields float64 updates with float32 0d-array input
Just tracked down this strange behavior.

```Python
>>> import theano
>>> import lasagne
>>> import numpy as np
>>> import theano.tensor as T
>>> x = theano.shared(np.ones((10,), dtype=theano.config.floatX))
>>> f = T.sum(x**2)
>>> u = lasagne.updates.adam(f, [x], beta2=.99)
>>> print [v.type for v in u.values()]
[TensorType(float32, vector),
 TensorType(float32, vector),
 TensorType(float32, vector),
 TensorType(float32, scalar)]
>>> u = lasagne.updates.adam(
...     f, [x], beta2=np.array(.99, dtype=theano.config.floatX))
>>> print [v.type for v in u.values()]
[TensorType(float32, vector),
 TensorType(float64, vector),
 TensorType(float64, vector),
 TensorType(float32, scalar)]
```

If the resulting updates are passed into `theano.function`, a type error is raised because the update for a `float32` values is `float64`.  In this particular case, what's happening is that in `lasagne.adams.updates`, the following:
```
v_t = beta2*v_prev + (1-beta2)*g_t**2
```
`(1 - beta2)` will be given a dtype `float64`:
```Python
>>> (1 - np.array(0., dtype='float32')).dtype
dtype('float64')
```

So, given this, I think we have three options:

1. Implicitly cast things using `utils.floatX` everywhere, which would make the line in question become
`beta2*v_prev + (utils.floatX(1.)-beta2)*g_t**utils.floatX(2.)` (puke)

1. Convert arguments to updates functions to `float`, because if a float is passed instead of a 0d-array with the correct dtype, everything works out fine thanks to Theano's willingness to cast Python scalars to the correct type

1. Tell users (like me) to RTFM, i.e. `beta2 : float` is specified in the docstring, so if I pass anything else (like a numpy scalar), I'm going to have a bad time.

### LINKED ISSUE #563: dropout layers leak float64 into graph

sarin1991:
These changes fix the upcasting issue in lasagne.update functions rmsprop, adadelta, adam and adamax where float32 inputs are given float64 updates 
A test method, test_update_returntype has also been added to the class TestUpdateFunctions which checks for this error.
Closes #402. Closes #563. Closes #578.

### DIFF
--- a/lasagne/tests/layers/test_noise.py
+++ b/lasagne/tests/layers/test_noise.py
@@ -60,6 +60,12 @@ class TestDropoutLayer:
         assert 0.9 < result_eval.mean() < 1.1
         assert (numpy.round(numpy.unique(result_eval), 2) == [0., 1.25]).all()
 
+    def test_get_output_for_p_float32(self, input_layer):
+        from lasagne.layers.noise import DropoutLayer
+        layer = DropoutLayer(input_layer, p=numpy.float32(0.5))
+        input = theano.shared(numpy.ones((100, 100), dtype=numpy.float32))
+        assert layer.get_output_for(input).dtype == input.dtype
+
     def test_specified_rng(self, input_layer):
         from lasagne.layers.noise import DropoutLayer
         input = theano.shared(numpy.ones((100, 100)))
--- a/lasagne/tests/test_updates.py
+++ b/lasagne/tests/test_updates.py
@@ -65,6 +65,54 @@ class TestUpdateFunctions(object):
         assert np.allclose(A.get_value(), B.get_value())
         assert np.allclose(A.get_value(), self.torch_values[method])
 
+    @pytest.mark.parametrize('method, kwargs', [
+        ['sgd', {'learning_rate': 0.1}],
+        ['momentum', {'learning_rate': 0.1,
+                      'momentum': 0.5}],
+        ['nesterov_momentum', {'learning_rate': 0.1,
+                               'momentum': 0.5}],
+        ['adagrad', {'learning_rate': 0.1,
+                     'epsilon': 1e-6}],
+        ['rmsprop', {'learning_rate': 0.01,
+                     'rho': 0.9,
+                     'epsilon': 1e-6}],
+        ['adadelta', {'learning_rate': 0.01,
+                      'rho': 0.9,
+                      'epsilon': 1e-6}],
+        ['adam', {'learning_rate': 0.01,
+                  'beta1': 0.9,
+                  'beta2': 0.999,
+                  'epsilon': 1e-8}],
+        ['adamax', {'learning_rate': 0.01,
+                    'beta1': 0.9,
+                    'beta2': 0.999,
+                    'epsilon': 1e-8}],
+        ])
+    def test_update_returntype(self, method, kwargs):
+        '''Checks whether lasagne.updates handles float32 inputs correctly'''
+        floatX_ = theano.config.floatX
+        theano.config.floatX = 'float32'
+        try:
+            A = theano.shared(lasagne.utils.floatX([1, 1, 1]))
+            B = theano.shared(lasagne.utils.floatX([1, 1, 1]))
+            update_func = getattr(lasagne.updates, method)
+            updates = update_func(self.f(A) + self.f(B),
+                                  [A, B],
+                                  **kwargs)
+
+            assert all(v.dtype == 'float32' for v in updates)
+
+            # Checking for float32 arguments
+            for param in kwargs:
+                kwargs[param] = np.float32(kwargs[param])
+            updates = update_func(self.f(A) + self.f(B),
+                                  [A, B],
+                                  **kwargs)
+
+            assert all(v.dtype == 'float32' for v in updates)
+        finally:
+            theano.config.floatX = floatX_
+
 
 def test_get_or_compute_grads():
 
--- a/lasagne/updates.py
+++ b/lasagne/updates.py
@@ -443,11 +443,14 @@ def rmsprop(loss_or_grads, params, learning_rate=1.0, rho=0.9, epsilon=1e-6):
     grads = get_or_compute_grads(loss_or_grads, params)
     updates = OrderedDict()
 
+    # Using theano constant to prevent upcasting of float32
+    one = T.constant(1)
+
     for param, grad in zip(params, grads):
         value = param.get_value(borrow=True)
         accu = theano.shared(np.zeros(value.shape, dtype=value.dtype),
                              broadcastable=param.broadcastable)
-        accu_new = rho * accu + (1 - rho) * grad ** 2
+        accu_new = rho * accu + (one - rho) * grad ** 2
         updates[accu] = accu_new
         updates[param] = param - (learning_rate * grad /
                                   T.sqrt(accu_new + epsilon))
@@ -511,6 +514,9 @@ def adadelta(loss_or_grads, params, learning_rate=1.0, rho=0.95, epsilon=1e-6):
     grads = get_or_compute_grads(loss_or_grads, params)
     updates = OrderedDict()
 
+    # Using theano constant to prevent upcasting of float32
+    one = T.constant(1)
+
     for param, grad in zip(params, grads):
         value = param.get_value(borrow=True)
         # accu: accumulate gradient magnitudes
@@ -521,7 +527,7 @@ def adadelta(loss_or_grads, params, learning_rate=1.0, rho=0.95, epsilon=1e-6):
                                    broadcastable=param.broadcastable)
 
         # update accu (as in rmsprop)
-        accu_new = rho * accu + (1 - rho) * grad ** 2
+        accu_new = rho * accu + (one - rho) * grad ** 2
         updates[accu] = accu_new
 
         # compute parameter update, using the 'old' delta_accu
@@ -530,7 +536,7 @@ def adadelta(loss_or_grads, params, learning_rate=1.0, rho=0.95, epsilon=1e-6):
         updates[param] = param - learning_rate * update
 
         # update delta_accu (as accu, but accumulating updates)
-        delta_accu_new = rho * delta_accu + (1 - rho) * update ** 2
+        delta_accu_new = rho * delta_accu + (one - rho) * update ** 2
         updates[delta_accu] = delta_accu_new
 
     return updates
@@ -578,8 +584,11 @@ def adam(loss_or_grads, params, learning_rate=0.001, beta1=0.9,
     t_prev = theano.shared(utils.floatX(0.))
     updates = OrderedDict()
 
+    # Using theano constant to prevent upcasting of float32
+    one = T.constant(1)
+
     t = t_prev + 1
-    a_t = learning_rate*T.sqrt(1-beta2**t)/(1-beta1**t)
+    a_t = learning_rate*T.sqrt(one-beta2**t)/(one-beta1**t)
 
     for param, g_t in zip(params, all_grads):
         value = param.get_value(borrow=True)
@@ -588,8 +597,8 @@ def adam(loss_or_grads, params, learning_rate=0.001, beta1=0.9,
         v_prev = theano.shared(np.zeros(value.shape, dtype=value.dtype),
                                broadcastable=param.broadcastable)
 
-        m_t = beta1*m_prev + (1-beta1)*g_t
-        v_t = beta2*v_prev + (1-beta2)*g_t**2
+        m_t = beta1*m_prev + (one-beta1)*g_t
+        v_t = beta2*v_prev + (one-beta2)*g_t**2
         step = a_t*m_t/(T.sqrt(v_t) + epsilon)
 
         updates[m_prev] = m_t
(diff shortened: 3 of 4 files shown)


Seanny123:
LGTM.

benanne:
Awesome work, thanks a lot! Just two remarks, both fairly minor:

- I wonder if there's some way we can make this less verbose, all the `as_theano_expression` calls add quite a lot of cruft. Maybe it could be as simple as creating a simple multi-argument shortcut function so we could do something like `learning_rate, rho, epsilon = no_upcast(learning_rate, rho, epsilon)`.

- When adding a new updates function, the coverage checker will not tell us that we also need to add some parameters to this test as well. I wonder if there's a way to make things more general, so this is less likely to be forgotten.

Just some thoughts, I would be fine with seeing this merged as-is though. LGTM :)

lasagne/tests/test_updates.py
```diff
@@ -65,6 +65,53 @@ def test_updates(self, method, kwargs):
         assert np.allclose(A.get_value(), B.get_value())
         assert np.allclose(A.get_value(), self.torch_values[method])
 
+    @pytest.mark.parametrize('method, kwargs', [
+        ['sgd', {'learning_rate': 0.1}],
+        ['sgd', {'learning_rate': np.float32(0.1)}],
```
f0k: 1. Instead of listing everything twice, can you run it once with floats and then with np.float32 in the test function body?
2. Instead of listing all the arguments, would it be possible to just list the methods and then in the function body, use `inspect.getargspec` to get the default arguments and once call with the defaults and then with the defaults converted to np.float32? This would make it a lot easier to extend if more functions are added.

lasagne/updates.py
```diff
@@ -440,6 +440,12 @@ def rmsprop(loss_or_grads, params, learning_rate=1.0, rho=0.9, epsilon=1e-6):
            Neural Networks for Machine Learning, Lecture 6.5 - rmsprop.
            Coursera. http://www.youtube.com/watch?v=O3sxAc4hxZU (formula @5:20)
     """
+    # Converting input arguments to theano expressions so that
+    # float32 inputs arent upcasted to float 64 automatically
+    learning_rate = utils.as_theano_expression(learning_rate)
+    rho = utils.as_theano_expression(rho)
+    epsilon = utils.as_theano_expression(epsilon)
```
f0k: I think I'd prefer only converting what's needed. Something like:
```python
# Convert rho to theano expression so (1 - rho) will never be casted to float64
rho = utils.as_theano_expression(rho)
```
The others don't risk upcasting anyway. Similar for adam and adamax: only beta1, beta2 need to be converted.

sarin1991:
@benanne  I have added in a function no_upcast to utils pretty much along the lines you suggested. I was initially thinking of doing the same but wasn't sure if you guys were okay with me adding another function to utils.py. I probably need to improve the documentation. What do you think?

@f0k I can remove the variables which are not necessary if you think that's better but I was thinking we should do this check for all the arguments including the ones sent to layer instances. As it can cause weird errors in the future if the people developing it are not careful about this upcasting.

lasagne/utils.py
```diff
@@ -21,6 +21,29 @@ def floatX(arr):
     return np.asarray(arr, dtype=theano.config.floatX)
 
 
+def no_upcast(list_element):
```
craffel: Why not use `*args`?  This would save you from having to construct a list to call the function, which would make it cleaner, IMO:
```Python
def no_upcast(*args):
    result_list = []
     for element in args:
         if isinstance(element, np.ndarray):
             result_list.append(T.constant(element))
         else:
             result_list.append(element)
     return result_list
```
and the call signature would be e.g.
```
learning_rate, rho, epsilon = utils.no_upcast(
    learning_rate, rho, epsilon)
```

... or, better yet, make it a pretty one-liner with a list comprehension
```
def no_upcast(*args):
     return (T.constant(e) if isinstance(e, np.ndarray) else e for e in args)
```

craffel:
Thanks for doing this!  We need it.  Just made a comment re: your `no_upcast` function, I personally think it would be cleaner to just use `*args` rather than forcing an iterable to be passed in.  The function could also be written in a readable one-liner.  Otherwise LGTM after a cursory glance.

f0k:
> I have added in a function no_upcast to utils pretty much along the lines you suggested. I was initially thinking of doing the same but wasn't sure if you guys were okay with me adding another function to utils.py. I probably need to improve the documentation. What do you think?

To be honest, I liked it better before. Firstly, `no_upcast` is not a suitable name because it still only wraps things as Theano variables, we just happen to need it in one case to sidestep numpy's upcasting rules. And secondly, I don't see a need for an extra function just to save two or three lines in total.

> I can remove the variables which are not necessary if you think that's better but I was thinking we should do this check for all the arguments including the ones sent to layer instances.

Fair enough, but you're not doing it for all the arguments either, just for the update functions that were affected. And I wouldn't want to expand this to all attributes of all layers, we couldn't do any comparisons to constants then:
```
>>> 1 == 1
True
>>> np.array(1) == 1
True
>>> T.constant(1) == 1
False
```

> As it can cause weird errors in the future if the people developing it are not careful about this upcasting.

Seeing where this leads, isn't it easier to just wrap the few Python constants we have in `utils.floatX()`? People need to be careful in either case.

sarin1991:
@craffel Yes, using *args looks a lot cleaner. I will implement your suggestion if we end up using a function in utils. Thanks! :)

@f0k `utils.floatX()` doesn't work. It breaks when you give python float arguments.
```
>>>from lasagne import utils
>>>beta1 = 0.99
>>>one = utils.floatX(1)
>>>one.dtype
dtype('float32')
>>>(one-beta1).dtype
dtype('float64')
```

> Fair enough, but you're not doing it for all the arguments either, just for the update functions that were affected. And I wouldn't want to expand this to all attributes of all layers, we couldn't do any comparisons to constants

I didn't realize that we were doing comparison operations since I thought all the parameters could be replaced as a theano expression. You are right that would be a problem. We could use theano's get_scalar_constant_value function though but I guess it might be easier to just use `utils.floatX()` for dropout layer.

In any case, I don't mind whether we use a new function or not. I will wait for a bit to see what approach is acceptable by everyone here before making any further changes :)

f0k:
> `utils.floatX()` doesn't work. It breaks when you give python float arguments.

Bummer... now that's a good argument against it :) Guess I only checked the beta1=ndarray case.

> I didn't realize that we were doing comparison operations since I thought all the parameters could be replaced as a theano expression.

We'd need to check in detail. Some parameters will accept Theano expressions without problems, some with minor disadvantages (e.g., the dropout rate can be a Theano expression, but then it might not be optimized away if it's set to zero), some won't (e.g., the number of units in dense layers). What about turning the `1` into a Theano expression instead of the `beta` parameters?
```
>>> one = T.constant(1)
>>> (one - .99).dtype
'float32'
>>> (one - np.array(.99, dtype=np.float32)).dtype
'float32'
>>> (one - T.scalar()).dtype
'float32'
```
This would cover all cases and be less invasive. What do you think? What do others think?

 `T.constant(1.)` and `T.constant(1)` seem to give the same compiled graph (e.g., `T.matrix() - T.constant(1)` will be executed on GPU with `T.constant(1)` becoming a float32 CUDA array even if `T.constant(1)` is `int8` originally).

sarin1991:
That's a great idea! :) Yes, I agree it should work for all the cases and is less invasive

sarin1991:
I have updated the fix using T.constant in the expression. I have also added the fix for the Dropout Layer and added a unit test for it

lasagne/tests/layers/test_noise.py
```diff
@@ -60,6 +65,11 @@ def test_get_output_for_p_02(self, layer_p_02):
         assert 0.9 < result_eval.mean() < 1.1
         assert (numpy.round(numpy.unique(result_eval), 2) == [0., 1.25]).all()
 
+    def test_get_output_for_p_float32_dtype(self, layer_p_float32):
+        input = theano.shared(numpy.ones((100, 100), dtype=numpy.float32))
+        result = layer_p_float32.get_output_for(input)
+        assert result.dtype == input.dtype
```
f0k: As `layer_p_float32` is only used once, I'd prefer not to define it and write the test directly as:
```python
def test_get_output_for_float32_p(self):
    floatX_ = theano.config.floatX
    theano.config.floatX = 'float32'
    from lasagne.layers.noise import DropoutLayer
    layer = DropoutLayer((None, 100), p=numpy.float32(0.5))
    input = T.matrix()
    assert layer.get_ouput_for(input).dtype == input.dtype
    theano.config.floatX = floatX_
```

lasagne/updates.py
```diff
@@ -440,14 +440,17 @@ def rmsprop(loss_or_grads, params, learning_rate=1.0, rho=0.9, epsilon=1e-6):
            Neural Networks for Machine Learning, Lecture 6.5 - rmsprop.
            Coursera. http://www.youtube.com/watch?v=O3sxAc4hxZU (formula @5:20)
     """
+    # Using theano constant to prevent upcasting of float32
+    one = T.constant(1)
+
```
f0k: Nitpick, but I'd prefer to move this *after* `grads = ` and `updates = ` so all update functions start the same.

sarin1991:
@f0k What do you think? Everything okay now?

lasagne/updates.py
```diff
@@ -440,14 +440,18 @@ def rmsprop(loss_or_grads, params, learning_rate=1.0, rho=0.9, epsilon=1e-6):
            Neural Networks for Machine Learning, Lecture 6.5 - rmsprop.
            Coursera. http://www.youtube.com/watch?v=O3sxAc4hxZU (formula @5:20)
     """
+
```
f0k: Please remove this extra whitespace.

lasagne/updates.py
```diff
@@ -508,9 +512,13 @@ def adadelta(loss_or_grads, params, learning_rate=1.0, rho=0.95, epsilon=1e-6):
            ADADELTA: An Adaptive Learning Rate Method.
            arXiv Preprint arXiv:1212.5701.
     """
+
```
f0k: And this.

lasagne/updates.py
```diff
@@ -574,12 +582,16 @@ def adam(loss_or_grads, params, learning_rate=0.001, beta1=0.9,
            Adam: A Method for Stochastic Optimization.
            arXiv preprint arXiv:1412.6980.
     """
+
```
f0k: And this.

lasagne/updates.py
```diff
@@ -633,12 +645,16 @@ def adamax(loss_or_grads, params, learning_rate=0.002, beta1=0.9,
...
```
f0k: And this.

lasagne/tests/test_updates.py
```diff
@@ -65,6 +65,52 @@ def test_updates(self, method, kwargs):
...
+
+        # Checking for float32 arguments
+        for param in kwargs:
+            kwargs[param] = np.float32(kwargs[param])
+        updates = update_func(self.f(A) + self.f(B),
+                              [A, B],
+                              **kwargs)
+
+        assert all(v.dtype == 'float32' for v in updates)
+        theano.config.floatX = floatX_
```
f0k: Very good, thank you! Just one last thing: We should probably wrap this in a try..finally block, otherwise all other tests will be affected if this one fails:
```python
floatX_ = theano.config.floatX
theano.config.floatX = 'float32'
try:
    A = ...
    B = ...
    ...
    assert all(...)
finally:
    theano.config.floatX = floatX_
```
Sorry, I didn't think of this before.

sarin1991:
@f0k How about now?

f0k:
Perfect! Now we just need to reorganize the commits a bit. Can you do:
```bash
git reset 0440814  # that was were you branched off
git add lasagne/updates.py
git add lasagne/tests/test_updates.py
git commit -m 'Avoid upcast to float64 in lasagne.updates'
git add lasagne/layers/noise.py
git add lasagne/tests/layers/test_noise.py
git commit -m 'Avoid upcast to float64 in DropoutLayer'
git push --force
```
This will group all your changes into two coherent commits. Let me know when you're done and I can merge!

sarin1991:
@f0k Done!

f0k:
Yay, thanks a lot for all your work! Merging!

@benanne: coveralls is back, did you do anything or did that happen on its own?

sarin1991:
Sure no problem! 👍

benanne:
@f0k not sure, I think I "regenerated the repository token" a couple of days ago, whatever that means. Maybe that did it. Maybe not :)

PR #664, opened 2016-04-19
better GPU usage for TransformerLayer

kmyid:
Hi,

I noticed that the index computation for interpolation with the _interpolate function is doing integer math. This causes unnecessary overheads as it seems this cannot be done in the GPU.
Instead, if we keep the indices as floats until the very end, all this computation can be done on the GPU.

Cheers,
Kwang

### DIFF
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -461,7 +461,7 @@ def _transform_affine(theta, input, downsample_factor):
 
 
 def _interpolate(im, x, y, out_height, out_width):
-    # *_f are floats
+    # *_f are floats, othere may also be floats.
     num_batch, height, width, channels = im.shape
     height_f = T.cast(height, theano.config.floatX)
     width_f = T.cast(width, theano.config.floatX)
@@ -481,18 +481,21 @@ def _interpolate(im, x, y, out_height, out_width):
     y0_f = T.floor(y)
     x1_f = x0_f + 1
     y1_f = y0_f + 1
-    x0 = T.cast(x0_f, 'int64')
-    y0 = T.cast(y0_f, 'int64')
-    x1 = T.cast(T.minimum(x1_f, width_f - 1), 'int64')
-    y1 = T.cast(T.minimum(y1_f, height_f - 1), 'int64')
+    # do not cast to integers yet, to allow computation to take place in GPU
+    x0 = T.floor(x0_f)
+    y0 = T.floor(y0_f)
+    x1 = T.floor(T.minimum(x1_f, width_f - 1))
+    y1 = T.floor(T.minimum(y1_f, height_f - 1))
 
     # The input is [num_batch, height, width, channels]. We do the lookup in
-    # the flattened input, i.e [num_batch*height*width, channels]. We need
-    # to offset all indices to match the flat version
-    dim2 = width
-    dim1 = width*height
+    # the flattened input, i.e [num_batch*height*width, channels]. We need to
+    # offset all indices to match the flat version. We also use the float
+    # vertion to make sure that the computation can be carried on the GPU.
+    dim2 = width_f
+    dim1 = width_f*height_f
     base = T.repeat(
-        T.arange(num_batch, dtype='int64')*dim1, out_height*out_width)
+        T.arange(num_batch, dtype=theano.config.floatX)*dim1,
+        out_height*out_width)
     base_y0 = base + y0*dim2
     base_y1 = base + y1*dim2
     idx_a = base_y0 + x0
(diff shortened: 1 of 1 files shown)


f0k:
Thanks! But IIRC, I tried several different variants in #394 and the current one turned out to be pretty much optimal. Also see #408. Can you please profile your variant and check whether it's actually faster? It's possible that your version causes additional transfers.

kmyid:
Sure thing! didn't notice that you guys already had that discussion!

f0k:
> Do you still have those scripts by any chance?

Turns out I do! It's the one used for this comment: https://github.com/Lasagne/Lasagne/pull/394#issuecomment-134597468.
```python
import sys
import numpy as np
import theano
import theano.tensor as T
floatX = theano.config.floatX
from theano.sandbox.cuda import as_cuda_ndarray_variable
from theano.sandbox.cuda import synchronize as sync
import lasagne

batch_size = 64
channels = 3
image_size = (50, 50)
input_shape = (batch_size, channels)
factor = 1.5
if sys.argv[1] == 'fixed':
	input_shape = input_shape + image_size
elif sys.argv[1] == 'dynamic':
	input_shape = input_shape + (None, None)
else:
	raise NotImplementedError("First argument must be 'fixed' or 'dynamic'")
input_layer = lasagne.layers.InputLayer(input_shape)
loc_layer = lasagne.layers.InputLayer((batch_size, 6))
trans = lasagne.layers.TransformerLayer(input_layer, loc_layer,
		downsample_factor=factor)

x = input_layer.input_var
l = loc_layer.input_var
y = lasagne.layers.get_output(trans)

X = np.random.randn(batch_size, channels, *image_size).astype(floatX)
L = np.random.rand(batch_size, 6).astype(floatX)
fn_fwd = theano.function([], as_cuda_ndarray_variable(y), givens={x: X, l: L})

gy = T.tensor4()
Y = np.random.randn(batch_size, channels,
		int(image_size[0]/factor), int(image_size[1]/factor)).astype(floatX)
gx, gl = theano.grad(None, wrt=[x, l], known_grads={y: gy})
gx = as_cuda_ndarray_variable(gx)
gl = as_cuda_ndarray_variable(gl)
fn_bwd = theano.function([], [gx, gl], givens={x: X, l: L, gy: Y},
		on_unused_input='ignore')

if sys.argv[2] == 'fwd':
	fn = fn_fwd
elif sys.argv[2] == 'bwd':
	fn = fn_bwd
else:
	raise NotImplementedError("Second argument must be 'fwd' or 'bwd'")

import timeit
print min(timeit.repeat('fn(); fn(); fn(); fn(); fn(); sync()', 'from __main__ import fn, sync', repeat=50, number=1))
```
The first command line should not make a difference (fixed or dynamic), the second one changes between forward and backward pass (fwd and bwd).

kmyid:
Hi, I'm back :-)

So here are my timing results with Titan X. I ran your script with repeat = 1000 a couple of times and took the minimum.

|            |FORWARD     |BACKWARD   |
|------------|:----------:|:---------:|
|NEW         |0.01620     |0.03023    |
|OLD	     |0.01665     |0.03284    |
|GAIN  	     |2.703 %     |7.948 %    |

As you noted before [here](https://github.com/Lasagne/Lasagne/pull/394#issuecomment-134557572) `GpuAdvancedSubtensor1` is the main time consuming one and it's not like super duper. However, it's at least not slow, and when you profile the function, there's nolonger the pesky Elemwise{add} which was there before (it's now GpuElemwise{add}). There's a lot of Elemwise{Cast{int64}} still, but I guess that's inevitable.

I also checked that they give identical results. Of course when CuDNN is used, the back propagation might be slightly different (1e-5 when the max of data is 3e3).

f0k:
Thanks for timing! I'll run this check as well.

> Of course when CuDNN is used, the back propagation might be slightly different (1e-5 when the max of data is 3e3).

There's no cuDNN involved here at all, but that seems like a negligible difference.

The comments would need some improvements, though, I'll comment on the diff.

kmyid:
Hi, again.

> I'll comment on the diff.

I am waiting for your feedback :-)

f0k:
Sorry for not acting on this earlier, in the course of #793 I've now gotten familiar with the code again. A possible problem is that float32 can only losslessly represent integers up to 16777216:
```
>>> int(np.float32(16777217))
16777216
```
The indices in question are over batchsize * width * height, so already for a batch of 128 images of 400 times 400, there would be rounding errors. Inputs will often be smaller than this, but it's not an unrealistic input size either. So we should not defer the integer conversion as much as proposed in this PR! Thank you anyway!

PR #665, opened 2016-04-23
Have `output_shape` property error out for symbolic shapes

### LINKED ISSUE #645: Check that Layer.output_shape is not symbolic
Users creating custom layers may inadvertently have `get_output_shape_for()` return a (partly) symbolic shape (e.g., https://groups.google.com/forum/#!topic/lasagne-users/uEPYSSk6sRI). The `Layer.output_shape` property could check that it returns a valid shape, i.e., a tuple that only contains integers, or numpy integers, or `None`. If not, it could display a helpful error message.

f0k:
Closes #645.

### DIFF
--- a/lasagne/layers/base.py
+++ b/lasagne/layers/base.py
@@ -1,5 +1,7 @@
 from collections import OrderedDict
 
+import theano.tensor as T
+
 from .. import utils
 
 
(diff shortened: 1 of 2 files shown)


benanne:
There seems to be a PEP8 problem with the tests, LGTM other than that!

f0k:
Oh. New attempt.

PR #666, opened 2016-04-23
Document that corrmm, cuda_convnet and dnn layers need explicit imports

f0k:
Closes the last TODO item of #524.


benanne:
great, thanks!

PR #671, opened 2016-04-28
Fix SpatialPyramidPoolingDNNLayer mismatch between output shape and data

f0k:
Follow-up to #650, which produced output data not matching the reported output shape (batchsize, channels * features) vs. (batchsize, channels, features), see https://github.com/Lasagne/Lasagne/pull/650/files#r61360880. Changed the layer to return (batchsize, channels, features) and extended the tests to ensure the returned data shape matches the computed output shape.

Thanks to @kli-nlpr for reporting it!


f0k:
@PGrothaus, since you did the original PR, could you review this? @benanne is currently abroad. Cheers!

PGrothaus:
Sure. Everything's good. All changes make sense to me.

f0k:
Thanks for checking, merging!

PR #672, opened 2016-04-28
Name layers created using batch_norm() utility

### LINKED ISSUE #670: Name layers created using batch_norm() utility
This is a _very_ minor issue, but I think it would be nice for layers created using the `batch_norm()` utility to be named. This way all layers will be named when e.g. printing out network structures.

The created `BatchNormLayer` and `NonlinearityLayer` layers could be named by appending a suffix to the name of the wrapped layer, separated by a dot (like in the recurrent layers). In case the `kwargs` argument passed on to the BN layer already contains a name, it could be used to name the `BatchNormLayer` and as a base name for the `NonlinearityLayer`. If neither the wrapped layer is named, nor a name is specified in `kwargs`, the created layers could remain unnamed.

I could make a PR for this.

wuaalb:
Closes #670.

Using suffixes '_bn' and '_bn_nonlin' (rather than just '_nonlin' for consistency w.r.t. case where BN layer is named through kwargs) for `BatchNormLayer` and `NonlinearityLayer` respectively.

Did not modify the docstring to mention this behavior.

### DIFF
--- a/lasagne/layers/normalization.py
+++ b/lasagne/layers/normalization.py
@@ -365,8 +365,11 @@ def batch_norm(layer, **kwargs):
     if hasattr(layer, 'b') and layer.b is not None:
         del layer.params[layer.b]
         layer.b = None
-    layer = BatchNormLayer(layer, **kwargs)
+    bn_name = (kwargs.pop('name', None) or
+               (getattr(layer, 'name', None) and layer.name + '_bn'))
+    layer = BatchNormLayer(layer, name=bn_name, **kwargs)
     if nonlinearity is not None:
         from .special import NonlinearityLayer
-        layer = NonlinearityLayer(layer, nonlinearity)
+        nonlin_name = bn_name and bn_name + '_nonlin'
+        layer = NonlinearityLayer(layer, nonlinearity, name=nonlin_name)
     return layer
--- a/lasagne/tests/layers/test_normalization.py
+++ b/lasagne/tests/layers/test_normalization.py
@@ -299,3 +299,29 @@ def test_batch_norm_macro():
     bnstack = batch_norm(layer, name='foo')
     assert isinstance(bnstack, BatchNormLayer)
     assert bnstack.name == 'foo'
+
+    # check if created layers are named with kwargs name
+    layer = Mock(Layer, output_shape=input_shape, nonlinearity=obj)
+    layer.name = 'foo'
+    bnstack = batch_norm(layer, name='foo_bnorm')
+    assert isinstance(bnstack, NonlinearityLayer)
+    assert isinstance(bnstack.input_layer, BatchNormLayer)
+    assert bnstack.name == 'foo_bnorm_nonlin'
+    assert bnstack.input_layer.name == 'foo_bnorm'
+
+    # check if created layers are named with wrapped layer name
+    layer = Mock(Layer, output_shape=input_shape, nonlinearity=obj)
+    layer.name = 'foo'
+    bnstack = batch_norm(layer)
+    assert isinstance(bnstack, NonlinearityLayer)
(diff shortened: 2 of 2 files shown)


lasagne/layers/normalization.py
```diff
@@ -359,14 +359,32 @@ def batch_norm(layer, **kwargs):
...
+    if bn_name is not None:
+        nonlin_name = bn_name + '_nonlin'
+    # Else, use wrapped layer's name as base name for
+    # created BN and Nonlinearity layers
+    elif wrapped_name is not None:
+        bn_name = wrapped_name + '_bn'
+        nonlin_name = bn_name + '_nonlin'
+    # If neither BN layer is named, nor wrapped layer has name,
+    # leave all created layers unnamed
+
```
f0k: Functionality is good, but I'd like to keep this shorter:
```python
bn_name = kwargs.pop('name', None) or (layer.name and layer.name + '_bn')
nonlin_name = bn_name and bn_name + '_nonlin'
```
Two lines instead of ten! I'd also move these two lines right before the respective layer is created. E.g., `nonlin_name` is only needed if we actually create a `NonlinearityLayer`.

wuaalb:
Sorry about the tests..
I'm not familiar with mock, what would be the proper way to create a named layer?

This way seems to pass the test
```python
def test_batch_norm_macro(dummy_input_layer):
    from lasagne.layers import (Layer, DenseLayer, BatchNormLayer, batch_norm,
                                NonlinearityLayer)
    from lasagne.nonlinearities import identity
    input_shape = (2, 3)
    obj = object()

   ...

    # check if created layers are named with kwargs name
    layer = DenseLayer(dummy_input_layer, num_units=3, nonlinearity=obj,
                 name='foo')
    bnstack = batch_norm(layer, name='foo_bnorm')
    assert isinstance(bnstack, NonlinearityLayer)
    assert isinstance(bnstack.input_layer, BatchNormLayer)
    assert bnstack.name == 'foo_bnorm_nonlin'
    assert bnstack.input_layer.name == 'foo_bnorm'

    ...
```
... but I'm just guessing to be honest.

lasagne/layers/normalization.py
wuaalb: Great! Will change this and correct tests at the same time.

f0k:
> I'm not familiar with mock, what would be the proper way to create a named layer?

I'd have tried what you did (passing `name='foo'`) to the `Mock` constructor. But reading the documentation, `Mock` actually accepts a `name` keyword argument on its own, so you can't use this to set the `name` attribute of the `Mock`. You can do `layer = Mock(...)` and `layer.name = 'foo'` instead.

wuaalb:
Ok, fixed now. Thanks!

lasagne/layers/normalization.py
```diff
@@ -365,8 +365,11 @@ def batch_norm(layer, **kwargs):
     if hasattr(layer, 'b') and layer.b is not None:
         del layer.params[layer.b]
         layer.b = None
-    layer = BatchNormLayer(layer, **kwargs)
+    bn_name = kwargs.pop('name', None) or \
+        (getattr(layer, 'name', None) and layer.name + '_bn')
```
f0k: Looks good except that we usually do a double indent here. Or use visual indent with an extra pair of brackets:
```python
bn_name = (kwargs.pop('name', None) or
           getattr(layer, 'name', None) and layer.name + '_bn')
```
Sorry for nitpicking! It's just that a single indentation level looks like an indented code block rather than a continued line.

f0k:
Thanks, looks good to me now! Will merge tomorrow unless there are further comments.

f0k:
Merged, thanks again for idea and realization!

PR #673, opened 2016-04-29
Add DilatedConv2DLayer

f0k:
This adds a dilated convolution layer (as used in http://arxiv.org/abs/1511.07122). For the implementation, it uses the backward pass wrt. weights, using the fact that the convolution backward passes already support implicit dilation of the gradient wrt. output. By providing the weights as the output gradient and the input as the input, the convolution backward pass wrt. weights computes the dilated correlation we want.

The only caveat of this implementation is that we cannot support padding (it would pad the filters, resulting in cropped output) or filter flipping (it would flip the input, resulting in flipped output).
Performance-wise, another downside of the implementation is that we need to transpose the input and output, swapping batch size and channels.

An alternative implementation would reshape the input such that strided input excerpts become separate input examples, then perform a normal convolution. Something similar is possible for transposed convolution as well. The downside is that this would also incur a copy operation, and require input dimensions to be divisible by the filter dilation factor. We might include this as a `method='reshape'` option later on if it turns out to be advantageous for anything.

@fvisin, since you're into transposed convolution, do you feel competent to review this one as well?

### DIFF
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -13,6 +13,7 @@ __all__ = [
     "Conv2DLayer",
     "TransposedConv2DLayer",
     "Deconv2DLayer",
+    "DilatedConv2DLayer",
 ]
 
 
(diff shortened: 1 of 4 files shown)


lasagne/layers/conv.py
```diff
@@ -781,3 +782,153 @@ def convolve(self, input, **kwargs):
...
+                    "DilatedConv2DLayer requires flip_filters=False.")
+
+    def get_W_shape(self):
+        num_input_channels = self.input_shape[1]
+        # first two sizes are swapped compared to a forward convolution
+        return (num_input_channels, self.num_filters) + self.filter_size
+
+    def get_output_shape_for(self, input_shape):
+        # when called from the constructor, dilation is still called stride:
+        dilation = getattr(self, 'dilation', getattr(self, 'stride', None))
```
fvisin: Consider calling `super` in the constructor *after* renaming `stride`, so that you don't need this line.

lasagne/layers/conv.py
```diff
@@ -781,3 +782,153 @@ def convolve(self, input, **kwargs):
...
+        batchsize = input_shape[0]
+        return ((batchsize, self.num_filters) +
+                tuple(conv_output_length(input, (filter-1) * dilate + 1, 1, 0)
+                      for input, filter, dilate
+                      in zip(input_shape[2:], self.filter_size, dilation)))
+
+    def convolve(self, input, **kwargs):
+        # we perform a convolution backward pass wrt weights,
+        # passing kernels as output gradient
+        imshp = self.input_shape
```
fvisin: I am pretty sure it's never the case, but I'll double check: in no case `self.input_shape` can be symbolic, right?

fvisin:
@f0k LGTM, nice addition to the arsenal!

lasagne/layers/conv.py
f0k: We could, but then we need to copy all this logic: https://github.com/Lasagne/Lasagne/blob/3525dfb/lasagne/layers/conv.py#L283 -- ah, it's just a single line here. It was more for the `pad`/`crop` argument in the transposed convolution layer, and I copied the scheme here. Will change it to set the `dilation` argument before the `super` call and delete the `stride` arguments after the `super` call.

lasagne/layers/conv.py
f0k: Yes, that's forbidden and since #665 there's an explicit check to help users avoid this when writing custom layers.

f0k:
Thanks for reviewing! I've addressed your comment. Will leave it open until tomorrow and then merge if there are no further comments.

fvisin:
No problem, I'm happy to help when I can!  Thanks for the PR :)

benanne:
Thanks to both of you, very nice work!

lamblin:
Is it correct to say that this implementation does not support sub-sampling the output in addition to dilating the weights? Or would these options always cancel out, so it would not make sense?

f0k:
> Is it correct to say that this implementation does not support sub-sampling the output in addition to dilating the weights?

Yes, this implementation does not support strides in addition to dilation, because it uses one of the convolution backward passes. That's a possible disadvantage, but I haven't seen strided dilated convolution in a paper yet. On the other hand, we're the only ones using cuDNN for dilated convolution now :)

>  Or would these options always cancel out, so it would not make sense?

No, in theory we could have strided dilated convolution. The easiest solution will be extending im2col and col2im in GpuCorrMM and CorrMM: https://github.com/torch/nn/pull/797/commits/7e86ba4f2b687d8cfe1d34fc8770506610bb4f94#diff-9c7194a6647abb997cb04d4e37a864f1L6
This *could* even be done conditionally, so we compile a different version of the kernel for dilated convolution, to avoid the extra multiplications if they're unneeded (not sure how much of a difference in performance this would make).

Another reason for integrating dilation into im2col/col2im would be that the current implementation doesn't support padding, as explained in the PR description.

benanne:
Cool! In the long term it would be really nice to have a convolution API that supports arbitrary input and output strides (where input strides correspond to dilation).

f0k:
>  In the long term it would be really nice to have a convolution API that supports arbitrary [...]

That's what we've been discussing in Theano/Theano/issues/4481! Now somebody has to go and implement it :)

> input and output strides (where input strides correspond to dilation).

I think dilation is a better-fitting term: Traditional strided convolution is equivalent to subsampling the output (discarding elements), but here we're not discarding anything. Fractionally-strided convolution is somehow supersampling the output, but I already think that's a stretch. Filter dilation and input dilation (for "fractionally-strided convolution") explain things quite nicely, or am I missing something?

benanne:
I find the input/output stride terminology the clearest. The stride is literally the size of the step between successive computations. The dilation is the size of the step taken w.r.t. the input, and what is traditionally called the stride is the size of the step taken w.r.t. the output.

Note also that when you compute the gradient of a convolution operation, the input/output strides will switch places. (Which is why we can implement a convolution with arbitrary input stride and output stride fixed to 1 by using the gradient of a convolution with arbitrary output stride and input stride fixed to 1.)

But considering that "stride" is now commonly used to refer to the output stride, using a different name like "dilation" might be the most pragmatic solution.

f0k:
> I find the input/output stride terminology the clearest. The stride is literally the size of the step between successive computations. The dilation is the size of the step taken w.r.t. the input, and what is traditionally called the stride is the size of the step taken w.r.t. the output.

How does transposed convolution fit into the picture then? "fractionally-strided convolution" pretends you could move the kernel by half a pixel, but that's not defined. Dilating the input is very clear about it: it's upsampling in a nail-of-beds fashion prior to the convolution. Or would this be the filter stride, because you're only using every n'th filter element per output pixel? Okay, I can relate to the stride terminology, but still think dilation is clearer, mainly because output stride (strided convolution) is so different from input stride (dilated convolution) and filter stride (transposed convolution).

madratman:
Hi. Thanks for this. Thought I'll play but can't import.
```
    from lasagne.layers import DilatedConv2DLayer as DilatedConv2DLayer
ImportError: cannot import name DilatedConv2DLayer
```
I am on `Lasagne==0.2.dev1`. Am I missing something?

benanne:
This was merged very recently, so you may have to update your installation of Lasagne.

madratman:
I did a `pip install --upgrade --user https://github.com/Lasagne/Lasagne/archive/master.zip`, the version is 0.2.dev1. Could you please confirm the version? Or should I do a [development install](http://lasagne.readthedocs.io/en/latest/user/installation.html#development-installation) instead (of bleeding edge)?

f0k:
Unfortunately we didn't update the version number since shortly after the 0.1 release... we never agreed on a schedule. Your `pip install --upgrade ...` should have done the trick, though. Maybe you need to add a `--force-reinstall` to enforce upgrading?

madratman:
hmm thanks I ll try that. I copy pasted the relevant code which works. Question (if you guys don't mind heh) - in the [paper](http://arxiv.org/pdf/1511.07122v3.pdf) they mention how they initialize the "context module" (the appendage with dilated layers). I don't exactly get what they mean by *t* in eqn 4 (bottom of page 3)
Does that simply mean to initialize all weights to 1 which is `lasagne.init.Constant(val=1.0)` or something else?

f0k:
So you're referring to this part:
> We found an alternative initialization with clear semantics to be much more effective:
> k^b(t; a) = 1_[t=0] 1_[a=b]
> where a is the index of the input feature map and b is the index of the output map. This is a form
of identity initialization, which has recently been advocated for recurrent networks (Le et al., 2015).
This initialization sets all filters such that each layer simply passes the input directly to the next.

The 1_[...] parts are indicator functions, they are 1 if and only if the condition (in square brackets) is true. So they initialize the weights with mostly zeros, but ones where the input channel index equals the output channel index *and* the position in the filter is exactly in the center (or maybe in the corner, you'll have to check the definition for t further above -- the main idea is to have a single 1 in the filter so the filter copies the input verbatim, as they note two sentences later).

madratman:
thanks! got it.

f0k:
We could add `lasagne.init.Identity` for this (both for dense and convolutional layers), with a second initializer for the case num_out > num_in.

madratman:
exactly. I was thinking of filing a pr in a couple of days, sorry travelling then

benanne:
That would be most welcome, thanks :)

PR #679, opened 2016-05-11
Update RReLU tests for updated T.nnet.relu function

f0k:
Two of the RandomizedRectifierLayer tests used the fact that T.nnet.relu() worked both with symbolic expressions and with numpy arrays. With Theano/Theano/pull/4484, it always returns a symbolic expression. This updates the tests in Lasagne to work with Theano versions both before and after this change.


benanne:
LGTM, merging

PR #682, opened 2016-05-13
Added sparse submodule and test for sparse input layers

hmendozap:
As mentioned on #596 that solves issue #317. I took the code from @eduardo4jesus and made some modifications:
 - all one file, as I think is more clean and we are reusing the dense and noise layers
- added tests to improve coverall

I also have doubts about implementation, maybe a quick discussion, so we can at least close this small change.

- Are the names of the layers clear enough? The idea is to have sparse input, but dense (in the sparse matrices sense) output in a fully connected layer. In the dropout layer the only real modification is regarding the rescaling.

- Is ok if it checks against Sparse Theano types? Or there is some standard for the project?

- Are the bounds for the ```test_get_output_for_non_deterministic``` correct, I have an intuition about them, but I'm not sure that they are tight enough.

I only tested against CSR sparse format. As I believe it's best suited for this [task](http://docs.scipy.org/doc/scipy-0.17.0/reference/sparse.html#matrix-vector-product).

Thanks in advanced,
HM

### DIFF
--- a/lasagne/layers/sparse.py
+++ b/lasagne/layers/sparse.py
@@ -1,0 +1,46 @@
+import theano.tensor as T
+import theano.sparse as S
+from .dense import DenseLayer
+from .noise import DropoutLayer
+
+
+__all__ = [
+    "SparseInputDenseLayer",
+    "SparseInputDropoutLayer"
+]
+
+
+class SparseInputDenseLayer(DenseLayer):
+    def get_output_for(self, input, **kwargs):
+        if not isinstance(input, (S.SparseVariable, S.SparseConstant,
+                                  S.sharedvar.SparseTensorSharedVariable)):
+            raise ValueError("Input for this layer must be sparse")
+
+        activation = S.dot(input, self.W)
+        if self.b is not None:
+            activation = activation + self.b.dimshuffle('x', 0)
+        return self.nonlinearity(activation)
+
+
+class SparseInputDropoutLayer(DropoutLayer):
+    def get_output_for(self, input, deterministic=False, **kwargs):
+        if not isinstance(input, (S.SparseVariable, S.SparseConstant,
+                                  S.sharedvar.SparseTensorSharedVariable)):
+            raise ValueError("Input for this layer must be sparse")
+
+        if deterministic or self.p == 0:
+            return input
+        else:
+            # Using Theano constant to prevent upcasting
+            one = T.constant(1, name='one')
+            retain_prob = one - self.p
+
+            if self.rescale:
+                input = S.mul(input, one/retain_prob)
+
+            input_shape = self.input_shape
+            if any(s is None for s in input_shape):
+                input_shape = input.shape
+
+            return input * self._srng.binomial(input_shape, p=retain_prob,
+                                               dtype=input.dtype)
--- a/lasagne/tests/layers/test_sparse.py
+++ b/lasagne/tests/layers/test_sparse.py
@@ -1,0 +1,95 @@
+from mock import Mock
+import numpy as np
+import pytest
+import theano.sparse as S
+import scipy.sparse as sp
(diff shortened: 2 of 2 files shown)


stablum:
are the gradients of the weight matrix calculated only on the basis of the observed features or are they calculated considering them as being set to 0?
In other words, a single update from a partially filled datapoint is going to affect all weights or only the weights that are connected to the filled entries in the input?

stablum:
@hmendozap , @f0k : I have another question : as a I was trying to use the module proposed in this pull request, I noticed something about the naming: `SparseInputDenseLayer` and `SparseInputDropoutLayer` : why include the word *Input* in the class name? This makes me think that they are related to `InputLayer`, but they are not. Was that a typo?

hmendozap:
Hi @stablum,

it's the second one (missing entries are densified, in this case filled with zeros), so that answers the next questions, an sparsed input affects all the weights.

afshinrahimi:
I'm using the sparse module, the code works for a small dataset and also works for a large dataset where deterministic = True (no dropout). But when I use dropout during training I get this error:

```
  File "tensormain.py", line 311, in <module>
    regularization=args.regularization)
  File "tensormain.py", line 215, in main_mlpconv
    clf.fit(X_train, Y_train, X_dev, Y_dev, X_test, Y_test, H)
  File "/home/user1/git/localism/mlpconv.py", line 279, in fit
    l_train, acc_train = self.f_train(x_batch, y_batch, self.train_indices)
  File "/home/user1/.local/lib/python2.7/site-packages/theano/compile/function_module.py", line 871, in __call__
    storage_map=getattr(self.fn, 'storage_map', None))
  File "/home/user1/.local/lib/python2.7/site-packages/theano/gof/link.py", line 314, in raise_with_op
    reraise(exc_type, exc_value, exc_trace)
  File "/home/user1/.local/lib/python2.7/site-packages/theano/compile/function_module.py", line 859, in __call__
    outputs = self.fn()
**ValueError: rng_mrg cpu-implementation does not support more than (2**31 -1) samples**
Apply node that caused the error: mrg_uniform{TensorType(float32, matrix),inplace}(<TensorType(int32, matrix)>, MakeVector{dtype='int64'}.0)
Toposort index: 22
Inputs types: [TensorType(int32, matrix), TensorType(int64, vector)]
Inputs shapes: [(15360, 6), (2,)]
Inputs strides: [(24, 4), (8,)]
Inputs values: ['not shown', array([449200,  95085])]
Outputs clients: [['output'], [Elemwise{Composite{Cast{float32}(LT(i0, i1))}}[(0, 0)](mrg_uniform{TensorType(float32, matrix),inplace}.1, TensorConstant{(1, 1) of 0.5})]]
```

Note that I'm using a full batch and so when the dropout matrix is being built it becomes a huge matrix. Is there a more efficient way of creating the dropout matrix rather than binomial function?

nouiz:
This is a limitation of the random number implementation that can't make
more then 2**31-1 samples per call. You can lower the batch size to request
less samples at a time or use the numpy random number generator instead.
Maybe it don't have this restriction. But I'm not sure of that.

f0k:
> Note that I'm using a full batch and so when the dropout matrix is being built it becomes a huge matrix. Is there a more efficient way of creating the dropout matrix rather than the _srng.binomial function?

You can also try batchwise dropout (using the same dropout mask for all items in a minibatch). You'd just need to pass `shared_axes=0` to the `DropoutLayer` constructor. (Alternatively, batchwise dropout can also be implemented by subsampling the weight matrices, using only the rows and columns of the non-dropped input and output units, respectively: https://arxiv.org/abs/1502.02478.)

afshinrahimi:
Thanks f0k (and nouiz), I'll try the suggestions.

PR #683, opened 2016-05-16
Clarifying comment for iterate_minibatches in mnist example

mehdidc:
Fix iterate_minibatches function in case nb of inputs < batchsize, if nb inputs < batchsize it does not iterate through available inputs, I know it does not make sense to have nb inputs < batchsize but it should do the correct thing if it is.

### DIFF
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -209,6 +209,9 @@ def build_cnn(input_var=None):
 # own custom data iteration function. For small datasets, you can also copy
 # them to GPU at once for slightly improved performance. This would involve
 # several changes in the main program, though, and is not demonstrated here.
+# Notice that this function returns only mini-batches of size `batchsize`.
+# If the size of the data is not a multiple of `batchsize`, it will not
+# return the last (remaining) mini-batch.
 
 def iterate_minibatches(inputs, targets, batchsize, shuffle=False):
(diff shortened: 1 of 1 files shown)


f0k:
The promise of the `iterate_minibatches` function is to always return batches of `batchsize` items, and no half-filled batches. Your proposal would drop the last (half-filled) batch for datasets that have more than `batchsize` items, but return it for all smaller datasets. That's inconsistent. The current function always drops the last half-filled batch, even if it would have been the only one.
If anything, we could change the code to return half-filled batches for validation/testing, but for the MNIST example we've tailored the batch size to the size of the validation and test set to avoid this case altogether.
I'm leaning towards leaving such modifications to users basing their code on the MNIST example.

benanne:
> I'm leaning towards leaving such modifications to users basing their code on the MNIST example.

Agreed, we could maybe add some comments to clarify what's happening though, so people can decide for themselves whether they need to modify this behaviour or not.

mehdidc:
Thanks for the feedback, @f0k and @benanne, Ok I see, actually my proposal should have been  'for start_idx in range(0, len(inputs), batchsize)', anyway, indeed it would be nice indeed  to add some comments explaining what is happening.

f0k:
Perfect, even squashed into a single commit already. Merging, thank you!

Issue #693, opened 2016-06-06
Bridge PyMC3 and Lasagne


twiecki:
I've been working a bit on Bayesian Neural Networks in PyMC3 (here is the blog post: http://twiecki.github.io/blog/2016/06/01/bayesian-deep-learning/). I wonder if it would be possible to use Lasagne for construction of the layers. The only requirement would be to place priors (theano expressions) on the weight parameters and feed the output into a likelihood function. Not sure if and how easy it would be to adapt Lasagne in that way.

f0k:
Judging from the first code snippet in that blog post, as you just pass `weights_in_1` to `T.dot`, you can just as well pass it for the weights when constructing a Lasagne layer (`DenseLayer(..., W=weights_in_1)`). To obtain the output expression to pass to `pm.Bernoulli`, you'd then call `lasagne.layers.get_output(...)` for your output layer. As far as I see, there's nothing you'd need to change!
We'd welcome a PR to [Lasagne/Recipes](/Lasagne/Recipes) if you have a working example.

twiecki:
@f0k Perfect, thanks for the quick response. I'll give that a try and report back.

twiecki:
Does not seem to like theano expressions:
```python
    act_in = lasagne.layers.InputLayer(X_train.shape, input_var=ann_input)
    act_1 = lasagne.layers.DenseLayer(act_in, n_hidden, W=weights_in_1, nonlinearity=lasagne.nonlinearities.tanh)
    act_2 = lasagne.layers.DenseLayer(act_1, n_hidden, W=weights_1_2, nonlinearity=lasagne.nonlinearities.tanh)
    act_out = lasagne.layers.DenseLayer(act_1, n_hidden, W=weights_2_out, nonlinearity=lasagne.nonlinearities.sigmoid)
    net_out = lasagne.layers.get_output(act_out)
```

```
RuntimeError                              Traceback (most recent call last)
<ipython-input-17-519be30e2063> in <module>()
     32     # Build neural-network using tanh activation function
     33     act_in = lasagne.layers.InputLayer(X_train.shape, input_var=ann_input)
---> 34     act_1 = lasagne.layers.DenseLayer(act_in, n_hidden, W=weights_in_1, nonlinearity=lasagne.nonlinearities.tanh)
     35     act_2 = lasagne.layers.DenseLayer(act_1, n_hidden, W=weights_1_2, nonlinearity=lasagne.nonlinearities.tanh)
[...]

PR #695, opened 2016-06-09
Make lasagne.utils.create_param() get callable that returns Theano expressions or shared variables

erfannoury:
This issues comments raised in https://github.com/Lasagne/Lasagne/issues/693

### DIFF
--- a/lasagne/utils.py
+++ b/lasagne/utils.py
@@ -289,45 +289,66 @@ def create_param(spec, shape, name=None):
 
     Parameters
     ----------
-    spec : numpy array, Theano expression, or callable
+    spec : scalar number, numpy array, Theano expression, or callable
         Either of the following:
 
-        * a numpy array with the initial parameter values
+        * a scalar or a numpy array with the initial parameter values
         * a Theano expression or shared variable representing the parameters
         * a function or callable that takes the desired shape of
           the parameter array as its single argument and returns
-          a numpy array.
+          a numpy array, a Theano expression, or a shared variable
+          representing the parameters.
 
     shape : iterable of int
         a tuple or other iterable of integers representing the desired
         shape of the parameter array.
 
     name : string, optional
-        If a new variable is created, the name to give to the parameter
-        variable. This is ignored if `spec` is already a Theano expression
-        or shared variable.
+        The name to give to the parameter variable. Ignored if `spec`
+        is or returns a Theano expression or shared variable that
+        already has a name.
+
 
     Returns
     -------
     Theano shared variable or Theano expression
         A Theano shared variable or expression representing layer parameters.
-        If a numpy array was provided, a shared variable is initialized to
-        contain this array. If a shared variable or expression was provided,
-        it is simply returned. If a callable was provided, it is called, and
-        its output is used to initialize a shared variable.
+        If a scalar or a numpy array was provided, a shared variable is
+        initialized to contain this array. If a shared variable or expression
+        was provided, it is simply returned. If a callable was provided, it is
+        called, and its output is used to initialize a shared variable.
 
     Notes
     -----
     This function is called by :meth:`Layer.add_param()` in the constructor
     of most :class:`Layer` subclasses. This enables those layers to
-    support initialization with numpy arrays, existing Theano shared variables
-    or expressions, and callables for generating initial parameter values.
+    support initialization with scalars, numpy arrays, existing Theano shared
+    variables or expressions, and callables for generating initial parameter
+    values, Theano expressions, or shared variables.
     """
+    import numbers  # to check if argument is a number
     shape = tuple(shape)  # convert to tuple if needed
     if any(d <= 0 for d in shape):
         raise ValueError((
             "Cannot create param with a non-positive shape dimension. "
             "Tried to create param with shape=%r, name=%r") % (shape, name))
+
+    err_prefix = "cannot initialize parameter %s: " % name
+    if callable(spec):
+        spec = spec(shape)
+        err_prefix += "the %s returned by the provided callable"
+    else:
+        err_prefix += "the provided %s"
+
+    if isinstance(spec, numbers.Number) or isinstance(spec, np.generic) \
+            and spec.dtype.kind in 'biufc':
+        spec = np.asarray(spec)
+
+    if isinstance(spec, np.ndarray):
+        if spec.shape != shape:
+            raise ValueError("%s has shape %s, should be %s" %
+                             (err_prefix % "numpy array", spec.shape, shape))
+        spec = theano.shared(spec)
 
     if isinstance(spec, theano.Variable):
         # We cannot check the shape here, Theano expressions (even shared
@@ -337,34 +358,20 @@ def create_param(spec, shape, name=None):
         # `name` attribute of the variable, but the user may have already
         # named the variable and we don't want to override this.
         if spec.ndim != len(shape):
-            raise RuntimeError("parameter variable has %d dimensions, "
-                               "should be %d" % (spec.ndim, len(shape)))
+            raise ValueError("%s has %d dimensions, should be %d" %
+                             (err_prefix % "Theano variable", spec.ndim,
+                              len(shape)))
+        if not spec.name:
+            spec.name = name
         return spec
 
-    elif isinstance(spec, np.ndarray):
-        if spec.shape != shape:
-            raise RuntimeError("parameter array has shape %s, should be "
-                               "%s" % (spec.shape, shape))
-        return theano.shared(spec, name=name)
-
-    elif hasattr(spec, '__call__'):
-        arr = spec(shape)
-        try:
-            arr = floatX(arr)
-        except Exception:
-            raise RuntimeError("cannot initialize parameters: the "
-                               "provided callable did not return an "
-                               "array-like value")
-        if arr.shape != shape:
-            raise RuntimeError("cannot initialize parameters: the "
-                               "provided callable did not return a value "
-                               "with the correct shape")
-        return theano.shared(arr, name=name)
-
     else:
-        raise RuntimeError("cannot initialize parameters: 'spec' is not "
-                           "a numpy array, a Theano expression, or a "
-                           "callable")
+        if "callable" in err_prefix:
+            raise TypeError("%s is not a numpy array or a Theano expression" %
+                            (err_prefix % "value"))
+        else:
+            raise TypeError("%s is not a numpy array, a Theano expression, "
+                            "or a callable" % (err_prefix % "spec"))
 
 
 def unroll_scan(fn, sequences, outputs_info, non_sequences, n_steps,
--- a/lasagne/tests/test_utils.py
+++ b/lasagne/tests/test_utils.py
@@ -137,16 +137,16 @@ def test_compute_norms_ndim6_raises():
 def test_create_param_bad_callable_raises():
     from lasagne.utils import create_param
 
-    with pytest.raises(RuntimeError):
+    with pytest.raises(TypeError):
         create_param(lambda x: {}, (1, 2, 3))
-    with pytest.raises(RuntimeError):
+    with pytest.raises(ValueError):
         create_param(lambda x: np.array(1), (1, 2, 3))
 
 
 def test_create_param_bad_spec_raises():
     from lasagne.utils import create_param
 
-    with pytest.raises(RuntimeError):
+    with pytest.raises(TypeError):
         create_param({}, (1, 2, 3))
 
 
@@ -161,7 +161,7 @@ def test_create_param_numpy_bad_shape_raises_error():
     from lasagne.utils import create_param
 
     param = np.array([[1, 2, 3], [4, 5, 6]])
-    with pytest.raises(RuntimeError):
+    with pytest.raises(ValueError):
         create_param(param, (3, 2))
 
 
@@ -187,7 +187,7 @@ def test_create_param_shared_bad_ndim_raises_error():
     from lasagne.utils import create_param
 
     param = theano.shared(np.array([[1, 2, 3], [4, 5, 6]]))
-    with pytest.raises(RuntimeError):
+    with pytest.raises(ValueError):
         create_param(param, (2, 3, 4))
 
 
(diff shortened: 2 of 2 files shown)


erfannoury:
BTW, I have a question about naming in the PR that isn't yet resolved. What should I do with the name parameter if the callable method returns a shared variable or a Theano expression?

erfannoury:
I have no idea why there are too many failures 😕

benanne:
There seem to be a bunch of failures, and not just in tests for `create_param`, so that's odd. Have you run the tests locally?

Regarding the name parameter: our assumption is that, when a shared variable is passed in, it may or may not already have been named. If it already has a name, we do not want to overwrite it, so we do nothing. We could technically check if the specified variable has a name, and then insert a name if it doesn't have one, but that would get a bit hairy, I think. Also it wouldn't work for expressions.

erfannoury:
I have only run the test locally for the parts of the code that I have changed. Maybe the failures are affected by the update to the `create_param`. I will re-check everything now.

Yes, it's written in the comments. My question was specifically for the case that a callable has been passed to the `create_param` which returns a Theano expression or a shared variables. In this case, should we ignore the name or set it? I think checking for the presence of name is a good solution.

erfannoury:
@benanne One particular thing I encountered when running tests locally vs. Travis CI is that some tests won't pass when Theano is configured to use GPU. What should be done with these test? Some of the tests that previously were in the repository (not added by this PR) are prone to this problem, too.

erfannoury:
@benanne I wonder how these two tests passed before:
* `lasagne/tests/layers/test_special.py::TestParametricRectifierLayer::test_alpha_init`
* `lasagne/tests/layers/test_special.py::TestParametricRectifierLayer::test_get_output_for`

In both tests, a scalar is passed to `create_param`, however I think (and I even ran the abovementioned tests using the original source code) these two tests should fail anyway. What am I missing here?

Now the build failures are down to these two and a doctest error on the EmbeddingLayer (`lasagne/layers/embedding.py: lasagne.layers.embedding.EmbeddingLayer`)

f0k:
> What am I missing here?

You changed the behaviour of `create_param()`. Before, when it got a callable, it called the callable and converted the result with `lasagne.utils.floatX`. Now you call the callable and run the result through the `if` cases, which don't grip for Python scalars. You'll have to keep your modifications to the original third `if` case or change the function to also accept Python scalars. Something like this:
```python
if spec is callable:
    spec = spec(shape)
if spec is Python scalar:
    spec = floatX(spec)
if spec is numpy array:
    check for correct shape
    spec = theano.shared(spec)
if spec is theano variable:
    check for correct dimensionality
    add name if unset
else:
    complain that the spec is unsupported
```
Note that the `if` cases are not `elif` cases.

Instead of setting a boolean flag to spit out the correct error message in the end, you could gradually build up a prefix string for the error message inside the if cases ("the numpy array", "the Theano variable", "the numpy array returned by the callable", ...).

erfannoury:
Ok, @f0k thanks for the suggestions. All tests pass now except the `EmbeddingLayer`'s doctest, which I'm looking into now. I'll try updating the way error messages are handled, too.

erfannoury:
I can't reproduce the doctest failure in `EmbeddingLayer` in either CPU or GPU mode on my computer.

lasagne/utils.py
```diff
@@ -321,14 +322,31 @@ def create_param(spec, shape, name=None):
...
         raise ValueError((
             "Cannot create param with a non-positive shape dimension. "
             "Tried to create param with shape=%r, name=%r") % (shape, name))
 
+    callable_prefix = ""
+    if hasattr(spec, '__call__'):
+        spec = spec(shape)
+        callable_prefix = "cannot initialize parameters: the provided " \
+                          "callable did not return a value with the " \
+                          "correct shape or dimensions."
```
f0k: Just to make the errors a little nicer:
```python
err_prefix = "cannot initialize parameter %s: " % name
if callable(spec):
    spec = spec(shape)
    err_prefix += "the %s returned by the provided callable"
else:
    err_prefix += "the provided %s"
...
if isinstance(spec, np.ndarray):
    if spec.shape != shape:
        raise ValueError("%s has shape %s, should be %s") % (err_prefix % "numpy array", spec.shape, shape)
    ...

if isinstance(spec, theano.Variable):
    if spec.ndim != len(shape):
        raise ValueError("%s has %d dimensions, should be %d" % (err_prefix % "Theano variable", spec.ndim, len(shape))
else:
    if "callable" in err_prefix:
        raise ValueError("%s is not a numpy array or a Theano expression" % (err_prefix % "value"))
    else:
        raise ValueError("%s is not a numpy array, a Theano expression, or a callable" % (err_prefix % "value"))
```

erfannoury:
Changed the error handling to what @f0k suggested. Tests are also updated accordingly and all pass. Only the `EmbeddingLayer`'s doctest fails, however I can't reproduce it locally.

lasagne/utils.py
```diff
@@ -321,14 +322,31 @@ def create_param(spec, shape, name=None):
...
+    Theano expressions, or shared variables.
     """
     shape = tuple(shape)  # convert to tuple if needed
     if any(d <= 0 for d in shape):
         raise ValueError((
             "Cannot create param with a non-positive shape dimension. "
             "Tried to create param with shape=%r, name=%r") % (shape, name))
 
+    err_prefix = "cannot initialize parameter %s: " % name
+    if hasattr(spec, '__call__'):
```
f0k: I'd prefer `if callable(spec)`.

lasagne/utils.py
```diff
@@ -337,105 +355,91 @@ def create_param(spec, shape, name=None):
...
-            raise RuntimeError("cannot initialize parameters: the "
-                               "provided callable did not return a value "
-                               "with the correct shape")
-        return theano.shared(arr, name=name)
-
     else:
-        raise RuntimeError("cannot initialize parameters: 'spec' is not "
-                           "a numpy array, a Theano expression, or a "
-                           "callable")
+        if hasattr(spec, '__call__'):
```
f0k: This is not going to work, `spec` has been replaced by then. As in my proposal, it should be `if "callable" in err_prefix`. It's a bit ugly, but otherwise the error message is less precise.

lasagne/utils.py
```diff
@@ -337,105 +355,91 @@ def create_param(spec, shape, name=None):
...
+            raise TypeError("%s is not a numpy array or a Theano expression" %
+                            (err_prefix % "value"))
+        else:
+            raise TypeError("%s is not a numpy array, a Theano expression, "
+                            "or a callable" % (err_prefix % "spec"))
 
 
 def unroll_scan(fn, sequences, outputs_info, non_sequences, n_steps,
                 go_backwards=False):
-        """
```
f0k: Interesting that nobody noticed this before. But anyway, this shouldn't be part of your PR! Please leave the indentation for `unroll_scan` as it was.

lasagne/utils.py
```diff
@@ -337,105 +355,91 @@ def create_param(spec, shape, name=None):
...
-                               "with the correct shape")
-        return theano.shared(arr, name=name)
-
     else:
-        raise RuntimeError("cannot initialize parameters: 'spec' is not "
-                           "a numpy array, a Theano expression, or a "
-                           "callable")
+        if hasattr(spec, '__call__'):
+            raise TypeError("%s is not a numpy array or a Theano expression" %
+                            (err_prefix % "value"))
```
f0k: This exception is missing in the tests now. We'll need a `test_create_param_callable_returns_wrong` with a callable that returns a string, for example.

f0k:
> I can't reproduce the doctest failure in `EmbeddingLayer` in either CPU or GPU mode on my computer.

This is still a mystery. Does the new `create_param` change the dtype of a plain numpy array (i.e., is it passed through floatX)? Did you try running the test locally with `THEANO_FLAGS=device=cpu,floatX=float64`?

lasagne/utils.py
```diff
@@ -321,14 +322,31 @@ def create_param(spec, shape, name=None):
...
 
+    err_prefix = "cannot initialize parameter %s: " % name
+    if hasattr(spec, '__call__'):
+        spec = spec(shape)
+        err_prefix += "the %s returned by the provided callable"
+    else:
+        err_prefix += "the provided %s"
+
+    if np.isscalar(spec):
+        spec = floatX(spec)
```
f0k: `np.isscalar("asdf")` returns True. We need to explicitly check for numbers.

lasagne/utils.py
```diff
@@ -321,14 +322,31 @@ def create_param(spec, shape, name=None):
...
+
+    if isinstance(spec, np.ndarray):
+        if spec.shape != shape:
+            raise ValueError("%s has shape %s, should be %s" %
+                             (err_prefix % "numpy array", spec.shape, shape))
+        spec = theano.shared(floatX(spec))
```
f0k: Ah, you sneaked in a `floatX` here. This wasn't present before in the codepath for an ndarray. That's why the doctest for the EmbeddingLayer fails. We should have a more direct test that checks whether the dtype of an ndarray is retained in `create_param`!

f0k: Checking the original code again, there is a conversion to `floatX` for values returned by a callable, but not for values directly passed as a numpy array. This probably wasn't originally intended, but I don't know if any code relies on this. As far as I see, all initializers provided by Lasagne produce values of the correct dtype, but third-party code might not.

lasagne/utils.py
erfannoury: Yes you're right. I've messed it up here.

lasagne/utils.py
erfannoury: Ok. I'll change this, too.

lasagne/utils.py
erfannoury: So if the code has to retain the original type of the input ndarray, then the code should change a bit. I'll add a test to be sure to retain the type in this case.

erfannoury:
All tests are passing now. @f0k could you please review the code once more?

lasagne/utils.py
```diff
@@ -321,14 +322,35 @@ def create_param(spec, shape, name=None):
...
 
+    err_prefix = "cannot initialize parameter %s: " % name
+    if callable(spec):
+        spec = spec(shape)
+        err_prefix += "the %s returned by the provided callable"
+    else:
+        err_prefix += "the provided %s"
+
+    if isinstance(spec, numbers.Number):
+        spec = np.asarray(spec)
```
f0k: Nice find! This seems slightly too strict, though (I'm sorry that this PR turns out to be much more intricate than it seemed at first glance):
```
>>> x = np.float32(123)
>>> isinstance(spec, numbers.Number)
False
```
Maybe:
```python
if isinstance(spec, numbers.Number) or isinstance(spec, np.generic) and spec.dtype.kind in 'biufc':
```
Although this is getting silly. Let's leave it at what you have.

lasagne/utils.py
```diff
@@ -321,14 +322,35 @@ def create_param(spec, shape, name=None):
...
+
+    if isinstance(spec, np.ndarray):
+        if spec.shape != shape:
+            raise ValueError("%s has shape %s, should be %s" %
+                             (err_prefix % "numpy array", spec.shape, shape))
+        if "callable" in err_prefix:
+            spec = floatX(spec)
```
f0k: Yes, this restores original behaviour. I'm leaning towards breaking compatibility, though, and announcing it on the mailing list.
@benanne: Would you be fine with *not* converting results of callables to `floatX`? We never did this for numpy arrays passed as initial values, just for numpy arrays returned by callables passed as initial values. This is inconsistent.

lasagne/utils.py
erfannoury: > (I'm sorry that this PR turns out to be much more intricate than it seemed at first glance)

No, it's ok. Sorry that I have been slow in implementing the correct code. This is my first PR for Lasagne 😄 

> ```
> >>> x = np.float32(123)
> >>> isinstance(spec, numbers.Number)
> False
> ```

@f0k I think this code isn't completely correct. `numbers.Number` in fact correctly handles the case you have provided. 
```python
>>> x = np.float32(123)
>>> isinstance(x, numbers.Number)
True
```

As far as I tested multiple cases, `numbers.Number` correctly handles scalars of multiple types and kinds.

lasagne/utils.py
erfannoury: Ok, I'll wait for the final decision.

lasagne/utils.py
f0k: > I think this code isn't completely correct. `numbers.Number` in fact correctly handles the case you have provided.

Interesting. Not for me on Python 2.7.6 or 3.4.2, numpy 1.8.2.

erfannoury: You've passed `spec` as argument to `isinstance`, instead of `x`. I'm working on Python 2.7.11, NumPy 1.11.0.

f0k: No, I had just written it wrongly, I passed `x`. It depends on the numpy/Python version than. In any case, it indicates we don't need to do such a stretch here and keep it as you have it.

erfannoury: But it won't work correctly in those combinations of NumPy/Python.
So I should keep them as is?

f0k: > But it won't work correctly in those combinations of NumPy/Python.

It's a corner case. Using scalars was not actually documented, it just happened to be used in one of the tests. Using numpy scalars instead of Python scalars is even more undocumented.

> So I should keep them as is?

You can also add my modified `if` statement, as you please. Either way is fine by me!

erfannoury: Yes, I think that's a more robust solution. I'll update the if statement.

lasagne/utils.py
benanne: Yeah, I think it would be fine to change this.

erfannoury: @f0k done.

lasagne/utils.py
```diff
@@ -289,14 +289,15 @@ def create_param(spec, shape, name=None):
 
     Parameters
     ----------
-    spec : numpy array, Theano expression, or callable
+    spec : scalar number, numpy array, Theano expression, or callable
         Either of the following:
 
-        * a numpy array with the initial parameter values
+        * an scalar or a numpy array with the initial parameter values
```
f0k: *a scalar

lasagne/utils.py
```diff
@@ -305,30 +306,50 @@ def create_param(spec, shape, name=None):
     name : string, optional
         If a new variable is created, the name to give to the parameter
         variable. This is ignored if `spec` is already a Theano expression
-        or shared variable.
+        or shared variable and which already has a name.
```
f0k: Needs to be changed a bit because it doesn't only apply when a new variable is created now:
```
The name to give to the parameter variable. Ignored if `spec` is or returns a Theano expression or shared variable that already has a name.
```

lasagne/utils.py
```diff
@@ -305,30 +306,50 @@ def create_param(spec, shape, name=None):
...
 
     Returns
     -------
     Theano shared variable or Theano expression
         A Theano shared variable or expression representing layer parameters.
-        If a numpy array was provided, a shared variable is initialized to
-        contain this array. If a shared variable or expression was provided,
-        it is simply returned. If a callable was provided, it is called, and
-        its output is used to initialize a shared variable.
+        If an scalar or a numpy array was provided, a shared variable is
```
f0k: *a scalar

lasagne/utils.py
```diff
@@ -305,30 +306,50 @@ def create_param(spec, shape, name=None):
...
     This function is called by :meth:`Layer.add_param()` in the constructor
     of most :class:`Layer` subclasses. This enables those layers to
-    support initialization with numpy arrays, existing Theano shared variables
-    or expressions, and callables for generating initial parameter values.
+    support initialization with scalars, numpy arrays, existing Theano shared
+    variables or expressions, and callables for generating initial parameter
+    values, Theano expressions, or shared variables.
     """
+    # to check if argument is a number
+    import numbers
```
f0k: If you're fixing the spelling above, could you make this a line comment?
```python
import numbers  # to check if argument is a scalar number
```
Otherwise it looks as if this comment applies to the whole following code block.

f0k:
Very cool! Just three minor requests. In the end, could you please squash this into a single commit? It should go like this:
```
git reset d5fbd64  # set HEAD to the first commit of this PR
git commit -a --amend -m 'Rewrite create_param() to allow callables returning Theano variables'  # amend all other changes to that commit
git push --force origin HEAD  # overwrite PR with the new clean history
```

erfannoury:
@f0k squashed as you said, and all tests are passing.

f0k:
Great, thanks a lot, everything looks fine! Will merge in the evening if there are no further comments.

f0k:
Merged, thanks again!

twiecki:
:+1:

Issue #411, opened 2015-09-13
Docker images


Kaixhin:
I've created [CPU-backend](https://hub.docker.com/r/kaixhin/lasagne/) and [CUDA-backend](https://hub.docker.com/r/kaixhin/cuda-lasagne/) (v6.5, v7.0 and v7.5) Docker images for Lasagne. I've checked the `mnist` and `recurrent` examples on both of these, so they seem to run fine. You may wish to add these to the [installation docs](https://lasagne.readthedocs.org/en/latest/user/installation.html) as an alternative (like [Torch7](https://github.com/torch/torch7/wiki/Cheatsheet#docker-images) or [neon](http://neon.nervanasys.com/docs/0.9.0/installation.html#docker-images)).

benanne:
Sounds awesome! I am completely unfamiliar with Docker, could someone with some more knowledge about this stuff comment as well? From what I can see it would be a very useful addition.

f0k:
I'm totally unfamiliar with this, too. Would you like to send a PR updating the installation docs to link to your images? I'd suggest to just add an additional section at the very end, titled "Docker images", with something like "Instead of manually installing Theano and Lasagne on your machine as described above, you may want to try a pre-made Docker image: [Lasagne Docker images (CPU)](https://hub.docker.com/r/kaixhin/lasagne/), [Lasagne Docker images (CUDA)](https://hub.docker.com/r/kaixhin/cuda-lasagne/)"
Feel free to add one or two sentences about Docker, of course, or a link to a good introduction.

> I've checked the `mnist` and `recurrent` examples on both of these

Bonus points if you can also run the test suite! The basic explanations are in the documentation ([install requirements-dev.txt](https://lasagne.readthedocs.org/en/latest/user/development.html#development-setup) and [run py.test](https://lasagne.readthedocs.org/en/latest/user/development.html#testing)), let us know if you need any help with that.

PR #697, opened 2016-06-10
Update Docker instructions

Kaixhin:
The CUDA Docker images are moving to use the NVIDIA Docker program, so I've updated the docs appropriately.

### DIFF
--- a/docs/user/installation.rst
+++ b/docs/user/installation.rst
@@ -228,11 +228,11 @@ To check whether it is found by Theano, run the following command:
 It will print ``True`` if everything is fine, or an error message otherwise.
 There are no additional steps required for Theano to make use of cuDNN.
 
-Docker images
-=============
+Docker
+======
 
-Instead of manually installing Theano and Lasagne on your machines as described
-above, you may want to use a pre-made `Docker <https://www.docker.com/whatisdocker>`_
+Instead of manually installing Theano and Lasagne on your machines as described above,
+you may want to use a pre-made `Docker <https://www.docker.com/what-docker>`_
 image: `Lasagne Docker (CPU) <https://hub.docker.com/r/kaixhin/lasagne/>`_ or
 `Lasagne Docker (CUDA) <https://hub.docker.com/r/kaixhin/cuda-lasagne/>`_. These
 are updated on a weekly basis with bleeding-edge builds of Theano and Lasagne.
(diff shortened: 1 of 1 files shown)


nouiz:
Just a note that if you want GPU docker with Theano, there is some there:

https://github.com/nouiz/Theano-Docker

docs/user/installation.rst
```diff
@@ -228,21 +228,21 @@ To check whether it is found by Theano, run the following command:
...
-Docker images
-=============
+Docker
+======
 
-Instead of manually installing Theano and Lasagne on your machines as described
-above, you may want to use a pre-made `Docker <https://www.docker.com/whatisdocker>`_
-image: `Lasagne Docker (CPU) <https://hub.docker.com/r/kaixhin/lasagne/>`_ or
+Builds of Lasagne are available as `Docker <https://www.docker.com>`_
+images: `Lasagne Docker (CPU) <https://hub.docker.com/r/kaixhin/lasagne/>`_ or
```
f0k: Why this change? The previous sentence and section title made explicit that this is an alternative way of installing/using Theano and Lasagne.

Kaixhin: With other docs I wrote in this more formal way (see Theano for example). Now that you mention it the reason must be that I literally took your wording from https://github.com/Lasagne/Lasagne/issues/411. Would you like to have the wording changed back or not? I can just amend this commit to keep it tidy.

f0k: > Now that you mention it the reason must be that I literally took your wording from #411.

Ah, that must be why I liked the original wording better ;)

> Would you like to have the wording changed back or not? I can just amend this commit to keep it tidy.

Yes, amending the comment to restore the original formulation would be appreciated! Also I think the direct link to "what is docker" is more useful for people unfamiliar with it than just linking to docker.com. The latter is easy to find from the former, but not vice versa!

Kaixhin: OK will do. In terms of the link, it's actually changed from https://www.docker.com/whatisdocker to https://www.docker.com/what-docker, so even though there's a redirect the URL seems like it could keep changing (unlike the top-level website). But I'll put in the new address as they will hopefully keep redirects in place.

f0k:
I think you forgot the `git add` before `git commit --amend` and `git push --force`.

f0k:
Thank you, merging!

benanne:
Thanks :)

PR #701, opened 2016-06-15
SpatialDropoutLayer

gzuidhof:
Hi there,

I think a spatial dropout layer (which is also called feature dropout) would make a nice addition to Lasagne. It was already discussed half a year ago [here](https://groups.google.com/forum/#!topic/lasagne-users/qOV7i4SFUH8). 

This implementation is an adaptation of the standard dropout layer found in lasagne.

### DIFF
--- a/lasagne/layers/noise.py
+++ b/lasagne/layers/noise.py
@@ -90,6 +90,7 @@ class DropoutLayer(Layer):
 
 dropout = DropoutLayer  # shortcut
 
+
 class SpatialDropoutLayer(Layer):
     """Spatial dropout layer
     Sets whole filter activations to zero with probability p. See notes for
@@ -124,7 +125,7 @@ class SpatialDropoutLayer(Layer):
            2016.
     """
     def __init__(self, incoming, p=0.5, rescale=True, **kwargs):
-        super(DropoutLayer, self).__init__(incoming, **kwargs)
+        super(SpatialDropoutLayer, self).__init__(incoming, **kwargs)
         self._srng = RandomStreams(get_rng().randint(1, 2147462579))
         self.p = p
         self.rescale = rescale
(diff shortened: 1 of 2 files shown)


f0k:
Thank you! I'd rather extend the `DropoutLayer` with a `shared_axes` argument (defaulting to the empty tuple) so you can also use it to dropout rows or columns (or any combination of that). Would you be willing to do this instead? Otherwise I'll do it myself when I've got time.

gzuidhof:
Sure, I can contribute that :+1:.

benanne:
Thanks for the PR! I'm fine with either. Maybe extending the current `DropoutLayer` is a bit more natural, at the expense of making the code for that slightly less readable.

gzuidhof:
I was thinking about how to implement this. It is very straightforward if the first few dimensions are "dropped out altogether". But is there a use case for dropping out say the second and the third dimension (given `<image, feature_channel, x, y>`)?

Making this `shared_axes` an integer which denotes from which (shape-)index  everything should either be dropt or kept would keep the implementation a lot simpler.

f0k:
> But is there a use case for dropping out say the second and the third dimension?

I can think of use cases for many different `shared_axes` values:
* shared_axes=(): standard dropout
* shared_axes=(0): batch-wise dropout (use the same dropout mask across axis 0)
* shared_axes=(2,3): drop full feature maps (use the same dropout mask across axes 2 and 3)
* shared_axes=(1,2): drop full frequency bands of spectrograms (use the same dropout mask across axes 1 and 2)

We wouldn't want a separate dropout layer class for each variant. We could have a `spatial_dropout()` shortcut, though.

> Making this `shared_axes` an integer which denotes from which (shape-)index everything should either be dropt or kept would keep the implementation a lot simpler.

I can do the implementation myself if you're not under time pressure! This has been on my TODO list for a while.

PR #708, opened 2016-06-23
Fix documentation of dilation factor in DilatedConv2DLayer

### LINKED ISSUE #707: DilatedConv2DLayer: dilation or dilation-1
In the DilatedConv2DLayer docstring, for the `dilation` parameter [it says](https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/conv.py#L812-L814): 
```
An integer or a 2-element tuple specifying the dilation factor of the filters, 
i.e., the number of zeros inserted between filter elements
```
Which implies that a dilation factor of e.g. 1 would insert 1 zero between every pair of filter elements. However, the note at the bottom [says](https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/conv.py#L868-L874):
```
The dilated convolution is implemented as the backward pass of a
convolution wrt. weights, passing the filters as the output gradient.
It can be thought of as dilating the filters (by adding ``dilation - 1``
zeros between adjacent filter elements) and cross-correlating them with the
input. See [1]_ for more background.
```
Which implies that a dilation factor of 1 would result in no zeros being added between filter elements.

Looking at the code I think the latter is what actually happens (?), but it would be helpful if this were clarified in the docstring for the dilation parameter.

f0k:
Closes #707.

### DIFF
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -811,7 +811,8 @@ class DilatedConv2DLayer(BaseConvLayer):
 
     dilation : int or iterable of int
         An integer or a 2-element tuple specifying the dilation factor of the
-        filters, i.e., the number of zeros inserted between filter elements.
+        filters. A factor of :math:`x` corresponds to :math:`x - 1` zeros
+        inserted between adjacent filter elements.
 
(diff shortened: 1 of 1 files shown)


benanne:
Thanks!

PR #715, opened 2016-06-30
Have create_param() set the broadcast pattern

f0k:
This PR makes create_param() set the broadcast pattern when creating a shared variable. We assume our network parameters never change shape, so we can at least fix the broadcast pattern (shared variables don't allow setting a fixed shape). This saves some headaches when users have a DenseLayer of 1 unit and try to do something fancy with the output (as here: https://groups.google.com/d/msg/lasagne-users/MQfjYI-tV2I/pf-z_NnyBQAJ), since now Theano can infer that the output is broadcastable on the last axis.

### DIFF
--- a/lasagne/tests/test_utils.py
+++ b/lasagne/tests/test_utils.py
@@ -274,6 +274,15 @@ def test_create_param_retain_ndarray_dtype():
     assert (result.dtype == param.dtype)
 
 
+def test_create_param_broadcast_pattern():
+    from lasagne.utils import create_param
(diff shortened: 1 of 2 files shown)


benanne:
Sounds good, should we maybe have a comment in the code that explains why we are doing this (basically what you just said)? broadcastable patterns don't come up too often so it might baffle some people who read that piece of code.

f0k:
> Sounds good, should we maybe have a comment in the code that explains why we are doing this (basically what you just said)?

Good point, did that!

benanne:
LGTM, will merge soon if there are no further comments :)

PR #725, opened 2016-07-26
Updated docs for Conv3DDNNLayer

rubenvereecken:
The incoming dimensions were in the wrong order. Looked around and found the correct order [in a recipe](https://github.com/Lasagne/Recipes/blob/master/modelzoo/c3d.py) and [in the Theano source](https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/dnn.py#L1201).

### DIFF
--- a/lasagne/layers/dnn.py
+++ b/lasagne/layers/dnn.py
@@ -398,7 +398,7 @@ class Conv3DDNNLayer(BaseConvLayer):
     incoming : a :class:`Layer` instance or a tuple
         The layer feeding into this layer, or the expected input shape. The
         output of this layer should be a 5D tensor, with shape ``(batch_size,
-        num_input_channels, input_rows, input_columns, input_depth)``.
+        num_input_channels, input_depth, input_rows, input_columns)``.
 
     num_filters : int
         The number of learnable convolutional filters this layer has.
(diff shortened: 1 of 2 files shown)


f0k:
> The incoming dimensions were in the wrong order.

Thank you! The choice is actually arbitrary, there's nothing preventing you from using the last dimension for depth and it shouldn't make a difference for the implementation either as long all spatial dimensions are grouped together. That's why we chose to call it `rows` and `columns` instead of `height` and `width` or `width` and `height`. If this is the more common layout, I'm fine with changing it, though. It makes sense to be able to index a single image as `input[a,b,c]` instead of `input[a,b,:,:,c]`. @benanne, any opinion?

We cannot merge it this way, though, it's inconsistent with the docstring of `W` and `b` now, and also with the tests (they use `b, c, h, w, d`). Can you amend your commit (not add a second one) to make everything consistent?

rubenvereecken:
Wasn't trying to be pedantic, was genuinely concerned the implementation might behave differently across different dimensions.

I think I have all the required changes included now.

lasagne/layers/recurrent.py
```diff
@@ -944,6 +944,7 @@ def get_output_for(self, inputs, **kwargs):
         """
         # Retrieve the layer input
         input = inputs[0]
+        print input
```
f0k: This shouldn't be here.

f0k:
> Wasn't trying to be pedantic, was genuinely concerned the implementation might behave differently across different dimensions.

Okay, I can assure you there are just three anonymous spatial dimensions, we're free to give them any meaning we want :) I still agree it's probably a good suggestion to keep the traditional image dimensions in the end.

lasagne/layers/recurrent.py
rubenvereecken: Uhm, yeah, my bad, didn't realize I was still on that branch while snooping around.

f0k:
Thank you, looks good now! I'll merge tomorrow if nobody objects.

benanne:
Fine by me :)

f0k:
Merging, thanks again!

PR #741, opened 2016-09-01
Fixed theano error on truedivision of integers

Erotemic:
The TestTransformLayer.test_transform_affine_identity and TestTPSTransformLayer.test_transform_thin_plate_spline_variable_input tests were  failing on my machine due to my .theanorc file:

```
[global]
...
int_division = raise
...
```

They both raised this theano error:

```
                raise IntegerDivisionError(
>                   "With `config.int_division` set to 'raise', dividing two "
                    "integer types with '/' is forbidden to avoid confusion "
                    "between integer and floating point divisions. Please "
                    "use // for integer division, or if you want a float result "
                    "either cast one of the arguments to a float or directly call "
E                   IntegerDivisionError: With `config.int_division` set to 'raise', dividing two integer types with '/' is forbidden to avoid confusion between integer and floating point divisions. Please use // for integer division, or if you want a float result either cast one of the arguments to a float or directly call `x.__truediv__(y)`.
```

The tracebacks were: 

```
lasagne/layers/special.py:432: in get_output_for
    return _transform_affine(theta, input, self.downsample_factor)
lasagne/layers/special.py:440: in _transform_affine
    out_height = T.cast(height / downsample_factor[0], 'int64')
../Theano/theano/tensor/var.py:171: in __div__
    return theano.tensor.basic.div_proxy(self, other)
../Theano/theano/tensor/basic.py:3265: in div_proxy
    as_tensor_variable(y).dtype in discrete_dtypes)
```

and 

```
lasagne/layers/special.py:689: in get_output_for
    self.precompute_grid, self.downsample_factor)
lasagne/layers/special.py:716: in _transform_thin_plate_spline
    out_height = T.cast(height / downsample_factor[0], 'int64')
../Theano/theano/tensor/var.py:171: in __div__
    return theano.tensor.basic.div_proxy(self, other)
../Theano/theano/tensor/basic.py:3265: in div_proxy
    as_tensor_variable(y).dtype in discrete_dtypes)
```

I was able to fix these tests on my machine by simply using integer division. This should not cause any change in behavior because the outputs of the previous true-division were always cast to integers.

### DIFF
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -423,7 +423,7 @@ class TransformerLayer(MergeLayer):
     def get_output_shape_for(self, input_shapes):
         shape = input_shapes[0]
         factors = self.downsample_factor
-        return (shape[:2] + tuple(None if s is None else int(s / f)
+        return (shape[:2] + tuple(None if s is None else int(s // f)
                                   for s, f in zip(shape[2:], factors)))
 
     def get_output_for(self, inputs, **kwargs):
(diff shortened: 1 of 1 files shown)


f0k:
> This should not cause any change in behavior because the outputs of the previous true-division were always cast to integers.

There is a theoretical change in behavior: Casting a division result to `int` rounds towards zero, while floor division (`//`) rounds towards minus infinity:
```
>>> -13. / 4
-3.25
>>> int(-13. / 4)
-3
>>> int(-13. // 4)
-4
```
But the layer would break for negative downsampling factors anyway, so we can ignore this.

Will merge tomorrow if there are no further comments. Thank you for your contribution!

PR #745, opened 2016-09-24
Add parameter to transposed conv to control the output shape

### LINKED ISSUE #711: TransposedConv2DLayer: output_size argument
The `TransposedConv2DLayer` currently produces an output of the minimum size that could have produced its input in a forward convolution. Sometimes that's not what you want: For a encoder/decoder architecture, you'll want each transposed convolution in the decoder to recreate the input shape of the corresponding forward convolution in the encoder. For example, convolving a 32x32 input with 3x3 filters, stride 2x2 and "same" padding will produce a 16x16 output, but applying the corresponding transposed convolution will produce a 31x31 output only (because a 31x31 input would have resulted in a 16x16 output as well).

To solve this, as agreed in #610, we should add an optional `output_size` argument to the `TransposedConv2DLayer` constructor that allows to override the inferred minimum output size.

fvisin:
This PR adds a parameter `alpha` to the transposed convolution layer to let the user specify which of the  possible output shapes to return (when stride is not 1).

Closes #711.

### DIFF
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -3,7 +3,7 @@ import theano.tensor as T
 from .. import init
 from .. import nonlinearities
 from ..utils import as_tuple
-from ..theano_extensions import conv, padding
+from ..theano_extensions import conv
 
 from .base import Layer
 
@@ -704,6 +704,11 @@ class TransposedConv2DLayer(BaseConvLayer):
         correlation (this is the default). Note that this flag is inverted
         compared to a non-transposed convolution.
 
+    output_size : int or iterable of int or symbolic tuple of ints
+        The output size of the transposed convolution. Allows to specify
+        which of the possible output shapes to return when stride > 1.
+        If not specified, the smallest shape will be returned.
+
     **kwargs
         Any additional keyword arguments are passed to the `Layer` superclass.
 
@@ -744,7 +749,12 @@ class TransposedConv2DLayer(BaseConvLayer):
                  crop=0, untie_biases=False,
                  W=init.GlorotUniform(), b=init.Constant(0.),
                  nonlinearity=nonlinearities.rectify, flip_filters=False,
-                 **kwargs):
+                 output_size=None, **kwargs):
+        # output_size must be set before calling the super constructor
+        if (not isinstance(output_size, T.Variable) and
+                output_size is not None):
+            output_size = as_tuple(output_size, 2, int)
+        self.output_size = output_size
         super(TransposedConv2DLayer, self).__init__(
                 incoming, num_filters, filter_size, stride, crop, untie_biases,
                 W, b, nonlinearity, flip_filters, n=2, **kwargs)
@@ -758,6 +768,13 @@ class TransposedConv2DLayer(BaseConvLayer):
         return (num_input_channels, self.num_filters) + self.filter_size
 
     def get_output_shape_for(self, input_shape):
+        if self.output_size is not None:
+            size = self.output_size
+            if isinstance(self.output_size, T.Variable):
+                size = (None, None)
+            return input_shape[0], self.num_filters, size[0], size[1]
+
+        # If self.output_size is not specified, return the smallest shape
         # when called from the constructor, self.crop is still called self.pad:
         crop = getattr(self, 'crop', getattr(self, 'pad', None))
         crop = crop if isinstance(crop, tuple) else (crop,) * self.n
@@ -776,7 +793,9 @@ class TransposedConv2DLayer(BaseConvLayer):
             subsample=self.stride, border_mode=border_mode,
             filter_flip=not self.flip_filters)
         output_size = self.output_shape[2:]
-        if any(s is None for s in output_size):
+        if isinstance(self.output_size, T.Variable):
+            output_size = self.output_size
+        elif any(s is None for s in output_size):
             output_size = self.get_output_shape_for(input.shape)[2:]
         conved = op(self.W, input, output_size)
         return conved
--- a/lasagne/tests/layers/test_conv.py
+++ b/lasagne/tests/layers/test_conv.py
@@ -2,6 +2,7 @@ import numpy as np
 import pytest
 import importlib
 import theano
+from theano import tensor as T
 
 import lasagne
 from lasagne.utils import floatX, as_tuple
@@ -88,7 +89,7 @@ def dilate(input, factors):
     return output
 
 
-def transposed_convNd(input, kernel, crop, stride=1, n=None):
+def transposed_convNd(input, kernel, crop, stride=1, n=None, extend=None):
     if n is None:
         n = input.ndim - 2
     if crop == 'valid':
@@ -102,6 +103,10 @@ def transposed_convNd(input, kernel, crop, stride=1, n=None):
         pad = tuple(f - 1 - c for f, c in zip(kernel.shape[2:], crop))
     stride = as_tuple(stride, n, int)
     dilated_input = dilate(input, (1, 1) + stride)
+    if extend is not None:
+        extend = as_tuple(extend, n, int)
+        extend = [(0, p) for p in (0, 0) + extend]
+        dilated_input = np.pad(dilated_input, extend, mode='constant')
     return convNd(dilated_input, kernel, pad, stride=1, n=n)
 
 
@@ -195,6 +200,21 @@ def transp_conv2d_test_sets():
     # flip_filters=False case
     output = transposed_convNd(input, kernel[:, :, ::-1, ::-1], 'valid')
     yield _convert(input, kernel, output, {'flip_filters': False})
+    # extend (w/ and w/out symbolic output shape)
+    for symbolic in [False, True]:
+        input_shape = (4, 3, 7, 9)
+        input = np.random.random(input_shape)
+        kernel = np.random.random((16, 3, 2, 3))
+        stride = (2, 3)
+        for extend in [(0, 1), (1, 2)]:
+            output = transposed_convNd(input, kernel, 0, stride, extend=extend)
+            kwargs = {'stride': stride, 'flip_filters': True}
+            if symbolic:
+                kwargs['output_size'] = theano.shared(
+                    np.array(output.shape[2:]))
+            else:
+                kwargs['output_size'] = output.shape[2:]
+            yield _convert(input, kernel, output, kwargs)
 
 
 def dilated_conv2d_test_sets():
@@ -549,8 +569,15 @@ class TestTransposedConv2DLayer:
                 **kwargs)
         actual = layer.get_output_for(input).eval()
         assert actual.shape == output.shape
-        assert actual.shape == layer.output_shape
+        # layer.output_shape == actual.shape or None
+        assert all([s1 == s2 for (s1, s2) in
+                    zip(actual.shape, output.shape) if s2])
         assert np.allclose(actual, output)
+        # Check get_output_shape_for for symbolic output
+        if 'output_size' in kwargs and isinstance(kwargs['output_size'],
+                                                  T.Variable):
+            assert all(el is None for el in
+                       layer.get_output_shape_for(input.shape)[2:])
 
     @pytest.mark.parametrize(
         "input, kernel, output, kwargs", list(transp_conv2d_test_sets()))
@@ -566,10 +593,20 @@ class TestTransposedConv2DLayer:
                 filter_size=kernel.shape[2:],
                 W=kernel.transpose(1, 0, 2, 3),
                 **kwargs)
-        assert layer.output_shape == (None, output.shape[1], None, None)
+        if 'output_size' not in kwargs or isinstance(kwargs['output_size'],
+                                                     T.Variable):
+            assert layer.output_shape == (None, output.shape[1], None, None)
         actual = layer.get_output_for(input).eval()
         assert actual.shape == output.shape
         assert np.allclose(actual, output)
+        # Check get_output_shape_for for non symbolic output
+        if 'output_size' in kwargs and not isinstance(kwargs['output_size'],
+                                                      T.Variable):
+            assert layer.get_output_shape_for(input.shape) == output.shape
+            # The layer should report the output size even when it
+            # doesn't know most of the input size
+            assert layer.output_shape == (
+                None, output.shape[1]) + kwargs['output_size']
 
 
 class TestDilatedConv2DLayer:


fvisin:
I should probably add a test for alpha. Is there something else you think is missing?

benanne:
This is great, thanks! I have two discussion points:

* I'm not sure about the name "alpha", where does this come from? I think I'd prefer something more descriptive, since it is not a very common use case. Although admittedly it might be hard to come up with something descriptive that is not too long.

* This may be naive, but would it not be more straightforward to simply let the user specify the desired output shape directly? I haven't used transposed convolutions much myself, but that would seem more natural to me. With this alpha parameter I think I would find myself doing a bunch of arithmetic to work out the right values, and that is really something the computer should be figuring out for itself ;)

fvisin:
Thanks for your answer!

Both questions have the same answer: I mimicked the guide to convolution arithmetic! :)
I picked "alpha" to keep it close to the "a" we used in the guide, but I agree it's not very descriptive and that using the output shape might be more intuitive. 

Is there any setting where one might not care about the shape but rather only about the padding? E.g., if one wants to pad the output by a fixed amount? I guess using the input symbolic shape would not make much of a difference in that case either.. Ok, I cannot find any counter argument. Unless somebody comes up with a case where alpha is more convenient, I will modify it to use the output shape directly.

f0k:
> This may be naive, but would it not be more straightforward to simply let the user specify the desired output shape directly?

+1 for specifying the output size directly, as planned in https://github.com/Lasagne/Lasagne/issues/711.

> Unless somebody comes up with a case where alpha is more convenient, I will modify it to use the output shape directly.

Specifying the extra padding amount is the cleaner interface, but inconvenient in most use cases. Would be great if you could tackle this! There should be an `output_size` argument similar to the `filter_size`, i.e., passed through `as_tuple` in the constructor to easily support square images.

fvisin:
I modified the code to accept the `output_size` parameter instead of `alpha`. I tried to use it in my own project and it seems to work. 

I am beginning to work on a proper test, do you have suggestions on how to modify `convNd` to account for the variable output size?

lasagne/layers/conv.py
```diff
@@ -125,9 +125,7 @@ def conv_input_length(output_length, filter_size, stride, pad=0):
     int or None
         The smallest input size corresponding to the given convolution
         parameters for the given output size, or ``None`` if `output_size` is
-        ``None``. For a strided convolution, any input size of up to
-        ``stride - 1`` elements larger than returned will still give the same
-        output size.
+        ``None``.
```
f0k: Why did you remove this? The functionality of `conv_input_length` stays the same.

lasagne/layers/conv.py
```diff
@@ -744,7 +747,11 @@ def __init__(self, incoming, num_filters, filter_size, stride=(1, 1),
                  crop=0, untie_biases=False,
                  W=init.GlorotUniform(), b=init.Constant(0.),
                  nonlinearity=nonlinearities.rectify, flip_filters=False,
-                 **kwargs):
+                 output_size=None, **kwargs):
+        if (not isinstance(output_size, T.TensorVariable) and
+                output_size is not None):
+            output_size = as_tuple(output_size, 2, int)
+        self.output_size = output_size
```
f0k: Maybe add a comment: `# output_size must be set before calling the super constructor`

lasagne/layers/conv.py
```diff
@@ -757,7 +764,18 @@ def get_W_shape(self):
...
-    def get_output_shape_for(self, input_shape):
+    def get_output_shape_for(self, input_shape, symbolic=False):
+        if self.output_size is not None:
+            if isinstance(self.output_size, T.TensorVariable):
+                if not symbolic:
+                    size = (None, None)
+                else:
+                    size = self.output_size
+            else:
+                size = as_tuple(self.output_size, 2, int)
```
f0k: `self.output_size` already is a tuple, no need to convert again.

lasagne/layers/conv.py
```diff
@@ -757,7 +764,18 @@ def get_W_shape(self):
...
         return (num_input_channels, self.num_filters) + self.filter_size
 
-    def get_output_shape_for(self, input_shape):
+    def get_output_shape_for(self, input_shape, symbolic=False):
+        if self.output_size is not None:
+            if isinstance(self.output_size, T.TensorVariable):
+                if not symbolic:
+                    size = (None, None)
+                else:
+                    size = self.output_size
```
f0k: I'd prefer to have the `symbolic` distinction in `get_output_for`. It's a nice idea, but you need to read more code to understand it.

lasagne/layers/conv.py
```diff
@@ -777,7 +795,7 @@ def convolve(self, input, **kwargs):
             filter_flip=not self.flip_filters)
         output_size = self.output_shape[2:]
         if any(s is None for s in output_size):
-            output_size = self.get_output_shape_for(input.shape)[2:]
+            output_size = self.get_output_shape_for(input.shape, True)[2:]
```
f0k: Add here:
```python
if isinstance(self.output_size, T.TensorVariable):
    output_size = self.output_size
elif any(s is None for s in output_size):
    output_size = self.get_output_shape_for(input.shape)[2:]
```
This avoids the `symbolic` distinction in `get_output_shape_for()`.

lasagne/layers/conv.py
```diff
@@ -704,6 +702,11 @@ class TransposedConv2DLayer(BaseConvLayer):
         correlation (this is the default). Note that this flag is inverted
         compared to a non-transposed convolution.
 
+    output_size : int or iterable of int or symbolic tuple of ints
+        The output size of the transposed convolution. Allows to specify
+        which of the possible output shapes to return when stride > 1.
+        If not specified, the smaller shape will be returned by default.
```
f0k: *the smallest shape
(and omit "by default").

lasagne/layers/conv.py
fvisin: You are definitely right. For some reason I was convinced I added it in the previous commit.

lasagne/layers/conv.py
fvisin: I agree, it makes much more sense to have it here. I modified it in:
```python
    output_size = self.output_shape[2:]
    if (not isinstance(self.output_size, T.TensorVariable) and
            any(s is None for s in output_size)):
        output_size = self.get_output_shape_for(input.shape)[2:]
```

fvisin:
I implemented the modifications you asked and added some tests. 
I modified the existing test functions to cover asymmetric strides and generalize to the case with extension. If everything is right I can squash the last 3 commits together (I would not squash the first "alpha" version in case it's needed in the future)

lasagne/layers/conv.py
```diff
@@ -794,8 +793,9 @@ def convolve(self, input, **kwargs):
             subsample=self.stride, border_mode=border_mode,
             filter_flip=not self.flip_filters)
         output_size = self.output_shape[2:]
-        if any(s is None for s in output_size):
-            output_size = self.get_output_shape_for(input.shape, True)[2:]
+        if (not isinstance(self.output_size, T.TensorVariable) and
+                any(s is None for s in output_size)):
+            output_size = self.get_output_shape_for(input.shape)[2:]
```
f0k: You're ignoring a symbolic `self.output_size` now. My previous suggestion was already correct: https://github.com/Lasagne/Lasagne/pull/745#discussion_r80759218

lasagne/tests/layers/test_conv.py
```diff
@@ -102,6 +102,10 @@ def transposed_convNd(input, kernel, crop, stride=1, n=None):
         pad = tuple(f - 1 - c for f, c in zip(kernel.shape[2:], crop))
     stride = as_tuple(stride, n, int)
     dilated_input = dilate(input, (1, 1) + stride)
+    if extend is not None:
+        extend = as_tuple(extend, 2, int)
+        extend = ((0, 0), (0, 0), (0, extend[0]), (0, extend[1]))
+        dilated_input = np.pad(dilated_input, extend, mode='constant')
```
f0k: Please rewrite to support arbitrary `n`. Let me know if you need help.

lasagne/tests/layers/test_conv.py
```diff
@@ -571,6 +589,25 @@ def test_with_nones(self, DummyInputLayer, input, kernel, output, kwargs):
         assert actual.shape == output.shape
         assert np.allclose(actual, output)
 
+    @pytest.mark.parametrize(
+        "input, kernel, output, kwargs", list(transp_conv2d_test_sets(True)))
+    def test_with_output_size(self, DummyInputLayer, input, kernel, output,
+                              kwargs):
```
f0k: As I said (https://github.com/Lasagne/Lasagne/pull/745#pullrequestreview-1789166), I'd prefer to just add a few test cases to `transp_conv2d_test_sets`, so they're run by the existing `test_defaults` and `test_with_nones`. You don't need the `with_output` flag in `transp_conv2d_test_sets` then. Just add a few manual test cases at the end of the function, and make sure they include the `output_size` argument -- both with a fixed shape tuple, and with a shared variable (like T.shared(output.shape)). The idea of using the size of the ground truth output for the output_size is good, makes it easier.

lasagne/tests/layers/test_conv.py
f0k:
```python
extend = as_tuple(extend, n, int)
extend = [(0, p) for p in (0, 0) + extend]
```

lasagne/layers/conv.py
fvisin: Ok. I modified your suggestion in:
```python
if isinstance(self.output_size, T.TensorVariable):
    output_size = self.output_size
elif self.output_size is None or any(s is None for s in self.output_size):
    output_size = self.get_output_shape_for(input.shape)[2:]
```
to take into account the case when `self.output_size` is `None`

f0k: That's also wrong. When `self.output_size` is `None`, the code should do the same as before (i.e., before your PR). I think my suggestion is fully correct. Of course, you need to keep the existing `output_size = self.output_shape[2:]` before the part that's modified here.

fvisin: I modified your suggestion because when `self.output_size` is `None` the `any(s is None for s in output_size)` would fail, but I see now that the code I proposed would instead fail when `input.shape` is symbolic.

What if I add a condition to the `elif` not to enter when `self.output_shape` is `None`?
```python
output_size = self.output_shape[2:]
if isinstance(self.output_size, T.TensorVariable):
    output_size = self.output_size
elif self.output_size is not None and any(s is None for s in self.output_size):
    output_size = self.get_output_shape_for(input.shape)[2:]
```

f0k: > I modified your suggestion because when self.output_size is None the any(s is None for s in output_size) would fail

No, why? There's `output_size = self.output_shape[2:]` before, not `output_size = self.output_size`. You can use my code as-is.

fvisin: Oh my bad! I misread `any(s is None for s in output_size)` as `any(s is None for s in self.output_size)`. Ok I finally got it, sorry it took so long! :)

lasagne/layers/conv.py
```diff
@@ -776,8 +793,11 @@ def convolve(self, input, **kwargs):
             subsample=self.stride, border_mode=border_mode,
             filter_flip=not self.flip_filters)
         output_size = self.output_shape[2:]
-        if any(s is None for s in output_size):
+        if isinstance(self.output_size, T.Variable):
+            output_size = self.output_size
+        elif any(s is None for s in output_size):
```
f0k: Yay! :)

lasagne/layers/conv.py
```diff
@@ -776,8 +793,11 @@ def convolve(self, input, **kwargs):
...
             output_size = self.get_output_shape_for(input.shape)[2:]
+
```
f0k: Was this intentional?

lasagne/tests/layers/test_conv.py
```diff
@@ -100,8 +100,12 @@ def transposed_convNd(input, kernel, crop, stride=1, n=None):
     else:
         crop = as_tuple(crop, n, int)
         pad = tuple(f - 1 - c for f, c in zip(kernel.shape[2:], crop))
-    stride = as_tuple(stride, n, int)
+    stride = as_tuple(stride, n)
```
f0k: Was this intentional?

lasagne/tests/layers/test_conv.py
```diff
@@ -181,20 +185,40 @@ def _convert(input, kernel, output, kwargs):
...
-                                           'flip_filters': True})
+                                           'flip_filters': True,
+                                           'stride': stride})
     # flip_filters=False case
-    output = transposed_convNd(input, kernel[:, :, ::-1, ::-1], 'valid')
-    yield _convert(input, kernel, output, {'flip_filters': False})
+    output = transposed_convNd(input, kernel[:, :, ::-1, ::-1], 'valid',
+                               stride)
+    yield _convert(input, kernel, output, {'flip_filters': False,
+                                           'stride': stride})
```
f0k: I think we can leave the other tests without strides, as they were before.

lasagne/tests/layers/test_conv.py
```diff
@@ -181,20 +185,40 @@ def _convert(input, kernel, output, kwargs):
...
+        stride = np.random.randint(2, 3, 2)
+        extend = [np.random.randint(0, s-1) for s in stride]
+        output = transposed_convNd(input, kernel, 0, stride, extend=extend)
+        kwargs = {'stride': stride, 'flip_filters': True}
+        if symbolic:
+            kwargs['output_size'] = theano.shared(np.array(output.shape[2:]))
+        yield _convert(input, kernel, output, kwargs)
+
     # bias-less case
+    input_shape = (3, 1, 6, 13)
```
f0k: Maybe just move your `extend` tests to the end of the function, so you don't have to modify any existing tests.

lasagne/tests/layers/test_conv.py
```diff
@@ -181,20 +185,40 @@ def _convert(input, kernel, output, kwargs):
                                                        'stride': stride,
                                                        'flip_filters': True})
 
+    # extend (w/ and w/out symbolic output shape)
+    for symbolic in [False, True]:
+        input_shape = (4, 3, 7, 9)
+        input = np.random.random(input_shape)
+        kernel = np.random.random((16, 3, 2, 3))
+        stride = np.random.randint(2, 3, 2)
```
f0k: This is equivalent to `stride = np.array((2, 2))` (check the documentation). I'd prefer a fixed setting anyway. We don't want the test suite's ability to spot problems be dependent on the random seed. (Technically we are already depending on the random seed for the numbers, but that's probably not as likely to expose a bug than changing parameters.)

lasagne/tests/layers/test_conv.py
```diff
@@ -181,20 +185,40 @@ def _convert(input, kernel, output, kwargs):
...
+        extend = [np.random.randint(0, s-1) for s in stride]
```
f0k: Again, I'm not sure about using randomness here. Maybe fix the stride to (2,3) and loop over two choices for extend, like (0,1) and (1,2).

lasagne/tests/layers/test_conv.py
```diff
@@ -181,20 +185,42 @@ def _convert(input, kernel, output, kwargs):
...
+        output = transposed_convNd(input, kernel, 0, stride, extend=extend)
+        kwargs = {'stride': stride, 'flip_filters': True}
+        if symbolic:
+            kwargs['output_size'] = theano.shared(np.array(output.shape[2:]))
+        else:
+            kwargs['output_size'] = np.array(output.shape[2:])
```
f0k: Ah, you found this yourself while I was reviewing. :+1:

lasagne/tests/layers/test_conv.py
```diff
@@ -181,20 +185,42 @@ def _convert(input, kernel, output, kwargs):
...
+        yield _convert(input, kernel, output, kwargs)
+
     # bias-less case
+    input_shape = (3, 1, 6, 13)
```
f0k: If you move your `extend` test to the end of the function, you don't need to redefine the input shape here. That way, existing tests will stay the same as before.

lasagne/layers/conv.py
```diff
@@ -281,7 +281,7 @@ def __init__(self, incoming, num_filters, filter_size, stride=1, pad=0,
         self.num_filters = num_filters
         self.filter_size = as_tuple(filter_size, n, int)
         self.flip_filters = flip_filters
-        self.stride = as_tuple(stride, n, int)
+        self.stride = as_tuple(stride, n)
```
fvisin: Yeah...it's admittedly a not elegant solution to a python3 problem. Fact is, I don't have python3 installed and I went for a lazy solution. :)

For some reason in python3 `numpy.int32` and `numpy.int64` fail to be recognized as `int` (`isinstance(var, int)` returns `False`). This triggers [this exception](https://github.com/Lasagne/Lasagne/blob/996bf64c0aec6d481044495800b461cc62040041/lasagne/utils.py#L194-L196).

It will probably work if I replace the `int` with `(int, np.int32, np.int64)`. I can try

lasagne/layers/conv.py
fvisin: :)

lasagne/tests/layers/test_conv.py
fvisin: Ditto

lasagne/tests/layers/test_conv.py
fvisin: Right. I'll change it.

lasagne/tests/layers/test_conv.py
fvisin: OK

lasagne/tests/layers/test_conv.py
fvisin: :)

lasagne/tests/layers/test_conv.py
fvisin: Agreed

lasagne/layers/conv.py
f0k: I see. It's not limited to Python 3, actually:
```
In [1]: isinstance(np.int32(3), int)
Out[1]: False

In [2]: isinstance(np.int64(3), int)
Out[2]: True
```
In Python 3, both are False.

---

Anyway, I think this is outside the scope of this PR. It affects a lot of existing layers and should be addressed in a separate issue. Please better adapt the test not to pass a numpy array here.

fvisin: OK. I actually think that removing the `np.random.randInt` already fixed that issue, I pushed a commit to go back to having `int` in the `as_tuple`.

lasagne/layers/conv.py
```diff
@@ -1,3 +1,4 @@
+import numpy as np
```
f0k: !

lasagne/tests/layers/test_conv.py
```diff
@@ -566,10 +593,17 @@ def test_with_nones(self, DummyInputLayer, input, kernel, output, kwargs):
...
+                                                     T.Variable):
+            assert layer.output_shape == (None, output.shape[1], None, None)
         actual = layer.get_output_for(input).eval()
         assert actual.shape == output.shape
         assert np.allclose(actual, output)
+        # Check get_output_shape_for for symbolic output
+        if 'output_size' in kwargs and isinstance(kwargs['output_size'],
+                                                  T.Variable):
+            assert all(el is None for el in
+                       layer.get_output_shape_for(input)[2:])
```
f0k: You tested the symbolic case above already. What we'd need here would be a test for a nonsymbolic output size, to check if it's reported to the outside world.

lasagne/tests/layers/test_conv.py
```diff
@@ -566,10 +593,16 @@ def test_with_nones(self, DummyInputLayer, input, kernel, output, kwargs):
...
+        if 'output_size' not in kwargs or isinstance(kwargs['output_size'],
+                                                     T.Variable):
+            assert layer.output_shape == (None, output.shape[1], None, None)
         actual = layer.get_output_for(input).eval()
         assert actual.shape == output.shape
         assert np.allclose(actual, output)
+        # Check get_output_shape_for for non symbolic output
+        if 'output_size' in kwargs and not isinstance(kwargs['output_size'],
+                                                      T.Variable):
+            assert layer.get_output_shape_for(input.shape) == output.shape
```
f0k: That's an interesting test as well, but I had though of:
```python
assert layer.output_shape = (None, output.shape[1]) + kwargs['output_size']
```
I.e., the layer should report the output size even when it doesn't know most of the input size. Can you add this? (Not replace your current line, just add this one. Both seem useful.)

fvisin: > Sorry for the wealth of requests. 

No problem at all!

I added your test and squashed. This should be it!
In case it's needed in the future, the alpha version was in bf594106eb977f38b87d253d82960687aeeb7fa2

fvisin:
That's great, thanks for reviewing the PR!

f0k:
Wow, this thread has grown long with the new review interface and coveralls posting a comment after each push.
@benanne: I've disabled the coveralls comments now -- there's the "All checks have passed" / "Some checks have failed" status anyway.

f0k:
Merging, thanks again for your contribution!

fvisin:
That's great, thank you for reviewing!

benanne:
Sweet, thanks a lot for this :) re: coveralls, good call. Can we still easily see the coverage percentage in case it goes down?

f0k:
> Can we still easily see the coverage percentage in case it goes down?

I think so, but I'm not sure. Let's see. Either way, I'd be fine with having to click the "Details" link in exchange for not being spammed by coveralls comments.

PR #748, opened 2016-10-05
Add mode bed_of_nails to Upscale1DLayer, Upscale2DLayer, Upscale3DLayer

libfun:
This PR adds a parameter mode to the upscaling pool layer and 'bed of nails' mode for all upscaling layers. This modification does not break compatibility as the default mode is 'repeat' as it was before.

Examples of this method used for unpooling:
https://arxiv.org/abs/1411.5928
https://arxiv.org/abs/1506.02753
https://arxiv.org/abs/1602.02644

Example of differences in modes:
With 'repeat' and old behaviour
```
[[1, 2], 
 [3, 4]] 
```
becomes 
```
[[1, 1, 2, 2], 
 [1, 1, 2, 2],
 [3, 3, 4, 4],
 [3, 3, 4, 4]]
```
With 'bed_of_nails' it becomes
```
[[1, 0, 2, 0], 
 [0, 0, 0, 0],
 [3, 0, 4, 0],
 [0, 0, 0, 0]]
```

### DIFF
--- a/lasagne/layers/pool.py
+++ b/lasagne/layers/pool.py
@@ -5,7 +5,6 @@ from ..utils import as_tuple
 
 from theano.tensor.signal.pool import pool_2d
 
-
 __all__ = [
     "MaxPool1DLayer",
     "MaxPool2DLayer",
@@ -392,12 +391,16 @@ class Upscale1DLayer(Layer):
     scale_factor : integer or iterable
         The scale factor. If an iterable, it should have one element.
 
+    mode : {'repeat', 'dilate'}
+        Upscaling mode: repeat element values or upscale leaving zeroes between
+        upscaled elements. Default is 'repeat'.
+
     **kwargs
         Any additional keyword arguments are passed to the :class:`Layer`
         superclass.
     """
 
-    def __init__(self, incoming, scale_factor, **kwargs):
+    def __init__(self, incoming, scale_factor, mode='repeat', **kwargs):
         super(Upscale1DLayer, self).__init__(incoming, **kwargs)
 
         self.scale_factor = as_tuple(scale_factor, 1)
@@ -405,6 +408,11 @@ class Upscale1DLayer(Layer):
         if self.scale_factor[0] < 1:
             raise ValueError('Scale factor must be >= 1, not {0}'.format(
                 self.scale_factor))
+
+        if mode not in {'repeat', 'dilate'}:
+            msg = "Mode must be either 'repeat' or 'dilate', not {0}"
+            raise ValueError(msg.format(mode))
+        self.mode = mode
 
     def get_output_shape_for(self, input_shape):
         output_shape = list(input_shape)  # copy / convert to mutable list
@@ -415,8 +423,14 @@ class Upscale1DLayer(Layer):
     def get_output_for(self, input, **kwargs):
         a, = self.scale_factor
         upscaled = input
-        if a > 1:
-            upscaled = T.extra_ops.repeat(upscaled, a, 2)
+        if self.mode == 'repeat':
+            if a > 1:
+                upscaled = T.extra_ops.repeat(upscaled, a, 2)
+        elif self.mode == 'dilate':
+            if a > 1:
+                output_shape = self.get_output_shape_for(input.shape)
+                upscaled = T.zeros(shape=output_shape, dtype=input.dtype)
+                upscaled = T.set_subtensor(upscaled[:, :, ::a], input)
         return upscaled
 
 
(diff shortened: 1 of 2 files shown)


lasagne/layers/pool.py
```diff
@@ -406,6 +409,14 @@ def __init__(self, incoming, scale_factor, **kwargs):
             raise ValueError('Scale factor must be >= 1, not {0}'.format(
                 self.scale_factor))
 
+        self.mode = mode
+        if self.mode is None:
+            self.mode = 'repeat'
+
+        if self.mode not in {'repeat', 'bed_of_nails'}:
+            msg = "Mode must be either 'repeat' or 'bed_of_nails', not {0}"
+            raise ValueError(msg.format(self.mode))
```
f0k: Why should we support `None` here? I wouldn't add this unless required. Minor nitpick: I'd check `mode` before assigning to `self.mode`, that's how most layers do it.
```python
if mode not in ...:
    ...
self.mode = mode
```

lasagne/layers/pool.py
```diff
@@ -415,8 +426,18 @@ def get_output_shape_for(self, input_shape):
...
+                upscaled = T.extra_ops.repeat(upscaled, a, 2)
+        elif self.mode == 'bed_of_nails':
+            if a > 1:
+                input_shape = input.shape
+                output_shape = self.get_output_shape_for(input_shape)
+                _, _, i = T.nonzero(input)
+                upscaled = T.zeros(shape=output_shape, dtype=input.dtype)
+                upscaled = T.set_subtensor(
+                    upscaled[:, :, i * self.scale_factor[0]],
+                    input[:, :, i])
```
f0k: I think the `T.nonzero` will make this unnecessarily expensive. Can't you just do:
```python
output_shape = self.get_output_shape_for(input.shape)
upscaled = T.zeros(shape=output_shape, dtype=input.dtype)
upscaled = T.set_subtensor(upscaled[:, :, ::a], input)
```
Note that I'm reusing `a` for consistency with the `repeat` version.

lasagne/tests/layers/test_pool.py
```diff
@@ -60,6 +60,13 @@ def upscale_1d(data, scale_factor):
     return upscaled
 
 
+def upscale_1d_bed_of_nails(data, scale_factor):
+    upscaled = np.zeros(upscale_1d_shape(data.shape, scale_factor))
+    _, _, i = np.nonzero(data)
+    upscaled[:, :, i * scale_factor[0]] = data[:, :, i]
```
f0k: Again, the `nonzero` makes it more expensive. Just use ordinary slicing: `upscaled[..., ::scale_factor[0]] = data`

lasagne/tests/layers/test_pool.py
```diff
@@ -691,14 +713,19 @@ def scale_factor_test_sets():
         for scale_factor in [2, 3]:
             yield scale_factor
 
+    def mode_test_sets():
+        for mode in [None, 'repeat', 'bed_of_nails']:
+            yield mode
```
f0k: Ah, so removing `mode=None` support will even make the tests run faster :)

lasagne/layers/pool.py
```diff
@@ -464,6 +458,12 @@ class Upscale2DLayer(Layer):
     **kwargs
         Any additional keyword arguments are passed to the :class:`Layer`
         superclass.
+
+    Notes
+    -----
+    Using ``mode='bed_of_nails'`` followed by a convolution can be
+    realized more efficiently with a transposed convolution.
+    lasagne.layers.TransposedConv2DLayer
```
f0k: for the correct markup:
```
realized more efficiently with a transposed convolution, see
:class:`lasagne.layers.TransposedConv2DLayer`.
```

lasagne/layers/pool.py
```diff
@@ -433,11 +430,8 @@ def get_output_for(self, input, **kwargs):
             if a > 1:
                 input_shape = input.shape
                 output_shape = self.get_output_shape_for(input_shape)
```
f0k: Since `input_shape` is neither reused nor shorter than `input.shape`, I'd prefer
```python
output_shape = self.get_output_shape_for(input.shape)
```

libfun:
Thank you for review, I've fixed problems you've pointed at.

There is still one issue I haven't been able to tackle. The `make html` doesn't seem to catch the changes in the docstrings. 

Sorry for the confusion, I don't have experience working with sphinx.

f0k:
> The `make html` doesn't seem to catch the changes in the docstrings.

Try `make clean html` to delete existing files before recompiling. Sometimes this helps.

f0k:
Nice, thank you for the clean PR and for being so quick incorporating the requested changes! Will merge tomorrow if there are no further comments.

f0k:
Merged, thanks again for your contribution!

Issue #626, opened 2016-02-29
add spatial softmax helper?


benanne:
It seems like semantic segmentation is becoming a popular task, so there are quite a few questions on the mailing list lately about doing per-pixel classification. This currently requires some trickery because Theano's softmax function expects 2D input, so you have to shuffle and reshape your 'output maps' for this to work (latest discussion: https://groups.google.com/forum/#!topic/lasagne-users/6P1bodBbhWs )

Perhaps we should add a helper function or layer `SpatialSoftmax`, that does this sequence of operations? Or more generally we could have `Softmax1D`, `Softmax2D` etc.?

One thing we have to be careful about is that it is not confused with a softmax across spatial locations (rather than a per-pixel softmax operation), which is sometimes used for attention mechanisms. Perhaps we should provide a helper for that use case as well.

f0k:
Maybe we should have a general `nonlinearities.Softmax` class that accepts a list of axes and performs the necessary reshaping, and then some predefined instances such as the original `softmax` that accommodate the most common use cases.

> One thing we have to be careful about is that it is not confused with a softmax across spatial locations (rather than a per-pixel softmax operation), which is sometimes used for attention mechanisms. Perhaps we should provide a helper for that use case as well.

Good point. We should be careful about the naming then. What about `softmax_per_location()` (e.g., one classification per pixel, or per voxel) and `softmax_over_locations()` (normalizing the softmax over spatial locations)? Should the original `softmax()` stay limited to flattened input and give a meaningful error for too many input dimensions?

benanne:
All good point, clearly this issue can use some more brainstorming so we get it right :)

krikru:
Until there is a working spatial softmax layer, [here is a workaround](https://groups.google.com/d/msg/lasagne-users/OQR19fuPzbM/5p1lF-PJAQAJ) that uses reshaping of the 'output maps'.

f0k:
> What about `softmax_per_location()` (e.g., one classification per pixel, or per voxel) and `softmax_over_locations()` (normalizing the softmax over spatial locations)?

In #710, we agreed on `softmax_per_location()` and `softmax_per_channel()`.

ahundt:
I came across this, and thought you may be interested in this relevant tf issue: https://github.com/tensorflow/tensorflow/issues/6271

f0k:
> I came across this and thought you may be interested in this relevant tf issue

[...]

PR #750, opened 2016-10-09
Add num_leading_axes argument to DenseLayer

### LINKED ISSUE #719: More flexible DenseLayer
A common use case for a dense layer is to stack it on top of a recurrent layer to compute a prediction at every time step. Since the recurrent layer output is of shape `(batchsize, seq_len, feats)`, this requires three layers: A `ReshapeLayer(..., (-1, feats))`, the `DenseLayer(..., num_units)`, and possibly another `ReshapeLayer(..., (batchsize, seq_len, num_units))`. The latter requires either the `batchsize` or `seq_len` to be known at compile time, or obtained as symbolic variables.

The operation inside the `DenseLayer` is quite simple: It consists of `T.flatten()` followed by `T.dot()`.
Now `T.dot(x, y)` already supports arbitrary dimensionality for `x`, performing the dot product over the last axis only. The only reason we cannot directly use it on the output of a recurrent layer is that it flattens the input to be a matrix, in this case of shape `(batchsize, seq_len * feats)`.

We could make the `DenseLayer` much more flexible simply by allowing to specify the number of axes to be flattened (or to be kept). I propose to add a keyword argument with the following semantics:
* If it's a positive integer, this gives the number of axes not to be flattened. E.g., setting it to `1` (the default) means to keep the first axis and flatten the remaining ones. Put differently, it gives the index of the axis all following ones are to be collapsed under.
* If it's a negative integer, it also gives the index of the axis all following ones are to be collapsed under, counted from the end as is customary in Python for negative indices. E.g., setting it to `-2` means to flatten the last two axes, and `-1` means to only flatten the last axis, which does nothing.
* If it's zero, we throw an error because we think this is never intended and based on a misunderstanding? Otherwise, this would flatten all the input into a single vector, losing the batch dimension on the way.

I'm not too sure about the name of the keyword argument, though. It could be `keep_axes`, or `distribute_over` (as the dot product is distributed over the given number of axes).

Alternatively, we could also have a full-blown `shared_axes` argument listing the axes to distribute the dot product over, but this would involve moving all the others to the end before the dot product and restoring the order afterwards (e.g., a `(3,4,5,6)` tensor with `shared_axes=(0,-1)` would have a `(20,num_units)` weight matrix and produce a `(3,num_units,6)` output) -- maybe that's too much flexibility.

f0k:
As proposed in #719, this adds a `leading_axes` argument to `DenseLayer` which allows to specify how many axes of the input tensor should be treated as leading axes, only collapsing the remaining axes for the matrix product. For example, with `leading_axes=2`, this corresponds to the time-distributed dense layer of Keras.
I've slightly changed the dense layer tests to use a parameterized fixture, so existing tests run with different settings for `leading_axes`.
I've also added a check to the constructor to give a meaningful error if any trailing axes have an unspecified size. Before, it would just run into an exception in `np.prod`.

Closes #719.

### DIFF
--- a/lasagne/layers/dense.py
+++ b/lasagne/layers/dense.py
@@ -17,7 +17,7 @@ class DenseLayer(Layer):
     """
     lasagne.layers.DenseLayer(incoming, num_units,
     W=lasagne.init.GlorotUniform(), b=lasagne.init.Constant(0.),
-    nonlinearity=lasagne.nonlinearities.rectify, **kwargs)
+    nonlinearity=lasagne.nonlinearities.rectify, num_leading_axes=1, **kwargs)
 
     A fully connected layer.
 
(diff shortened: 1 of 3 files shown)


lasagne/layers/dense.py
```diff
@@ -17,7 +17,7 @@ class DenseLayer(Layer):
     """
     lasagne.layers.DenseLayer(incoming, num_units,
     W=lasagne.init.GlorotUniform(), b=lasagne.init.Constant(0.),
-    nonlinearity=lasagne.nonlinearities.rectify, **kwargs)
+    nonlinearity=lasagne.nonlinearities.rectify, leading_axes=1, **kwargs)
```
benanne: I'm wondering if perhaps we should call this `num_leading_axes` instead, for consistency? The name `leading_axes` implies that you can specify a list of axes, imo.

benanne:
Looks good to me :) Just one concern regarding naming.

f0k:
Renamed, merging! Thanks for reviewing!

benanne:
Thanks for doing all the work ;)

aripakman:
Thanks for the new option!
But note that using `leading_axes=2`, `nonlinearity=lasagne.nonlinearities.softmax` gives an error because the theano sofmax function only accepts 1- or 2-dim tensors.

f0k:
> the theano sofmax function only accepts 1- or 2-dim tensors

Yes, this is another issue: #626

juancamilog:
In python3 this breaks:

    File "/home/juancamilog/.local/lib/python3.5/site-packages/lasagne/layers/dense.py", line 111, in get_output_shape_for
        return input_shape[:self.num_leading_axes] + (self.num_units,)
    TypeError: can only concatenate list (not "tuple") to list

PR #751, opened 2016-10-09
Add missing layers to layer overview

f0k:
`ScaleLayer`, `standardize` and `TPSTransformerLayer` are listed in `doc/layers/special.rst`, but were missing from `doc/layers.rst`.


benanne:
Good catch, merging!

PR #752, opened 2016-10-09
Add BatchNormDNNLayer and convenience function

### LINKED ISSUE #743: Batchnorm with CuDNN v5
Hey guys,

@marcociccone and I would like to use the [new cuDNN 5 Theano batchnorm](https://github.com/Theano/Theano/blob/2711f775d34e453611c797c09cb3da1734f19c6e/theano/sandbox/cuda/dnn.py#L2659) with Lasagne. 

I read here and there some discussion about it (e.g. https://github.com/Lasagne/Lasagne/issues/575#issuecomment-228396513) but I coulnd't find any implementation.

I was wondering:

1. Is there any code for cuDNN 5 batchnorm anywhere?
2. Should we work on it, would you be interested in a PR?
3. My understanding is that you decided to keep the current BN interface. I didn't have much time to look into the theano Op nor your previous implementation of BN yet. Can you give us a quick idea of what/how you would like the new BN to be implemented?

Thanks!

f0k:
This adds `lasagne.layers.dnn.BatchNormDNNLayer` and `batch_norm_dnn`, as drop-in replacements for the ordinary batch normalization layers. They improve training performance a lot over the vanilla Theano layers, and also reduce memory usage. For a 40-layer DenseNet, I observed speed improvements of 25% to 35% depending on the GPU. Closes #743.

cuDNN-based batch normalization currently requires the bleeding-edge version of Theano (specifically, after this PR: https://github.com/Theano/Theano/pull/4582). If the Theano version is too old, both the class and the convenience function are deleted from `lasagne.layers.dnn`. This allows for easy compatibility in user code such as:
```python
try:
    from lasagne.layers.dnn import batch_norm_dnn as batch_norm
except ImportError:
    from lasagne.layers import batch_norm
```
This code does a seamless fallback when no GPU is available, cuDNN is not available, Lasagne is too old or Theano is too old.

### DIFF
--- a/docs/modules/layers.rst
+++ b/docs/modules/layers.rst
@@ -210,4 +210,6 @@
     dnn.MaxPool3DDNNLayer
     dnn.Pool3DDNNLayer
     dnn.SpatialPyramidPoolingDNNLayer
+    dnn.BatchNormDNNLayer
+    dnn.batch_norm_dnn
 
--- a/lasagne/layers/dnn.py
+++ b/lasagne/layers/dnn.py
@@ -7,6 +7,7 @@ from .base import Layer
 
 from .conv import conv_output_length, BaseConvLayer
 from .pool import pool_output_length
+from .normalization import BatchNormLayer
 from ..utils import as_tuple
 
 if not theano.sandbox.cuda.cuda_enabled:
@@ -28,6 +29,8 @@ __all__ = [
     "Conv2DDNNLayer",
     "Conv3DDNNLayer",
     "SpatialPyramidPoolingDNNLayer",
+    "BatchNormDNNLayer",
+    "batch_norm_dnn",
 ]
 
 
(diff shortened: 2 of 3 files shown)


benanne:
Nice work :) again, no time for a detailed review but it looks good.

f0k:
Thank you for skimming, merged!

fvisin:
Great job, thanks!!

I just skimmed through it too, if I'm not wrong you might want to add a check on the number of dimensions, to fall back to the previous implementation in case it's more than 4 (the max currently supported by the dnn BN in theano).

f0k:
> I just skimmed through it too, if I'm not wrong you might want to add a check on the number of dimensions, to fall back to the previous implementation in case it's more than 4

I thought about this as well while hovering above the "Merge" button... I wouldn't add an automatic fallback, just let it fail more gracefully. It doesn't automatically fallback if epsilon is too small or if the axes argument is not convertible to one of the two supported modes either.

> the max currently supported by the dnn BN in theano

IIRC the maximum is 5, cuDNN supports 3D convolution throughout their API.

fvisin:
> I thought about this as well while hovering above the "Merge" button... I wouldn't add an automatic fallback, just let it fail more gracefully. It doesn't automatically fallback if epsilon is too small or if the axes argument is not convertible to one of the two supported modes either.

Sounds reasonable.

> IIRC the maximum is 5, cuDNN supports 3D convolution throughout their API.

It's the maximum supported by cuDNN, but not by Theano. It is being implemented in https://github.com/Theano/Theano/pull/4948, which should be merged soon.

f0k:
> Sounds reasonable.

The current implementation fails with a reasonable error message on `get_output`, before compilation or running. Feel free to send a PR adding a check in the constructor (along with tests), but I think it's not absolutely required.

> It's the maximum supported by cuDNN, but not by Theano.

Oops, sorry, my bad. Nice to see it has already been ported to the new backend, and 5D inputs are supported now!

Issue #756, opened 2016-10-21
Is there interest for flexible training loops, batch iteration schemes, etc


Britefury:
Would the Lasagne community / developers be interested in adding code for training loops, basic DNN construction, etc?

I found that code for training loops with features such as early termination, etc would involve a lot of copy+paste, so I put together a library that can be found at:

http://github.com/Britefury/britefury-lasagne-helpers

I find it to be very helpful when I need to throw a network together and train/use it.

I would like to know if the Lasagne developers would be interested in taking some of this code and incorporating it into Lasagne?

My code has no documentation beyond the doc-strings; this would need to be worked on.

What's there:

## `data_souce` module
Defines the `batch_iterator` function that extracts mini-batches from a data set which can be NumPy arrays or other sources of data; the protocol is defined in the doc-strings.
Code for iterating though a dataset looks like:
```
for (batch_X, batch_y) in data_source.batch_iterator([trainX, train_y], batchsize=128)
    ...
```
where `train_X` and `train_y` can be NumPy arrays or other objects that support `__len__` and `__getitem__`; for more info see the protocols defined in `data_source`

## `trainer` module
Defines the `Trainer` class that implements a reasonably flexible training loop. It monitors the 
validation score of the network and can terminate early if the score doesn't improve within a certain number of epochs. It also reports training progress and saves the state of the network when validation score improves. Its train method accepts data sources and uses the `data_source` module to get batches. Setting up the trainer looks like:
```
train_batch = theano.function(...) # should update params and return train loss
eval_batch = theano.function(...) # should evaluate the performance and return loss/err rate

trainer = trainer.Trainer()

# Provide training function
trainer.train_with(train_batch_fn=train_batch)

# Provide evaluation function
trainer.evaluate_with(eval_batch_fn=eval_batch)

# Set the verbosity; can be VERBOSITY_NONE (nothing), VERBOSITY_MINIMAL (final performance only),
VERBOSITY_EPOCH (report every epoch), VERBOSITY_BATCH (report each batch for those large datasets)
trainer.report(verbosity=trainer.VERBOSITY_EPOCH)

# Have the trainer save the state of the network after each improvement in validation score
trainer.retain_best_scoring_state_of_network(final_layer)

# Train for a maximum of 100 epochs, minimum of 50, if training proceeds for 35 epochs without any improvement in validation score, stop early
trainer.train_for(num_epochs=100, min_epochs=50, val_improve_num_epochs=35)

# Train it
[...]

PR #759, opened 2016-10-23
Trainer

Britefury:
New `batch` and `trainer` modules with batch iteration helper and `Trainer` class respectively, as per issue [#756](https://github.com/Lasagne/Lasagne/issues/756).


Britefury:
Ready for review (@f0k).
There is a slight drop in coverage due to one RuntimeError that is raised in conditions that should never arise and a couple of logging statements that aren't run.

Britefury:
@f0k, @benanne This PR contains the code that I am offering with respect to issue #756.

@benanne: I like your proposal with respect to having a `train` function to which you provide a model, a loss to minimise, data and a function for computing updates (adam/nesterov/etc).

This PR does not go that far; it does the `fit`-loop in Keras terminology.

You provide functions to train and evaluate a single mini-batch, along with the number of epochs, batch size, callbacks, etc. The `train` function supplied (perhaps it should be renamed `train_loop`) runs the loop (optionally perming early termination), monitors the performance on validation and test sets, etc.

Would you like to review this code and guide me as to how to proceed?

benanne:
I'm currently abroad so it might be a while until I can take a look at it, unfortunately... so if anyone else wants to review be my guest. I'll make a note on my calendar to take a look when I'm back, though!

examples/mnist_trainer.py
```diff
@@ -0,0 +1,325 @@
...
+    # - layer to restore; this way the network will be set back to its
+    #   state when it achieved the best validation score
+    # - random number generator to shuffle training samples (just for show
+    #   here since it would use `lasagne.random.get_rng()` anyway
+    res = lasagne.trainer.train(
+        [X_train, y_train], [X_val, y_val], [X_test, y_test], batchsize=500,
+        train_batch_func=train_fn, train_log_func=train_log_func,
+        eval_batch_func=eval_fn, eval_log_func=eval_log_func,
+        num_epochs=num_epochs, verbosity=lasagne.trainer.VERBOSITY_EPOCH,
+        layer_to_restore=network, shuffle_rng=lasagne.random.get_rng())
```
benanne: This seems like a pretty complicated API, especially with the log functions. I'm not a big fan. It hides the for-loop, which is nice if you're happy with a lot of defaults, but in this case it ends up requiring a bunch of lambdas to be able to customise things, which is a lot less readable than just having that stuff in the body of a for loop. I think this is probably not the best API for this use case.

lasagne/trainer.py
```diff
@@ -0,0 +1,687 @@
...
+          train_log_func=None, train_epoch_results_check_func=None,
+          train_pass_epoch_number=False, eval_batch_func=None,
+          eval_log_func=None, val_improved_func=None, val_interval=None,
+          batchsize=128, num_epochs=100, min_epochs=None,
+          val_improve_patience=1, val_improve_patience_factor=0.0,
+          epoch_log_func=None, pre_epoch_callback=None,
+          post_epoch_callback=None, progress_iter_func=None,
+          verbosity=VERBOSITY_EPOCH, log_stream=sys.stdout,
+          log_final_result=True, get_state_func=None, set_state_func=None,
+          layer_to_restore=None, updates_to_restore=None, shuffle_rng=None):
```
f0k: Agree with @benanne, this may be a bit heavy on options. My idea was to have a relatively simple `train` function that calls a `train_epoch` function in a loop, and when you need anything more complex, you'd call the `train_epoch` function yourself and extend the loop body with whatever you need (learning rate updates, early stopping, ...). I'd need more time to think this through, though, and I won't have that time soon. Sorry for delaying this for so long, we certainly appreciate your efforts and I would love to continue this discussion in one or two months!

Britefury:
@benanne @f0k With a view to (maybe only partially) addressing some of your concerns, I have made some changes.

There are the new `batch_map` and `mean_batch_map` functions in the `batch` module that are mini-batch driven versions of `map`. `mean_batch_map` differs in that it returns the mean result across all samples, e.g. mean of the losses or error rates. I have modified the `mnist` example as a demo; the batch loops are gone and replaced by calls to `mean_batch_map`. I think that this is some way towards @f0k's `train_epoch` suggestion.

I have simplified the `train_log_func` and `epoch_log_func` arguments so that you can pass a string (template) whose `format` method is invoked to generate the log message. These arguments are not strictly necessary; the defaults will report the results just fine. Customising the output makes things more readable though. The `mnist_trainer` example has been simplified as the `verbosity` and `shuffle_rng` arguments were redundant as the values passed were the defaults anyway.

f0k:
> I realise that this is not quite what you want, as is. I understand that you are looking for an API that works along the lines of `train_network(final_layer, loss_to_minimise, metrics_to_evaluate, training_set, val_set, test_set, batch_size, ...)`.

Yes, I think that's about what I had in mind. Although maybe the user should still provide the updates dictionary. The API for creating the updates dictionary is not too complicated, and it's difficult to customize if it's created inside `train_network`. But a similar argument can be made for the compiled functions -- maybe `train_network` shouldn't do everything, but accept the compiled functions (`train_fn`, `test_fn`), and there should be an API for creating those. We really need to sit down and think through different options and use cases. But I don't know when I'll come to this.

> Perhaps a good way forward would be to split this into two PRs? The first PR could contain just the `batch` module with `batch_map` and `mean_batch_map`, with the second containing the `trainer` module that builds on top of `batch`.

I'm not convinced yet these should be separate modules. The only argument is that the batch iteration code can also be used for testing, so it doesn't really fit the `lasagne.trainer` module name.

> `train` certainly does have a lot of arguments. In many cases you won't need to use most of them as the defaults will do. I have however found early stopping, network state saving and restoring and various other features there to be very useful, hence their inclusion.

A nicer API for this might be something like:
```python
for epoch, errors in training_loop(...):
    print whatever you want
    save whatever you want
    update the learning rate when you want
    early stop when you want
```
There could be convenience functions for some of these. And well, yes, there could also be a function that encapsulates all this (that's the multiple levels of abstraction Sander and me were referring to). But then maybe the training loop should become a class that you can register such callbacks with, so you don't have a function with a myriad of arguments, but something like:
```python
trainer = Trainer(...)
trainer.after_epoch(LogResults(your_format_string))
trainer.after_epoch(ScaleParam(eta, 0.95))
trainer.at_epoch(100, ScaleParam(eta, 0.1))
trainer.run()
```
Again, the design space is huge, and we need to properly sit down to decide what to include in Lasagne.

Britefury:
@f0k, I hope you don't mind, but I decided to go ahead with the separate PR approach. I think that the design space is sufficiently large that it warrants building the training loop up in small pieces. The `data_source` module in PR #801 works toward this. I would like to take this 'smaller steps' approach as getting there in one leap could involve quite a lot of back and forth, with all the refactoring that goes along the way. If we can build and agree on it piece by piece, I think that I will find the process more manageable. :)

PR #760, opened 2016-10-24
Add LocallyConnected2DLayer, including documentation and tests.

alexlee-gk:
This PR follows up on the discussion from [here](https://groups.google.com/forum/#!topic/lasagne-users/G-LF7JCu9PM). I believe there is no issue relating to this though.

### DIFF
--- a/lasagne/layers/local.py
+++ b/lasagne/layers/local.py
@@ -1,0 +1,202 @@
+import theano.tensor as T
+
+from .. import init
+from .. import nonlinearities
+
+from .conv import Conv2DLayer
+
+
+__all__ = [
+    "LocallyConnected2DLayer",
+]
+
+
+class LocallyConnected2DLayer(Conv2DLayer):
+    """
+    lasagne.layers.LocallyConnected2DLayer(incoming, num_filters, filter_size,
+    stride=(1, 1), pad='same', untie_biases=False,
+    W=lasagne.init.GlorotUniform(), b=lasagne.init.Constant(0.),
+    nonlinearity=lasagne.nonlinearities.rectify, flip_filters=True,
+    channelwise=False, **kwargs)
+
+    2D locally connected layer
+
+    Performs an operation similar to a 2D convolution but without the weight
+    sharing, then optionally adds a bias and applies an elementwise
+    nonlinearity.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape. The
+        output of this layer should be a 4D tensor, with shape
+        ``(batch_size, num_input_channels, input_rows, input_columns)``.
+
+    num_filters : int
+        The number of learnable convolutional filters this layer has.
+
+    filter_size : int or iterable of int
+        An integer or a 2-element tuple specifying the size of the filters.
+
+    stride : int or iterable of int
+        An integer or a 2-element tuple specifying the stride of the
+        convolution operation.
+        This implementation only supports unit stride, the argument is
+        provided for compatibility to convolutional layers only.
+
+    pad : int, iterable of int, or 'valid' (default: 'same')
+        The amount of implicit zero padding of the input.
+        This implementation only supports 'same' padding, the argument is
+        provided for compatibility to other convolutional layers only.
+
+    untie_biases : bool (default: False)
+        If ``False``, the layer will have a bias parameter for each channel,
+        which is shared across all positions in this channel. As a result, the
+        `b` attribute will be a vector (1D).
+
+        If True, the layer will have separate bias parameters for each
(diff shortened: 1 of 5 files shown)


lasagne/layers/local.py
```diff
@@ -0,0 +1,196 @@
...
+
+    num_filters : int
+        The number of learnable convolutional filters this layer has.
+
+    filter_size : int or iterable of int
+        An integer or a 2-element tuple specifying the size of the filters.
+
+    stride : int or iterable of int
+        An integer or a 2-element tuple specifying the stride of the
+        convolution operation.
```
f0k: Add a note that "This implementation only supports unit stride, the argument is provided for compatibility to convolutional layers only."

lasagne/layers/local.py
```diff
@@ -0,0 +1,196 @@
...
+    def get_W_shape(self):
+        num_input_channels = self.input_shape[1]
+        output_shape = self.get_output_shape_for(self.input_shape)
+        if self.channelwise:
+            if self.channelwise and self.num_filters != num_input_channels:
+                raise ValueError("num_filters and the number of input "
+                                 "channels should match when channelwise is "
+                                 "true, but got num_filters=%r and %d input "
+                                 "channels" %
+                                 (self.num_filters, num_input_channels))
```
f0k: Hmm. Actually this should be in the constructor, but `self.input_shape` is only set when calling the super constructor, and the super constructor already calls `get_W_shape`. Probably it's okay here.

benanne:
This looks neat, thanks for contributing :)

lasagne/layers/local.py
alexlee-gk: Sure, I just updated the pull request to include this comment.

lasagne/layers/local.py
alexlee-gk: Yeah, I thought about that. At least the exception is still raised during construction.

lasagne/layers/local.py
```diff
@@ -0,0 +1,198 @@
...
+            return (self.num_filters,) + self.filter_size + output_shape[-2:]
+        else:
+            return (self.num_filters, num_input_channels) + \
+                   self.filter_size + output_shape[-2:]
+
+    def convolve(self, input, **kwargs):
+        output_shape = self.get_output_shape_for(self.input_shape)
```
f0k: `output_shape = self.output_shape` would be enough here -- we're past the constructor, so this has been set already. It was only required in `get_W_shape()`.

lasagne/layers/local.py
```diff
@@ -0,0 +1,198 @@
...
+            raise NotImplementedError(
+                "LocallyConnected2DLayer requires stride=1 / (1, 1), but got "
+                "%r." % (stride,))
+        # require same convolution
+        if self.pad != 'same':
+            raise NotImplementedError(
+                "LocallyConnected2DLayer requires pad='same', but got %r." %
+                (pad,))
+
+    def get_W_shape(self):
```
f0k: Would be cool to have a check against variable input shape here. Similar to a dense layer, we require the input shape to be known in a locally-connected layer. Something like:
```python
if any(s is None for s in self.input_shape[1:]):
    raise ValueError(
            "A LocallyConnected2DLayer requires a fixed input shape "
            "(except for the batch size). Got %r." % (self.input_shape,))
```
Otherwise, there'll be an error in creating the weight matrix that's more difficult to interpret.

f0k:
Thanks for fixing the docstring! I've had time to read through the implementation now, and have two minor comments -- if you don't want to address them, let me know and I'll address them myself.

alexlee-gk:
Thanks for the feedback! I incorporated your suggestions and updated the PR.

lasagne/tests/layers/test_local.py
```diff
@@ -0,0 +1,153 @@
...
+            LocallyConnected2DLayer(input_layer, 2, 3, stride=2)
+        assert "requires stride=1 / (1, 1)" in exc.value.args[0]
+
+    def test_invalid_settings(self, DummyInputLayer):
+        from lasagne.layers import LocallyConnected2DLayer
+        input_layer = DummyInputLayer((10, 2, 4, 4))
+        with pytest.raises(ValueError) as exc:
+            LocallyConnected2DLayer(input_layer, 4, 3, channelwise=True)
+        assert "num_filters and the number of input channels should match"\
+               in exc.value.args[0]
```
f0k: Thanks for your quick update! You'll need to add a test with `DummyInputLayer((10, None, 4, 4))` or something like this to trigger the new check for a fixed input size.

alexlee-gk: Alright, just added a test for that. The PR passes all tests now.

f0k:
Perfect! Thanks again for your clean contribution, finally merging!

PR #769, opened 2016-11-20
Support InputLayer shape given as list

### LINKED ISSUE #763: Convert InputLayer input_shape to tuple if needed
At least one user got trapped when passing a shape list instead of a shape tuple to the `InputLayer` constructor: https://github.com/Lasagne/Lasagne/commit/bc963cd12b2b7b51d81225c50077521d40aee75c#commitcomment-19594546
Many `get_output_shape_for()` implementations do not rely on the shape being a tuple, but if they do, it is hard to understand where the problem originated.
We should probably just unconditionally convert the shape to a tuple in the `InputLayer` constructor.

f0k:
Closes #763.

### DIFF
--- a/lasagne/layers/input.py
+++ b/lasagne/layers/input.py
@@ -48,7 +48,7 @@ class InputLayer(Layer):
     >>> l_in = InputLayer((100, 20))
     """
     def __init__(self, shape, input_var=None, name=None, **kwargs):
-        self.shape = shape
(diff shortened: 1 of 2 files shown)


benanne:
LGTM, merging :)

f0k:
Thank you for reviewing!

benanne:
There wasn't exactly a lot to review there :)

PR #770, opened 2016-11-20
Allow vector targets for column vector predictions in objectives

### LINKED ISSUE #755: squared_error: Warn if comparing column vector against row vector
At least one user got trapped by setting target_var to a vector for a regression task, and then passing it to lasagne.objectives.squared_error along with the network predictions for a network of a single output unit: https://groups.google.com/forum/#!topic/lasagne-users/DsSY1cHCC2M

Due to #715, Theano knows that the network predictions are a column vector (a matrix with broadcast pattern (False, True)), so it will happily broadcast the network predictions and the targets into a square matrix of (batchsize, batchsize) and compute the squared differences between every prediction and every target (as opposed to computing the squared differences between corresponding predictions and targets only). The mean of the loss is still a scalar, so training will appear to work, but not learn anything.

We should catch this case and warn the user to use a column vector for the squared error targets.

f0k:
This changes behaviour of lasagne.objectives.* for a special case: If predictions are a column vector (i.e., a matrix that is broadcastable in the second dimension) and targets are a 1D vector, implicitly turn the targets into a column vector as well. This avoids the current trap (opened in #715) that combining a network with a single output unit and a T.vector() target broadcasts predictions and targets to compare all predictions against all targets. Closes #755, but we need to discuss whether we want this. I'd argue that it seems so natural to use T.vector() for regression or binary classification targets that we should just support it. The alternative would be to warn the user to use a T.matrix() or T.col() instead.

Trapped users:
https://groups.google.com/forum/#!topic/lasagne-users/DsSY1cHCC2M
https://groups.google.com/forum/#!topic/lasagne-users/_JAgvyhpf7Q

### DIFF
--- a/lasagne/objectives.py
+++ b/lasagne/objectives.py
@@ -77,7 +77,7 @@ This gives a loss expression good for monitoring validation error.
 
 import theano.tensor.nnet
 
-from lasagne.layers import get_output
+from .utils import as_theano_expression
 
 __all__ = [
     "binary_crossentropy",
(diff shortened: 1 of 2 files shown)


benanne:
Looks good, I think this is a very common mistake so I'm in favour. Can you think of any reasons why we wouldn't want this? Are there any use cases where the current behaviour is desired, or does it complicate things when one wants to achieve the current behaviour at all? If there are any they are probably very rare.

It adds a little bit of complexity, but it's only one additional line and the purpose is clear. So I have no objections.

f0k:
> Are there any use cases where the current behaviour is desired, or does it complicate things when one wants to achieve the current behaviour at all?

No, not really. To achieve the current behaviour, you would need to define the target vector as a `T.row()` or dimshuffle it accordingly, or use e.g. `T.nnet.binary_crossentropy()` directly. But this is only needed for special use cases, when you want to define some kind of contrastive loss, for example (where you'd contrast the main diagonal against off-diagonal elements). I think closing the trap is more important.

> Feel free to merge :)

Thank you!

PR #771, opened 2016-11-20
binary hinge loss: Add log_odds argument

### LINKED ISSUE #744: binary hinge loss
Hi,
In the code for binary hinge loss, the targets are normalized to {-1,1}. 
Shouldn't the same apply for the predictions if they are in (0,1)? I mean, the predictions should also be normalized to (-1,1).

f0k:
The `binary_hinge_loss` looks like it would be a drop-in replacement for `binary_crossentropy`, but it isn't: It requires the log odds of the network predictions (i.e., the sigmoid input, not the sigmoid output). This PR documents this and adds a `log_odds` argument. Similar to the existing `binary` argument which allows using the objective both with 0/1 and +/-1 targets, it allows to use the objective both with sigmoid outputs and sigmoid inputs. Closes #744.

### DIFF
--- a/lasagne/objectives.py
+++ b/lasagne/objectives.py
@@ -254,7 +254,8 @@ def binary_hinge_loss(predictions, targets, binary=True, delta=1):
     Parameters
     ----------
     predictions : Theano tensor
-        Predictions in (0, 1), such as sigmoidal output of a neural network.
+        Log-odds of predictions, such as the input to a sigmoidal layer in
+        a neural network.
     targets : Theano tensor
         Targets in {0, 1} (or in {-1, 1} depending on `binary`), such as
         ground truth labels.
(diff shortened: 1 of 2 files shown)


benanne:
I'm on the fence as to whether we want to change the default behaviour. I guess it's more natural, but as you point out it can also be unstable. Torch and TensorFlow both seem to prefer having the network output log-odds, even when paired with cross-entropy. In Torch you put a "LogSoftmax" on top of it, in TF you just use the "softmax_crossentropy_with_logits" loss function or its variants, which takes care of things.

For beginners this is probably less straightforward, but maybe it's something that people are getting used to?

A related concern is that this potential instability introduces a failure case that will be hard to understand precisely for these beginners, so maybe we're not doing them a favour by changing this default after all...

Anyway, as I said, I'm on the fence, I think I have no strong opinion either way. Just something worth thinking about :)

f0k:
> actually one thing I didn't take into account is your plan to add an optimisation to Theano to take care of the instability. In that case I'm definitely in favour.

Yes, we'd only change the default when Theano can cope with it (and then we'd update the requirements.txt along with changing the default). Sorry, I should have linked https://github.com/Lasagne/Lasagne/issues/744#issuecomment-261814604 here!

> Torch and TensorFlow both seem to prefer having the network output log-odds, even when paired with cross-entropy. In Torch you put a "LogSoftmax" on top of it, in TF you just use the "softmax_crossentropy_with_logits" loss function or its variants, which takes care of things.

In Theano, this happens under the hood. Cross-entropy of a softmax or sigmoid gets optimized to use log-softmax or log-sigmoid, respectively. Since all our objective functions expect the softmax or sigmoid output, it would be nice to have the hinge losses accept that, too!

By the way, we should look at the categorical hinge loss as well...

Issue #777, opened 2016-12-02
`CustomRecurrentLayer` does not allow hidden-to-hidden sub-net to have arbitrary depth


Britefury:
When using a `CustomRecurrentLayer` the hidden-to-hidden network can only have one layer besides the input, unless the second last layer has the same output dimensions as output shape. 

PR #778 addresses this problem.

Here is some code based on the unit tests that triggers the bug:

```py
    # Check that the custom recurrent layer can handle a hidden-to-hidden
    # network with an arbitrary depth
    n_batch, n_steps, n_channels, width, height = (2, 3, 4, 5, 6)
    n_out_filters = 7
    n_in_hid_filters_0 = 11
    n_hid_hid_filters_0 = 13
    filter_shape = (3, 3)
    l_in = lasagne.layers.InputLayer(
        (n_batch, n_steps, n_channels, width, height))

    # Expect the output shape of `l_in` as input shape for input-to-hidden
    l_in_to_hid = lasagne.layers.InputLayer((None, n_channels, width, height))
    # Two conv layers; first to `n_hid_filters_0` channels
    l_in_to_hid = lasagne.layers.Conv2DLayer(
        l_in_to_hid, n_in_hid_filters_0, filter_shape, pad='same')
    # then to `n_out_filters` channels
    l_in_to_hid = lasagne.layers.Conv2DLayer(
        l_in_to_hid, n_out_filters, filter_shape, pad='same')

    # Expect the output shape of `l_in_to_hid` as input shape for
    # hidden-to-hidden
    l_hid_to_hid = lasagne.layers.InputLayer((None, n_out_filters,
                                              width, height))
    # Two conv layers; first to `n_hid_hid_filters_0` channels
    l_hid_to_hid = lasagne.layers.Conv2DLayer(
        l_hid_to_hid, n_hid_hid_filters_0, filter_shape, pad='same')
    # then to `n_out_filters` channels
    l_hid_to_hid = lasagne.layers.Conv2DLayer(
        l_hid_to_hid, n_out_filters, filter_shape, pad='same')

    # --- ValueError raised HERE ---
    l_rec = lasagne.layers.CustomRecurrentLayer(
        l_in, l_in_to_hid, l_hid_to_hid)

    assert l_rec.output_shape == (n_batch, n_steps, n_out_filters, width,
                                  height)

    out = theano.function([l_in.input_var], lasagne.layers.get_output(l_rec))
[...]

PR #778, opened 2016-12-02
`CustomRecurrentLayer` arbitrary depth layer stacks

Britefury:
This PR addresses issue #777, allowing the input-to-hidden and hidden-to-hidden layer stacks to have arbitrary depth (with unit tests).

### DIFF
--- a/lasagne/layers/recurrent.py
+++ b/lasagne/layers/recurrent.py
@@ -222,6 +222,23 @@ class CustomRecurrentLayer(MergeLayer):
 
         super(CustomRecurrentLayer, self).__init__(incomings, **kwargs)
 
+        input_to_hidden_in_layers = \
+            [layer for layer in helper.get_all_layers(input_to_hidden)
+             if isinstance(layer, InputLayer)]
+        if len(input_to_hidden_in_layers) != 1:
+            raise ValueError(
+                '`input_to_hidden` must have exactly one InputLayer, but it '
+                'has {}'.format(len(input_to_hidden_in_layers)))
+
+        hidden_to_hidden_in_lyrs = \
+            [layer for layer in helper.get_all_layers(hidden_to_hidden)
+             if isinstance(layer, InputLayer)]
+        if len(hidden_to_hidden_in_lyrs) != 1:
+            raise ValueError(
+                '`hidden_to_hidden` must have exactly one InputLayer, but it '
+                'has {}'.format(len(hidden_to_hidden_in_lyrs)))
+        hidden_to_hidden_in_layer = hidden_to_hidden_in_lyrs[0]
+
         self.input_to_hidden = input_to_hidden
         self.hidden_to_hidden = hidden_to_hidden
         self.learn_init = learn_init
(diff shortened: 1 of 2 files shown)


f0k:
Good spot! But if I see correctly, the actual problem seems to be this line only: https://github.com/Lasagne/Lasagne/blob/0abf57f/lasagne/layers/recurrent.py#L299
Can't we fix this more easily by just computing the input shape of `hidden_to_hidden` in a correct way, without adding more attributes and changing `get_output_for`?

Britefury:
Good call. That additional code was in there because I also attempted to allow the input-to-hidden and hidden-to-hidden graphs to have arbitrary structure, including branching back to the input graph. This turned out not to work as it was a bad idea.

There is still a little additional sanity checking code at the beginning of the constructor that checks that the input-to-hidden and hidden-to-hidden graphs contain a unique `InputLayer`, just in case anyone tries anything unwise. It can be a bugger to find the root cause if you do...

lasagne/layers/recurrent.py
```diff
@@ -222,6 +222,23 @@ def __init__(self, incoming, input_to_hidden, hidden_to_hidden,
 
         super(CustomRecurrentLayer, self).__init__(incomings, **kwargs)
 
+        input_to_hidden_layers = helper.get_all_layers(input_to_hidden)
+        input_to_hidden_in_lyrs = [lyr for lyr in input_to_hidden_layers
+                                   if isinstance(lyr, InputLayer)]
```
f0k: Can you combine this into one call? The fewer local variables are initialized, the less I need to worry about when reading the code. I'd also prefer not to use abbreviations:
```python
input_to_hidden_input_layers = [layer for layer
                                in helper.get_all_layers(input_to_hidden)
                                if isinstance(layer, InputLayer)]
```

lasagne/layers/recurrent.py
```diff
@@ -222,6 +222,23 @@ def __init__(self, incoming, input_to_hidden, hidden_to_hidden,
...
+        if len(input_to_hidden_in_lyrs) != 1:
+            raise ValueError('input_to_hidden_in_layer '
+                             'not provided and input_to_hidden does not '
+                             'have a unique InputLayer')
```
f0k: This indicates you could provide the input layer in some way, but you can't. I'd say:
```
`input_to_hidden` must have exactly one InputLayer, but it has %d
```

lasagne/layers/recurrent.py
```diff
@@ -294,16 +311,16 @@ def __init__(self, incoming, input_to_hidden, hidden_to_hidden,
...
-                             "input_to_hidden.output_shape={} and "
-                             "hidden_to_hidden.input_shape={}".format(
-                                 input_to_hidden.output_shape,
-                                 hidden_to_hidden.input_shape))
+                                     h_to_h_input_shape[1:])):
+            raise ValueError(
+                "The output shape for input_to_hidden must be equal to the "
+                "input shape of hidden_to_hidden after the first dimension, "
+                "but input_to_hidden.output_shape={} and "
+                "hidden_to_hidden:input_layer.output_shape={}".format(
```
f0k: Maybe `hidden_to_hidden:input_layer.shape` is easier to grasp? I don't know. `InputLayer` has a `shape` property that may feel more natural than `output_shape`.

lasagne/tests/layers/test_recurrent.py
```diff
@@ -189,7 +189,129 @@ def test_custom_recurrent_arbitrary_shape():
...
+
+    out = theano.function([l_in.input_var], lasagne.layers.get_output(l_rec))
+    out_shape = out(np.zeros((n_batch, n_steps, n_channels, width, height),
+                             dtype=theano.config.floatX)).shape
+    assert out_shape == (n_batch, n_steps, n_out_filters, width, height)
+
+
+def test_custom_recurrent_non_unique_inputs():
+    # Check that the custom recurrent layer can handle a hidden-to-hidden
+    # network with an arbitrary depth
```
f0k: This doesn't seem like the correct comment for this test.

Britefury:
@f0k I have addressed your comments and change requests; I hope its in a good enough state now.

f0k:
Looks good, thank you! Can you squash into a single commit again? We're trying to have a clean history. Cheers!

Britefury:
Done.

f0k:
Merged, thanks again!

PR #785, opened 2016-12-23
Fixed mask shape to suit arbitrary number of feature dimensions

TobyPDE:
If you use the `CustomRecurrentLayer` in order to perform a recurrent convolution, then the masking did not work properly because the masking variable assumed input to have the shape `(batch_size, seq_len, num_features)`. This PR fixes this bug and makes masking work with an arbitrary number of feature dimensions.

### DIFF
--- a/lasagne/layers/recurrent.py
+++ b/lasagne/layers/recurrent.py
@@ -450,7 +450,13 @@ class CustomRecurrentLayer(MergeLayer):
             return [hid_out]
 
         if mask is not None:
-            mask = mask.dimshuffle(1, 0, 'x')
+            # After processing, the input has shape
+            # (seq_len, batch_size, feature_dim_0, feature_dim_1, ..., feature_dim_N)
+            # We have to determine N, which is the number of feature dimensions in order to adjust the shape of the
+            # masking variable to
+            # (seq_len, batch_size, 1, ..., 1)
+            num_feature_dims = input.ndim - 2
+            mask = mask.dimshuffle((1, 0) +  ('x',) * num_feature_dims)
             sequences = [input, mask]
             step_fun = step_masked
         else:
--- a/lasagne/layers/recurrent.py
+++ b/lasagne/layers/recurrent.py
@@ -451,12 +451,13 @@ class CustomRecurrentLayer(MergeLayer):
 
         if mask is not None:
             # After processing, the input has shape
-            # (seq_len, batch_size, feature_dim_0, feature_dim_1, ..., feature_dim_N)
-            # We have to determine N, which is the number of feature dimensions in order to adjust the shape of the
-            # masking variable to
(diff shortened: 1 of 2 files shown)


lasagne/layers/recurrent.py
```diff
@@ -450,7 +450,14 @@ def step_masked(input_n, mask_n, hid_previous, *args):
...
         if mask is not None:
-            mask = mask.dimshuffle(1, 0, 'x')
+            # After processing, the input has shape
+            # (seq_len, batch_size, f_0, f_1, ..., f_N)
+            # We have to determine N, which is the number of feature dimensions
+            # in order to adjust the shape of the masking variable to
+            # (seq_len, batch_size, 1, ..., 1)
+            #                       |---N---|
+            num_feature_dims = input.ndim - 2
+            mask = mask.dimshuffle((1, 0) + ('x',) * num_feature_dims)
```
f0k: Good catch! To make the implementation simpler, though, could you instead do:
```python
if mask is not None:
    # expand mask from (seq_len, batch_size) to cover all feature dimensions
    mask = T.shape_padright(mask, input.ndim - 2)
```
This is easier to read and also doesn't require a long comment to explain!

Reading again, your implementation is actually wrong -- the mask needs to match the dimensionality of the hidden states, not the input states. So instead of `input.ndim - 2`, use `len(self.input_to_hidden.output_shape) - 2`. Putting this together:
```python
if mask is not None:
    # expand mask from (seq_len, batch_size) to cover all hidden dimensions
    hid_ndim = len(self.input_to_hidden.output_shape)
    mask = T.shape_padright(mask, hid_ndim - 2)
```

Could you update your PR accordingly? This would be great!

lasagne/tests/layers/test_recurrent.py
```diff
@@ -189,6 +189,34 @@ def test_custom_recurrent_arbitrary_shape():
...
+    filter_shape = (3, 3)
+    l_in = lasagne.layers.InputLayer(
+        (n_batch, n_steps, n_channels, width, height))
+    l_mask = lasagne.layers.InputLayer((n_batch, n_steps))
+    l_in_to_hid = lasagne.layers.Conv2DLayer(
+        lasagne.layers.InputLayer((None, n_channels, width, height)),
+        n_out_filters, filter_shape, pad='same')
+    l_hid_to_hid = lasagne.layers.Conv2DLayer(
+        lasagne.layers.InputLayer((None, n_out_filters, width, height)),
+        n_out_filters, filter_shape, pad='same')
```
f0k: Could you modify this test such that the hidden dimensionality is different from the input dimensionality? This should uncover that your current fix of the mask dimensionality is wrong. I'm not sure there's a sensible use case for this -- maybe just reshape l_in_to_hid to 5 dimensions and then use a DenseLayer(..., num_leading_dims=-1) for hid_to_hid so the dimensionality stays at 5.

Issue #782, opened 2016-12-13
TransformerLayer: unexpected shear artifact for non-rectangular images


v-ablavsky:
Given a non-square input (see attached examples) and a transform vector that specifies rotation and translation (i.e., a rigid transform), the transformed image computed by TransformerLayer appears to exhibit a shear artifact.

Attached files: 
(1) expected_sampling_grid.pdf <- sampling grid computed using my hand-crafted
method that differs from the implementation in lasagne/layers/special.py

(2) unexpected_transformed_image.pdf <-- an output of SpatialTransformer that shows shear, even though the input transform was a rigid transform.

I conjecture that the issue stems from the implementation of scaling
in _interpolate() "# scale coordinates from [-1, 1] to [0, width/height -1]"

[expected_sampling_grid.pdf](https://github.com/Lasagne/Lasagne/files/649488/expected_sampling_grid.pdf)
[unexpected_transformed_image.pdf](https://github.com/Lasagne/Lasagne/files/649489/unexpected_transformed_image.pdf)

f0k:
Interesting. Possibly it wasn't originally developed for rectangular images. A network should be able to just learn to counter that, but it's still annoying you can't easily use it to apply a certain fixed transformation then (e.g., for data augmentation). Can you provide a simple script demonstrating the problem? If you have already looked into the details of how to create the sampling grid, do you know how this would be fixed?

v-ablavsky:
Thanks for getting back to me so quickly! I am attaching a stand-alone script [issue_782.py packaged in a .zip file] that demonstrates this unexpected behavior. I will follow up tomorrow with some suggestions on how to address this issue.
[issue_782.zip](https://github.com/Lasagne/Lasagne/files/650634/issue_782.zip)

v-ablavsky:
Here is my code to generate a sampling grid that does not exhibit unexpected/undesired shear artifacts.
The key idea is to apply scaling of grid coordinates from [-1,1] to pixel space *before* applying Theta (the affine transform input to the TransformerLayer).
The code below is not elegant, and any suggestions on making it simpler would be appreciated.
As to integrating this computation into special.py, I'd propose to put it in _transform_affine( ), and be sure to remove the scaling computations from _interpolate( )

=========================================
def compute_sampling_grid(theta, imgs, downsample_factor):
    """
    Deep-dive into the computations performed by
    lasagne.layers.special.TransformerLayer in particular, this function
    pulls together code from
[...]

PR #793, opened 2017-01-28
Additional border_modes for _interpolate.

### LINKED ISSUE #689: Smearing issue with TransformerLayer
It seems to me as if the pixels that lie on the boundary of the input to the TransformerLayer are smeared in the output.  I noticed this when I was playing around with spatial transformers by transforming MNIST digits and trying to let the transformer module learn to map the digit back to its original pose.

In the image below, the first row shows the input, the second row shows the output of the spatial transformer, and the third row shows the output of the spatial transformer after the boundary of the input has been set to zero.  Notice that in the second row smearing occurs wherever the input digit was touching the boundary, whereas the third row does not exhibit this effect.
smearing

I suspect that the issue arises when the coordinates are [clipped](https://github.com/Lasagne/Lasagne/blob/8fe645d28b66f991d547e9b6a314251b8e84446a/lasagne/layers/special.py#L470), because this leads to several locations in the output grid all mapping back to the same boundary location in the  input.

Here is some code to reproduce the issue.  You can get a sample of transformed MNIST digits [here](https://www.dropbox.com/s/ayxt5jkxtvekri3/transformed_mnist.pickle?dl=0).

```
import cv2
import cPickle as pickle
import numpy as np
import theano
import theano.tensor as T
from lasagne.layers import get_output
from lasagne.layers import DenseLayer
from lasagne.layers import InputLayer
from lasagne.layers import TransformerLayer
from lasagne.init import Constant
from lasagne.nonlinearities import linear

with open('transformed_mnist.pickle', 'rb') as f:
    X = pickle.load(f)
Y = np.copy(X)
Y[:, :, 0, :] = 0
Y[:, :, :, 0] = 0
Y[:, :, -1, :] = 0
Y[:, :, :, -1] = 0

l_in = InputLayer((None, 1, 28, 28))

M = np.array([[1.5, 0, 0.0], [0, 1.5, 0.0]])
l_loc = DenseLayer(
    l_in, num_units=6, W=Constant(0.0), b=M.flatten(), nonlinearity=linear
)
l_trans = TransformerLayer(l_in, l_loc)

X_tensor = T.tensor4()
X_hat = get_output(l_trans, X_tensor)
infer = theano.function([X_tensor], X_hat)

output1 = infer(X)
output2 = infer(Y)
row1 = (255. * np.hstack(X.transpose(0, 2, 3, 1)))
row2 = (255. * np.hstack(output1.transpose(0, 2, 3, 1)))
row3 = (255. * np.hstack(output2.transpose(0, 2, 3, 1)))
cv2.imwrite('smearing.png', np.vstack((row1, row2, row3)))
```

hjweide:
Closes #689.  Add additional border modes for _interpolate to be consistent with scipy.

### DIFF
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -376,6 +376,12 @@ class TransformerLayer(MergeLayer):
         output image (in both spatial dimensions). A value of 1 will keep the
         original size of the input. Values larger than 1 will downsample the
         input. Values below 1 will upsample the input.
+    border_mode : 'nearest', 'mirror', or 'wrap'
+        Determines how border conditions are handled during interpolation.  If
+        'nearest', points outside the grid are clipped to the boundary. If
+        'mirror', points are mirrored across the boundary. If 'wrap', points
+        wrap around to the other side of the grid.  See
+        http://stackoverflow.com/q/22669252/22670830#22670830 for details.
 
     References
     ----------
@@ -405,10 +411,11 @@ class TransformerLayer(MergeLayer):
     >>> l_trans = lasagne.layers.TransformerLayer(l_in, l_loc)
     """
     def __init__(self, incoming, localization_network, downsample_factor=1,
-                 **kwargs):
+                 border_mode='nearest', **kwargs):
         super(TransformerLayer, self).__init__(
             [incoming, localization_network], **kwargs)
         self.downsample_factor = as_tuple(downsample_factor, 2)
+        self.border_mode = border_mode
 
         input_shp, loc_shp = self.input_shapes
 
@@ -429,10 +436,11 @@ class TransformerLayer(MergeLayer):
     def get_output_for(self, inputs, **kwargs):
         # see eq. (1) and sec 3.1 in [1]
         input, theta = inputs
-        return _transform_affine(theta, input, self.downsample_factor)
+        return _transform_affine(theta, input, self.downsample_factor,
+                                 self.border_mode)
 
 
-def _transform_affine(theta, input, downsample_factor):
+def _transform_affine(theta, input, downsample_factor, border_mode):
     num_batch, num_channels, height, width = input.shape
     theta = T.reshape(theta, (-1, 2, 3))
 
@@ -452,7 +460,7 @@ def _transform_affine(theta, input, downsample_factor):
     input_dim = input.dimshuffle(0, 2, 3, 1)
     input_transformed = _interpolate(
         input_dim, x_s_flat, y_s_flat,
-        out_height, out_width)
+        out_height, out_width, border_mode)
 
     output = T.reshape(
         input_transformed, (num_batch, out_height, out_width, num_channels))
@@ -460,15 +468,29 @@ def _transform_affine(theta, input, downsample_factor):
     return output
 
 
-def _interpolate(im, x, y, out_height, out_width):
+def _interpolate(im, x, y, out_height, out_width, border_mode):
     # *_f are floats
     num_batch, height, width, channels = im.shape
     height_f = T.cast(height, theano.config.floatX)
     width_f = T.cast(width, theano.config.floatX)
 
     # clip coordinates to [-1, 1]
-    x = T.clip(x, -1, 1)
-    y = T.clip(y, -1, 1)
+    if border_mode == 'nearest':
+        x = T.clip(x, -1, 1)
+        y = T.clip(y, -1, 1)
+    # 0.9 1.0 1.1 -> 0.9 1.0 0.9
+    elif border_mode == 'mirror':
+        xa = T.mod(x + 1, 4) - 1
+        ya = T.mod(y + 1, 4) - 1
+        x = T.minimum(xa, 2 - xa)
+        y = T.minimum(ya, 2 - ya)
+    # 0.9 1.0 1.1 -> 0.9 1.0 -0.9
+    elif border_mode == 'wrap':
+        x = T.mod(x + 1, 2) - 1
+        y = T.mod(y + 1, 2) - 1
+    else:
+        raise ValueError("border_mode must be one of "
+                         "'nearest', 'mirror', 'wrap'")
 
     # scale coordinates from [-1, 1] to [0, width/height - 1]
     x = (x + 1) / 2 * (width_f - 1)
@@ -590,6 +612,12 @@ class TPSTransformerLayer(MergeLayer):
         computed as part of the Theano computational graph, which is
         substantially slower as this computation scales with
         num_pixels*num_control_points. Default is 'auto'.
+    border_mode : 'nearest', 'mirror', or 'wrap'
+        Determines how border conditions are handled during interpolation.  If
+        'nearest', points outside the grid are clipped to the boundary'. If
+        'mirror', points are mirrored across the boundary. If 'wrap', points
+        wrap around to the other side of the grid.  See
+        http://stackoverflow.com/q/22669252/22670830#22670830 for details.
 
     References
     ----------
@@ -633,10 +661,12 @@ class TPSTransformerLayer(MergeLayer):
     """
 
     def __init__(self, incoming, localization_network, downsample_factor=1,
-                 control_points=16, precompute_grid='auto', **kwargs):
+                 control_points=16, precompute_grid='auto',
+                 border_mode='nearest', **kwargs):
         super(TPSTransformerLayer, self).__init__(
                 [incoming, localization_network], **kwargs)
 
+        self.border_mode = border_mode
         self.downsample_factor = as_tuple(downsample_factor, 2)
         self.control_points = control_points
 
@@ -686,12 +716,12 @@ class TPSTransformerLayer(MergeLayer):
         return _transform_thin_plate_spline(
                 dest_offsets, input, self.right_mat, self.L_inv,
                 self.source_points, self.out_height, self.out_width,
-                self.precompute_grid, self.downsample_factor)
+                self.precompute_grid, self.downsample_factor, self.border_mode)
 
 
 def _transform_thin_plate_spline(
         dest_offsets, input, right_mat, L_inv, source_points, out_height,
-        out_width, precompute_grid, downsample_factor):
+        out_width, precompute_grid, downsample_factor, border_mode):
 
     num_batch, num_channels, height, width = input.shape
     num_control_points = source_points.shape[1]
@@ -732,7 +762,7 @@ def _transform_thin_plate_spline(
     input_dim = input.dimshuffle(0, 2, 3, 1)
     input_transformed = _interpolate(
             input_dim, x_transformed, y_transformed,
-            out_height, out_width)
+            out_height, out_width, border_mode)
 
     output = T.reshape(input_transformed,
                        (num_batch, out_height, out_width, num_channels))
--- a/lasagne/tests/layers/test_special.py
+++ b/lasagne/tests/layers/test_special.py
@@ -416,6 +416,46 @@ class TestTransformLayer():
                                         constant(thetas)]).eval()
         np.testing.assert_allclose(inputs, outputs, rtol=1e-6)
 
+    def test_transform_border_modes(self):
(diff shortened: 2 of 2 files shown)


lasagne/layers/special.py
```diff
@@ -633,10 +649,12 @@ class TPSTransformerLayer(MergeLayer):
     """
 
     def __init__(self, incoming, localization_network, downsample_factor=1,
-                 control_points=16, precompute_grid='auto', **kwargs):
+                 border_mode='nearest', control_points=16,
+                 precompute_grid='auto', **kwargs):
```
f0k: Please add the new argument in the end, to keep compatibility with existing code that uses positional arguments.

lasagne/layers/special.py
```diff
@@ -452,23 +454,37 @@ def _transform_affine(theta, input, downsample_factor):
...
     width_f = T.cast(width, theano.config.floatX)
 
     # clip coordinates to [-1, 1]
-    x = T.clip(x, -1, 1)
-    y = T.clip(y, -1, 1)
+    if border_mode == 'nearest':
+        x = T.clip(x, -1, 1)
+        y = T.clip(y, -1, 1)
+    # 0.9 1.0 1.1 -> 0.9 1.0 0.9
+    elif border_mode == 'reflect':
```
f0k: Looking at http://stackoverflow.com/a/22670830, I think this should be `'mirror'` instead. `reflect` is more difficult to implement since it depends on the discrete pixel positions, we won't need this.

lasagne/layers/special.py
```diff
@@ -452,23 +454,37 @@ def _transform_affine(theta, input, downsample_factor):
...
+        xa = T.maximum(x, -(2 + x))
+        ya = T.maximum(y, -(2 + y))
+        x = T.minimum(xa, 2 - xa)
+        y = T.minimum(ya, 2 - ya)
```
f0k: Almost! It needs to be periodic just like `wrap`. E.g., when you go further, 2.9 should be mapped to -.9 (which works), 3.0 should be mapped to -1 (which works), and 3.1 should be mapped to -.9 again (mapped to -1.1 now).

hjweide:
Getting ```mirror``` right was a little trickier than I first thought, but I think I managed to make it consistent with scipy.  I've also changed the order of the arguments and added docstrings.  If those all look good to you I can add some tests too.

lasagne/layers/special.py
```diff
@@ -376,6 +376,12 @@ class TransformerLayer(MergeLayer):
         output image (in both spatial dimensions). A value of 1 will keep the
         original size of the input. Values larger than 1 will downsample the
         input. Values below 1 will upsample the input.
+    border_mode : 'nearest', 'mirror', or 'wrap'
+        Determines how border conditions are handled during interpolation.  If
+        'nearest', points outside the grid are clipped to the boundary. If
+        'mirror', points are mirrored across the boundary. If 'wrap', points
+        wrap around to the other side of the grid.  See
+        http://stackoverflow.com/q/22669252/22670830#22670830 for details.
```
f0k: Not sure if we want to refer to stackoverflow there... but if we do, we could use a shorter link: `See http://stackoverflow.com/a/22670830 for an illustration.`

lasagne/layers/special.py
```diff
@@ -452,23 +460,37 @@ def _transform_affine(theta, input, downsample_factor):
...
-    y = T.clip(y, -1, 1)
+    if border_mode == 'nearest':
+        x = T.clip(x, -1, 1)
+        y = T.clip(y, -1, 1)
+    # 0.9 1.0 1.1 -> 0.9 1.0 0.9
+    elif border_mode == 'mirror':
+        xa = T.mod(x + 1, 4) - 1
+        ya = T.mod(y + 1, 4) - 1
+        x = T.minimum(xa, 2 - xa)
+        y = T.minimum(ya, 2 - ya)
```
f0k: Perfect! 👍

hjweide:
Any suggestion how to test the ```_interpolate``` function?  It returns the image itself, so to test it we'd need to reproduce the effect of the border modes on the input image.

f0k:
> It returns the image itself, so to test it we'd need to reproduce the effect of the border modes on the input image.

Hmm, that would basically require to reimplement `_interpolate` in numpy. Maybe it's easier to test the TransformerLayer instead, with a fixed transform so the output image is very easy to compute in numpy. Something like an input of [0 0 0 0 0 0 0 0 1 1 1 1 1 1 1 1], horizontally squashed to 25% (but still filling 16 pixels), thresholded to make it binary again, and then we can check if the border mode was respected (each border mode will result in another simple pattern). Does that sound reasonable?

hjweide:
Thanks for the suggestion.  I've added tests to get the coverage back up to 100%.

f0k:
> I've added tests to get the coverage back up to 100%.

Yay! Nice and clean. I'm surprised the patterns become a bit irregular, but we probably can't do anything about this. Must be some unfortunate rounding.

As a sanity check, I tried an MNIST digit (loaded as single-image batch `x`):
```python
trans = np.array([[4, 0, 0, 0, 2, 0]], dtype=np.float32)
for border_mode in 'nearest', 'mirror', 'wrap':
    plt.figure()
    plt.imshow(lasagne.layers.TransformerLayer(x.shape, trans.shape, border_mode=border_mode, downsample_factor=(.5, .25)).get_output_for((T.constant(x), T.constant(trans))).eval()[0,0].T, cmap='gray', interpolation='none')
```
Looks nice (in order: nearest, mirror, wrap):
nearest mirror wrap

Again, upon close inspection, the mirrored/wrapped versions are not exactly identical to the original in the center (and also the version in the center is not exactly identical to the untransformed digit), but that's not the business of this PR.

Very good job! :+1: Will merge tomorrow if there are no further comments.

f0k:
Merged, thanks again for your great work!

hjweide:
Happy to help!  I think setting pixels outside the grid to a constant value would be a useful border mode to have, but I'm not sure how that would be implemented.  Perhaps I'll get back to it in the future.

f0k:
> I think setting pixels outside the grid to a constant value would be a useful border mode to have, but I'm not sure how that would be implemented.

I guess you just wouldn't clip the coordinates, and then use a `T.switch` when fetching the pixel values to replace out-of-image values with a given constant value here: https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/special.py#L527-L530

f0k:
Hmm, and looking at the code again, wouldn't we have to respect the border mode here as well? https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/special.py#L508-L509

hjweide:
Indeed you're right, no idea how I missed that.  I've implemented what I think is the correct fix, and will be pushing it momentarily.  This also seems to take care of the irregular patterns we were seeing.

f0k:
> Indeed you're right, no idea how I missed that.

Well, I also missed it, your solution was very convincing :)

> Turns out getting mirror right is a little tricky. I'll get back to it soon.

Thank you for taking this! I think the correct solution would be to change the `x0 = T.cast(x0_f)` and the other three conversions to respect the border mode, and I think you can use the code you have, just using `width_f` or `height_f` instead of `1` (and `4 * width_f` instead of `4`). Placing the border mode code there will also make it easier to change the grid domain as suggested in #800 to fix #782.

hjweide:
My first attempt was just to replace the previous width/height of 2 (from -1...1) with width_f/height_f as appropriate, but that leads to strange results for the mirror case.
```
x0a = T.mod(x0_f + width_f / 2, 2 * width_f) - width_f / 2
y0a = T.mod(y0_f + height_f / 2, 2 * height_f) - height_f / 2
x0 = T.minimum(x0a, width_f - x0a)
y0 = T.minimum(y0a, height_f - y0a)

x1a = T.mod(x1_f + width_f / 2, 2 * width_f) - width_f / 2
y1a = T.mod(y1_f + height_f / 2, 2 * height_f) - height_f / 2
x1 = T.minimum(x1a, width_f - x1a)
y1 = T.minimum(y1a, height_f - y1a)
```
Not sure if I'm missing something obvious but I may need to redo the calculations.
mirror

f0k:
You're right, the formulas need to change a bit. I have the following:
```python
    # obtain indices of the 2x2 pixel neighborhood surrounding the coordinates;
    # we need those in floatX for interpolation and in int64 for indexing.
    x0_f = T.floor(x)
    y0_f = T.floor(y)
    x1_f = x0_f + 1
    y1_f = y0_f + 1

    # for indexing, we need to take care of the border mode for outside pixels.
    if border_mode == 'nearest':
        x0 = T.clip(T.cast(x0_f, 'int64'), 0, width - 1)
        x1 = T.clip(T.cast(x1_f, 'int64'), 0, width - 1)
        y0 = T.clip(T.cast(y0_f, 'int64'), 0, height - 1)
        y1 = T.clip(T.cast(y1_f, 'int64'), 0, height - 1)
    elif border_mode == 'mirror':
        xa0 = T.mod(x0_f, 2 * width_f)
        xa1 = T.mod(x1_f, 2 * width_f)
        ya0 = T.mod(y0_f, 2 * height_f)
        ya1 = T.mod(y1_f, 2 * height_f)
        x0 = T.cast(T.minimum(xa0, 2 * width_f - 1 - xa0), 'int64')
        x1 = T.cast(T.minimum(xa1, 2 * width_f - 1 - xa1), 'int64')
        y0 = T.cast(T.minimum(ya0, 2 * height_f - 1 - ya0), 'int64')
        y1 = T.cast(T.minimum(ya1, 2 * height_f - 1 - ya1), 'int64')
    elif border_mode == 'wrap':
        x0 = T.cast(T.mod(x0_f, width_f), 'int64')
        x1 = T.cast(T.mod(x1_f, width_f), 'int64')
        y0 = T.cast(T.mod(y0_f, height_f), 'int64')
        y1 = T.cast(T.mod(y1_f, height_f), 'int64')
    else:
        raise ValueError("border_mode must be one of "
                         "'nearest', 'mirror', 'wrap'")
```
Not 100% sure about the `-1`.

hjweide:
Previously the grid ended at ```1```, so we mirrored about ```2``` before taking the minimum.  Now the grid ends at ```width_f - 1```, so does that mean we should mirror about ```2 * (width_f - 1)```?

f0k:
> Now the grid ends at `width_f - 1`, so does that mean we should mirror about `2 * (width_f - 1)`?

Well, there's a clear definition for the `mirror` mode at http://stackoverflow.com/a/22670830. My solution above might be the `reflect` mode instead, I only briefly tried it with a picture of a parrot, which was nice to look at, but not the best choice for scrutiny ;)
Let's just go through the formula for a toy image of three pixels. We would want the grid positions `-1, 0, 1` to be mapped to `0, 1, 2` (since the grid spans from the left corner of the first pixel (0) to the left corner of the third pixel (2)), this is catered for by the conversion to `x`. Grid positions `2, 3, 4, 5, 6` will be mapped to pixel positions `3, 4, 5, 6, 7`. We want the border mode code for `mirror` to map this to `1, 0, 1, 2, 1` (`reflect` would be `2, 1, 0, 0, 1` instead).
My solution above was `min(x mod 2*width, 2*width - 1 - x)`. With `width = 3` and `x = [3, 4, 5, 6, 7]`, this would give `[2, 1, 0, -1, -2]`. So that's wrong, the second part needs a `mod` as well. With `min(x mod 6, (5 - x) mod 6)` we'll have `reflect`:
```
>>> x = np.arange(12)
>>> np.minimum(x % 6, (5 - x) % 6)
array([0, 1, 2, 2, 1, 0, 0, 1, 2, 2, 1, 0])
```
The correct solution for `mirror` would be:
```
>>> np.minimum(x % 4, (4 - x) % 4)
array([0, 1, 2, 1, 0, 1, 2, 1, 0, 1, 2, 1])
```
Do you want to continue with this? (I'd suggest to not include `reflect`, as before, even if the two are both easy to implement now that we're in pixel space. I think `reflect` is usually not what you want, and including both makes it confusing for users.)

PR #797, opened 2017-02-09
use inspect.signature instead of inspect.getargspec if possible

danstender:
This comes from Debian packaging, the test failed building on Python 3.5 (<https://bugs.debian.org/834910>). The patch is written by Ole Streicher.
Best,
DS

### DIFF
--- a/lasagne/layers/helper.py
+++ b/lasagne/layers/helper.py
@@ -1,6 +1,5 @@
 from collections import deque
 from difflib import get_close_matches
-from inspect import getargspec
 from itertools import chain
 from warnings import warn
 
@@ -190,13 +189,11 @@ def get_output(layer_or_layers, inputs=None, **kwargs):
                                  % layer)
             all_outputs[layer] = layer.get_output_for(layer_inputs, **kwargs)
             try:
-                names, _, _, defaults = getargspec(layer.get_output_for)
+                accepted_kwargs |= set(utils.inspect_kwargs(
+                        layer.get_output_for))
             except TypeError:
                 # If introspection is not possible, skip it
                 pass
-            else:
-                if defaults is not None:
-                    accepted_kwargs |= set(names[-len(defaults):])
             accepted_kwargs |= set(layer.get_output_kwargs)
     unused_kwargs = set(kwargs.keys()) - accepted_kwargs
     if unused_kwargs:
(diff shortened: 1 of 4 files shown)


f0k:
Thank you for the report, and for the fix (and for the debian package!). Unfortunately, it fails pep8 (needs two blank lines before a function definition) and reduces test coverage. Make sure this isn't a problem for the package.

Reading up on `getargspec`, a more compatible replacement is `getfullargspec`: https://docs.python.org/3.6/library/inspect.html#inspect.getfullargspec
Unfortunately, while this is allowed in Python 3.6, it was deprecated in Python 3.5, and so it will still trigger the deprecation warnings (that's a fun way to fail the tests, by the way -- they expect to get a single warning about the unused keyword in `get_output`, but then get two deprecation warnings extra). So we're stuck with `signature`.

If we're making this a separate function, I'd suggest to have it return the keyword arguments right away:
```python
def getkwargs(func):
   try:
       params = inspect.signature(func).parameters
       return [p.name for p in params.values() if p.default is not p.empty]
   except AttributeError:  # pragma: no cover
       spec = inspect.getargspec(func)
       return spec.args[-len(spec.defaults):] if spec.defaults else []
```
This would simplify the remaining code a bit, and make it easier to write a test for this helper function.

Do you want to finish this or shall I take over?

danstender:
Hi Jan,
thanks for quick reply. If you like, yes please, that's welcome.
Dan

f0k:
I decided to move the helper function to `lasagne.utils` and properly document and test it. I also extended the `get_output` unused kwargs test to actually make use of inspection (it relied on the `get_output_kwargs` attribute only. Hmm, actually, why don't we fill in that attribute in the base class constructor using inspection to simplify `get_output`? Anyway, not the business of this PR). I'll wait a bit with merging to see if anybody has comments, but feel free to already backport this for the Debian package!

danstender:
Yes, o.k. You can just decline the pull request if you like, then.

Debian is currently in the freeze, and the new package (with the posted patch), is already in, so I can't pull that after.
But that fulfills the purpose of having the test passing through during the build, so from that standpoint ...

Thanks for appreciation of the Debian package! We have Caffe and Torch, Theano and some subpackages in now. Unfortunately
I can't maintain my share any longer due to the lack of time. But I'll keep the packages up-to-date as long as nobody
else has taken them up. As soon as Squeeze is out an new Snapshot could go in.

Best,
Daniel

f0k:
> You can just decline the pull request if you like, then.

No, I had updated your PR. Merging now, thanks again!

> As soon as Squeeze is out an new Snapshot could go in.

Sounds good. We should really try to gather people for a small code sprint to release 0.2 and begin working on 0.3.

danstender:
All right. When Squeeze is out and the freeze is over we can push new packages for Debian Testing, Backports, Ubuntu etc.
Till then,
DS

PR #799, opened 2017-02-16
Implement SpatialPyramidPooling without cuDNN; new mode='kaiming'. Addresses #681

### LINKED ISSUE #681: There are anomalous dimensions for SPPLayer
Hi,

the spatial pyramid pooling layer has some dimensions for which the window size and stride size are inconsistent with the number of output features.
E.g.:

input_shape = (batch_size,3,11,11)
pool_dims=[1,2,4]

For pool_dim=4, we will get:

win_size=(3,3)
str_size=(2,2).

We would pool at pixels 0,2,4,6,8 which will give in total 25 features, whereas there should be only 16.
The authors of the original paper must have not run in any of these issues, but I think their recipe to pool is not quite correct, or do I miss something?

To solve this, one could do sth like:

```
            pool = pool.flatten(3)[:,:,:pool_dim*pool_dim]
            pool_list.append(pool)
```

such that the number of dimensions is restricted in these cases.

luizgh:
This pool request addresses issue #681. 

It implements the layer lasagne.layers.pool.SpatialPyramidPoolingLayer. This layer includes both the same functionality as dnn.SpatialPyramidPoolingDNNLayer (without relying on cudnn), and adds a new mode (mode='kaiming') that implements the pooling operation as described in the original paper [1]. This new mode resolves #681, and works with any input size. However, this implementation is slower since it does not use pool_2d, and instead manually slices the input and compute the max of the regions (which may not be a big overhead, given that the number of output "bins" is usually small: 21 in the default implementation). 
Since it is slower, and also to keep the same functionality as the dnn version, it is not the default mode.

[1] He, Kaiming et al (2015): Spatial Pyramid Pooling in Deep Convolutional Networks for Visual Recognition. http://arxiv.org/pdf/1406.4729.pdf.

### DIFF
--- a/lasagne/layers/pool.py
+++ b/lasagne/layers/pool.py
@@ -14,6 +14,7 @@ __all__ = [
     "FeaturePoolLayer",
     "FeatureWTALayer",
     "GlobalPoolLayer",
+    "SpatialPyramidPoolingLayer",
 ]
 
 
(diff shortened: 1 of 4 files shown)


nouiz:
What is the link with this PR in Theano?

https://github.com/Theano/Theano/pull/5395/files

Just in case combining both would be useful.

f0k:
> What is the link with this PR in Theano?

It seems the PR in Theano is just the current Lasagne implementation, which has the problems outlined in #681. The implementation proposed here works correctly for any combination of input size and pooling region count, but is slower.

lasagne/layers/pool.py
```diff
@@ -14,6 +14,7 @@
     "FeaturePoolLayer",
     "FeatureWTALayer",
     "GlobalPoolLayer",
+    "SpatialPyramidPoolingLayer"
```
f0k: Please add a `,` in the end. It's optional, but it allows to add lines without touching the ones that are there (just like you could add `"SpatialPyramidPoolingLayer"` without having to touch the `"GlobalPoolLayer",` line).

lasagne/layers/pool.py
```diff
@@ -755,3 +756,161 @@ def get_output_shape_for(self, input_shape):
...
+        The layer feeding into this layer, or the expected input shape.
+
+    pool_dims : list of integers
+        The list of :math:`n_i`'s that define the output dimension of each
+        pooling level :math:`i`. The length of pool_dims is the level of
+        the spatial pyramid.
+
+    mode : string
+        Pooling mode, one of 'max', 'average_inc_pad', 'average_exc_pad' or
+        'kaiming'. Defaults to 'max'.
```
f0k: Sorry, didn't have a look at the interface before. If `mode` is already used to distinguish between max- and mean-pooling, then `'kaiming'` vs. `'naive'` (or whatever we want to call the current implementation) should become a different keyword argument. Maybe `implementation`. `implementation='kaiming`' should also allow mean-pooling (but it doesn't need to distinguish between `'average_inc_pad'` and `'average_exc_pad'`).

lasagne/layers/pool.py
```diff
@@ -755,3 +756,161 @@ def get_output_shape_for(self, input_shape):
 
     def get_output_for(self, input, **kwargs):
         return self.pool_function(input.flatten(3), axis=2)
+
+
+def pooling_fixed_output_size(inputs, output_size, pooling_op=T.max):
```
f0k: Not sure about the name... I'll see if I can find a better one. It's good to move this into a separate function, though.

lasagne/layers/pool.py
```diff
@@ -755,3 +756,161 @@ def get_output_shape_for(self, input_shape):
...
+        for col in range(output_size):
+            start_h = T.floor(row / n * h).astype('int32')
+            end_h = T.ceil((row + 1) / n * h).astype('int32')
+            start_w = T.floor(col / n * w).astype('int32')
+            end_w = T.ceil((col + 1) / n * w).astype('int32')
+
+            pooling_region = inputs[:, :, start_h:end_h, start_w:end_w]
+            this_result = pooling_op(pooling_region, axis=(2, 3))
+            result.append(this_result.dimshuffle(0, 1, 'x'))
+    return T.concatenate(result, axis=2)
```
f0k: This will result in cascaded concatenation in the pyramid pooling graph. I'm not sure if this is optimized away by Theano. If not, it might be faster to just return the list and have a single concatenation in SpatialPyramidPoolingLayer.

lasagne/layers/pool.py
```diff
@@ -755,3 +756,161 @@ def get_output_shape_for(self, input_shape):
...
+                # Check if the running theano version supports symbolic
+                # variables as arguments for pool_2d. This is required
+                # unless using mode='kaiming'
+                try:
+                    pool_2d(T.tensor4(),
+                            ws=T.ivector(),
+                            stride=T.ivector(),
+                            ignore_border=True,
+                            pad=None)
+                except (ValueError, TypeError):
```
f0k: Why the `TypeError`?

lasagne/layers/pool.py
luizgh: ok

lasagne/layers/pool.py
luizgh: I agree that the name is not the best. It is in fact one possible implementation of pooling with fixed output sizes. 

What do you mean by moving this into a separate function?

lasagne/layers/pool.py
luizgh: Ok, I addressed this in the new check-in

lasagne/layers/pool.py
luizgh: ok

lasagne/layers/pool.py
luizgh: I had put it there while I was performing the check with T.signal.pool.pool_2d. A TypeError was raised if the theano version had the old parameter names for pool_2d. Now that we're using lasagne.pool.pool_2d (which does this check), this is no longer necessary. I have removed it in the new commit

luizgh:
So, I noticed that coverage is not 100%. Part of it is because I need to test passing an invalid "mode" when using "implementation=kaiming" - I'll do that.

However, part of the code is not being tested due to the Theano version being too old. It seems that this will be the case for any new feature that depend on a Theano version more recent than described in the requirements.txt. Do you have a recommendation for how to handle this?

f0k:
> Part of it is because I need to test passing an invalid "mode" when using "implementation=kaiming" - I'll do that.

Yes, that's a nuisance, but that's the way to go.

> However, part of the code is not being tested due to the Theano version being too old. It seems that this will be the case for any new feature that depend on a Theano version more recent than described in the requirements.txt.

Exactly.

> Do you have a recommendation for how to handle this?

We need to exclude the non-kaiming implementation from the coverage until we can update the requirements. This is done by adding `# pragma: no cover` as a line comment for the `if` or `else` case in question (one in the constructor, one in `get_output_for`).

Looks really good otherwise!

luizgh:
I have added the new test. I guess the only pending item now is the name of the function "pooling_fixed_output_size". Should we keep it like this, or for instance include "kaiming" on the name (since it refers to this particular paper)?

f0k:
Looks good!

> I guess the only pending item now is the name of the function "pooling_fixed_output_size".

What about `pool2d_n_regions` or `pool2d_nxn_regions`? It should have `2d` somewhere, unless we want to reserve the option to extend it to 1d and 3d (distinguished by the input dimensionality). Maybe that'd be good, actually. I'd like something that says "pool with given number of regions", but can't find something short enough. `pool_to_fixed_size`? Reads slightly better than the current name. What do you think?

If and when you changed the name, could you please squash your commits? `git rebase -i master` should be the easiest way. I'd suggest one commit for adding the tests to the existing DNN-based layer, and then everything else as a single commit that adds the new layer along with the tests and documentation.

luizgh:
I think "pool_2d_nxn_regions" is good for now. It is hard to now if this implementation would be useful for more dimensions, since the number of pooling regions becomes much larger (e.g. 64 outputs for pool_dim= 4 with 3 dimensions), and the overhead of  manually slicing the tensors may be too large. Either way, this is something that can be reconsidered if someone wants SPP with more dimensions.

As requested, I squashed the commits, separating the new SPP implementation, and the new tests for the  SPP cuDNN version. Let me know if there is anything else missing.

f0k:
> Let me know if there is anything else missing.

Actually, yes: It should be added to `docs/modules/layers.rst` as well. Sorry, didn't notice before. Could you amend the first commit to do so?

Other than that, I'm fine with merging! Thanks a lot for your work! I'll leave this open until Monday in case anybody else has a comment.

luizgh:
Ok, I have updated the "Layers" documentation as well. 
Thank you for the help on getting this done. I really like this library, and it was nice to contribute a little to it!

f0k:
Merging! Thanks again!

> it was nice to contribute a little to it!

I'd be glad to see you again! :) We could definitely use some help to finally get 0.2 finished (https://github.com/Lasagne/Lasagne/milestone/2).

PR #800, opened 2017-02-16
Fix #782 TransformerLayer for rectangle images

### LINKED ISSUE #782: TransformerLayer: unexpected shear artifact for non-rectangular images

artvive:
Fix issue #782 
and add the corresponding test.

A few comments about this PR:
- I chose to only modify the function _transform_affine so that it won't interfere with the TPS Transformer which uses also _interpolate and _meshgrid.
- This is maybe at the cost of performance because I feel like doing and undoing things that could be taken care of in those two functions.
- I am not quite satisfied by the test which is very sensitive to the rounding of the indices, so every idea is welcomed.

PS: this is my very first PR so I'd gladly receive any comment and advice.

### DIFF
--- a/lasagne/layers/special.py
+++ b/lasagne/layers/special.py
@@ -442,9 +442,12 @@ def _transform_affine(theta, input, downsample_factor):
     grid = _meshgrid(out_height, out_width)
 
     # Transform A x (x_t, y_t, 1)^T -> (x_s, y_s)
+    # Make the transformation in a correctly scaled space
+    scaler = T.as_tensor([width, height, 1]).reshape((3, 1))
+    grid = scaler * grid
     T_g = T.dot(theta, grid)
-    x_s = T_g[:, 0]
-    y_s = T_g[:, 1]
+    x_s = T_g[:, 0] / width
+    y_s = T_g[:, 1] / height
     x_s_flat = x_s.flatten()
     y_s_flat = y_s.flatten()
 
--- a/lasagne/tests/layers/test_special.py
+++ b/lasagne/tests/layers/test_special.py
@@ -416,6 +416,34 @@ class TestTransformLayer():
                                         constant(thetas)]).eval()
         np.testing.assert_allclose(inputs, outputs, rtol=1e-6)
 
+    def test_transform_rotation(self):
+        # Check that a 90° rotation on a rectangle image is correct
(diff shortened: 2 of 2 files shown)


f0k:
Nice, thank you!

> This is maybe at the cost of performance because I feel like doing and undoing things that could be taken care of in those two functions.

I guess the performance impact won't be too bad, but we could measure it. It's nice that the change is so local.

With this change, Theta is applied in pixel space. This means the translation must be given in pixels, not as a fraction of the input size. As [I argued before](https://github.com/Lasagne/Lasagne/issues/782#issuecomment-267324977), this might not be ideal, as it will require a very different value range for translations compared to rotation and shearing. Also it might make it more difficult to process images of variable input size; there's no easy way to specify cropping the top left corner, for example.
So another solution might be to figure out the aspect ratio and scale either x_s or y_s (not both), depending on whether the width or height is smaller.
In any case, the operation space of Theta should be determined by an option given to the Layer constructor. Existing networks should just work as before. So we could have three alternatives for a new `grid_space` or `grid_domain` option: `pixels` (your new implementation), `unit_square` (the old implementation, staying the default), and `unit_rect` (a rectangle with the longer side bounded in `[0,1]` and the shorter side being shorter). I'm not very sure about the naming, but does this seem sensible to you?

artvive:
Yes I agree with your design principles, I'll implement them when I have time

Issue #762, opened 2016-10-26
Layer normalization for recurrent networks


justanothercoder:
Hello!
Recently, a new normalization [procedure](https://arxiv.org/abs/1607.06450) has been invented. It provided great results for improving training on both feed-forward and especially recurrent networks.  Training is accelerated several times. It would be great to see it implemented in Lasagne. 

Unfortunately, as I am looking at code for GRU and LSTM it seems to me, that it would be necessary to add another layers like `LN_LSTMLayer` because LN requires to change the step function in `scan` procedure. Also, it seems like it is another use case for [this](https://github.com/Lasagne/Lasagne/issues/425) proposal. Also, it is not that hard to change existing layers to choose to use or not use LN.

There is implementation from paper authors. It is not in Lasagne, but rather in Theano.
https://github.com/ryankiros/layer-norm

Waiting for your replies.

benanne:
Unfortunately the current interface for recurrent layers is not particularly modular. We've been talking for a long time about fixing this, but so far the plans haven't really materialised. Have a look at https://github.com/Lasagne/Lasagne/issues/425 if you're interested (other issues may be relevant as well but this is the only one I can find right now).

vincent341:
Hi. I wrote a LNLSTM implementing Layer Normalization. Actually I modified the original LSTM Layer by referencing this code. I want to give great thanks to Jan Schlüter for helping me check that and make it elegant.

vincent341:
Hi. I wrote a LNLSTM (https://github.com/vincent341/LayerNormLSTM/blob/master/LayerNormLSTM.py) implementing Layer Normalization(https://arxiv.org/abs/1607.06450). Actually I modified the original LSTM Layer by referencing this code(https://github.com/ryankiros/layer-norm). I want to give great thanks to Jan Schlüter for helping me check that and make it elegant.

nikostr:
Hey, there were some minor issues with the implementation by @vincent341 linked above. In case anyone else finds this interesting, I've submitted a [pull request](https://github.com/vincent341/LayerNormLSTM/pull/1) to amend that implementation. This implementation runs even when `unroll_scan=False`, and we've checked the code against the original paper, amending some minor math issues.

Issue #738, opened 2016-08-26
Layer normalization 


skaae:
I'm implementing layernormalization from http://arxiv.org/pdf/1607.06450v1.pdf
 Should i create a pr for this to lasagne? 

The feedforward variant can be implemented as separate layer similar to `BatchNormalizationLayer` and the `batch_norm` helpder. 

For GRU and LSTM i think you would need a lot of if statements to handle layer normalization because the gate_biases are no longer used and because of the 'precompute' option.

f0k:
Duplicate of #736. Short result: I'd suggest the feedforward implementation to be a superclass for `BatchNormalizationLayer`, to avoid code duplication. I wouldn't name it after the paper (`LayerNormalizationLayer`?), but in a way that conveys it's normalizing to zero mean and unit variance. Maybe `StandardizationLayer`, with a `layer_norm` helper?

Issue #736, opened 2016-08-20
Call for Layer Layer Normalization


icemansina:
I am trying to implement the layer normalization and LNRNN. Is there anybody interested in this implementation in Lasagne?

PS: Keras code: https://gist.github.com/udibr/7f46e790c9e342d75dcbd9b1deb9d940
      Tensorflow code: https://github.com/pbhatia243/tf-layer-norm/blob/master/layers.py

@f0k

erfannoury:
I have a working implementation of Layer Normalization for LSTM [(LN-LSTM)](https://github.com/erfannoury/seq2seq-lasagne/blob/master/CustomLSTMLayer.py#L17-L650) that you may take a look. We can further improve it if you notice anything that needs to be changed.

f0k:
> Is there anybody interested in this implementation in Lasagne?

The existing `BatchNormLayer` can already do layer normalization by just changing the `axis` specification, but then you'd always need to call `get_output()` with `batch_norm_use_averages=False, batch_norm_compute_averages=False`. So we could either add a `LayerNormLayer` (uh...) or make the running averages of `BatchNormLayer` completely optional and just add a `layernorm()` convenience function that instantiates a `BatchNormLayer`. Hmm, or maybe add a `NormLayer` for layer normalization and derive `BatchNormLayer` from it, just adding the running average statistics on top and changing the default `axis`. I think that might be the cleanest option. Sorry for thinking aloud.

For recurrent models, currently the only solution are separate classes like an LNLSTMLayer, and I'm not sure we want to add this to Lasagne, as it duplicates a lot of code. We'll be able to include a much more modular solution with #629, but this is blocked by #678, which is blocked by https://github.com/Lasagne/Lasagne/milestone/2 -- any help with this would be appreciated!

icemansina:
Thanks a lot. @f0k @erfannoury 

- For normal model, i.e., CNN or Dense layers, 

> So we could either add a `<LayerNormLayer>` (uh...) or make the running averages of `<BatchNormLayer>` completely optional and just add a `<layernorm()>` convenience function that instantiates a `<BatchNormLayer>`. 

[...]

PR #808, opened 2017-03-02
Implementation of Instance Normalization

SimonKohl:
Added a layer called 'InstanceNormLayer' that implements instance normalization of its inputs as according to 'Instance Normalization: The Missing Ingredient for Fast Stylization'      (https://arxiv.org/pdf/1607.08022.pdf).

While the existing BatchNormLayer is able to implement instance normalization by setting the axes, beta and gamma arguments accordingly, it still requires a fixed batch size as learned parameters need to be initialized. 

The proposed InstanceNormLayer gets away with that, allowing for more flexibility.

### DIFF
--- a/docs/modules/layers.rst
+++ b/docs/modules/layers.rst
@@ -166,8 +166,9 @@
     LocalResponseNormalization2DLayer
     BatchNormLayer
     batch_norm
-    InstanceNormLayer
+    StandardizationLayer
     instance_norm
+    layer_norm
 
 
 .. rubric:: :doc:`layers/embedding`
--- a/docs/modules/layers.rst
+++ b/docs/modules/layers.rst
@@ -166,6 +166,8 @@
     LocalResponseNormalization2DLayer
     BatchNormLayer
     batch_norm
+    InstanceNormLayer
+    instance_norm
 
 
 .. rubric:: :doc:`layers/embedding`
--- a/docs/modules/layers/normalization.rst
+++ b/docs/modules/layers/normalization.rst
@@ -13,3 +13,8 @@ Normalization layers
 
 .. autofunction:: batch_norm
 
+.. autoclass:: InstanceNormLayer
+    :members:
+
+.. autofunction:: instance_norm
+
--- a/docs/modules/layers/normalization.rst
+++ b/docs/modules/layers/normalization.rst
@@ -13,8 +13,10 @@ Normalization layers
 
 .. autofunction:: batch_norm
 
-.. autoclass:: InstanceNormLayer
+.. autoclass:: StandardizationLayer
     :members:
 
 .. autofunction:: instance_norm
+
+.. autofunction:: layer_norm
 
--- a/lasagne/layers/normalization.py
+++ b/lasagne/layers/normalization.py
@@ -50,6 +50,9 @@ __all__ = [
     "LocalResponseNormalization2DLayer",
     "BatchNormLayer",
     "batch_norm",
+    "StandardizationLayer",
+    "instance_norm",
+    "layer_norm",
 ]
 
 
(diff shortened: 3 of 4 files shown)


f0k:
Hey, thank you for the PR! This technique has also been termed "Layer normalization" before, see #736, #738, #762. In that case, it also contains the learnable scale and shift as in batch normalization.

> While the existing BatchNormLayer is able to implement instance normalization by setting the axes, beta and gamma arguments accordingly, it still requires a fixed batch size as learned parameters need to be initialized.

Yes, that's a bummer. Another disadvantage of using the batch normalization layer for this is that you'll need to pass extra arguments to `lasagne.layers.get_output` to disable learning and using the exponential running average statistics.
As argued in https://github.com/Lasagne/Lasagne/issues/736#issuecomment-241374360, I'd prefer to have a `NormLayer` as a new base class for `BatchNormLayer`, such that the `BatchNormLayer` only adds the running average statistics on top of the `NormLayer` implementation and changes the default axes. This would reduce some code duplication. I like the term "instance normalization", though, we could still use `instance_norm` for the convenience function. (I think I wouldn't use `InstanceNormLayer` for the class, because depending on the axes, it could do something else, and it would sound a bit weird to derive batch normalization from instance normalization. On the other hand, depending on the axes, `BatchNormLayer` also doesn't necessarily normalize a batch. I'm open for other arguments.)

Would you like to update your PR such that `NormLayer` includes `mean` and `gamma` and becomes a base class for `BatchNormLayer`?

SimonKohl:
I'd be happy to. Wouldn't mind some guidance in the process though. So, basically, have a base class :NormLayer: that implements a basic __init__() and get_output_for() (in line with what is currently :InstanceNormLayer:, i.e. gamma and beta being non-trainable) and then have :BatchNormLayer: derive derive from it  and expand on the two methods accordingly?

f0k:
> Wouldn't mind some guidance in the process though.

Sure! :)

> What what about the initilialization of `gamma` and `beta`?

Ah. There was a catch. We want `gamma` and `beta` to be shared across the batch, but not normalize data across the batch. This means the interface is actually very different.

Okay, I think we can forget about the common base class and go with an `InstanceNormLayer`.

Now the question is what to do about layer normalization. Checking their paper again (https://arxiv.org/abs/1607.06450), they normalize across all but the batch axis (i.e., across all features) and learn a scalar shift and scale. I'm leaning towards omitting this. Layer normalization gives each neuron its own scale and shift, so it's not tied to the normalization axes as in batch normalization. It's probably easier to just use a sequence of instance normalization, bias and scale layers for this.

Great, I'll just review your current PR now!

lasagne/layers/normalization.py
```diff
@@ -373,3 +375,144 @@ def batch_norm(layer, **kwargs):
...
+    convenience function :func:`instance_norm` modifies an existing layer to
+    insert instance normalization in front of its nonlinearity.
+    The behavior can be controlled by passing keyword arguments to
+    :func:`lasagne.layers.get_output()` when building the output expression
+    of any network containing this layer.
+    During training, [1]_ normalize each instance in an input mini-batch
+    by its statistics.
+    See also
+    --------
+    batch_norm : Convenience function to apply batch normalization to a layer
```
f0k: `instance_norm` and `instance normalization`

lasagne/layers/normalization.py
```diff
@@ -373,3 +375,144 @@ def batch_norm(layer, **kwargs):
...
+    This layer should be inserted between a linear transformation (such as a
+    :class:`DenseLayer`, or :class:`Conv2DLayer`) and its nonlinearity. The
+    convenience function :func:`instance_norm` modifies an existing layer to
+    insert instance normalization in front of its nonlinearity.
+    The behavior can be controlled by passing keyword arguments to
+    :func:`lasagne.layers.get_output()` when building the output expression
+    of any network containing this layer.
+    During training, [1]_ normalize each instance in an input mini-batch
+    by its statistics.
+    See also
```
f0k: Missing newline before

lasagne/layers/normalization.py
```diff
@@ -373,3 +375,144 @@ def batch_norm(layer, **kwargs):
...
+    --------
+    batch_norm : Convenience function to apply batch normalization to a layer
+    References
```
f0k: Missing newline before

lasagne/layers/normalization.py
```diff
@@ -373,3 +375,144 @@ def batch_norm(layer, **kwargs):
...
+           Instance Normalization: The Missing Ingredient for Fast Stylization.
+           https://arxiv.org/pdf/1607.08022.pdf.
+    """
+    def __init__(self, incoming, axes='auto', epsilon=1e-4, **kwargs):
+        super(InstanceNormLayer, self).__init__(incoming, **kwargs)
+
+        if axes == 'auto':
+            # default: normalize over spatial dimensions only,
+            # i.e. separate for each instance in the batch
+            axes = (2, 3)
```
f0k: You don't know if these axes are there. Should be `tuple(range(2, len(self.input_shape)))`.

lasagne/layers/normalization.py
```diff
@@ -373,3 +375,144 @@ def batch_norm(layer, **kwargs):
...
+        elif isinstance(axes, int):
+            axes = (axes,)
+        self.axes = axes
+
+        self.epsilon = epsilon
+
+    def get_output_for(self, input, **kwargs):
```
f0k: All this code can be simplified to:
```python
mean = input.mean(self.axes, keepdims=True)
std = T.sqrt(input.var(self.axes, keepdims=True) + self.epsilon)
return (input - mean) / std
```

lasagne/layers/normalization.py
```diff
@@ -373,3 +375,144 @@ def batch_norm(layer, **kwargs):
...
+
+
+def instance_norm(layer, **kwargs):
+    """
+    Apply instance normalization to an existing layer. This is a convenience
+    function modifying an existing layer to include instance normalization: It
+    will steal the layer's nonlinearity if there is one (effectively
+    introducing the normalization right before the nonlinearity), remove
+    the layer's bias if there is one (because it would be redundant), and add
+    a :class:`InstanceNormLayer` and :class:`NonlinearityLayer` on top.
```
f0k: Hmm, we need to have a word on that bias. Currently you remove it without giving the network a chance to introduce an offset afterwards. What do the instance normalization authors do in their paper?

lasagne/layers/normalization.py
```diff
@@ -373,3 +375,144 @@ def batch_norm(layer, **kwargs):
...
+    Parameters
```
f0k: Your missing the newline before.

lasagne/layers/normalization.py
```diff
@@ -373,3 +375,144 @@ def batch_norm(layer, **kwargs):
...
+    ----------
+    layer : A :class:`Layer` instance
+        The layer to apply the normalization to; note that it will be
+        irreversibly modified as specified above
+    **kwargs
+        Any additional keyword arguments are passed on to the
+        :class:`InstanceNormLayer` constructor.
+    Returns
```
f0k: Missing newline before.

lasagne/layers/normalization.py
```diff
@@ -373,3 +375,144 @@ def batch_norm(layer, **kwargs):
...
+        or a nonlinearity layer stacked on top of both
+        if `layer` was nonlinear.
+    Examples
+    --------
+    Just wrap any layer into a :func:`instance_norm` call on creating it:
+    >>> from lasagne.layers import InputLayer, Conv2DLayer, instance_norm
+    >>> from lasagne.nonlinearities import rectify
+    >>> l1 = InputLayer((10, 3, 28, 28))
+    >>> l2 = instance_norm(Conv2DLayer(l1, num_filters=64, filter_size=3,
+    nonlinearity=rectify))
```
f0k: Needs continuation characters `... ` at the beginning of the line.

lasagne/layers/normalization.py
SimonKohl: You're right, this is an important issue to converge on. The instance normalization paper does not mention the bias whatsoever, suggesting that they don't remove it and plainly apply instance norm on top of the layers output. As you mentioned earlier, the 'layer normalization' (e.g. https://arxiv.org/pdf/1607.06450.pdf) is conceptually slightly different to instance norm in that they retain a gain and bias (gamma and beta in `BatchNormLayer`).  We could delete the bias of the preceeding (e.g. conv) layer and retain gain and bias as learnable (basically scalar) parameters in `InstanceNormLayer`?

SimonKohl: Ok, so my latest commit is a suggestion to cover both ideas, 'layer normalization'  and 'instance normalization'. I added `beta` and `gamma` as learnable parameters, which which are broadcast across the batch-size as well as the trailing dimensions beyond the 2nd dimension. This way 'layer normalization' as in https://arxiv.org/pdf/1607.06450.pdf should be available, as well as 'instance normalization' as in https://arxiv.org/pdf/1607.06450.pdf, when setting  `beta` and `gamma` to None or using the default initialization.

SimonKohl: Ah, there's one issue remaining. If `beta` is set to None, we shouldn't delete the incoming layer's bias. Fixed this in `instance_norm()` convenience function.

f0k: > The instance normalization paper does not mention the bias whatsoever, suggesting that they don't remove it and plainly apply instance norm on top of the layers output. 

But that would subtract the bias again. I doubt that's a good idea. Did they provide any code?

> I added beta and gamma as learnable parameters, which are broadcast across the batch-size as well as the trailing dimensions beyond the 2nd dimension.

This fits the layer normalization paper, but it's a bit sad we can't control the axes then. I'm torn between leaving it as it is, or not providing beta/gamma at all and offloading this to a ScaleLayer and BiasLayer on top (which could be added automatically by the convenience function if asked for). This would simplify the layer and tests. For the batch normalization layer it was good to have everything in a single layer, because this allows to use cuDNN's batch normalization primitives. For instance normalization, I doubt this will play a role.

SimonKohl: Hi Jan, sorry I was bit absorbed with other things these past days. Concerning your first comment, that would mean we'd always want to have a learnable bias, just the gamma factor would optionally be learnable, right? With respect to whether or not to include gamma/beta in the way it is currenlty done: Did you converge on a preferred implementation? If so, I could go ahead with the implementation and the corresponding tests.

lasagne/layers/normalization.py
```diff
@@ -373,3 +375,158 @@ def batch_norm(layer, **kwargs):
...
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape
+    axes : 'auto', int or tuple of int
+        The axis or axes to normalize over. If ``'auto'`` (the default),
+        normalize over all axes except for the first two:
+        this will normalize over all spatial dimensions for
+        convolutional layers.
```
f0k: With the new `StandardizationLayer`, it might make sense to have `'auto'` normalize over the second axis for 2D input, and over all but the first two axes for 3D+ input. This would do something useful for dense layers, recurrent layers and convolutional layers.

lasagne/layers/normalization.py
```diff
@@ -373,3 +377,225 @@ def batch_norm(layer, **kwargs):
...
+        Any additional keyword arguments are passed to the :class:`Layer`
+        superclass.
+        
+    Notes
+    -----
+    This layer should be inserted between a linear transformation (such as a
+    :class:`DenseLayer`, or :class:`Conv2DLayer`) and its nonlinearity. The
+    convenience functions :func:`instance_norm` :func:`layer_norm`
+    modify an existing layer to insert instance normalization or
+    layer normalization in front of its nonlinearity.
```
f0k: I think this doesn't apply here. For a generic standardization layer, we shouldn't state any rules.

lasagne/layers/normalization.py
```diff
@@ -373,3 +377,225 @@ def batch_norm(layer, **kwargs):
...
+
+    See also
+    --------
+    instance_norm : Convenience function to apply instance normalization to a layer
+    References
+    layer_norm : Convenience function to apply layer normalization to a layer
+    References
+
+    ----------
```
f0k: You messed up something with the `References` title here :)

lasagne/layers/normalization.py
```diff
@@ -373,3 +377,225 @@ def batch_norm(layer, **kwargs):
...
+    .. [1] Ulyanov, D., Vedaldi, A., & Lempitsky, V. (2016):
+           Instance Normalization: The Missing Ingredient for Fast Stylization.
+           https://arxiv.org/pdf/1607.08022.pdf.
+
+    .. [2] Ba, J., Kiros, J., & Hinton, G. (2016):
+           Layer normalization. arXiv preprint arXiv:1607.06450.
```
f0k: Let's unify this and give the arxiv link in both cases, but using `abs` instead of `pdf`.

lasagne/layers/normalization.py
```diff
@@ -373,3 +377,225 @@ def batch_norm(layer, **kwargs):
...
+
+        mean = input.mean(self.axes, keepdims=True)
+        std = T.sqrt(input.var(self.axes, keepdims=True) + self.epsilon)
+
+        return (input - mean) / std
+
+
+def instance_norm(layer, learn_scale=True, learn_bias=True, **kwargs):
+    """
+	Apply instance normalization to an existing layer. This is a convenience
```
f0k: The indentation is wrong (4 characters too much).

lasagne/layers/normalization.py
```diff
@@ -373,3 +377,225 @@ def batch_norm(layer, **kwargs):
...
+	:class:`BiasLayer` will be inserted inbetween.
+
+	Parameters
+	----------
+	layer : A :class:`Layer` instance
+		The layer to apply the normalization to; note that it will be
+		irreversibly modified as specified above
+	**kwargs
+		Any additional keyword arguments are passed on to the
+		:class:`StandardizationLayer` constructor.
```
f0k: `learn_scale` and `learn_bias` need to be described.

lasagne/layers/normalization.py
SimonKohl: Sounds good. It's implemented in the latest commit.

docs/modules/layers.rst
```diff
@@ -163,6 +163,8 @@
     LocalResponseNormalization2DLayer
     BatchNormLayer
     batch_norm
+    InstanceNormLayer
+    instance_norm
```
f0k: This doesn't match the code any more. Should have StandardizationLayer, instance_norm and layer_norm instead.

docs/modules/layers/normalization.rst
```diff
@@ -13,3 +13,8 @@ Normalization layers
 
 .. autofunction:: batch_norm
 
+.. autoclass:: InstanceNormLayer
+    :members:
+
+.. autofunction:: instance_norm
+
```
f0k: This doesn't match the code any more. Should have StandardizationLayer, instance_norm and layer_norm instead.

f0k:
Finally finished and merged -- thanks again and sorry for the long pause.

Issue #720, opened 2016-07-07
Simple "reapply" functionality


justheuristic:
Greetings!

It's probably a small issue in most image-CNN-related cases, but when dealing with text, multi-input NNs, reinforcement learning or long-term memory networks some layers should be applied at multiple spots with same weights.

Some usage cases:
* Text processing: Apply the same embedding and text convolution / recurrent processing to advertisement title and description
* Reinforcement learning: Apply critic subnetwork in deterministic policy graident algorithm should evaluate both "optimal" and actual(i.e. with exploration) actor output.

As far as i understood the docs, one natural way to do so in Lasagne is by passing other layer's params on creation, e.g.

```
l_d0 = DenseLayer(bottom_layer,10)
l_d0_again = DenseLayer(other_bottom_layer, 10,  W = l_d0.W, b = l_d0.b)
```
or alternatively, create several networks and use .get_output
```
lasagne.layers.get_output(my_nn, {my_nn_input:input1}
lasagne.layers.get_output(my_nn, {my_nn_input:input2}
```

The problem with the first approach is that it takes immensely many lines of code to do so with larger networks, especially with e.g. LSTM with a lot more parameters. Not to mention it's easy to make a mistake that way.

The problem with the second approach is that it forces you to break outside lasagne network and makes routine tasks e.g. getting all params, regularizing, applying flags more complicated and verbose.

It may be a good idea to introduce some simpler method to do that.

For example, in [blocks](https://github.com/mila-udem/blocks) layers can be .apply --ed to as many spots as necessary with all params shared.
```
>>> from blocks.bricks import Linear
>>> from blocks.initialization import IsotropicGaussian, Constant
>>> linear = Linear(input_dim=10, output_dim=5,
...                 weights_init=IsotropicGaussian(),
...                 biases_init=Constant(0.01))
>>> y = linear.apply(x)
```
This complicates the code a lot and i personally don't think this is the best solution since most of layers are only applied once.

Another approach is to make some .reapply() method for layers with params that would work like a cloning constructor that reuses the original layer's weights, but that too makes cloning large nets complicated.
```
l_d0 = DenseLayer(bottom_layer)
l_d0_again = l_d0.reapply(other_bottom_layer)
```

In one of our libraries for reinforcement learning we use a generic function to clone a lasagne network with or without keeping the params. 
The source can be found here - https://github.com/yandexdataschool/AgentNet/blob/master/agentnet/utils/clone.py 
[...]

PR #811, opened 2017-03-05
A easy way to reuse lasagne networks

guoxuesong:
The idea to create join.py is that some time I want reuse the networks I built using lasagne, or join two network together. 

For example, I have a deep conv network which map a 3x64x64 image to a 64 feature point 64x1x1, and a transform network which map a 64x1x1 point to another 64x1x1 point, I want to join them together. I can do it this way:

joined_network=lasagne.layers.Input(shape=(None,64,1,1),input_var=lasagne.get_output(transform_network,lasagne.layers.get_output(conv_network))

There is a problem: lasagne.layers.get_all_params(joined_network) does not return the params in conv_network and transform_network. I think lasagne miss something, something just like lasagne.layers.get_output() but return a new network instead of a theano expression, that is why I created lasagne.layers.join.join_layer()

Such a network returned by lasagne.layers.join.join_layer() shares params with the original layers, generally BatchNormLayer should not share params, for that you can call lasagne.layers.join.copy_batch_norm() after all the lasagne.layers.join.join_layer() calls.

### DIFF
--- a/lasagne/layers/join.py
+++ b/lasagne/layers/join.py
@@ -1,0 +1,140 @@
+# coding:utf-8
+# vi:tabstop=4:shiftwidth=4:expandtab:sts=4
+
+import lasagne
+
+__all__ = [
+        'BatchNormLayer',
+        'copy_batch_norm',
+        'join_layer'
+        ]
+
+
+class BatchNormLayer(lasagne.layers.BatchNormLayer):
+    def __init__(self, *args, **kwargs):
+        self.init_kargs = kwargs
+        self.init_args = args
+        super(BatchNormLayer,  self).__init__(*args, **kwargs)
+
+    def copy(self):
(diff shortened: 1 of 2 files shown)


f0k:
> joined_network=lasagne.layers.Input(shape=(None,64,1,1),input_var=lasagne.get_output(transform_network,lasagne.layers.get_output(conv_network))

You can also do:
```python
lasagne.layers.get_all_layers(transform_network)[1].input_layer = conv_network
```
This will replace the input layer of the transform network with the output layer of the CNN.

Other than that, your PR basically is an implementation of theano.clone for Lasagne layer graphs instead of Theano expressions. There might be use cases for that, but we'd need to find some examples to warrant inclusion into the library, and a simpler implementation (it's possible to use `copy.deepcopy` for that). There's another draft of such functionality in #720. Let's continue the discussion there.

Issue #816, opened 2017-03-18
Jacaard_similarity_accuracy to the objectives


saeedizadi:
One of widely-used metrics for evaluationg the results of a segmentation task is the Jaccard_similarity_coefficients. It is as simple as diving the intersection region over the union region between two binary maps. 

Adding this accuracy metric to the objectives (just like binary_accuracy) can be so helpful. 
For more information, please see this Wikipedia page: 
https://en.wikipedia.org/wiki/Jaccard_index

f0k:
> One of widely-used metrics for evaluationg the results of a segmentation task

Do you have some references at hand?

> Adding this accuracy metric to the objectives (just like binary_accuracy) can be so helpful.

I'd be fine with this. Feel free to submit a PR. There are some things to consider:
* For binary classification, when binarizing the predictions, the Jaccard index is the same as the accuracy.
* One could go for a non-binarized version that computes `minimum(A, B).sum() / maximum(A, B).sum()`. This would even be differentiable. Probably there's literature for that as well.
* For multi-class problems, this should support both one-hot encoded and integer targets, like `categorical_accuracy`.
* We should think about whether we want `jaccard_index()` (`jaccard_similarity_coefficient()` is a bit long for my taste) or `jaccard_distance()`. Depends on whether we advertise this as a loss function or evaluation measure.

PR #818, opened 2017-03-21
Jaccard Similarity Coefficient (objective accuracy)

saeedizadi:
Fix the #816 
@f0k 
This is my try to implement the Jaccard similarity coefficient (Jaccard index) in the objectives. This metric is highly used in evaluating the performance of medical segmentation models where a two-class prediction (probablistic map) is compared with the binary ground truth. 
Please see this as the [reference ](http://www.ee.cuhk.edu.hk/~snli/rshi_ICIP2014.pdf)

Thanks
Saeed

### DIFF
--- a/lasagne/objectives.py
+++ b/lasagne/objectives.py
@@ -121,8 +121,6 @@ def align_targets(predictions, targets):
 def jaccard_similarity_accuracy(predictions,targets):
     """Computes the Jaccard similarity coefficient (accuracy) between predictions and targets.
 
-    .. math:: L_i = \\mathbb{I}(t_i = \mathbb{I}(p_i \\ge \\alpha))
-
     Parameters
     ----------
     predictions : Theano tensor
--- a/lasagne/objectives.py
+++ b/lasagne/objectives.py
@@ -118,14 +118,38 @@ def align_targets(predictions, targets):
         targets = as_theano_expression(targets).dimshuffle(0, 'x')
     return predictions, targets
 
-def jaccard_similarity_accuracy(predictions,targets,threshold=0.5):
+def jaccard_similarity_accuracy(predictions,targets):
+    """Computes the Jaccard similarity coefficient (accuracy) between predictions and targets.
+
+    .. math:: L_i = \\mathbb{I}(t_i = \mathbb{I}(p_i \\ge \\alpha))
+
+    Parameters
+    ----------
+    predictions : Theano tensor
+        Predictions in [0, 1], such as a sigmoidal output of a neural network,
+        giving the probability of the positive class
+    targets : Theano tensor
+        Targets in {0, 1}, such as ground truth labels.
+
+    Returns
+    -------
+    Theano tensor
+        An expression for the jaccard similarity coefficient(accuracy) in [0, 1]
+
+    Notes
+    -----
+    This objective is also known as Jaccard index.
+    This objective function should not be used with a gradient calculation;It is intended as a convenience for
+    validation and testing, not training.
+
+    To obtain the average accuracy, call :func:`theano.tensor.mean()` on the
+    result, passing ``dtype=theano.config.floatX`` to compute the mean on GPU.
+    """
     predictions,targets = align_targets(predictions,targets)
-    predictions = theano.tensor.ge(predictions,threshold)
-    predictions = theano.tensor.cast(predictions,'uint8')
-    targets = theano.tensor.cast(targets,'uint8')
-    intersection = theano.tensor.and_(predictions,targets)
-    union = theano.tensor.or_(predictions,targets)
+    intersection = theano.tensor.minimum(predictions,targets)
+    union = theano.tensor.maximum(predictions,targets)
     return intersection.sum()/union.sum()
+
 def binary_crossentropy(predictions, targets):
     """Computes the binary cross-entropy between predictions and targets.
 
--- a/lasagne/objectives.py
+++ b/lasagne/objectives.py
@@ -87,7 +87,8 @@ __all__ = [
     "binary_hinge_loss",
     "multiclass_hinge_loss",
     "binary_accuracy",
-    "categorical_accuracy"
+    "categorical_accuracy",
+    "jaccard_similarity_accuracy"
 ]
 
 
@@ -117,7 +118,14 @@ def align_targets(predictions, targets):
         targets = as_theano_expression(targets).dimshuffle(0, 'x')
     return predictions, targets
 
-
+def jaccard_similarity_accuracy(predictions,targets,threshold=0.5):
+    predictions,targets = align_targets(predictions,targets)
+    predictions = theano.tensor.ge(predictions,threshold)
+    predictions = theano.tensor.cast(predictions,'uint8')
+    targets = theano.tensor.cast(targets,'uint8')
+    intersection = theano.tensor.and_(predictions,targets)
+    union = theano.tensor.or_(predictions,targets)
+    return intersection.sum()/union.sum()
 def binary_crossentropy(predictions, targets):
     """Computes the binary cross-entropy between predictions and targets.
 


saeedizadi:
> The code has some PEP8 violations (missing spaces around , and /).

Done!

> The docstring should include the actual formula that is used.

Done!

> The name jaccard_similarity_accuracy is a bit weird, I'd suggest jaccard_index instead.

I changed the name to` binary_jaccard_index`. I think this name is meaningful as the multi-class version is a bit more complicated so we can address it in another function say `multi_jaccard_index`

> I'm not sure about the note on not using it for training; the current form would at least be differentiable.

As far as I know, it is not used for training, while differentiable. See [this](https://arxiv.org/pdf/1702.08699.pdf) which introduces a loss function for training which is near `jaccard_index` but not exactly the same. 

> The docstring does not make clear if the objective also applies to multi-class problems. 

I emphasized in the docstring that this objective is for binary. We can address the multi-class versions in another try. 

> The summation should probably be across all axes but the batch dimension, to obtain a per-item loss.

You do not provide per-item loss in `binary_accuracy` as well. How come it may be needed?

> The loss needs tests.

Give me some hints. What should I do?

> The loss needs to be included in the module docstring at the top of the file.

Done!

> It would be nice to include references to a paper or two that use it.

Jaccard index is the same as Intersection-over-Union (IoU) metric which is being used in semantic segmentation/Object Detection tasks. PASCAL VOC uses this metric for evaluation.
See [this](http://www.bmva.org/bmvc/2013/Papers/paper0032/paper0032.pdf) for reference.

f0k:
> I changed the name to `binary_jaccard_index`. I think this name is meaningful as the multi-class version is a bit more complicated so we can address it in another function

Sounds good to me. Even if the implementations turn out to be similar, it's easier to separate it into two functions -- especially if we want to retain the ability of providing either one-hot or integer ground truth.

> As far as I know, it is not used for training, while differentiable. See this which introduces a loss function for training which is near `jaccard_index` but not exactly the same.

Interesting. Well, the original definition wouldn't be differentiable, only the min(A,B).sum()/max(A,B).sum() formulation is.

> You do not provide per-item loss in `binary_accuracy` as well. How come it may be needed?

Right, this depends on whether we want to advertise this as an evaluation measure or as a training objective.
For an evaluation measure, the binary jaccard index *with binarized predictions* is exactly the same as the binary accuracy, so there wouldn't be a benefit of including it. With real-valued predictions it's different, but I don't know how useful/interpretable that actually is. Anyway, in this case we should compute the average, as you said, to match the interface of `binary_accuracy`.
For a training objective, it should return the item-wise loss just like `binary_crossentropy`. But in that case we'd want it to behave like a loss function, so we'd want to implement the [Jaccard distance](https://en.wikipedia.org/wiki/Jaccard_index) instead (and call it `binary_jaccard_distance`).

Ah, just by chance I found that the formulation I suggested is actually given in the Wikipedia article as [Generalized Jaccard similarity and distance](https://en.wikipedia.org/wiki/Jaccard_index#Generalized_Jaccard_similarity_and_distance).

https://arxiv.org/abs/1701.03056 (Section 2.2) use yet another formulation which they call Jaccard index:
`sqrt(sum((A*B)**2)) / (sum(A**2) + sum(B**2))`
They then minimize the Jaccard distance defined as 1 - Jaccard index.

https://arxiv.org/abs/1606.04797 use a similar formulation (Dice coefficient, Section 3):
`2 * sum(A*B) / (sum(A**2) + sum(B**2))`

@FabianIsensee had [another implementation](https://github.com/Lasagne/Recipes/issues/99#issuecomment-281600662) of a dice coefficient:
`2 * sum(A*B) / (sum(A) + sum(B))`

Hmm, so I'm definitely leaning towards making this a loss function for training. But I'd be happier with the min/max formulation if we found a paper using it for training. There seems no real consensus on what "the real" Jaccard distance or Dice loss should be. If we want to add one to Lasagne, we should do more literature research to collect what people have been using and what works well. Otherwise we may want to leave it up to the user to implement the variant they fancy.

> Jaccard index is the same as Intersection-over-Union (IoU) metric which is being used in semantic segmentation/Object Detection tasks. PASCAL VOC uses this metric for evaluation.

Okay, that's good, but again, it assumes discrete predictions, not real-valued ones.

FabianIsensee:
Very interesting discussion. I think a dice/jaccard loss would make a great addition to lasagne. My implementation of the dice coefficients helped me beat the current state of the art in BraTS 2015 (2016 has no online evaluation platform available). 
You can see my results here: https://www.smir.ch/BRATS/Start2015 in section testing evaluation, my submission is placed 1st, Kayalibay et al (which is the second arxiv paper you mentioned) is placed 4th, although I would argue they should actually be placed second since the other two submissions do not cover all test cases and is it easy to cheat by not submitting everything. The difference between their and my results are very small, I guess both loss formulations yield approximately equivalent results.
You are right that we need to find a consensus about these formulations. Unfortunately, the literature is already pretty much covered by what you already posted (there is another MICCAI paper on the importance of skip connections that makes use of a dice loss) and there is no consensus here. I always boils down to the definition of 'union', 'intersection' and `|A|` in case of either A or B (or maybe even both) being non-binary. I think we should discuss these three separately.
Here is a repetition of the definition Jan wrote earlier, just sorted by measure:

'intersection'
1) `intersection(A,B) = sum(min(A,B))`
2) `intersection(A,B) = sqrt(sum((A*B)**2))`
3) `intersection(A,B) = sum(A*B)`
If either A or B is a binary vector (which is the case here) then 3) is equivalent to 1). I am not sure which formulation between 1)/3) and 2) is better. They basically differ in the way the pixel-wise similarities between A and B are accumulated.

'union'
1) `union(A,B) = sum(max(A,B))`
2) `union(A,B) = (sum(A**2) + sum(B**2))`
here I would also like to add my own implementation:
3) 'union(A,B) = sum(A) + sum(B) - intersect(A,B)' where `intersect(A,B)` follows 3) above
Again, if either a or B is binary, 1) and 3) should be equivalent.
Here I disagree with 2) since it follows more the definition of dice than jaccard. I would reformulate 2) as `union(A,B) = sqrt(sum(A**2)) + sqrt(sum(B**2)) - intersect(A,B)` where intersect is case 2) from above. I added sqrt to be consistent with the intersect

`|A|`
1) `|A| = sum(A**2)`
2) `|A| = sum(A)`
or you could even use
3) `|A| = sqrt(sum(A**2))`


If we consider 1) and 3) for intersect and union to be equivalent then there are only two alternatives left
a) `jacc(A,B) = intersect(A,B)/union(A,B) = sqrt(sum((A*B)**2)) / (sqrt(sum(A**2)) + sqrt(sum(B**2)) - sqrt(sum((A*B)**2)))'  (here we could also argue whether the sqrt is necessary or not)
b) `jacc(A,B) = intersect(A,B)/union(A,B) = sum(A*B) / (sum(A) + sum(B) - sum(A*B))'

Using the UNet road classification example I could run the experiment using both terms and compare performance, however that would take a while since I am going on vacation on friday and need all my GPU power before that. If you can wait that long (~3 weeks) then I will happily do it (I have in interest in findung the optimal solution myself). If someone else does it until then I will be happy as well ;-)

f0k:
> Very interesting discussion.

Thanks a lot for chiming in! This is really helpful.

> You are right that we need to find a consensus about these formulations.

Either that, or offer multiple variants. If there is no consensus in literature, we are not in a position to change that. There also seems to be no consensus on whether to use Jaccard or Dice, so maybe we should offer both.

> the definition of 'union', 'intersection' and |A| in case of either A or B (or maybe even both) being non-binary. I think we should discuss these three separately.

Yes, good idea!

> If we consider 1) and 3) for intersect and union to be equivalent then there are only two alternatives left

I think we should not dismiss the case where both A and B are non-binary. Imagine you're trying to predict a non-binary mask, with a non-binary target (e.g., for spectrogram-based source separation).

> `union(A,B) = sum(A) + sum(B) - intersect(A,B)`

That's a good idea to obtain a consistent definition. Note that if `intersect(A,B) = sum(min(A,B))`, this leads to `union(A,B) = sum(A) + sum(B) - sum(min(A,B)) = sum(max(A,B))`.

Also note that your lineup doesn't discuss the definitions of `union` and `intersection`, but of `|union|` and `|intersection|`. We could also try to split this up. Then we have:

* jaccard = |intersect(A,B)| / |union(A,B)|
* dice = 2 * |intersect(A,B)| / (|A| + |B|)
* intersect(A,B):
  * min(A,B)
  * A*B
* union(A,B):
  * max(A,B)
* |A|:
  * sum(A)
  * sum(A^2)
  * sqrt(sum(A^2))

To cover all definitions we had before, we still need:
* |union(A,B)|:
  * |A| + |B| - |intersect(A,B)|
  * |A|^2 + |B|^2 - |intersect(A,B)|

The latter is needed for https://arxiv.org/abs/1701.03056. The `|union(A,B)| = sum(A^2) + sum(B^2)` definition I had in https://github.com/Lasagne/Lasagne/pull/818#issuecomment-291927147 was incomplete, sorry -- they also have the `- |intersect(A,B)|` part in Eq 3 of https://arxiv.org/abs/1701.03056. The weird part is that they mix the squared and non-squared L2 norms in the denominator. It kind of makes sense if you assign a virtual unit to A and B, to have the virtual units of `sum(sqr(A))`, `sum(sqr(B))` and `sqrt(sum(sqr(A*B)))` match. I'm not sure what to make of this.

Now as I noted above, the `|union(A,B)|` definition in terms of `|A| + |B| + |intersect(A,B)|` recovers the `union(A,B) = max(A,B)` definition for `|A| = sum(A)` and `intersect(A,B) = min(A,B)`, so maybe there's no good reason to have `union(A,B) = max(A,B)` in the list. If we combine `union(A,B) = max(A,B)` with anything else than `|A| = sum(A)` and `intersect(A,B) = min(A,B)`, it will become inconsistent.

This leaves us with two choices for the intersection, three choices for the cardinality, and two choices for the union, as well as the choice between Jaccard and Dice.

Another way to see it would be to directly define `|intersect(A,B)|` (6 choices if we allow all combinations we had above, 4 choices if we avoid combining min(A,B) with L2, or more choices if we want to come up with something crazy), and then define `|union(A,B)| = |intersect(A,A)| + |intersect(B,B)| - |intersect(A,B)|` (for Jaccard) and `|A| + |B| = |intersect(A,A)| + |intersect(B,B)|` (for Dice). This would still exclude https://arxiv.org/abs/1701.03056.

Phew... is there anything we can argue against / rule out? I personally don't see why to use an L2 norm at all... for binary vectors, the squared L2 norm is the same as the L1 norm, but the square root seems to come out of the blue.

lazypoet:
@FabianIsensee Can you point me to your Dice loss implementation as I wanted to see it too. Btw, I saw it once somewhere and figured that you were using the dice loss including background. Am i right? Thanks for your help and congrats for your win!

Issue #814, opened 2017-03-15
Adding Huber Loss to objectives


Sentient07:
The huber loss is the least squared error if the absolute difference is less than a threshold(which is defined to be 1 for SmoothL1Loss), else it is `δ (|d|−δ/2)` where d is the absolute error. This is vastly used in regression problems and is useful in building models that generates bounding boxes over different objects in an image. 
The draft of the above described is [here](https://gist.github.com/Sentient07/bade6167bc923a4f0bf4a14196065cf7#file-theano-fastrcnn-py-L20-L24). If it is good to have this is Lasagne, I will open a PR :)

f0k:
Sounds good to me, with reference to Huber and to Fast RCNN!

PR #819, opened 2017-03-21
Added Huber loss

Sentient07:
For #814

### DIFF
--- a/lasagne/objectives.py
+++ b/lasagne/objectives.py
@@ -2,7 +2,7 @@
 Provides some minimal help with building loss expressions for training or
 validating a neural network.
 
-Five functions build element- or item-wise loss expressions from network
+Six functions build element- or item-wise loss expressions from network
 predictions and targets:
 
 .. autosummary::
@@ -13,6 +13,7 @@ predictions and targets:
     squared_error
     binary_hinge_loss
     multiclass_hinge_loss
+    huber_loss
 
 A convenience function aggregates such losses into a scalar expression
 suitable for differentiation:
@@ -86,6 +87,7 @@ __all__ = [
     "aggregate",
     "binary_hinge_loss",
     "multiclass_hinge_loss",
+    "huber_loss",
     "binary_accuracy",
     "categorical_accuracy"
 ]
@@ -336,6 +338,53 @@ def multiclass_hinge_loss(predictions, targets, delta=1):
                                  (-1, num_cls-1))
     rest = theano.tensor.max(rest, axis=1)
     return theano.tensor.nnet.relu(rest - corrects + delta)
+
+
+def huber_loss(predictions, targets, delta=1):
+    """ Computes the huber loss between predictions and targets.
+
+    .. math:: L_i = \\frac{(p - t)^2}{2},  |p - t| \\le \\delta
+
+        L_i = \\delta (|p - t| - \\frac{\\delta}{2} ), |p - t| \\gt \\delta
+
+    Parameters
+    ----------
+    predictions : Theano 2D tensor or 1D tensor
+        Prediction outputs of a neural network.
+
+    targets : Theano 2D tensor or 1D tensor
+        Ground truth to which the prediction is to be compared
+        with. Either a vector or 2D Tensor.
+
+    delta : scalar, default 1
+        This delta value is defaulted to 1, for `SmoothL1Loss`
+        described in Fast-RCNN paper [1]_ .
+
+    Returns
+    -------
+    Theano tensor
+        An expression for the element-wise huber loss [2]_ .
+
+    Notes
+    -----
+    This is an alternative to the squared error for
+    regression problems.
+
+    References
+    ----------
+    .. [1] Ross Girshick et al (2015):
+           Fast RCNN
+           https://arxiv.org/pdf/1504.08083.pdf
+
+    .. [2] Huber, Peter et al (1964)
+           Robust Estimation of a Location Parameter
+           https://projecteuclid.org/euclid.aoms/1177703732
+    """
+    predictions, targets = align_targets(predictions, targets)
+    abs_diff = abs(targets - predictions)
+    ift = 0.5 * squared_error(targets, predictions)
+    iff = delta * (abs_diff - delta / 2.)
+    return theano.tensor.switch(abs_diff <= delta, ift, iff)
 
 
 def binary_accuracy(predictions, targets, threshold=0.5):
--- a/lasagne/tests/test_objectives.py
+++ b/lasagne/tests/test_objectives.py
@@ -153,6 +153,30 @@ def test_binary_hinge_loss(colvect):
 
 
 @pytest.mark.parametrize('colvect', (False, True))
+@pytest.mark.parametrize('delta', (0.5, 1.0))
+def test_huber_loss(colvect, delta):
+    from lasagne.objectives import huber_loss
+    if not colvect:
+        a, b = theano.tensor.matrices('ab')
+        l = huber_loss(a, b, delta)
+    else:
+        a, b = theano.tensor.vectors('ab')
+        l = huber_loss(a.dimshuffle(0, 'x'), b, delta)[:, 0]
+
+    # numeric version
+    floatX = theano.config.floatX
+    shape = (10, 20) if not colvect else (10,)
+    x = np.random.rand(*shape).astype(floatX)
+    y = np.random.rand(*shape).astype(floatX)
+    abs_diff = abs(x - y)
+    ift = 0.5 * abs_diff ** 2
+    iff = delta * (abs_diff - delta / 2.)
+    z = np.where(abs_diff <= delta, ift, iff)
+    # compare
+    assert np.allclose(z, l.eval({a: x, b: y}))
+
+
+@pytest.mark.parametrize('colvect', (False, True))
 def test_binary_hinge_loss_not_binary_targets(colvect):
     from lasagne.objectives import binary_hinge_loss
     p = theano.tensor.vector('p')
--- a/docs/modules/objectives.rst
+++ b/docs/modules/objectives.rst
@@ -12,6 +12,7 @@ Loss functions
 .. autofunction:: squared_error
 .. autofunction:: binary_hinge_loss
 .. autofunction:: multiclass_hinge_loss
+.. autofunction:: huber_loss
 
 
 Aggregation functions


lasagne/objectives.py
```diff
@@ -338,6 +338,50 @@ def multiclass_hinge_loss(predictions, targets, delta=1):
...
+
+    Parameters
+    ----------
+    predictions : Theano 2D tensor
+        Predictions in (0, 1), such as softmax output of a neural network,
+        with data points in rows and class probabilities in columns.
+    targets : Theano 2D tensor or 1D tensor
+        Either a vector of int giving the correct class index per data point
+        or a 2D tensor of one-hot encoding of the correct class in the same
+        layout as predictions (non-binary targets in [0, 1] do not work!)
```
f0k: Wasn't this meant for regression? You copied the description from a multi-class loss.

lasagne/objectives.py
```diff
@@ -338,6 +338,50 @@ def multiclass_hinge_loss(predictions, targets, delta=1):
     return theano.tensor.nnet.relu(rest - corrects + delta)
 
 
+def huber_loss(predictions, targets, delta=1):
+    """ Computes the huber loss between predictions and targets.
+
+    .. math:: L_\delta (diff) = \frac{diff^2}{2} & \text{if  } |diff| \le \
+                                \delta
+    .. math:: L_\delta (diff) & = & \delta (|diff| - \frac{\delta}{2} ),\
+                                     &\text{else}
```
f0k: Haven't checked what this renders like, but shouldn't this use a single `.. math::` statement? http://www.sphinx-doc.org/en/stable/ext/math.html#directive-math
It seems rendering cases with a one-sided bracket is not easily possible in Sphinx, so using two lines is a good workaround.

lasagne/objectives.py
```diff
@@ -338,6 +338,50 @@ def multiclass_hinge_loss(predictions, targets, delta=1):
...
+    predictions : Theano 2D tensor
+        Predictions in (0, 1), such as softmax output of a neural network,
+        with data points in rows and class probabilities in columns.
+    targets : Theano 2D tensor or 1D tensor
+        Either a vector of int giving the correct class index per data point
+        or a 2D tensor of one-hot encoding of the correct class in the same
+        layout as predictions (non-binary targets in [0, 1] do not work!)
+    delta : scalar, default 1
+        This delta value is defaulted to 1, for SmoothL1Loss
+        described in Fast-RCNN paper[1].
```
f0k: Should be ` [1]_`, with an underscore in the end and a space before, if I recall the syntax correctly.

lasagne/objectives.py
```diff
@@ -338,6 +338,50 @@ def multiclass_hinge_loss(predictions, targets, delta=1):
...
+
+    Returns
+    -------
+    Theano 1D tensor
+        An expression for the item-wise huber loss.
+
+    Notes
+    -----
+    This is an alternative to the Least Squared loss for
+    regression problems.
```
f0k: `least squares`, or better `squared error` because that's what the function is called in Lasagne.

lasagne/objectives.py
```diff
@@ -338,6 +338,50 @@ def multiclass_hinge_loss(predictions, targets, delta=1):
...
+        or a 2D tensor of one-hot encoding of the correct class in the same
+        layout as predictions (non-binary targets in [0, 1] do not work!)
+    delta : scalar, default 1
+        This delta value is defaulted to 1, for SmoothL1Loss
+        described in Fast-RCNN paper[1].
+
+    Returns
+    -------
+    Theano 1D tensor
+        An expression for the item-wise huber loss.
```
f0k: Again, this was copied from a multi-class objective docstring. Copy it from the squared error instead (it should be a tensor of arbitrary dimensionality, and the element-wise loss).

lasagne/objectives.py
```diff
@@ -338,6 +338,50 @@ def multiclass_hinge_loss(predictions, targets, delta=1):
...
+    ----------
+    .. [1] Ross Girshick et al (2015):
+           Fast RCNN
+           https://arxiv.org/pdf/1504.08083.pdf
+    """
+    predictions, targets = align_targets(predictions, targets)
+    diff = targets - predictions
+    ift = 0.5 * squared_error(targets, predictions)
+    iff = delta * (abs(diff) - delta / 2.)
+    return theano.tensor.switch(abs(diff) <= delta, ift, iff).sum()
```
f0k: Don't sum in the end, it should return the element-wise loss.

lasagne/objectives.py
```diff
@@ -338,6 +338,50 @@ def multiclass_hinge_loss(predictions, targets, delta=1):
...
+    regression problems.
+
+    References
+    ----------
+    .. [1] Ross Girshick et al (2015):
+           Fast RCNN
+           https://arxiv.org/pdf/1504.08083.pdf
+    """
+    predictions, targets = align_targets(predictions, targets)
+    diff = targets - predictions
```
f0k: Use `abs_diff = abs(targets - predictions)` instead. Maximizes reuse of Theano expressions, so Theano doesn't have to merge them later.

lasagne/tests/test_objectives.py
```diff
@@ -153,6 +153,35 @@ def test_binary_hinge_loss(colvect):
 
 
 @pytest.mark.parametrize('colvect', (False, True))
+def test_huber_loss(colvect):
+    from lasagne.objectives import huber_loss
+    delta = [0.5, 1.0]
```
f0k: You can add `delta` to the test parameters and add another `parametrize`. Looks a little cleaner.

lasagne/objectives.py
```diff
@@ -338,6 +338,50 @@ def multiclass_hinge_loss(predictions, targets, delta=1):
...
+    Notes
+    -----
+    This is an alternative to the Least Squared loss for
+    regression problems.
+
+    References
+    ----------
+    .. [1] Ross Girshick et al (2015):
+           Fast RCNN
+           https://arxiv.org/pdf/1504.08083.pdf
```
f0k: Can you also cite Huber, maybe as the first reference? https://en.wikipedia.org/wiki/Huber_loss#cite_note-1

f0k:
Oh, and I forgot, the new function should be included in `__all__` at the top of the file, and in the module docstring at the top of the file, and in `docs/modules/objectives.rst`.

lasagne/objectives.py
Sentient07: I used two here because there are two different expressions that this will return, based on the value of delta. To represent that, I used two math.

lasagne/objectives.py
Sentient07: Ah! Yes, I will revert.

lasagne/objectives.py
f0k: Usually you'd use a single expression with a conditional bracket, but that's not easily possible in Sphinx (http://tex.stackexchange.com/questions/122407/writing-conditional-equations-with-braces-in-sphinx).
Using two lines seems like a good alternative, but you can write two lines in a single `.. math::` statement, as shown at http://www.sphinx-doc.org/en/stable/ext/math.html#directive-math. I guess that would be cleaner.

Sentient07:
I have added a new commit that addresses all the comments to the best of my knowledge. I haven't verified myself what sphinx renders. There seems to be some problem with my browser(safari). If the sphinx isn't fine, please let me know I will change it and try to build that on another system and push one more commit.
Thanks

lasagne/objectives.py
```diff
@@ -86,6 +86,7 @@
     "aggregate",
     "binary_hinge_loss",
     "multiclass_hinge_loss",
+    "huber_loss",
```
f0k: Please also add it to the module docstring further above.

lasagne/objectives.py
```diff
@@ -341,45 +342,50 @@ def multiclass_hinge_loss(predictions, targets, delta=1):
 def huber_loss(predictions, targets, delta=1):
     """ Computes the huber loss between predictions and targets.
 
-    .. math:: L_\delta (diff) = \frac{diff^2}{2} & \text{if  } |diff| \le \
-                                \delta
-    .. math:: L_\delta (diff) & = & \delta (|diff| - \frac{\delta}{2} ),\
+    .. math:: 
+        L_\delta (diff) = \frac{diff^2}{2} & \text{if  } |diff| \le \ delta \\
+
+        L_\delta (diff) & = & \delta (|diff| - \frac{\delta}{2} ),\ \\
```
f0k: The alignment characters (`&`) still seem off -- does this render correctly in Sphinx?

lasagne/objectives.py
```diff
@@ -341,45 +342,50 @@ def multiclass_hinge_loss(predictions, targets, delta=1):
...
-        with data points in rows and class probabilities in columns.
+    predictions : Theano 2D tensor or 1D tensor
+        Predictions output of a neural network
+
     targets : Theano 2D tensor or 1D tensor
-        Either a vector of int giving the correct class index per data point
-        or a 2D tensor of one-hot encoding of the correct class in the same
-        layout as predictions (non-binary targets in [0, 1] do not work!)
+        Ground truth to which the prediction is to be compared
+        with.
```
f0k: Please copy the `predictions` and `targets` from `binary_crossentropy`. This is still not correct.

lasagne/objectives.py
```diff
@@ -341,45 +342,50 @@ def multiclass_hinge_loss(predictions, targets, delta=1):
...
+    This is an alternative to the squared error for
     regression problems.
 
     References
     ----------
     .. [1] Ross Girshick et al (2015):
            Fast RCNN
            https://arxiv.org/pdf/1504.08083.pdf
+
+    .. [2] Huber, Peter et al (1964)
```
f0k: This line should have a colon `:` in the end. Sorry for the apparent nitpick, but note that the line break is only visible when viewing the docstring in Python, not in the Sphinx rendering. Also please add a period `.` after the title. Have a look at how this renders to be sure. (Last bullet point: https://github.com/Lasagne/Lasagne/blob/master/.github/PULL_REQUEST_TEMPLATE.md)

lasagne/objectives.py
```diff
@@ -341,45 +342,50 @@ def multiclass_hinge_loss(predictions, targets, delta=1):
...
+           Robust Estimation of a Location Parameter
+           https://projecteuclid.org/euclid.aoms/1177703732
     """
     predictions, targets = align_targets(predictions, targets)
-    diff = targets - predictions
+    ab_diff = abs(targets - predictions)
```
f0k: I'd strongly prefer `abs_diff`.

lasagne/tests/test_objectives.py
```diff
@@ -153,17 +153,17 @@ def test_binary_hinge_loss(colvect):
...
+@pytest.mark.parametrize('delta', (0.5, 1.0))
+def test_huber_loss(colvect, delta):
     from lasagne.objectives import huber_loss
-    delta = [0.5, 1.0]
     if not colvect:
         a, b = theano.tensor.matrices('ab')
-        l1 = huber_loss(a, b, delta[0])
-        l2 = huber_loss(a, b, delta[1])
+        l1 = huber_loss(a, b, delta)
+        l2 = huber_loss(a, b, delta)
```
f0k: You're computing the same thing twice now. The idea of making `delta` a test parameter was to remove the duplication :)

lasagne/objectives.py
f0k: And to the `.rst` file as I mentioned in https://github.com/Lasagne/Lasagne/pull/819#issuecomment-291599153. Thank you!

lasagne/objectives.py
f0k: Ah, didn't see the comment about your browser problem. Do you have another browser you can try? (Using another computer seems overkill!) Otherwise I can also check out your PR and test the documentation here, but of course it's easier if you can debug and fix things directly!

lasagne/tests/test_objectives.py
Sentient07: Ah! Yeah, sorry. I rushed through that PR. I will check every comment again before the next commit.

lasagne/objectives.py
Sentient07: It was a typo. I don't have anything against `abs_diff`, but yeah I understand the convention :)

lasagne/objectives.py
Sentient07: I am sorry, it's a little unclear to me. You mean to say, i should be having predictions and targets in a single line separated by `,` similar to squared_error?

Sentient07: Something like this?
```
    Parameters
    ----------
    predictions,  targets : Theano tensors
```

f0k: Yes, or you can also call it `a` and `b`, since the loss is completely symmetric. Your current docstring says they should be 1D or 2D tensors, which is too strict! They can be anything -- it's a drop-in replacement for the squared error.

Sentient07:
I very much apologise for the delay. The two months were really hectic in seeking positions. I have addressed the comments that were left over. The sphinx now looks good in my local build. If there are anymore changes, I will make them right away! 
screenshot at jun 01 01-24-30

lasagne/tests/test_objectives.py
```diff
@@ -153,6 +153,30 @@ def test_binary_hinge_loss(colvect):
...
+
+    # numeric version
+    floatX = theano.config.floatX
+    shape = (10, 20) if not colvect else (10,)
+    x = np.random.rand(*shape).astype(floatX)
+    y = np.random.rand(*shape).astype(floatX)
+    abs_diff = abs(x - y)
+    ift = 0.5 * abs_diff ** 2
+    iff = delta * (abs_diff - delta / 2.)
+    z = np.where(abdiff <= delta, ift, iff)
```
f0k: I'd merge this right away, only all your tests failed with `NameError: global name 'abdiff' is not defined`.

Sentient07: Sorry about that, I will fix it.

f0k:
Looks good now, can you please squash everything into the first commit?

PR #820, opened 2017-03-27
Make use of expm1() to implement ELU

danstowell:
expm1(x) is equivalent to exp(x)-1 but should be more stable since the calculation can be more precise. Available since Theano 0.6rc1 (Oct 2012).

### DIFF
--- a/lasagne/nonlinearities.py
+++ b/lasagne/nonlinearities.py
@@ -266,7 +266,7 @@ def elu(x):
        Fast and Accurate Deep Network Learning by Exponential Linear Units
        (ELUs), http://arxiv.org/abs/1511.07289
     """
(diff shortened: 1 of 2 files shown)


f0k:
Oh, yes, of course! Thanks a lot, merging!

f0k:
By the way, Theano knows this as well:
```
>>> import theano
>>> x = theano.tensor.scalar('x')
>>> y = theano.tensor.exp(x) - 1
>>> theano.printing.debugprint(theano.function([x], y))
Elemwise{expm1,no_inplace} [id A] ''   0
 |x [id B]
```
So it always used `expm1` for ELU, but it's still nice to create a smaller graph to begin with.

danstowell:
Aha - thanks for pointing that out

nouiz:
It still make Theano optimizer do less work, so this is good.

Issue #698, opened 2016-06-13
Support Request: Add support to theano's new backend


Sentient07:
When i use the `lasagne.layers.dnn` module, it checks with theano's old backend for availability of cuDNN.
`theano.sandbox.cuda.cuda_available`
 As the old backend will be depreciated soon and developments are taking place in the new backend, i would really like to see the `lasagne.layers.dnn` module support new backend. 
For a workaround, I am currently using `lasagne.layers.Conv2DLayer` instead of `Conv2DDNNLayer` module. 
cc : @nouiz

benanne:
Is there some sort of roadmap for Theano we can consult to determine when this is scheduled to happen, so we can prepare? The new backend has been in development for a very long time (longer than Lasagne has existed), and as far as I know the majority of Theano users are still using the old backend right now.

f0k:
> For a workaround, I am currently using `lasagne.layers.Conv2DLayer` instead of `Conv2DDNNLayer` module.

That's recommended anyway. There is no benefit of using `Conv2DDNNLayer` since Theano's overhaul of the conv2d interface.
Is there anything else blocking you from using the new backend?

nouiz:
We have only 1 thing to fix to have the new back-end a little faster then
the old one (~10% speed up in one case). After that, only a few not
frequent op haven't been implemented.

So I would guess we are pretty close to use it extensively here (in
particular for float16 in storage).

f0k:
Then maybe our roadmap should be: Release Lasagne 0.2 for Theano 0.8, then switch things over to use the new backend explicitly (assuming that supporting both backends at once is a bit cumbersome).

nouiz:
I think it should be easy to support both. If not, tell us. Mostly, from
the user point of view, it is just a few Theano flags difference.

benanne:
Supporting both will by definition require separate code paths in some cases, won't it? (I would imagine especially for the convolutional layers.) If not then that's great, but then there wouldn't be an issue in the first place and people could just switch their flags from `device=gpu0` to `device=cuda0`. I think I agree with @f0k, we might want to do a "hard" switch and avoid littering the code with a bunch of if-statements to support both backends.

Also, if there are no use cases anymore where `Conv2DDNNLayer` offers an advantage over `Conv2DLayer`, we should probably deprecate it and encourage people to switch to the latter.

f0k:
> Supporting both will by definition require separate code paths in some cases, won't it?

[...]

PR #823, opened 2017-04-04
Gpuarray

borisfom:
This should address https://github.com/Lasagne/Lasagne/issues/698


f0k:
Thanks for your contribution! Unfortunately, this is (a) not entirely correct (see the failing tests) and (b) removes compatibility with the old backend. As detailed in https://github.com/Lasagne/Lasagne/issues/698#issuecomment-290702908, I'd prefer a PR that makes Lasagne compatible with both backends, and only drop compatibility to the old backend when Lasagne requires Theano 0.10 (which won't have the old backend). Furthermore, as detailed in https://github.com/Lasagne/Lasagne/issues/698#issuecomment-290538224, Lasagne already works with the new backend anyway, only the implementation-specific convolutional layers don't (but they're usually not needed).

I think @Sentient07 may also work on a PR for dual-backend compatibility. @Sentient07, Is that true? I'm just assuming based on your questions in #698.

Otherwise, are you interested in extending your PR towards both backends? This will require more than a simple search & replace; we'll need to figure out how to test for CUDA support in a backend-agnostic way (without importing sandbox.cuda in case the new backend is used), do conditional imports of the cuDNN and gemm Ops (in layers/dnn and layers/corrmm), and possibly adapt the faked CUDA support for Sphinx accordingly.

Sentient07:
@f0k yes, i started working on it yesterday, addressing the places that Pascal suggested. I still have a little bit of cleanup remaining before I open a PR. 

>  and possibly adapt the faked CUDA support for Sphinx accordingly.

I don't understand this. could you please elaborate? Is it about updating the documentation ?

f0k:
> I don't understand this. could you please elaborate? Is it about updating the documentation ?

No, it's about rendering the documentation. The online documentation is built at readthedocs, which doesn't have a GPU. In order to render the documentation for `lasagne.layers.dnn`, Sphinx still has to be able to import `lasagne.layers.dnn` at readthedocs, despite the missing GPU support. To do that, we had to fake GPU support in the Sphinx configuration for Lasagne: https://github.com/Lasagne/Lasagne/blob/master/docs/conf.py#L319-L340
If we change the GPU support check in `lasagne.layers.dnn` (https://github.com/Lasagne/Lasagne/blob/master/lasagne/layers/dnn.py#L13-L21), which we have to do in order to support both backends, then we may also have to update the faked GPU support in the Sphinx configuration (i.e., find another hack that allows to import `lasagne.layers.dnn` without a GPU).

Sentient07:
That made things much clear! Thank you

f0k:
#836 provides a solution that support both backends, so I'll close this. The suggested documentation/docstring changes will be useful later. Thank you!

PR #840, opened 2017-06-09
TransposedConv3DLayer

FabianIsensee:
Hey there,
since Lasagne now has a Conv3DLayer but not yet a TransposedConv3DLayer I thought I could share my implementation. I also added tests and a proper docstring (most of which was xopy pasted from TransposedConv2DLayer). All tests were successful on my machine.
Cheers,
Fabian

### DIFF
--- a/lasagne/layers/corrmm.py
+++ b/lasagne/layers/corrmm.py
@@ -8,19 +8,32 @@ from .base import Layer
 from .conv import conv_output_length, BaseConvLayer
 from ..utils import as_tuple
 
-from theano.sandbox.cuda.basic_ops import gpu_contiguous
-from theano.sandbox.cuda.blas import GpuCorrMM
-
+# check if Theano's new GPU backend is available and in use
+try:
+    from theano import gpuarray as gpu
+except ImportError:
+    from theano.sandbox import gpuarray as gpu
+gpu_enabled = gpu.pygpu_activated
+# if not, try to fall back to Theano's old GPU backend
+if not gpu_enabled:
+    try:
+        from theano.sandbox import cuda as gpu
+    except ImportError:
+        gpu_enabled = False
+    else:
+        gpu_enabled = gpu.cuda_enabled
+# if either of the backends is available, use it, otherwise bail out
+if gpu_enabled:
+    gpu_contiguous = gpu.basic_ops.gpu_contiguous
+    GpuCorrMM = gpu.blas.GpuCorrMM
+else:
+    raise ImportError(
+        "requires GPU support -- see http://lasagne.readthedocs.org/en/"
+        "latest/user/installation.html#gpu-support")  # pragma: no cover
 
 __all__ = [
     "Conv2DMMLayer",
 ]
-
-
-if not theano.sandbox.cuda.cuda_enabled:
-    raise ImportError(
-            "requires GPU support -- see http://lasagne.readthedocs.org/en/"
-            "latest/user/installation.html#gpu-support")  # pragma: no cover
 
 
 class Conv2DMMLayer(BaseConvLayer):
--- a/lasagne/layers/dnn.py
+++ b/lasagne/layers/dnn.py
@@ -1,5 +1,4 @@
 import theano
-from theano.sandbox.cuda import dnn
 
 from .. import init
 from .. import nonlinearities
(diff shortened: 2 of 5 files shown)


FabianIsensee:
Hmm tests in Travis CI are failing because T.nnet.abstract_conv.AbstractConv3d_gradInputs seems to be missing? Could it be that it requires the cutting edge theano version? On my system all tests passed. Here is my theano version:
In [2]: theano.__version__
Out[2]: '0.10.0dev1.dev-RELEASE'

FabianIsensee:
(Fraom travis CI output:)
from theano.tensor.nnet.abstract_conv import AbstractConv3d_gradInputs
ImportError: cannot import name AbstractConv3d_gradInputs

I don't see why this does not work. Any ideas? Could it really be the theano version?

f0k:
Thank you for the PR!

> Could it be that it requires the cutting edge theano version?

Yes, it's the same issue as with the Conv3DLayer itself. My solution was to hide it when Theano is too old, and skip the test when it is not available: https://github.com/Lasagne/Lasagne/commit/e0eac0e. Can you do the same?

After Lasagne 0.2, we can bump the requirement to Theano 0.9 and get rid of these conditionals. I wanted to have Lasagne 0.2 be compatible to Theano 0.8.2, the last one not deprecating the old backend -- but I wasn't planning to take so long for it. (I'm currently writing my thesis and can't work on the missing bits and pieces for the release.)

lasagne/layers/conv.py
```diff
@@ -1127,8 +1126,16 @@ def convolve(self, input, **kwargs):
         conved = op(self.W, input, output_size)
         return conved
 
+
 Deconv3DLayer = TransposedConv3DLayer
 
+if not hasattr(T.nnet.abstract_conv, 'AbstractConv3d_gradInputs'):
+    # pragma: no cover
```
f0k: The `# pragma: no cover` must be on the same line as the `if` case. It's caught by a regular expression that ignores all lines which have this string in them. The following should work:
```python
if not hasattr(T.nnet.abstract_conv,
               'AbstractConv3d_gradInputs'):  # pragma: no cover
```

lasagne/layers/conv.py
```diff
@@ -15,6 +15,8 @@
     "TransposedConv2DLayer",
     "Deconv2DLayer",
     "DilatedConv2DLayer",
+    "TransposedConv3DLayer",
+    "Deconv3DLayer"
```
f0k: Don't forget the comma in the end, so we can add more lines without touching the existing ones.

lasagne/layers/conv.py
```diff
@@ -930,9 +932,211 @@ def convolve(self, input, **kwargs):
         conved = op(self.W, input, output_size)
         return conved
 
+
```
f0k: The idea was to have the aliases closer to the class definition (with a single empty line) than to other class definitions (two empty lines). pep8 was happy with it.

lasagne/layers/conv.py
```diff
@@ -930,9 +932,211 @@ def convolve(self, input, **kwargs):
...
+        output_size = self.output_shape[2:]
+        if isinstance(self.output_size, T.Variable):
+            output_size = self.output_size
+        elif any(s is None for s in output_size):
+            output_size = self.get_output_shape_for(input.shape)[2:]
+        conved = op(self.W, input, output_size)
+        return conved
+
+
+Deconv3DLayer = TransposedConv3DLayer
```
f0k: Again, I'd prefer a single blank line before the alias, for consistency.

f0k:
Ah, and we need to unhide the layer for building the documentation. Basically we need to mock the attribute that you check for hiding the layer: https://github.com/Lasagne/Lasagne/commit/e0eac0e#diff-85987f48f1258d9ee486e3191495582dR352
It's all a bit weird, I know, but it's the best solution I found for supporting multiple Theano versions.

lasagne/layers/conv.py
FabianIsensee: I put that in the next line because I was not sure whether the pep8 tests will fail (long line). Changed it back to inline

lasagne/layers/conv.py
```diff
@@ -933,6 +935,207 @@ def convolve(self, input, **kwargs):
 Deconv2DLayer = TransposedConv2DLayer
 
 
+class TransposedConv3DLayer(BaseConvLayer):
```
f0k: Ah, we need a `# pragma: no cover` after the class definition as well, because it cannot be tested with the Theano version used on Travis.

f0k: That's also the reason why I had two commits in #791. Makes it easy to undo the compatibility changes when we drop Theano 0.8 compatibility.

Hmm, in the end I would like to have two commits for this PR as well -- are you okay with me rearranging and squashing the commits as needed? I can add the missing `# pragma: no cover` as well then. (You'll still be the author of all commits.)

FabianIsensee: Sure! Go ahead ;-)

f0k:
Oops, seems I accidentally pushed my `master` instead of the commit I wanted. Now github helpfully closed the PR because it is identical to Lasagne's `master`. It says pushing a new commit will reopen the PR, but it does not allow me to commit anything presumably because the PR is closed.

Can you just do a `git push --force` from your `master` again so I get back write access?

f0k:
If you feel adventurous, you can also do:
```bash
git checkout -b backup
git checkout master
git reset --hard upstream/master
git pull https://github.com/f0k/Lasagne.git FabianIsensee-master
git push --force
```
This should put the PR in the shape I intended it to be :)

FabianIsensee:
Sure. One sec

FabianIsensee:
git push --force does not work for me. Git says 'everything up to date'. Also tried your second suggestion to no avail. I will just do a dummy commit (add empty file and then a second commit to delete it)

FabianIsensee:
That did also not reopen the pull request. I will just make a new one from the backup branch i created

f0k:
> That did also not reopen the pull request.

Might be a github bug then. When it automatically closed the request, it said it will be reopened when pushing something. Maybe it doesn't work because it's your `master` branch, or maybe it's just a feature that isn't tested too well...

PR #841, opened 2017-06-09
TransposedConv3DLayer

FabianIsensee:
Second attempt ;-)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -352,6 +352,9 @@ except NameError:
 if not hasattr(theano.tensor.nnet, 'conv3d'):
     theano.tensor.nnet.conv3d = Mock()
     reload(lasagne.layers.conv)
+if not hasattr(theano.tensor.nnet.abstract_conv, 'AbstractConv3d_gradInputs'):
+    theano.tensor.nnet.abstract_conv.AbstractConv3d_gradInputs = Mock()
+    reload(lasagne.layers.conv)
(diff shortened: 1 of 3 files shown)


f0k:
Pushed mine again. I wonder why the first commit is not attributed to you any more. It's authored by "Fabian Ibensee" in the log, using your heidelberg email address.

FabianIsensee:
Hm yeah I would very much like to retain the authorship of that commit. Do you know how we could fix this? I am not much of a git ninja

f0k:
I've changed it such that both the author and committer are your name and heidelberg email address. Is that a different one than you registered on github?

FabianIsensee:
Oh I did the new pull request form a laptop at home where I did not have github set properly. The email address is correct though (gmail and heidelberg are both fine). I now added the heidelberg address to github as well

FabianIsensee:
Hm now you don't show up anymore... Or am I missing something?

f0k:
> I now added the heidelberg address to github as well

You always had the heidelberg address in the commits, but probably github knew it was you because you pushed them.

> Hm now you don't show up anymore...

That was on purpose. I just rearranged things and fixed a small error (you removed one of the checks in docs/conf.py and duplicated another). I think it's good to merge now. Will merge in the evening or so if there are no other comments.

FabianIsensee:
OK great! Thank you!

PR #842, opened 2017-06-13
Extend TransposedConvLayer

frapa:
Hi, I partly rewrote the TransposedConv2DLayer in order to support both 1D and 3D transposed convolutions. I saw there is already a pull request for it (namely #841), but I hope my implementation is more complete.

I don't know how to run tests and how to run tests as I get the following error:

```
usage: py.test [options] [file_or_dir] [file_or_dir] [...]
py.test: error: unrecognized arguments: --cov=lasagne --cov-report=term-missing --pep8
  inifile: /usr/stud/pasa/Documents/Lasagne/setup.cfg
  rootdir: /usr/stud/pasa/Documents/Lasagne
```

I also tried updating the documentation. Is it possible to see a preview somewhere?

### DIFF
--- a/lasagne/nonlinearities.py
+++ b/lasagne/nonlinearities.py
@@ -159,7 +159,7 @@ def rectify(x):
 
 # leaky rectify
 class LeakyRectify(object):
-    """Leaky rectifier :math:`\\varphi(x) = \\max(\\alpha \\cdot x, x)`
+    """Leaky rectifier :math:`\\varphi(x) = (x > 0)? x : \\alpha \\cdot x`
 
     The leaky rectifier was introduced in [1]_. Compared to the standard
     rectifier :func:`rectify`, it has a nonzero gradient for negative input,


frapa:
Ok, the error was due to missing dependencies. My code fails some shape tests on TransposedConv2DLayer...

f0k:
Hey, good move! When looking at #840/#841, I also took a mental note that we'll need to introduce a common base class later to avoid duplicated code.

> I don't know how to run tests and how to run tests as I get the following error:

Seems you figured out this was because you missed the `pip install -r requirements-dev.txt` step in http://lasagne.readthedocs.io/en/latest/user/development.html#development-setup.

> I also tried updating the documentation. Is it possible to see a preview somewhere?

Yes, see http://lasagne.readthedocs.io/en/latest/user/development.html#documentation.

> My code fails some shape tests on TransposedConv2DLayer...

That's because you do not pass on the `output_size` argument in your `super` call (for all three layers). It's good the tests exist :)

> There are also some import error that on my machine I am unable to reproduce.

This is because Lasagne is currently meant to be compatible to Theano 0.8, which lacks the abstract conv3d interface (it only has abstract conv2d). See https://github.com/Lasagne/Lasagne/pull/840#issuecomment-307390362 and following. The solution entails hiding the TransposedConv3DLayer when it's unavailable, mocking it for Sphinx so it's still included in the documentation, and skipping the tests when it's unavailable: https://github.com/Lasagne/Lasagne/pull/841/commits/d0a3916

I'm not 100% sure how to continue -- I'd still like to merge #841 to give Fabian credit for his work, but I'd also like to have a base class for the transposed convolutions in the end. Shall we merge #841 and then rebase this one? It's extra work, but then it will also include the necessary workaround for Theano 0.8 and additional tests (which will have to be refactored as well in the end, so they cover all three dimensionalities).

frapa:
I think merging #841 and rebasing is a good way of handling this, too bad I did not check the issues before extending. Then I can check the two implementations and fix them taking advantage of what Fabian has already done.

frapa:
The issue is now #850

PR #843, opened 2017-06-13
Added SELU nonlinearity

partobs-mdp:
Implementing SELU nonlinearity (inspired by [this](https://arxiv.org/abs/1706.02515) arXiv paper)

This is my first commit to Lasagne, so I don't quite know the rules (yet). I understand, though, that just implementing a nonlinearity is not good enough to make the cut, so I would like to ask what else should I do? (extra examples, docs, ...)

Looking forward to your feedback ;)


f0k:
Welcome, and thank you for the PR!

> I understand, though, that just implementing a nonlinearity is not good enough to make the cut

Indeed! It also needs tests so the coverage stays at 100%, and must be added to the `nonlinearities.rst` files so it shows up in the documentation. See the changeset of the closely related ELU (https://github.com/Lasagne/Lasagne/pull/518/files) and scaled tanh (https://github.com/Lasagne/Lasagne/pull/430/files).

See http://lasagne.readthedocs.io/en/latest/user/development.html#how-to-contribute on how to build the documentation locally, and how to run the test suite locally. This allows you to check if everything looks fine and coverage is at 100%.

Besides, there are some other issues with your PR:
* The reference must be referenced with `[1]_`, not `[1]`
* The reference should include all authors
* `elu` is a special case of `SELU` now -- we could implement it as such, but actually it may not be worth the effort since `elu` is just a single line of code. We did the same when introducing scaled tanh; there was no point in making `tanh` a `ScaledTanh` instance.
* it's not clear what values to use for the two parameters, I'd suggest to add a `selu` instance with the parameters suggested by the authors (https://github.com/bioinf-jku/SNNs/blob/49c5298/selu.py#L23-L24). I.e., after defining the class, add `selu = SELU(..., ...)`, give it a docstring as we did here: https://github.com/Lasagne/Lasagne/blob/6eb2f86/lasagne/nonlinearities.py#L223-L227, and reference it from the `SELU` docstring as we did here: https://github.com/Lasagne/Lasagne/blob/6eb2f86/lasagne/nonlinearities.py#L204
* The implementation is not in line with the docstring and paper (`>=` vs `>`). Take the `elu` implementation as a starting point.
* `SELU` should come directly after `elu` in both the code and documentation since they're related.

gyglim:
Hi. Thanks for starting this PR

I it should also include AlphaDropout, which keeps the mean and variance bounded, in contrast to normal Dropout. See:
https://github.com/fchollet/keras/pull/6924#issuecomment-307573440

f0k:
> I it should also include AlphaDropout,

Yes, but I'd treat this as a separate PR, it's okay to add one after the other.

@partobs-mdp: Will you still work on this or should somebody else take it? The paper is getting old ;)

partobs-mdp:
@f0k Working (right now trying to build docs) - just had a hard time doing both university exams and GSoC project 😞

f0k:
> just had a hard time doing both university exams and GSoC project

No worries! Just wanted to check back. Let me know if you need help.

partobs-mdp:
Spkeaing about docs, I can't run `make html`, beacuse I run into some weird bug - Theano crashes during building docs. Log: [https://gist.github.com/partobs-mdp/04b70d4a80d229c541931e8b22324999](https://gist.github.com/partobs-mdp/04b70d4a80d229c541931e8b22324999)

Running `make html` from my venv (`make html SPHINXBUILD='python /usr/bin/sphinx-build'` doesn't help either)

partobs-mdp:
Aside from that problem and tests, I seem to have implemented everything from @f0k's initial review comment - so now help on that doc issue would be very appreciated ;-)

The current code (the one that reproduces the error I described above) is in the last commit.

FWIW, I also attach output of `make html` ran *without* venv: [https://gist.github.com/partobs-mdp/8517691031c93035fee50dd2a101271d](https://gist.github.com/partobs-mdp/8517691031c93035fee50dd2a101271d)

f0k:
> Spkeaing about docs, I can't run `make html`

Sorry, this was fixed in #846, you'll need to rebase:
```bash
git fetch upstream master
git rebase upstream/master
git push --force
```
Or a shorter version:
```bash
git pull --rebase upstream master
git push --force
```
Let me know if it works!

docs/modules/nonlinearities.rst
```diff
@@ -14,6 +14,7 @@
    leaky_rectify
    very_leaky_rectify
    elu
+   SELU
```
f0k: also add selu here

docs/modules/nonlinearities.rst
```diff
@@ -33,6 +34,8 @@ Detailed description
 .. autofunction:: leaky_rectify
 .. autofunction:: very_leaky_rectify
 .. autofunction:: elu
+.. autoclass:: SELU
+   :members:
```
f0k: and `.. autofunction:: selu`

lasagne/nonlinearities.py
```diff
@@ -269,6 +269,70 @@ def elu(x):
     return theano.tensor.switch(x > 0, x, theano.tensor.expm1(x))
 
 
+
+# selu
+class SELU(object):
+    """Scaled Exponential Linear Unit
+    :math:`\\varphi(x) = \\lambda (x > 0 ? x : \\alpha(e^x-1)`
```
f0k: You're missing a closing bracket in the end.

lasagne/nonlinearities.py
```diff
@@ -269,6 +269,70 @@ def elu(x):
...
+        self.scale_neg = scale_neg
+
+    def __call__(self, x):
+        return self.scale * theano.tensor.switch(
+            x > 0.0,
+            x,
+            self.scale_neg * (theano.tensor.exp(x) - 1))
+
+
+selu = SELU(scale=1.0507, scale_neg=1.6733)
```
f0k: Just because we can, I'd use all digits given by the authors: https://github.com/bioinf-jku/SNNs/blob/49c5298/selu.py#L23-L24.

lasagne/nonlinearities.py
```diff
@@ -269,6 +269,70 @@ def elu(x):
...
+selu.__doc__ = """selu(x)
+    Instance of :class:`SELU` with :math:`\\alpha=1.6733, \\lambda=1.0507`
+    (the values that were recommended in the original paper or self-normalzing neworks)
```
f0k: Should be two sentences, because the first sentence is used for the overview table at the beginning of the module documentation. I'd suggest:
```
Instance of :class:`SELU` with :math:`\\alpha \\approx 1.6733, \\lambda \\approx 1.0507`.
This has a stable and attracting fixed point of :math:`\\mu=0`, :math:`\\sigma=1` under
the assumptions of the original paper on self-normalizing neural networks.
```

lasagne/nonlinearities.py
```diff
@@ -269,6 +269,70 @@ def elu(x):
...
+
+    >>> from lasagne.layers import InputLayer, DenseLayer
+    >>> l_in = InputLayer((None, 100))
+    >>> from lasagne.nonlinearities import SELU
+    >>> selu = SELU(2, 3)
+    >>> l1 = DenseLayer(l_in, num_units=200, nonlinearity=selu)
+
+    See Also
+    --------
+    selu: Instance with :math:`\\alpha=1.6733, \\lambda=1.0507`, as recommended in [1]_.
```
f0k: Let's stay more neutral and say `as used in [1]_.` They don't necessarily recommend it, at least not so generally.

lasagne/nonlinearities.py
```diff
@@ -269,6 +269,70 @@ def elu(x):
...
+
+    def __init__(self, scale=1, scale_neg=1):
+        self.scale = scale
+        self.scale_neg = scale_neg
+
+    def __call__(self, x):
+        return self.scale * theano.tensor.switch(
+            x > 0.0,
+            x,
+            self.scale_neg * (theano.tensor.exp(x) - 1))
```
f0k: We generally use 8 spaces for hanging indent, not 4. Also I'd prefer `self.scale_neg * theano.tensor.expm1(x)` for the last case.

partobs-mdp:
Looks like I've implemented everything from @f0k's review - now swtiching to writing tests.

f0k:
Thanks, there are several PEP8 violations, though. But you'll also see them when you run the tests locally. Just remember to fix them as well.

partobs-mdp:
Fixed PEP8 issues + added several tests for `SELU`. There is one weird issue in doctest (but it also arises in other class-based nonlinearities, e.g. `ScaledTanH`), don't really know what I have to do with it.

partobs-mdp:
Why review is still red, by the way? I have implemented everything from it (or at least I think so)

partobs-mdp:
The Travis CI build is over - it crashed on those three classes (`LeakyRectify`, `ScaledTanh` and `SELU`). What have I messed up?

ebenolson:
You need blank lines after the ">>>" doctest sections, otherwise it thinks
the next line should be the output.

partobs-mdp:
@ebenolson, thanks, looks like *now* I managed to get it right :)

lasagne/nonlinearities.py
```diff
@@ -9,12 +9,10 @@
 # sigmoid
 def sigmoid(x):
     """Sigmoid activation function :math:`\\varphi(x) = \\frac{1}{1 + e^{-x}}`
-
```
f0k: In commit https://github.com/Lasagne/Lasagne/pull/843/commits/58904e3d9de56c4a0de068addc72ac96429bdb6f, you removed almost all blank lines in lasagne/nonlinearities.py. Please undo!

partobs-mdp:
First of all, sorry for the late reply - I did something dumb while `pacman -Syua`. As a result, right now I'm back to a fresh Ubuntu install without `anaconda` and other good stuff. However, I do have `git` and `vim` now :smiley: so I have returned blanklines as you reviewed.

Right now: waiting for Travic CI build to find some test / pep8 crashes to fix.

partobs-mdp:
Well, _one_ blank line was indeed extra (`nonlinearities.py:273`). Fixed that, and I think that this check will run smoothly.

lasagne/nonlinearities.py
```diff
@@ -107,7 +107,6 @@ class ScaledTanH(object):
 
     By carefully matching :math:`\\alpha` and :math:`\\beta`, the nonlinearity
     can also be tuned to preserve the mean and variance of its input:
-
```
f0k: You missed this one... it will break the rendering of the documentation. Please always check the complete diff!

lasagne/nonlinearities.py
```diff
@@ -269,6 +268,76 @@ def elu(x):
...
+    >>> from lasagne.layers import InputLayer, DenseLayer
+    >>> l_in = InputLayer((None, 100))
+    >>> from lasagne.nonlinearities import SELU
+    >>> selu = SELU(2, 3)
+    >>> l1 = DenseLayer(l_in, num_units=200, nonlinearity=selu)
+
+    See Also
+    --------
+    selu: Instance with :math:`\\alpha\\approx 1.6733,
+    \\lambda\\approx 1.0507`, as used in [1]_.
```
f0k: I think you cannot break this into two lines. Please check how the HTML documentation renders. Possibly use `=` instead of `\\approx`. It's not entirely correct, but close enough.

lasagne/nonlinearities.py
```diff
@@ -269,6 +268,76 @@ def elu(x):
...
+
+    References
+    ----------
+    .. [1] Günter Klambauer, Thomas Unterthiner,
+           Andreas Mayr, Sepp Hochreiter (2017):
+       Self-Normalizing Neural Networks,
+       https://arxiv.org/abs/1706.02515
```
f0k: I think you cannot mix indentations here, but I may be mistaken. Please check how the HTML documentation renders.

lasagne/nonlinearities.py
```diff
@@ -269,6 +268,76 @@ def elu(x):
...
+    """
+
```
f0k: There shouldn't be a blank line here, compare to the other classes.

lasagne/tests/test_nonlinearities.py
```diff
@@ -55,6 +72,16 @@ def test_nonlinearity(self, nonlinearity):
         elif nonlinearity == 'scaled_tanh_p':
             from lasagne.nonlinearities import ScaledTanH
             theano_nonlinearity = ScaledTanH(scale_in=0.5, scale_out=2.27)
+        elif nonlinearity.startswith('selu'):
+            from lasagne.nonlinearities import SELU, selu
+            if nonlinearity == 'selu':
+                theano_nonlinearity = SELU(scale=1, scale_neg=1)
```
f0k: I'd even remove the `scale=1, scale_neg=1` here, so the test also ensures nobody can change the default values without us noticing.

lasagne/nonlinearities.py
```diff
@@ -269,6 +268,76 @@ def elu(x):
...
+                x > 0.0,
+                x,
+                self.scale_neg * (theano.tensor.expm1(x)))
+
+
+selu = SELU(scale=1.0507009873554804934193349852946,
+            scale_neg=1.6732632423543772848170429916717)
+selu.__doc__ = """selu(x)
+    Instance of :class:`SELU` with :math:`\\alpha\\approx 1.6733,
+    \\lambda\\approx 1.0507`.
```
f0k: Can you please check whether this ends up in a single line in the overview table of the HTML documentation of the nonlinearities module? I'm not sure if the line break is a problem here. Cheers!

partobs-mdp:
Trying to get Lasagne dev version running on my `<sigh>` new Ubuntu 16.04 LTS, but I ran into two strange bugs:

- Py.test bug - py.test run fails to collect any items. Error log: https://gist.github.com/partobs-mdp/e488fb531726055625b2a67e2c67f266. 
- Sphinx bug - about the same crash as https://github.com/Lasagne/Lasagne/pull/843#issuecomment-310133416, but with other theano parts failing. Error log: https://gist.github.com/partobs-mdp/04b70d4a80d229c541931e8b22324999 (Yes, I tried fetching the `upstream` version)

f0k:
> Py.test bug - py.test run fails to collect any items.

Hmm, I don't immediately see what could be the reason.

> Sphinx bug

Can you try editing `lasagne/docs/conf.py` so before `sys.modules['pylearn2'] = Mock()` you have:
```python
theano.gpuarray = Mock()
sys.modules['theano.gpuarray'] = theano.gpuarray
sys.modules['theano.gpuarray.dnn'] = theano.gpuarray.dnn
theano.gpuarray.pygpu_activated = True
theano.gpuarray.dnn.dnn_present = lambda: True
```
I don't know why it works on my machine and fails on yours, though. It must be taking a different code path somewhere.

partobs-mdp:
> Did you install everything from the `requirements-dev.txt` file?

No (wtf?), `mock` module was indeed missing. The test pool have ran smoothly.

> Sphinx bug

After editing the file `make html` works, thanks!

lasagne/nonlinearities.py
partobs-mdp: I tried to resolve the problem, but I simply can't push the line
``` .. [1] Günter Klambauer, Thomas Unterthiner, Andreas Mayr, Sepp Hochreiter (2017):  ```
shorter than 80 characters, invoking PEP8 check.

Otherwise, I solved all the problems you mentioned. Waiting for your feedback ;)

**Comment by partobs-mdp** [comment]
What is the status of this PR? I seem to have done everything for the PR, maybe I am missing something?

**Comment by reynoldscem** [comment]
To me it's not very pleasant that it has a different API to the other nonlinearities.

I think it would be nicer to just have the two parameters as named parameters instead of set in the constructor. By default they could be as specified by the authors, and if people wish to do the equivalent to constructing the object as you do they can use `functools.partial` or similar.

**Comment by f0k** [comment]
> What is the status of this PR?

Thank you for the updates and sorry for the delay, I'm finishing up my thesis and cannot check up on Lasagne too frequently. I'll review the code in a minute.

> I think it would be nicer to just have the two parameters as named parameters instead of set in the constructor.

`functools.partial` is not commonly known, and `my_selu = functools.partial(lasagne.nonlinearities.selu, scale=a, scale_neg=b)` is not any easier than `my_selu = SELU(scale=a, scale_neg=b)`. So we chose to use classes to implement nonlinearities that have additional parameters such as leaky relu or scaled tanh. For common use cases, there are predefined instances that can be used directly (as you noticed yourself afterwards).

File: docs/conf.py [hunk 15]
```diff
@@ -334,6 +334,12 @@ def setup(app):
 theano.sandbox.cuda.cuda_enabled = True
 theano.sandbox.cuda.dnn.dnn_available = lambda: True
 
+theano.gpuarray = Mock()
+sys.modules['theano.gpuarray'] = theano.gpuarray
+sys.modules['theano.gpuarray.dnn'] = theano.gpuarray.dnn
+theano.gpuarray.pygpu_activated = True
+theano.gpuarray.dnn.dnn_present = lambda: True
+
```
f0k: Technically, this should be a separate PR, but let's not overdo this, a separate commit will be good enough.

lasagne/nonlinearities.py
```diff
@@ -269,6 +269,75 @@ def elu(x):
     return theano.tensor.switch(x > 0, x, theano.tensor.expm1(x))
 
 
+# selu
+class SELU(object):
+    """
+    Scaled Exponential Linear Unit
+    :math:`\\varphi(x)=\\lambda \\left[(x>0) ? x : \\alpha(e^x-1)\\right]`.
```
f0k: Just for consistence with the others, could you please remove the period in the end? (It's not a sentence.)

lasagne/nonlinearities.py
```diff
@@ -269,6 +269,75 @@ def elu(x):
...
+    self-normalizing neural networks.
+
+    Parameters
+    ----------
+    scale : float32
+        The scale parameter :math:`\\lambda` for scaling all output.
+
+    scale_neg  : float32
+        The scale parameter :math:`\\alpha`
+        for scaling output for negative argument values.
```
f0k: For consistency with the formula and implementation, it should be `nonpositive` instead of `negative` (nonpositive includes zero, negative doesn't).

lasagne/nonlinearities.py
```diff
@@ -269,6 +269,75 @@ def elu(x):
...
+
+    >>> from lasagne.layers import InputLayer, DenseLayer
+    >>> l_in = InputLayer((None, 100))
+    >>> from lasagne.nonlinearities import SELU
+    >>> selu = SELU(2, 3)
+    >>> l1 = DenseLayer(l_in, num_units=200, nonlinearity=selu)
+
+    See Also
+    --------
+    selu: Instance with :math:`\\alpha=1.6733,\\lambda=1.0507` as used in [1]_.
```
f0k: I just checked: It's possible to go the correct route with `\\approx` if it is indented correctly:
```
    See Also
    --------
    selu: Instance with :math:`\\alpha\\approx1.6733,\\lambda\\approx1.0507`
          as used in [1]_.
```
Could you do this?

lasagne/nonlinearities.py
```diff
@@ -269,6 +269,75 @@ def elu(x):
...
+                x,
+                self.scale_neg * (theano.tensor.expm1(x)))
+
+
+selu = SELU(scale=1.0507009873554804934193349852946,
+            scale_neg=1.6732632423543772848170429916717)
+selu.__doc__ = """selu(x)
+
+    Instance of :class:`SELU` with :math:`\\alpha\\approx 1.6733,
+    \\lambda\\approx 1.0507`.
```
f0k: Again please remove the period just for consistency in the overview table in the beginning (sorry for the nitpick).

partobs-mdp:
Squashed everything - I think *now* I've done everything for this to be merged :)

f0k:
Wohoo! Merging, thanks again!

f0k:
By the way, next time, I'd recommend to start with `git checkout -b selu` so you have a separate branch for the feature you're working on, instead of implementing it in your local `master` branch: https://guides.github.com/introduction/flow/. This will make it easier for you and also allow you to work on different ideas in parallel.

lasagne/nonlinearities.py
```diff
@@ -269,6 +269,76 @@ def elu(x):
...
+
+    def __call__(self, x):
+        return self.scale * theano.tensor.switch(
+                x > 0.0,
+                x,
+                self.scale_neg * (theano.tensor.expm1(x)))
+
+
+selu = SELU(scale=1.0507009873554804934193349852946,
+            scale_neg=1.6732632423543772848170429916717)
```
andreh7: I think it can be error prone that the default values for `selu()` are different than for `SELU()`. Why not make these values from the paper also the default for `SELU.__init__()` ?

f0k: I see your point. The `SELU` class implements the general concept of a scaled ELU, and `scale=1.0507009873554804934193349852946` is not exactly the scale that would come to mind first. Also I think it's not easy to confuse
```python
DenseLayer(layer, 123, nonlinearity=lasagne.nonlinearities.selu)
```
with
```python
DenseLayer(layer, 123, nonlinearity=lasagne.nonlinearities.SELU())
```
Note that the difference is not only in the capitalization, but also in the brackets (to create an instance of the class). I think this will hardly happen by accident.

I'd be fine with adding a note to the class docstring that it defaults to `elu`, and that `selu` provides the version from the paper.

PR #850, opened 2017-06-17
Transposed Convolution Layers from base class

frapa:
I made a bit of mess with the rebase (first time trying it...). I tried to merge the changes meaningfully. I'm uploading it because today I have no more time, then on Monday I'll see...

### DIFF
--- a/lasagne/layers/conv.py
+++ b/lasagne/layers/conv.py
@@ -12,8 +12,12 @@ __all__ = [
     "Conv1DLayer",
     "Conv2DLayer",
     "Conv3DLayer",
+    "TransposedConv1DLayer",
     "TransposedConv2DLayer",
+    "TransposedConv3DLayer",
+    "Deconv1DLayer",
     "Deconv2DLayer",
+    "Deconv3DLayer",
     "DilatedConv2DLayer",
     "TransposedConv3DLayer",
     "Deconv3DLayer",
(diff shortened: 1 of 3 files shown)


frapa:
I really do not understand the error message I get from Travis... some import problem. The funny thing is on my pc py.test runs fine.

lasagne/layers/conv.py
```diff
@@ -953,9 +1259,8 @@ class TransposedConv3DLayer(BaseConvLayer):  # pragma: no cover
     ----------
     incoming : a :class:`Layer` instance or a tuple
         The layer feeding into this layer, or the expected input shape. The
-        output of this layer should be a 5D tensor, with shape
-        ``(batch_size, num_input_channels, input_depth, input_rows,
-        input_columns)``.
+        output of this layer should be a 4D tensor, with shape
+        ``(batch_size, num_input_channels, input_rows, input_columns)``.
```
f0k: The previous one was correct.

lasagne/layers/conv.py
```diff
@@ -979,7 +1284,7 @@ class TransposedConv3DLayer(BaseConvLayer):  # pragma: no cover
         the `pad` argument in a non-transposed convolution.
 
         A single integer results in symmetric cropping of the given size on all
-        borders, a tuple of three integers allows different symmetric cropping
+        borders, a tuple of two integers allows different symmetric cropping
```
f0k: The previous one was correct.

lasagne/layers/conv.py
```diff
@@ -1005,7 +1310,7 @@ class TransposedConv3DLayer(BaseConvLayer):  # pragma: no cover
 
     W : Theano shared variable, expression, numpy array or callable
         Initial value, expression or initializer for the weights.
-        These should be a 5D tensor with shape
+        These should be a 4D tensor with shape
```
f0k: The previous one was correct.

lasagne/layers/conv.py
```diff
@@ -1005,7 +1310,7 @@ class TransposedConv3DLayer(BaseConvLayer):  # pragma: no cover
...
         ``(num_input_channels, num_filters, filter_rows, filter_columns)``.
```
f0k: Oh, this misses the filter_depth. Seems I overlooked this.

lasagne/layers/conv.py
```diff
@@ -1047,88 +1352,47 @@ class TransposedConv3DLayer(BaseConvLayer):  # pragma: no cover
 
     Notes
     -----
-    The transposed convolution is implemented as the backward pass of a
-    corresponding non-transposed convolution. It can be thought of as dilating
-    the input (by adding ``stride - 1`` zeros between adjacent input elements),
-    padding it with ``filter_size - 1 - crop`` zeros, and cross-correlating it
-    with the filters. See [1]_ for more background.
+    See `BaseTransposedConvLayer` for more.
```
f0k: Aha, I guess we will have to copy this note to all subclasses, as we cannot refer to the base class.

lasagne/layers/conv.py
```diff
@@ -1047,88 +1352,47 @@ class TransposedConv3DLayer(BaseConvLayer):  # pragma: no cover
...
 
     Examples
     --------
     To transpose an existing convolution, with tied filter weights:
 
     >>> from lasagne.layers import Conv3DLayer, TransposedConv3DLayer
-    >>> conv = Conv3DLayer((None, 1, 32, 32, 32), 16, 3, stride=2, pad=2)
+    >>> conv = Conv3DLayer((None, 1, 32, 64, 64), 16, 3, stride=2, pad=2)
```
f0k: I think the previous one was okay.

lasagne/layers/conv.py
```diff
@@ -1047,88 +1352,47 @@ class TransposedConv3DLayer(BaseConvLayer):  # pragma: no cover
...
     >>> deconv = TransposedConv3DLayer(conv, conv.input_shape[1],
     ...         conv.filter_size, stride=conv.stride, crop=conv.pad,
     ...         W=conv.W, flip_filters=not conv.flip_filters)
-
-    References
-    ----------
-    .. [1] Vincent Dumoulin, Francesco Visin (2016):
-           A guide to convolution arithmetic for deep learning. arXiv.
-           http://arxiv.org/abs/1603.07285,
-           https://github.com/vdumoulin/conv_arithmetic
```
f0k: Ah, I think we'll have to replicate this reference in all three subclasses as well.

lasagne/layers/conv.py
```diff
@@ -1047,88 +1352,47 @@ class TransposedConv3DLayer(BaseConvLayer):  # pragma: no cover
...
     """
-    def __init__(self, incoming, num_filters, filter_size,
-                 stride=(1, 1, 1), crop=0, untie_biases=False,
+    def __init__(self, incoming, num_filters, filter_size, stride=1,
+                 crop=0, untie_biases=False,
```
f0k: I'd prefer `(1, 1, 1)` just for consistency with the forward convolutional layers.

lasagne/layers/conv.py
```diff
@@ -1047,88 +1352,47 @@ class TransposedConv3DLayer(BaseConvLayer):  # pragma: no cover
...
+    del TransposedConv1DLayer, Deconv1DLayer
+    __all__.remove('TransposedConv1DLayer')
+    __all__.remove('Deconv1DLayer')
+    
+if not hasattr(T.nnet.abstract_conv,
+               'AbstractConv2d_gradInputs'):  # pragma: no cover
+    # Hide TransposedConv2DLayer for old Theano versions
+    del TransposedConv2DLayer, Deconv2DLayer
+    __all__.remove('TransposedConv2DLayer')
+    __all__.remove('Deconv2DLayer')
```
f0k: The 2D version is always available in the Theano versions we support. I'd prefer to only add the check for 1D, directly after the 1D definitions.

f0k:
> I really do not understand the error message I get from Travis... some import problem.

The exact error is `AttributeError: 'module' object has no attribute 'TransposedConv3DLayer'`, thrown on `from .conv import *`. A `*` import will import all names defined in `__all__`. I can't spot the mistake either; it should be removed from `__all__` just like before your PR.

frapa:
No, maybe theano 0.8 is the problem then. I tried with the bleeding-edge version.

I fixed all the things you mentioned (thanks a lot for the in-depth review). I'm sorry for the many mistakes but it's my first time rebasing and I made a mess...

frapa:
Ok, I tried adding test by copying and modifying the tests for the other cases. I am not at all sure how this whole unit-test thing works (I understand that it should check the output shape, but the code is a bit weird and right now I do not have time to delve into it.)

lasagne/tests/layers/test_conv.py
```diff
@@ -622,6 +672,61 @@ def test_get_params(self, Conv3DImpl, DummyInputLayer):
         assert layer.get_params(_nonexistent_tag=False) == [layer.W, layer.b]
 
 
+class TestTransposedConv1DLayer:
+    @pytest.mark.parametrize(
+        "input, kernel, output, kwargs", list(transp_conv1d_test_sets()))
+    def test_defaults(self, DummyInputLayer, input, kernel, output, kwargs):
+        from lasagne.layers import TransposedConv1DLayer
```
f0k: You'll need to wrap this in a `try..except`:
```python
try:
    from lasagne.layers import TransposedConv1DLayer
except ImportError:
    pytest.skip("TransposedConv1DLayer not available")
```
Same for the second test below. This ensures it will run through on Travis / on Theano 0.8.x.

lasagne/layers/conv.py
```diff
@@ -841,6 +846,11 @@ class TransposedConv2DLayer(BaseConvLayer):
         which of the possible output shapes to return when stride > 1.
         If not specified, the smallest shape will be returned.
 
+    n : int or None
+        The dimensionality of the transposed convolution (i.e., the number of
+        spatial dimensions of each feature map and each convolutional filter).
+        If ``None``, will be inferred from the input shape.
+
```
f0k: The `untie_biases` docstring still needs to be adapted (https://github.com/Lasagne/Lasagne/pull/850/files#r123304501). Unfortunately I cannot comment on the correct line for that because it's not part of the diff.

lasagne/layers/conv.py
```diff
@@ -743,18 +745,19 @@ def convolve(self, input, **kwargs):
...
+class BaseTransposedConvLayer(BaseConvLayer):
     """
-    lasagne.layers.TransposedConv2DLayer(incoming, num_filters, filter_size,
-    stride=(1, 1), crop=0, untie_biases=False,
+    lasagne.layers.BaseTransposedConvLayer(incoming, num_filters, filter_size,
+    stride=1, crop=0, untie_biases=False,
     W=lasagne.init.GlorotUniform(), b=lasagne.init.Constant(0.),
-    nonlinearity=lasagne.nonlinearities.rectify, flip_filters=False, **kwargs)
+    nonlinearity=lasagne.nonlinearities.rectify, flip_filters=False,
+    n=None, **kwargs)
```
f0k: `output_size` is mysteriously missing here (not your fault, but would be good to fix while we're at it, should come before `n`)

lasagne/layers/conv.py
```diff
@@ -919,23 +934,328 @@ def get_output_shape_for(self, input_shape):
...
+    convolution, fractionally-strided convolution or deconvolution in the
+    literature) on its input and optionally adds a bias and applies an
+    elementwise nonlinearity.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        The layer feeding into this layer, or the expected input shape. The
+        output of this layer should be a 4D tensor, with shape
+        ``(batch_size, num_input_channels, input_rows, input_columns)``.
```
f0k: should be 3D, without input_columns

lasagne/layers/conv.py
```diff
@@ -919,23 +934,328 @@ def get_output_shape_for(self, input_shape):
...
+
+    num_filters : int
+        The number of learnable convolutional filters this layer has.
+
+    filter_size : int
+        An integer or 1-element tuple specifying the size of the filters.
+
+    stride : int
+        An integeror 1-element tuple specifying the stride of the
```
f0k: integeror -> integer or

lasagne/layers/conv.py
```diff
@@ -919,26 +950,301 @@ def get_output_shape_for(self, input_shape):
...
+        the filter overlap by at least one position (a full convolution). When
+        ``stride=1``, this yields an output that is larger than the input by
+        ``filter_size - 1``. It can be thought of as a valid convolution padded
+        with zeros. The `crop` argument allows you to decrease the amount of
+        this zero-padding, reducing the output size. It is the counterpart to
+        the `pad` argument in a non-transposed convolution.
+
+        A single integer results in symmetric cropping of the given size on all
+        borders, a tuple of two integers allows different symmetric cropping
+        per dimension.
```
f0k: Still open. (https://github.com/Lasagne/Lasagne/pull/850/files#r123306264)

lasagne/layers/conv.py
```diff
@@ -919,26 +950,301 @@ def get_output_shape_for(self, input_shape):
...
+        integer values due to optimizations by Theano.
+
+    untie_biases : bool (default: False)
+        If ``False``, the layer will have a bias parameter for each channel,
+        which is shared across all positions in this channel. As a result, the
+        `b` attribute will be a vector (1D).
+
+        If True, the layer will have separate bias parameters for each
+        position in each channel. As a result, the `b` attribute will be a
+        3D tensor.
```
f0k: This is correct now.

lasagne/layers/conv.py
```diff
@@ -919,26 +950,301 @@ def get_output_shape_for(self, input_shape):
...
+
+    W : Theano shared variable, expression, numpy array or callable
+        Initial value, expression or initializer for the weights.
+        These should be a 4D tensor with shape
+        ``(num_input_channels, num_filters, filter_rows, filter_columns)``.
```
f0k: This is correct now. I wonder whether my previous comment (https://github.com/Lasagne/Lasagne/pull/850/files#r123304988) was originally attached to the 1D version and github messed up, or whether I accidentally thought it was about the 1D version.

lasagne/layers/conv.py
```diff
@@ -919,26 +950,301 @@ def get_output_shape_for(self, input_shape):
...
+        Note that the first two dimensions are swapped compared to a
+        non-transposed convolution.
+        See :func:`lasagne.utils.create_param` for more information.
+
+    b : Theano shared variable, expression, numpy array, callable or ``None``
+        Initial value, expression or initializer for the biases. If set to
+        ``None``, the layer will have no biases. Otherwise, biases should be
+        a 1D array with shape ``(num_filters,)`` if `untied_biases` is set to
+        ``False``. If it is set to ``True``, its shape should be
+        ``(num_filters, output_rows, output_columns)`` instead.
```
f0k: Still open (https://github.com/Lasagne/Lasagne/pull/850/files#r123305116).

lasagne/layers/conv.py
f0k: Still open (https://github.com/Lasagne/Lasagne/pull/850/files#r123307232).

lasagne/layers/conv.py
```diff
@@ -1006,7 +1325,8 @@ class TransposedConv3DLayer(BaseConvLayer):  # pragma: no cover
     W : Theano shared variable, expression, numpy array or callable
         Initial value, expression or initializer for the weights.
         These should be a 5D tensor with shape
-        ``(num_input_channels, num_filters, filter_rows, filter_columns)``.
+        ``(num_input_channels, num_filters, input_depth, input_rows,
+-        input_columns)``.
```
f0k: Looks like you left an initial `-` in there.

lasagne/layers/conv.py
```diff
@@ -1070,60 +1390,16 @@ class TransposedConv3DLayer(BaseConvLayer):  # pragma: no cover
            http://arxiv.org/abs/1603.07285,
            https://github.com/vdumoulin/conv_arithmetic
     """
-    def __init__(self, incoming, num_filters, filter_size,
-                 stride=(1, 1, 1), crop=0, untie_biases=False,
+    def __init__(self, incoming, num_filters, filter_size, stride=(1, 1, 1),
+                 crop=0, untie_biases=False,
```
f0k: Unnecessary change, better to avoid, but not a big deal.

lasagne/tests/layers/test_conv.py
frapa: Done

lasagne/layers/conv.py
frapa: This was also missing from the three subclasses. Fixed.

lasagne/layers/conv.py
frapa: I forgot about it in the base class, thanks.

lasagne/layers/conv.py
frapa: Another copy-paste error! Sorry!

PR #854, opened 2017-06-26
New merge broadcast

Sentient07:
Continuation of https://github.com/Lasagne/Lasagne/pull/633


lasagne/layers/input.py
```diff
@@ -67,6 +67,9 @@ def __init__(self, shape, input_var=None, name=None, **kwargs):
             if input_var.ndim != ndim:
                 raise ValueError("shape has %d dimensions, but variable has "
                                  "%d" % (ndim, input_var.ndim))
+            broad_pattern = [s == 1 for s in self.shape]
+            input_var = T.patternbroadcast(input_var, broad_pattern)
```
f0k: I thought about that as well, but am not sure if this could introduce problems. If you do:
```python
input_var = T.matrix()
net = InputLayer((100, 1), input_var)
```
then `net.input_var != input_var`. Users might rely on being able to find the input layer for a particular input variable by such a direct comparison. That's why I was leaning towards including this in `lasagne.layers.get_output()`, if at all. I'm unsure.

lasagne/layers/merge.py
```diff
@@ -275,9 +274,8 @@ def get_output_shape_for(self, input_shapes):
                         for sizes in zip(*input_shapes)]
 
         def match(shape1, shape2):
-            axis = self.axis if self.axis >= 0 else len(shape1) + self.axis
             return (len(shape1) == len(shape2) and
-                    all(i == axis or s1 is None or s2 is None or s1 == s2
+                    all(i == self.axis or s1 is None or s2 is None or s1 == s2
```
f0k: This will break the tests, you're removing support for a negative axis this way.

lasagne/layers/merge.py
```diff
@@ -233,8 +233,7 @@ def autocrop_array_shapes(input_shapes, cropping):
             if cr is None:
                 result.append(sh)
             elif cr in {'lower', 'center', 'upper'}:
-                min_sh = None if any(x is None for x in sh) else min(sh)
-                result.append([min_sh] * len(sh))
+                result.append([min(sh)] * len(sh))
```
f0k: This will break the tests, `None` needs to be handled explicitly since it cannot be compared to integers in Python 3.

lasagne/layers/merge.py
```diff
@@ -326,9 +323,36 @@ def __init__(self, incomings, merge_function, cropping=None, **kwargs):
...
+                return ((size_after_broadcast,)*len(input_dim), broadcast)
+            else:
+                return (input_dim, [None]*len(input_dim))
+
+        # if the dimension is broadcastable we replace 1's with the size
+        # after broadcasting.
+        input_dims, broadcastable = list(zip(
+            *[broadcasting(input_dim)for input_dim in zip(*input_shapes)]))
+
+        self.broadcastable = list(zip(*broadcastable))
```
f0k: We cannot set an attribute in `get_output_shape_for()`. There shouldn't be a need for that attribute in the first place.

lasagne/layers/merge.py
```diff
@@ -304,9 +302,8 @@ class ElemwiseMergeLayer(MergeLayer):
 
     Parameters
     ----------
-    incomings : a list of :class:`Layer` instances or tuples
-        the layers feeding into this layer, or expected input shapes,
-        with all incoming shapes being equal
+    incomings : Unless `cropping` is given, all shapes must be equal, except
+        for dimensions that are undefined (``None``) or broadcastable (``1``).
```
f0k: We still need the previous text, it tells what types are accepted. You can add your part in the end, after `being equal`.

lasagne/layers/merge.py
Sentient07: Ah! I ran it in python2 and it was fine. Will change it

lasagne/layers/merge.py
Sentient07: I think this was from the previous PR. It made sense to me as well, to keep this in `__init__`, since you hadn't made any comments there, I left it to be the way it was

lasagne/layers/input.py
Sentient07: I don't understand the problem here. You mean to say they both referred to the same object before(Prior to this PR) and now, they are different objects ? As in, if I understand the concern correctly, this does a deep copying whereas something like shallow copy is desired?

f0k: > You mean to say they both referred to the same object before(Prior to this PR) and now, they are different objects ?

Exactly.

> As in, if I understand the concern correctly, this does a deep copying whereas something like shallow copy is desired?

Hmm, no. It's just that it changes behaviour people might rely on.

```python
def build_model(input_var1, input_var2):
    net = something very complicated
    return net

input_var1, input_var2 = T.matrices('ab')
net = build_model(input_var1, input_var2)

# find layer associated with input_var1
input_layer1 = next(layer for layer in lasagne.layers.get_all_layers(network) if layer.input_var is input_var1)

# compute expression replacing it with something else
something = ...
output = lasagne.layers.get_output(net, {input_layer1: something})
```
This would work before, but not after this PR. I don't know if and how many users rely on this, but in my mental model, the input variable of an input layer is always what I constructed it with.

lasagne/layers/merge.py
f0k: Ah, I see. Don't take the previous PR as something that was exhaustively reviewed :)

lasagne/layers/input.py
Sentient07: > # find layer associated with input_var1
> input_layer1 = next(layer for layer in lasagne.layers.get_all_layers(network)
>                     if getattr(layer, 'input_var', None) is input_var1)

The code indeed would break after this PR, but could we have an extra parameter for the `InputLayer`, which by default is False and would broadcast the `Variable` if only set to True by the user? The above example wouldn't break. Also we can suggest in the error message from the `MergeLayer` in case if the input happens to have a `1` in shape and the MergeLayer fails because of a dimension mismatch. I confirmed with @nouiz that it is not possible to do a broadcasting without building a graph

lasagne/layers/merge.py
```diff
@@ -326,9 +323,36 @@ def __init__(self, incomings, merge_function, cropping=None, **kwargs):
...
         self.cropping = cropping
+        self.broadcastable = None
 
     def get_output_shape_for(self, input_shapes):
         input_shapes = autocrop_array_shapes(input_shapes, self.cropping)
+
+        input_dims = [len(shp) for shp in input_shapes]
+        if not all(input_dim == input_dims[0] for input_dim in input_dims):
+            raise ValueError('Input dimensions must be the same but were %s' %
+                             ", ".join(map(str, input_shapes)))
```
Sentient07: I thought I finished my comments(this and at a couple of more places). I came back to see the diff and realized it's not published on this page until I click the "submit review" button (after writing all the comments).

Sentient07:
One comment is yet to be addressed(moving the assignment of `self.broadcastable` to the constructor, I'll push that commit by tonight.

lasagne/layers/input.py
f0k: > The code indeed would break after this PR, but could we have an extra parameter for the `InputLayer`, which by default is False and would broadcast the Variable if only set to True by the user?

Mhm, but then what's the advantage over adding the broadcast in `get_output()`, when collecting the input variables? We cannot catch all cases in the `InputLayer` constructor anyway. If a user supplies the input variables to `lasagne.layers.get_output()` instead, which is another common use pattern, then the `InputLayer` cannot influence them in any way.

f0k:
> One comment is yet to be addressed

Let's quickly check whether this is really what we want. There are several options:
1. Ensuring the correct broadcast pattern in the `InputLayer` (as I commented above, this is not a good solution)
2. Ensuring the correct broadcast pattern in `lasagne.layers.get_output` when collecting the inputs (it further complicates the `get_output` function, but what can you do...)
3. Ensuring the correct broadcast pattern in `MergeLayer`

If we go for solution 2, do we still need solution 3 or can we rely on the MergeLayer getting correctly broadcastable tensors?

If we go for solution 3, we don't need a `self.broadcastable` attribute. `get_output_for()` can access `self.input_shapes` to compute the broadcast patterns on the fly. I think this is cleaner.

No matter whether we choose solution 2 or 3, I'd suggest a `lasagne.utils.ensure_broadcasts(var, shape)` function encapsulating the steps: (1) infer the broadcast pattern from the shape, (2) compare it to the broadcast pattern of the variable, (3) add a T.patternbroadcast if the patterns do not match.

nouiz:
For opt 2  is this the get output of the input layer? What about putting it
there?

Le ven. 30 juin 2017 08:44, Jan Schlüter <notifications@github.com> a
écrit :

Sentient07:
Very sorry for the late reply. 

> Ensuring the correct broadcast pattern in MergeLayer

So, we do the check which we have previously discussed, by iterating through the dimensions of the input Variables, checking if the dimension is `1` and setting the broadcastable attribute to `True`, right ? I am not sure of this case, but if there is a tie, that two variables have their Nth dimension to be `1`, which one do we make broadcastable ?

>  get_output_for() can access self.input_shapes to compute the broadcast patterns on the fly. 

But can't we ensure the broadcast pattern as and when the `MergeLayer` is called? There might be cases where `MergeLayer` is the output of a network and a possible error due to `Shape` mismatch would be thrown in the subsequent step. 

> I'd suggest a lasagne.utils.ensure_broadcasts(var, shape) 

Seems fine with me. Wouldn't break any existing code.

@nouiz 

> For opt 2  is this the get output of the input layer? What about putting it
there?

Unfortunately, the `InputLayer` doesn't have `get_output_for` method. I am also fine with adding the method to that class and having this functionality here, but again, I'm afraid doing this broadcasting work would make the variable not same as what was passed. If there is a way by which we can have the `same` Variable, but with altered broadcast pattern, that would be perfect IMO.

f0k:
> I am not sure of this case, but if there is a tie, that two variables have their Nth dimension to be 1, which one do we make broadcastable ?

All of them. Ideally, any variable that has a shape that we know to be `1` should be broadcastable in that dimension.

> But can't we ensure the broadcast pattern as and when the `MergeLayer` is called?

The constructor should check that the shapes match, but we should not store the broadcast pattern in an attribute to reuse it in `get_output_for`. This would suggest that it's something we can only compute in the constructor, but actually we can just as well compute it in `get_output_for`.

> Unfortunately, the `InputLayer` doesn't have `get_output_for` method.

Exactly. The reason was that the definition is `get_output_for(self, input, **kwargs)`, but the `InputLayer` does not process an `input`, it provides the first input. So we handle the input layers as special cases in `lasagne.layers.get_output`.

> I am also fine with adding the method to that class and having this functionality here

That's not really possible. We need to support `lasagne.layers.get_output(network, {somelayer: myvar})`, which assigns `myvar` as the output of `somelayer`. If we add a `get_output_for` method to an `InputLayer`, we still cannot call it when overriding its output with a custom variable. This would change the semantics to assigning `myvar` as the input of `somelayer`, which makes a huge difference if `somelayer` is anything else than an `InputLayer`.

> If there is a way by which we can have the same Variable, but with altered broadcast pattern, that would be perfect IMO.

There are three common usage patterns:
```python
network = InputLayer((1, 2, 3))
myvar = network.input_var
network = DenseLayer(network, 10)
output = lasagne.layers.get_output(network)
```
```python
myvar = T.tensor3('input')
network = InputLayer((1, 2, 3), myvar)
network = DenseLayer(network, 10)
output = lasagne.layers.get_output(network)
```
```python
network = InputLayer((1, 2, 3))
network = DenseLayer(network, 10)
myvar = T.tensor3('input')
output = lasagne.layers.get_output(network, myvar)
```
(Ignore the fact that this example doesn't actually require broadcastable input to the dense layer, I didn't want to create a more complex example.)
The first usage pattern is currently handled correctly by the InputLayer, since it creates the input variable that is used later. The second usage pattern could be handled by messing with the variable, but has the disadvantage that `network.input_layer.input_var is not myvar` afterwards. The third pattern can only be handled within `get_output`.

So I would suggest to add a helper function and prep the inputs when they're collected in `get_output`.

PR #855, opened 2017-06-28
Feature alpha dropout

rfratila:
This is my first PR to Lasagne so I'm definitely looking for suggestions. To complement SELU activation, I've made a new AlphaDropoutLayer as specified in the paper: https://arxiv.org/abs/1706.02515


lasagne/layers/noise.py
```diff
@@ -211,3 +212,88 @@ def get_output_for(self, input, deterministic=False, **kwargs):
...
+                mask_shape = tuple(1 if a in shared_axes else s
+                                   for a, s in enumerate(mask_shape))
+
+            mask = self._srng.uniform(mask_shape, dtype=input.dtype) < q
+
+            if self.shared_axes:
+                bcast = tuple(bool(s == 1) for s in mask_shape)
+                mask = T.patternbroadcast(mask, bcast)
+
+            a = T.pow(q + self.alpha ** 2 * q * (1 - q), -0.5)
```
f0k: Should use `p`, not `1 - q`, otherwise it might result in an unwanted upcast again. Should also use `T.square` instead of `** 2` to make the graph smaller.

lasagne/layers/noise.py
```diff
@@ -211,3 +212,88 @@ def get_output_for(self, input, deterministic=False, **kwargs):
...
+
+            b = -a * (1 - q) * self.alpha
```
f0k: Again, should use `p`, not `1-q`.

lasagne/layers/noise.py
```diff
@@ -211,3 +212,88 @@ def get_output_for(self, input, deterministic=False, **kwargs):
...
+           Self-Normalizing Neural Networks. arXiv preprint: 1706.02515
+    """
+
+    def __init__(self, incoming, p=0.1, rescale=True, shared_axes=(),
+                 **kwargs):
+        """Class initialization."""
+        super(AlphaDropoutLayer, self).__init__(incoming, **kwargs)
+        self._srng = RandomStreams()
+        self.p = p
+        self.alpha = -1.7580993408473766
```
f0k: Oh, and we should use all available digits, because we can. And maybe call it differently, since it's called alpha prime in the paper, computed as minus lambda times alpha? Not sure.

The default value would be `-1.0507009873554804934193349852946 * 1.6732632423543772848170429916717`.

rfratila: Good idea, I'll default it to that and create an additional parameter. The floating point precision is the difference between the calculated alpha and what I have.

rfratila:
I apologize for all of the commits @f0k I'm having some trouble squashing them together. Any help with this would be very much appreciated.

Thanks

f0k:
Thank you for continuing, and sorry for the delay!

> I'm having some trouble squashing them together. Any help with this would be very much appreciated.

I guess the easiest to get rid of all the merge commits in between will be to squash everything into a single commit for now. In the end I'd like two commits -- one refactoring the dropout layer and one introducing alpha dropout -- but we can take care of that later. To squash, do:
```bash
git reset --soft edbbc12  # set pointer to the first commit of the PR, leave other changes stages
git commit --amend --no-edit  # tack all other changes to the first commit
git push --force  # overwrite the current PR with the squashed version
```
I have made a local backup of the current state of your PR in case anything goes wrong.

I will review the code afterwards, as I don't know if the reviews will stick after squashing.

rfratila:
Thanks for the assist! I think everything is still working as I left it. Should be ready for another round of reviews.

rfratila:
@f0k Have you had a chance to look at the changes?

lasagne/layers/noise.py
```diff
@@ -211,3 +219,69 @@ def get_output_for(self, input, deterministic=False, **kwargs):
...
+                 alpha=None, **kwargs):
+        """Class initialization."""
+        super(AlphaDropoutLayer, self).__init__(incoming,
+                                                p=p,
+                                                rescale=rescale,
+                                                shared_axes=shared_axes,
+                                                **kwargs)
+        if alpha is None:
+            self.alpha = -1.0507009873554804934193349852946 * \
+                         1.6732632423543772848170429916717
```
andreh7: would it make sense to mention in a comment that these two values are fixed point solutions to equations (4) and (5) in [1] for zero mean and unit variance input and that the analytic expressions for them are given in equation (14) ?

Furthermore, these value for alpha should be consistent with a `SELU` input layer, would a possible option be to pass a `SELU` instance for the parameter alpha and take the value from there, e.g.:

```
elif isinstance(alpha, SELU):
    self.alpha = - alpha.scale * alpha.scale_neg
```
?

rfratila: You bring up good points. I will update the documentation accordingly. Regarding the option to pass a SELU instance and calculate the alpha based on that, I think that since AlphaDropout is meant to work with SELU, it only makes sense for it to be able to use SELU. I will make this change.

f0k:
Sorry, had to finish my thesis first. You're including the commits of the SELU PR in your PR, can you please rebase onto master instead? It's more difficult to review otherwise. Cheers!
```bash
git fetch upstream master
git rebase upstream/master
git push --force  # assuming that the rebase went through without conflicts, otherwise git rebase --abort and let me know
```

rfratila:
No worries and congrats! I rebased onto master. Resolved a few conflicts and pushed. Everything is as it was.

f0k:
Oops, sorry, introduced a conflict by merging #871. Will review and can then rebase your PR onto the latest master myself.

lasagne/tests/layers/test_noise.py
```diff
@@ -155,3 +155,103 @@ def test_specified_rng(self, input_layer):
...
+        result = AlphaDropoutLayer(input_layer).get_output_for(input)
+        result_eval2 = result.eval()
+
+        set_rng(rng)  # reset to original RNG for other tests
+        assert numpy.allclose(result_eval1, result_eval2)
+
+    def test_alpha(self, input_layer):
+        input = theano.shared(numpy.ones((100, 100)))
+        from lasagne.layers.noise import AlphaDropoutLayer
+        layer = AlphaDropoutLayer(input_layer, alpha=-1.7580993408473766)
```
f0k: But that's just the default value for `alpha`. It should test another value, otherwise you never know whether that value was actually respected or it took the default anyway :)

lasagne/tests/layers/test_noise.py
```diff
@@ -155,3 +155,103 @@ def test_specified_rng(self, input_layer):
...
+        assert 0.8 < layer.get_output_for(input).eval().mean() < 1.0
+
+    def test_SELU_alpha(self, input_layer):
+        input = theano.shared(numpy.ones((100, 100)))
+        from lasagne.layers.noise import AlphaDropoutLayer
+        from lasagne.nonlinearities import selu  # SELU instance
+        layer = AlphaDropoutLayer(input_layer, alpha=selu)
```
f0k: Same here, it should be a SELU instance with nonstandard scales. You can then test whether self.alpha was set correctly, in addition to checking the output.

lasagne/layers/noise.py
```diff
@@ -80,29 +81,36 @@ def get_output_for(self, input, deterministic=False, **kwargs):
...
-                mask = T.patternbroadcast(mask, bcast)
+            return self.apply_dropout(input, const=0)
+
+    def apply_dropout(self, input, const=0):
+        # Using theano constant to prevent upcasting
+        one = T.constant(1)
+
+        self.q = one - self.p
+        if self.rescale:
+            input /= self.q
```
f0k: `self.q` should not be set here, it should become a property of the layer (i.e., not an attribute):
```python
@property
def q(self):
    return T.constant(1) - self.p
```

Then you never need to set `self.q` and can use it in `apply_dropout()` and in `get_output_for()`.

lasagne/layers/noise.py
```diff
@@ -80,29 +81,36 @@ def get_output_for(self, input, deterministic=False, **kwargs):
...
+            mask_shape = tuple(1 if a in shared_axes else s
+                               for a, s in enumerate(mask_shape))
+        mask = self._srng.binomial(mask_shape, p=self.q,
+                                   dtype=input.dtype)
+        if self.shared_axes:
+            bcast = tuple(bool(s == 1) for s in mask_shape)
+            mask = T.patternbroadcast(mask, bcast)
+
+        if const != 0:
+            return (input * mask) + (const * (1 - mask))
```
f0k: This has to use `T.constant(1)` instead of `1` again.

lasagne/layers/noise.py
```diff
@@ -211,3 +219,76 @@ def get_output_for(self, input, deterministic=False, **kwargs):
             return input + self._srng.normal(input.shape,
                                              avg=0.0,
                                              std=self.sigma)
+
+
+class AlphaDropoutLayer(DropoutLayer):
+    """Dropout layer.
+
+    Sets values to alpha. Will also converge the remaining
+    neurons to mean 0 and a variance of 1 unit. See notes for disabling dropout
```
f0k: Sorry, this explanation is a bit imprecise. Can you be more explicit what the layer does? (If you can't find anything, I'll try that myself.)

lasagne/layers/noise.py
```diff
@@ -211,3 +219,76 @@ def get_output_for(self, input, deterministic=False, **kwargs):
...
+    during testing.
+
+    Parameters
+    ----------
+    incoming : a :class:`Layer` instance or a tuple
+        the layer feeding into this layer, or the expected input shape
+    p : float or scalar tensor
+        The probability of setting a value to zero
```
f0k: Only that it's not set to zero...

lasagne/layers/noise.py
```diff
@@ -211,3 +219,76 @@ def get_output_for(self, input, deterministic=False, **kwargs):
...
+        dropout is enabled, to keep the expected output mean the same.
+    shared_axes : tuple of int
+        Axes to share the dropout mask over. By default, each value can be
+        dropped individually. ``shared_axes=(0,)`` uses the same mask across
+        the batch. ``shared_axes=(2, 3)`` uses the same mask across the
+        spatial dimensions of 2D feature maps.
+    alpha : float or SELU instance
+        The default values are fixed point solutions to equations (4) and (5)
+        in [1] for zero mean and unit variance input. The analytic expressions
+        for them are given in equation (14) also in [1].
```
f0k: Should probably first state what role `alpha` plays for the operation of the layer. Also the reference must be given as `[1]_`, not `[1]`.

lasagne/layers/noise.py
```diff
@@ -211,3 +219,76 @@ def get_output_for(self, input, deterministic=False, **kwargs):
...
+
+    Notes
+    -----
+    The alpha dropout layer is a regularizer that randomly sets input values to
+    zero and also applies a normalization function to bring the mean to 0
+    and the variance to 1; see [1]_for why this might improve
```
f0k: That used to reference the original dropout paper. Either the reference or the text should be changed.

lasagne/layers/noise.py
```diff
@@ -211,3 +219,76 @@ def get_output_for(self, input, deterministic=False, **kwargs):
...
+    def get_output_for(self, input, deterministic=False, **kwargs):
+        """Apply alpha dropout."""
+        if deterministic or self.p == 0:
+            return input
+        else:
+            mask = self.apply_dropout(input, const=self.alpha)
+            a = T.pow(self.q + T.square(self.alpha) * self.q * self.p, -0.5)
+            b = -a * self.p * self.alpha
+
+            return T.cast(a * mask + b, input.dtype)
```
f0k: Why do you need this cast? Can you trace back why it is necessary? If possible, it would be better to cast earlier, otherwise Theano might execute part of the graph on CPU and then only cast and copy the results back to GPU afterwards.

rfratila: I believe it returns a float64 so I just want to make sure the dtype of the output is the same as the dtype of the input

rfratila:
@f0k I have made the changes that address all your comments. Let me know what you think.

f0k:
Thank you, the code looks good now! For the docstring I'll need to have a closer look, unfortunately I'll be abroad without web access for the next two weeks -- sorry to let you wait, I'll get back to this when I return!

rfratila:
Have you had a chance to review the docstring in more detail @f0k ? Please let me know if anything is inaccurate. Thanks!

PR #857, opened 2017-07-06
Add 2D bilinear upsampling layer

christopher-beckham:
Hi,

After being frustrated so many times with deconvolution artifacts, I decided to try out bilinear upsampling and it turns out this already existed in Theano, so I thought it would be reasonable to contribute this as a layer to Lasagne. I have tried this so far with some of my research involving pix2pix (conditional image-to-image translation with GANs), and the use of this + followed by a shape-preserving convolution (i.e., a conv with padding=same) gave me much better outputs than simply using deconv!

You can read more about it here:

http://distill.pub/2016/deconv-checkerboard/

Admittedly, my unit test `test_get_output_for` is a bit crude -- it simply checks the output shape, and not against a Numpy implementation of bilinear upsampling. Apart from that, everything seems to look ok. Feel free to offer suggestions, etc. I'd love to see this kind of layer upstream.

### DIFF
--- a/lasagne/layers/pool.py
+++ b/lasagne/layers/pool.py
@@ -12,6 +12,7 @@ __all__ = [
     "Pool3DLayer",
     "Upscale1DLayer",
     "Upscale2DLayer",
+    "UpscaleBilinear2DLayer",
     "Upscale3DLayer",
     "FeaturePoolLayer",
     "FeatureWTALayer",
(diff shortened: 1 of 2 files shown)


f0k:
Thank you for the contribution and sorry for the late reply, I've been busy with my thesis.

> gave me much better outputs than simply using deconv!

The easier / more efficient solution that popped up in between may be #862.

Still, bilinear upsampling would be nice. What about adding this as a different mode to `Upscale2DLayer` instead? We could have `bilinear` for all-in-one and `bilinear1d` for separate passes per dimension. It's nice that the complexity of creating the kernels and convolution expressions is hidden away in Theano.

christopher-beckham:
Interesting, yes, I think it would be cleaner to have this as an option for `Upscale2D`. I'll have a look at that paper too.

f0k:
> Interesting, yes, I think it would be cleaner to have this as an option for `Upscale2D`.

Would you like to work on this?

christopher-beckham:
Hi f0k, gahhh sorry for that one month delay... looks like we've both been swamped. Yes, absolutely, I'd be happy to. Let me try get back to you next week.

Issue #812, opened 2017-03-06
Update recurrent layer module docstring


f0k:
The [module docstring for `lasagne.layers.recurrent`](http://lasagne.readthedocs.io/en/latest/modules/layers/recurrent.html) has an example that uses reshapes to put a dense layer on top of a recurrent layer -- this should be updated to use the `num_leading_axes` argument of the `DenseLayer` instead.
In addition, it would be good to have a second example that uses a Conv1D layer to preprocess a sequence fed into an RNN, to highlight (a) that this is possible and (b) that you'll need a DimshuffleLayer in between (see https://groups.google.com/d/msg/lasagne-users/A5UCHKxo_Ys/GO2MV6XuDAAJ).

Sentient07:
I'll take up this. The first two(wrt to your yesterday's comment), we can probably have an `welcome contribution` tag ? I know a couple of people who might be interested and are a bit new to Open source contributions and can work on those two. Is that fine ? (if there is no one to take that before the end of the week, i'll finish it)

f0k:
> we can probably have an `welcome contribution` tag

I thought about this as well, but I'd actually welcome contributions for every issue on the tracker. So I put an `easy` tag on the ones most suitable for newcomers.

PR #869, opened 2017-07-25
Modification of RNN doc

Sentient07:
Addresses #812.

### DIFF
--- a/lasagne/layers/recurrent.py
+++ b/lasagne/layers/recurrent.py
@@ -49,10 +49,9 @@ network with variable batch size and number of time steps.
 >>> batchsize, seqlen, _ = l_inp.input_var.shape
 >>> l_lstm = LSTMLayer(l_inp, num_units=num_units)
 >>> # In order to connect a recurrent layer to a dense layer, we need to
->>> # flatten the first two dimensions (our "sample dimensions"); this will
->>> # cause each time step of each sequence to be processed independently
->>> l_shp = ReshapeLayer(l_lstm, (-1, num_units))
->>> l_dense = DenseLayer(l_shp, num_units=num_classes)
+>>> # process each time step of each sequence to be independently.
+>>> # this will be taken care of by setting the `num_leading_axes` to -1
+>>> l_dense = DenseLayer(l_lstm, num_units=num_classes, num_leading_axes=-1)
 >>> # To reshape back to our original shape, we can use the symbolic shape
 >>> # variables we retrieved above.
 >>> l_out = ReshapeLayer(l_dense, (batchsize, seqlen, num_classes))


Sentient07:
@f0k sorry about the delay, seems like this week, all the time that I said I'd be having are during the weekend. To my eyes, this PR is done. Would love to have some comments if I have missed anything

Thanks

lasagne/layers/recurrent.py
```diff
@@ -49,10 +49,18 @@
 >>> batchsize, seqlen, _ = l_inp.input_var.shape
 >>> l_lstm = LSTMLayer(l_inp, num_units=num_units)
 >>> # In order to connect a recurrent layer to a dense layer, we need to
->>> # flatten the first two dimensions (our "sample dimensions"); this will
->>> # cause each time step of each sequence to be processed independently
->>> l_shp = ReshapeLayer(l_lstm, (-1, num_units))
->>> l_dense = DenseLayer(l_shp, num_units=num_classes)
+>>> # process each time step of each sequence to be independently.
```
f0k: grammar broken!

lasagne/layers/recurrent.py
```diff
@@ -49,10 +49,18 @@
...
+>>> # this will be taken care of by setting the `num_leading_axes` to -1
```
f0k: this -> This

lasagne/layers/recurrent.py
```diff
@@ -49,10 +49,18 @@
...
+>>> l_dense = DenseLayer(l_lstm, num_units=num_classes, num_leading_axes=-1)
+>>> # An example involving Feeding the output of `LSTMLayer` to `Conv1DLayer`
```
f0k: That's not what I was talking about, it was about preprocessing the input. And it should be a second example, not mixed into the first, that makes it too confusing (especially since you don't use the result in the rest of the network).

lasagne/layers/recurrent.py
```diff
@@ -49,10 +49,18 @@
...
+>>> # input_length) and for a recurrent layer(`LSTMLayer` here) is
+>>> # (batch_size, sequence_length, num_input_channels). So, we need to 
+>>> # swap the last two dimension to ensure expected behaviour.
+>>> # Note : In this example, if the `Dimshuffle` layer is removed,
+>>> # `ValueError` will be thrown by the virtue of using `None` in the shape.
+>>> l_dshuf = DimshuffleLayer(l_lstm, (0, 2, 1))
+>>> l_conv1D = Conv1DLayer(l_dshuf, num_classes, 1)
 >>> # To reshape back to our original shape, we can use the symbolic shape
 >>> # variables we retrieved above.
 >>> l_out = ReshapeLayer(l_dense, (batchsize, seqlen, num_classes))
```
f0k: This is not required with `num_leading_axes`.

lasagne/layers/recurrent.py
Sentient07: I am not sure I understood this comment very well, do you mean I should create a new network from the start using an InputLayer for the second example?

Sentient07: @f0k  ping. I'm not sure if the change that I did in my recent commit addresses the concern that you raised. Please correct my changes if it's wrong.

Sentient07:
Updated grammar and fixed the Travis error. In case anything else needs to be done here, I'll fix it

PR #870, opened 2017-07-29
Added required doc for categorical cross entropy

### LINKED ISSUE #591: Documentation of categorial cross-entropy
It would be nice if the meaning of the indices i and j would be explained in http://lasagne.readthedocs.org/en/latest/modules/objectives.html#lasagne.objectives.categorical_crossentropy

Sentient07:
Fixes #591

### DIFF
--- a/lasagne/objectives.py
+++ b/lasagne/objectives.py
@@ -149,6 +149,9 @@ def categorical_crossentropy(predictions, targets):
 
     .. math:: L_i = - \\sum_j{t_{i,j} \\log(p_{i,j})}
 
+    :math:`p` are the predictions, :math:`t` are the targets, :math:`i`
+    denotes the data point and :math:`j` denotes the class.
+
     Parameters
     ----------
     predictions : Theano 2D tensor
@@ -157,6 +160,8 @@ def categorical_crossentropy(predictions, targets):
     targets : Theano 2D tensor or 1D tensor
         Either targets in [0, 1] matching the layout of `predictions`, or
         a vector of int giving the correct class index per data point.
+        In the case of an integer vector argument, each element
+        represents the position of the '1' in a one-hot encoding.
 
     Returns
     -------


lasagne/objectives.py
```diff
@@ -145,7 +145,9 @@ def binary_crossentropy(predictions, targets):
 
 
 def categorical_crossentropy(predictions, targets):
-    """Computes the categorical cross-entropy between predictions and targets.
+    """Computes the categorical cross-entropy between predictions $p_{i,j}$
+        and targets $t_{i,j}$, where $i$ denotes the data point and $j$
+        denotes the class.
```
f0k: Sounds good, but you completely messed up the formatting :)

a) the first sentence must be short since it's used for the overview table
b) the text must not be indented otherwise it's assumed to be a code block
c) the math syntax in Sphinx does not use $

Please always compile and look at the documentation when changing docstrings in a nontrivial way: http://lasagne.readthedocs.io/en/latest/user/development.html#documentation

So the explanation of p, t, i, j should come after the formula, or in a second sentence before the formula, not in the first sentence.

lasagne/objectives.py
```diff
@@ -157,6 +159,8 @@ def categorical_crossentropy(predictions, targets):
     targets : Theano 2D tensor or 1D tensor
         Either targets in [0, 1] matching the layout of `predictions`, or
         a vector of int giving the correct class index per data point.
+        In the case of an integer vector argument, each element
+        represents the position of the '1' in a 1-of-N encoding.
```
f0k: This makes it extra clear. Good!

Sentient07:
I've pushed a commit with the suggested changes. Here's how the doc looks, 
screenshot at aug 14 04-19-10

Sentient07:
Sorry about that, there was a PEP8 error. Fixed that and committed. Once this PR is LGTM(to @f0k ), I will squash all the commits into one

f0k:
Modified the formulation, corrected the punctuation and squashed it. Hope you don't mind, it was easier this way. Thanks again, merging!

PR #889, opened 2017-12-31
Better error message when importing cuda_convnet under new GPU backend

### LINKED ISSUE #866: Better error message when importing cuda_convnet
When importing `lasagne.layers.cuda_convnet` under Theano 0.10, it will blindly try to import theano.sandbox.cuda and Theano will throw a nasty error message about switching to the new GPU backend, even if the user actually uses the new backend.
`lasagne/layers/cuda_convnet.py` should first check if the new GPU backend is in use and give a better error message in this case (along the lines of "cuda_convnet is not supported under Theano's new GPU backend. Please either use the ordinary convolutional and pooling layers (they are usually faster), or downgrade Theano to version 0.9 and use its old GPU backend.").
This would have helped at least one user who had a `from lasagne.layers.cuda_convnet import Conv2DCCLayer` in his code for some reason: https://groups.google.com/forum/#!topic/lasagne-users/qgoa9BBamDA

f0k:
Fixes #866. Closes #882 (which had the check wrong).

### DIFF
--- a/lasagne/layers/cuda_convnet.py
+++ b/lasagne/layers/cuda_convnet.py
@@ -13,6 +13,18 @@ from .conv import conv_output_length, BaseConvLayer
 from .pool import pool_output_length
 from ..utils import as_tuple
 
+# bail out if using Theano's new GPU backend
+try:
+    from theano import gpuarray
+except ImportError:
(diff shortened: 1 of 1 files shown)


nouiz:
Should this PR be merged?

PR #898, opened 2018-02-19
Issue refactor upscale layer

### LINKED ISSUE #737: Refactor Upscale*DLayer
Upscale1DLayer, Upscale2DLayer, Upscale3DLayer share a lot of code (and tests). They could be refactored to be derived from a common UpscaleLayer that supports any number of spatial dimensions, or maybe even any set of axes (so it could be used to repeat channels or timesteps).

teorikraften:
closes #737

### DIFF
--- a/lasagne/layers/pool.py
+++ b/lasagne/layers/pool.py
@@ -626,6 +626,14 @@ class UpscaleLayer(Layer):
             Dimension must be less or equal to dimension of input.
             """
             raise ValueError(msg.format(dim))
+
+        if dim > 3:
+            msg = """
+            Invalid input dimension: {0}.
+            Number of spatial dimensions greater than 3 are not yet supported.
+            """
+            raise ValueError(msg.format(dim))
+
         self.dimension = dim
 
     def get_output_shape_for(self, input_shape):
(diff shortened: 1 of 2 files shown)


lasagne/layers/pool.py
```diff
@@ -579,14 +579,18 @@ def __init__(self, incoming, pool_size, stride=None, pad=(0, 0, 0),
...
+    Upscaling layer
 
-    Performs 1D upscaling over the trailing axis of a 3D input tensor.
+    Performs upscaling over the trailing axes of the input tensor.
 
     Parameters
     ----------
+
+    dim: integer
+        Dimension for upscaling. Must be smaller or equal to dimension of input
```
f0k: The first argument should be `incoming`, to be consistent with all the other layer classes. I'd see `UpscaleLayer` not only as a hidden class to implement `Upscale*DLayer`, but as a layer that can be used by itself.
Furthermore, I'd like to give it an `axes` argument instead of a dimensionality, so you can use it to scale a subset of the dimensions, and even to repeat instances or feature maps.

lasagne/layers/pool.py
```diff
@@ -601,46 +605,103 @@ class Upscale1DLayer(Layer):
...
         return upscaled
 
 
-class Upscale2DLayer(Layer):
+class Upscale1DLayer(UpscaleLayer):
+    """
+    1D upscaling layer
+
+    Calls the class UpscaleLayer with appropriate parameters to perform
+    1D upscaling over the trailing axis of a 3D input tensor
```
f0k: The docstring should be the same as before. We can point to the base class in a `See also` section (there are examples for the proper docstring syntax in layers/noise.py, for example).

PR #913, opened 2018-08-21
minor updates

some-0n3:
Hi,
I mainly replaced `pep8` with `codestyle` and tried to fix up all the errors.
I also added the new theano version (since 1.0.0 everything should work out).


f0k:
Thank you for your PR and sorry for the very late reply! Unfortunately, doing an all-over update of code formatting is unwanted, as this will make it hard to merge any pending PRs. Furthermore, we still support Theano versions older than 1.0, so we should not require 1.0. We only do integration testing against 0.8.2, although I know it works with 1.0 as well. If the `Theano==0.8.2` requirement is complicating things, I'd be fine changing it to `Theano>=0.8.2`. This should be a separate PR, though, with a single commit, and without changes to the code style.

PR #921, opened 2019-11-20
Restore compatibility with pytest

f0k:
pytest has deprecated using a `[pytest]` section in `setup.cfg`; it only supports `[tool:pytest]` now. This causes Travis builds to fail. We could either fix pytest to an old version in `requirements-dev.txt`, or fix `setup.cfg` -- this PR chooses the latter.

Furthermore, the default format for printing floating point arrays has changed in numpy; there was an additional leading space in front of every number in earlier versions. This causes one of the doctests to fail. Again, we could either fix a particular numpy dependency or fix the doctest. This PR chooses the latter, setting the formatting options inside the doctest (the NORMALIZE_WHITESPACE flag is not enough because it does not cover the change from `[ 1.0` to `[1.0`).


f0k:
Tests now fail because we pinned `Theano==0.8.2`, but use the latest version of numpy, and numpy >= 1.16 requires Theano>=1.0.4: https://github.com/Theano/Theano/pull/6671
We could simply change requirements to `Theano>=0.8.2`. We could also pin `Theano==1.0.4`, but Lasagne itself is compatible with earlier versions.

